Issue #117, opened 2026-01-31
D-flow profiles do not seem to import cleanly


skialpine:
D-flow profiles show as need to updated to be same as what’s on machine but do not update. Here is a video showing what I seeing.

PR #144, opened 2026-02-06
Fix D-Flow profiles always showing needs update after import

skialpine:
## Summary
- D-Flow profiles always showed as "different" (needs update) after import, and overwriting didn't fix it
- Root cause: `toJson()` only saved `max_flow_or_pressure_range` when `max_flow_or_pressure > 0`, but D-Flow profiles set range to `0.2` even when limiter value is `0`
- On reload, `fromJson()` defaulted range to `0.6`, so `compareProfileFrames()` always found a diff (`|0.2 - 0.6| > 0.1`)
- Fix: always save both limiter fields to JSON for lossless round-trip
- No machine behavior change — when limiter value is 0, no BLE extension frame is sent (matching de1app)

## Test plan
- [ ] Import a D-Flow TCL profile (e.g. D-Flow / Q)
- [ ] Verify it imports successfully
- [ ] Re-scan profiles — D-Flow should show as "identical", not "different"
- [ ] Import a profile with active limiter (e.g. any profile with max_flow_or_pressure > 0) and verify it still works correctly
- [ ] Run a shot with D-Flow profile and verify machine behavior is unchanged

### DIFF
--- a/src/profile/profileframe.cpp
+++ b/src/profile/profileframe.cpp
@@ -26,11 +26,10 @@ QJsonObject ProfileFrame::toJson() const {
     if (exitFlowUnder > 0) obj["exit_flow_under"] = exitFlowUnder;
     if (exitWeight > 0) obj["exit_weight"] = exitWeight;
 
-    // Limiter
-    if (maxFlowOrPressure > 0) {
-        obj["max_flow_or_pressure"] = maxFlowOrPressure;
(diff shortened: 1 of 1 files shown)


skialpine:
This is a fix to https://github.com/Kulitorum/Decenza/issues/117 which you incorrectly closed earlier as being fixed. This only seem to happen with D-FLow tcl imports.

PR #166, opened 2026-02-09
Fix data migration not exporting user-created profiles

skialpine:
## Summary

- Data migration (Settings → Data → Import from another device) was only exporting default/built-in profiles, not user-created ones like D-Flow profiles
- Root cause: commit 55c6070 ("Fix profile export to use correct storage paths") incorrectly assumed profiles are stored flat in the root directory. In reality, user-created profiles go in `user/` and downloaded profiles go in `downloaded/` subdirectories
- The original implementation from 1773928 was correct; this restores subdirectory scanning

## Changes

- **`shotserver_backup.cpp`**: All three backup handlers (`handleBackupManifest`, `handleBackupProfilesList`, `handleBackupProfileFile`) now scan root, `user/`, and `downloaded/` subdirectories
- **`shotserver.cpp`**: URL routing uses `lastIndexOf('/')` instead of `indexOf('/')` to correctly parse compound category paths like `external/user`
- Added subdirectory validation (only allows `user` and `downloaded`) for security

## Test plan

- [ ] Enable Remote Access on source device
- [ ] On target device, go to Settings → Data → Search
- [ ] Verify profile count in manifest matches actual number of profiles (including user-created ones)
- [ ] Import profiles and verify user-created profiles (e.g., D-Flow variants) are included
- [ ] Verify default/built-in profiles still import correctly

### DIFF
--- a/src/network/shotserver.cpp
+++ b/src/network/shotserver.cpp
@@ -778,8 +778,9 @@ void ShotServer::handleRequest(QTcpSocket* socket, const QByteArray& request)
     }
     else if (path.startsWith("/api/backup/profile/")) {
         // /api/backup/profile/{category}/{filename} - download individual profile
+        // Category can be compound (e.g., "external/user"), so split on LAST slash
         QString remainder = path.mid(20);  // After "/api/backup/profile/"
-        int slashIdx = remainder.indexOf('/');
+        int slashIdx = remainder.lastIndexOf('/');
         if (slashIdx > 0) {
             QString category = remainder.left(slashIdx);
             QString filename = QUrl::fromPercentEncoding(remainder.mid(slashIdx + 1).toUtf8());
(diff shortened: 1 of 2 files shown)


Kulitorum:
Thanks for the fix! Merged and applied a follow-up fix in 139f0fb:

**Bug found during review:** `handleBackupFull()` (the full `.dcbackup` archive export used by the web restore page) had the same root-only scanning bug — it was missing `user/` and `downloaded/` subdirectories. I've applied the same pattern you used in the other three handlers:

- `addProfilesFrom()` now takes a `basePath` + `subdir` parameter
- Scans root, `user/`, and `downloaded/` for both external and fallback paths
- Uses `subdir/filename` dedup key so profiles from different subdirs with the same filename are included

Everything else in the PR looked solid — the security allowlist, the `lastIndexOf` URL fix, and the lambda refactoring were all well done.

PR #203, opened 2026-02-17
Fix false temperature override after recipe editor changes

skialpine:
## Summary
- Fix `uploadRecipeProfile()` missing `setTemperatureOverride()` sync — after changing temperature in the recipe editor, the stale override caused wrong delta applied to machine frames and the shot plan showing an incorrect arrow (e.g., "84 → 91°C")
- Fix `TemperatureItem` showing "(override)" label whenever `hasTemperatureOverride` is true, which is always true after any profile load. Now only shows when the override actually differs from the profile temperature (>0.1°C), matching `ShotPlanText`'s existing logic

## Test plan
- [ ] Change temperature in recipe editor, save profile → shot plan should show just the new temp (no arrow)
- [ ] Change temperature in recipe editor, save profile → idle screen should show "Group Temp" without "(override)"
- [ ] Set a real override via BrewDialog (different from profile temp) → "(override)" label and arrow should appear
- [ ] Clear overrides → "(override)" label and arrow should disappear

### DIFF
--- a/qml/components/layout/items/TemperatureItem.qml
+++ b/qml/components/layout/items/TemperatureItem.qml
@@ -11,6 +11,8 @@ Item {
     readonly property double effectiveTargetTemp: Settings.hasTemperatureOverride
         ? Settings.temperatureOverride
         : MainController.profileTargetTemperature
+    readonly property bool isRealOverride: Settings.hasTemperatureOverride &&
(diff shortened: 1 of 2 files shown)


Kulitorum:
Merged locally into main. Thank you!

PR #217, opened 2026-02-19
Use dedicated AI report endpoint

skialpine:
## Summary
- Switch AI advice reporting from piggybacking on `POST /v1/crash-report` to the new dedicated `POST /v1/ai-report` endpoint
- Send structured fields (provider, model, system prompt, transcript, debug log) instead of concatenating everything into `crash_log`
- Add separate `aiReportSubmitted`/`aiReportFailed` signals so AI reports don't interfere with crash report signal handlers

## Depends on
- Server-side: https://github.com/Kulitorum/decenza-shotmap/pull/1

> ⚠️ **Deployment order**: server side must be deployed first. If this client ships before the server endpoint exists, users get a clear "Reporting service unavailable" error — no silent failure, no impact on crash reporting.

## Review fixes (commit 3fe9bc4)

| Severity | Issue | Fix |
|----------|-------|-----|
| Critical | SAW fix + ASan disable + version files bundled in PR | Rebased onto main, dropped unrelated commit (SAW is in PR #218) |
| Critical | Version files would regress on merge (1.5.1 vs main's 1.5.3) | Resolved by rebase — no version file changes in PR |
| Important | `m_isAiReport` flag is fragile — tied to object state not reply | Replaced with `reply->setProperty("isAiReport", true)`, eliminated member variable |
| Important | No network timeout — `m_submitting` stuck forever on flaky networks | Added `request.setTransferTimeout(30000)` on both endpoints |
| Important | Null reply guard leaves `m_submitting` permanently stuck | Now resets state and emits `failed()` signal |
| Minor | Non-JSON server response (502 proxy page) becomes "Unknown error" | Added `QJsonParseError` check with HTTP status code in error message |
| Minor | Platform fallback `"linux"` without explanation | Added comment explaining server enum validation workaround |

## Review fixes (commit 8b88fe4)

| Severity | Issue | Fix |
|----------|-------|-----|
| Critical | `submitAiReport()` early return (already submitting) silently returns — dialog locks in "submitting" state forever | Now emits `aiReportFailed(tr("A report is already being submitted..."))` before returning |
| Important | Null-reply guard can't read `reply->property("isAiReport")` when reply is null — always emitted `failed()` instead of `aiReportFailed()` | Added `m_pendingIsAiReport` member; set in both submit methods; null-reply guard uses it to pick the correct signal |
| Minor | `submitReport()` didn't explicitly set `isAiReport=false` on reply | Added explicit `reply->setProperty("isAiReport", false)` for defensive clarity |
| Minor | `user_notes` always sent in `submitAiReport()` even when empty | Wrapped in `if (!userNotes.isEmpty())` guard, consistent with other optional fields |

## Review fixes (commit 9825bc4)

| Severity | Issue | Fix |
|----------|-------|-----|
| Important | HTTP errors showed raw Qt string ("Error transferring https://... - server replied: Not Found") | Map known status codes to user-facing messages: 404 → service unavailable, 429 → rate limited, 5xx → server error |
| Important | `success:false` JSON with no `"error"` field showed "Unknown error" | Check `obj["error"]` first, then API Gateway's `obj["message"]` field, then fall back to HTTP status code |

## Test plan
- [ ] Open AI advisor, get advice, tap "Report Bad Advice"
- [ ] Enter notes and submit — verify GitHub issue is created with structured format and Gist link
- [ ] Verify crash reports still work independently (submit a crash report, confirm it uses `submitted`/`failed` signals)
- [ ] Test error path: disconnect network, submit report — verify descriptive error appears within 30s
- [ ] Test 404 path (server not yet deployed): verify "Reporting service unavailable. Please ensure the app is up to date." message
- [ ] Test double-tap "submit" quickly — verify `aiReportFailed` fires immediately with "already submitting" message, dialog doesn't lock
- [ ] Test retry from error state

### DIFF
--- a/src/network/crashreporter.cpp
+++ b/src/network/crashreporter.cpp
@@ -172,8 +172,18 @@ void CrashReporter::onReplyFinished()
     const bool wasAiReport = reply->property("isAiReport").toBool();
 
     if (reply->error() != QNetworkReply::NoError) {
-        QString error = reply->errorString();
-        qWarning() << "CrashReporter: Failed to submit -" << error;
+        int statusCode = reply->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+        QString error;
+        if (statusCode == 404) {
+            error = tr("Reporting service unavailable. Please ensure the app is up to date.");
+        } else if (statusCode == 429) {
+            error = tr("Too many reports submitted. Please try again in an hour.");
+        } else if (statusCode >= 500) {
+            error = tr("Server error (HTTP %1). Please try again later.").arg(statusCode);
+        } else {
+            error = reply->errorString();
+        }
+        qWarning() << "CrashReporter: Failed to submit - HTTP" << statusCode << reply->errorString();
         setLastError(error);
         if (wasAiReport)
             emit aiReportFailed(error);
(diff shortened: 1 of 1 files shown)


skialpine:
@Kulitorum  fixes for all review issues have been made and this is ready to merge when you are ready to merged the server side changes in the other project.

PR #230, opened 2026-02-21
Improve shot review page navigation and layout

catacgc:
## Summary

- **Return-to-page navigation**: When on PostShotReviewPage and performing steam/flush/water operations, the app now returns to the shot review page after the operation completes (instead of going to IdlePage)
- **Layout improvement**: Moved Rating and Notes fields to appear directly under the shot graph for quicker access to the most commonly edited fields

### Navigation
- Added `returnToPageName` and `returnToShotId` properties to track where to return after operations
- Modified `completionTimer` and `goToIdle()` to check for saved return page
- Added `saveReturnToPage()` function called before navigating to operation pages
- Handles chained operations (e.g., shot review → steam → flush → back to shot review)

### PostShotReviewPage Layout
New order:
1. Shot graph
2. Graph legend
3. **Rating** (moved up)
4. **Notes** (moved up)
5. Bean info, grinder, measurements, etc.

image

## Test plan

- [ ] Complete a shot, verify PostShotReviewPage appears
- [ ] From PostShotReviewPage, start steam → stop → verify returns to PostShotReviewPage
- [ ] From PostShotReviewPage, start flush → let it complete → verify returns to PostShotReviewPage
- [ ] From PostShotReviewPage, start hot water → stop → verify returns to PostShotReviewPage
- [ ] Verify rating and notes appear directly under the shot graph
- [ ] From IdlePage, start steam → stop → verify goes to IdlePage (not PostShotReviewPage)

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -22,6 +22,11 @@ ApplicationWindow {
     // Flag to open BrewDialog when IdlePage becomes active (set by AutoFavoritesPage)
     property bool pendingBrewDialog: false
 
+    // Track page to return to after steam/flush/water operations complete
+    // This allows returning to postShotReviewPage instead of always going to idlePage
+    property string returnToPageName: ""
+    property int returnToShotId: 0
+
     // True while the first-run restore dialog is active (prevents SettingsDataTab from also handling restore signals)
 
     // Global accessibility: find closest Text within radius of tap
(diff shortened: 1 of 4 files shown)


skialpine (approved):
Looks good — clean implementation of return-to-page navigation and sensible layout reorder.

**Minor notes (not blocking):**

1. **`getGroupTemperature()` debug logging** — logs on every call including the no-override path, which could be noisy during phase transitions. Consider removing the no-override log line if it becomes too chatty.

2. **Stale return tracking edge case** — if a user navigates away from an operation page mid-operation (e.g., to Settings), `returnToPageName` lingers until the next `goToIdle()` call. No user-visible impact since `showCompletion` won't fire anyway.

3. **Chained operations handled well** — `saveReturnToPage("steamPage")` correctly preserves existing return tracking, so postShotReview → steam → flush → postShotReview works as expected.

PR #257, opened 2026-02-23
Fix data migration SSL handshake and TOTP auth

skialpine:
## Summary
- Fix SSL handshake failure when connecting to a secured source device (self-signed TLS certificates are now accepted for LAN migration)
- Add TOTP authentication flow: when the source device has security enabled, the importing device prompts for the 6-digit authenticator code
- Session tokens are cached in QSettings so re-authentication isn't needed for 90 days

## Follow-up fixes
- Fix build error: renamed shadowed `QUrl serverUrl` variable to `parsedUrl`
- Fix auth error text immediately cleared by `onTextChanged` cascade (swap set order)
- Add network error check in `onAuthReply()` before checking status code
- Clear stale session token from QSettings (not just memory) on 401
- Replace `Component.onCompleted` with `onVisibleChanged` for TOTP field focus
- Remove distracting "000000" placeholder from TOTP input field

## Test plan
- [ ] Enable security on source device (Settings > Data > Enable Security + set up authenticator)
- [ ] On target device, search for and select the source device
- [ ] Verify TOTP code prompt appears instead of "SSL handshake failed"
- [ ] Enter correct code — should authenticate and show manifest
- [ ] Import data — all requests should use the cached session
- [ ] Restart the app and reconnect to same server — should not prompt for code again (cached)
- [ ] Enter wrong code — should show error message (not blank) and allow retry
- [ ] Disconnect network during auth — should show connection error, not "Authentication failed"

### DIFF
--- a/qml/pages/settings/SettingsDataTab.qml
+++ b/qml/pages/settings/SettingsDataTab.qml
@@ -788,7 +788,7 @@ KeyboardAwareContainer {
                             font.pixelSize: Theme.scaled(16)
                             font.bold: true
                             font.letterSpacing: Theme.scaled(3)
-                            placeholderText: "000000"
+                            placeholderText: ""
                             accessibleName: TranslationManager.translate("settings.data.migrationTotpFieldAccessible",
                                 "Six digit authenticator code for remote device")
 


skialpine:
### Code review

Found 1 issue:

1. `onAuthenticationFailed` sets error text then immediately clears it. Line 1020 sets `migrationAuthError.text = error`, but line 1021 sets `migrationTotpField.text = ""`, which triggers `onTextChanged` (line 795) that clears `migrationAuthError.text = ""`. The user never sees the error message.

https://github.com/Kulitorum/Decenza/blob/ff6e3f42047930cb4af5d54c2e3333fdae804987/qml/pages/settings/SettingsDataTab.qml#L1018-L1022

Triggered by:

https://github.com/Kulitorum/Decenza/blob/ff6e3f42047930cb4af5d54c2e3333fdae804987/qml/pages/settings/SettingsDataTab.qml#L794-L796

Fix: swap the order — clear the text field before setting the error, or guard `onTextChanged` so it only clears errors on user input (e.g., check `migrationTotpField.activeFocus`).

PR #268, opened 2026-02-25
Add tappable legend to live espresso graph

skialpine:
## Summary
- Add interactive `GraphLegend` to the EspressoPage so users can tap to toggle graph lines (pressure, flow, temp, weight, weight flow, resistance) on/off during live extraction
- Add `show*` visibility properties to `ShotGraph` matching `HistoryShotGraph`, binding all renderers and goal lines to persisted settings
- Remove the old static read-only legend overlay from `ShotGraph` (replaced by the tappable legend)

## Test plan
- [ ] Start an espresso shot (or use simulator) and verify the tappable legend appears above the info bar
- [ ] Tap each legend item and confirm the corresponding line toggles on/off
- [ ] Verify toggling persists across shots and app restarts
- [ ] Verify history graph views (ShotDetailPage, PostShotReviewPage) still work correctly with shared settings
- [ ] Check that goal lines (dashed) also hide when their corresponding line type is toggled off

### DIFF
--- a/qml/components/ShotGraph.qml
+++ b/qml/components/ShotGraph.qml
@@ -10,6 +10,14 @@ ChartView {
     plotAreaColor: Qt.darker(Theme.surfaceColor, 1.3)
     legend.visible: false
 
+    // Persisted visibility toggles (tappable legend)
+    property bool showPressure: Settings.value("graph/showPressure", true)
+    property bool showFlow: Settings.value("graph/showFlow", true)
+    property bool showTemperature: Settings.value("graph/showTemperature", true)
+    property bool showWeight: Settings.value("graph/showWeight", true)
+    property bool showWeightFlow: Settings.value("graph/showWeightFlow", true)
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

Found 1 issue:

1. `GraphLegend` is a `Flow` component with `Layout.fillWidth: true`, but in `EspressoPage.qml` it is placed as a direct child of `Page` (not inside a `ColumnLayout`/`RowLayout`). `Layout.fillWidth` is silently ignored outside layout managers. Without an explicit `width`, the `Flow` sizes to its `implicitWidth` (sum of all children in one row) and will not wrap on narrow screens — legend items will extend off-screen and be untappable on phones. Fix: add `width: parent.width` to the `GraphLegend` instance in `EspressoPage.qml`.

https://github.com/Kulitorum/Decenza/blob/331c479f06453790b041c4132a99097b7a850cbb/qml/pages/EspressoPage.qml#L324-L331

PR #298, opened 2026-02-28
Add SAW latency tracing and thread priority boost

skialpine:
## Summary

- **SAW latency instrumentation**: Monotonic timestamps trace the full stop-at-weight pipeline (worker trigger -> main thread dispatch -> BLE write -> BLE ack) with `[SAW-Latency]` log tags
- **WeightProcessor thread priority**: Boosted to `HighPriority` to reduce worker->main dispatch latency
- **Install intent dedup**: Event-based flag (`m_installIntentPending`) prevents repeated APK install intents on Android, cleared when app returns to foreground via `applicationStateChanged`

## Test plan

- [ ] Run an espresso with SAW enabled and check logcat for `[SAW-Latency]` lines showing dispatch/bleAck/total breakdown
- [ ] Verify SAW still stops correctly at target weight
- [ ] Trigger an update on Android and confirm no duplicate install prompts

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -223,7 +223,6 @@ int main(int argc, char *argv[])
     weightThread.setObjectName(QStringLiteral("WeightProcessor"));
     weightProcessor.moveToThread(&weightThread);
     weightThread.start(QThread::HighPriority);
-    qDebug() << "[SAW] WeightProcessor thread priority set to HighPriority";
 
     // Scale → WeightProcessor (main → worker, auto QueuedConnection)
     // Initially connected to FlowScale; reconnected when physical scale is found


skialpine:
### Code review

Found 2 issues:

1. **`m_lastSawWriteMs` becomes stale after BLE write retry, producing inflated latency numbers.** `m_lastSawWriteMs` is set once in `stopOperationUrgent()` when the first BLE write is issued. If that write times out and the retry path re-executes `m_lastCommand()` (line 122), the actual retried write goes out later but `m_lastSawWriteMs` is not updated. When the BLE ack finally arrives, `bleAckMs = ackMs - m_lastSawWriteMs` measures from the first attempt rather than the retry, inflating the reported latency. For example, a 5-second timeout + 50ms successful retry would report `bleAck=5050ms` instead of `50ms`.

https://github.com/Kulitorum/Decenza/blob/1f0125cb61ff21c385f93c4084bef4843f383447/src/ble/de1device.cpp#L1056-L1061

2. **Install intent cooldown blocks the "already downloaded" fast-path, regressing the fix from commit `6627396`.** The 5-second cooldown guard (lines 291-297) executes before the cached-APK fast-path (lines 306-318). That fast-path was explicitly added to handle the case where Android's "Install Unknown Apps" permission flow consumed the first install intent -- the user grants permission, the OS returns to Decenza, and the user taps install again. If that re-tap happens within 5 seconds (common on slow devices like the Decent tablet), the cooldown silently drops it with only a `qDebug` message and no user-visible feedback. Consider moving the cooldown check after the fast-path, or using an event-based guard (e.g., a boolean cleared when the install intent result callback fires) instead of a timestamp window.

https://github.com/Kulitorum/Decenza/blob/1f0125cb61ff21c385f93c4084bef4843f383447/src/core/updatechecker.cpp#L290-L318

skialpine (commented):
### Code review

Found 1 issue:

1. The install intent cooldown uses a timestamp-based guard (`m_lastInstallIntentLaunchMs` + 5-second window) to suppress duplicate taps. CLAUDE.md says "Never use timers as guards/workarounds. Timers are fragile heuristics that break on slow devices and hide the real problem. Use event-based flags and conditions instead." Consider replacing with a boolean flag (e.g., `m_installIntentInFlight`) set when the intent launches and cleared on a suitable event, rather than relying on elapsed time.

https://github.com/Kulitorum/Decenza/blob/ee9c228feaded9912db26a8b4ebcfabf7bacec28/src/core/updatechecker.cpp#L311-L319

PR #304, opened 2026-02-28
Defer scale dialogs until Ready, add disable option

### LINKED ISSUE #295: [FR] Option to turn-off info-box about scale not connected
Please implement a way to prevent the info box stating there's no scale connected from showing every time the scale disconnects or when the tablet awakens without the scale connected. This is expected behavior with most scales, as they either turn off after some time, or are turned off when the tablet wakes-up.

There should be either a "do not show again" checkbox on the info box, or a setting under "preferences" to not show that box. The latter is cleaner, as it makes it easier for anyone who wants to have it displayed again to do so.

skialpine:
## Summary

- Replace timer-based scale dialog suppression with event-driven `scaleDialogDeferred` flag that holds dialogs until the machine reaches Idle/Ready or an active phase
- Add `Settings.showScaleDialogs` toggle (Settings → Bluetooth → "Scale connection alerts") to disable modal scale dialogs entirely — the non-modal `ScaleWeightItem` indicator still shows status
- Queued scale popups are automatically discarded when the scale reconnects during warmup
- Replace `pendingPopupTimer` (500ms guard) with `Qt.callLater` — page transitions are instant so no delay needed

Resolves #295

## Test plan

- [ ] Wake from screensaver with scale off → no modal dialog during Idle/Heating, "Scale not found" shown in ScaleWeightItem
- [ ] Turn on scale during Heating → scale connects, no dialog ever shown
- [ ] Don't turn on scale → dialog appears when machine reaches Idle or Ready
- [ ] Already-warm machine wakes directly to Idle (skipping Heating/Ready) → deferral clears, dialog shows if scale is still disconnected
- [ ] Disable "Scale connection alerts" in Settings → Bluetooth → no modal dialog at any phase (ScaleWeightItem still shows warning)
- [ ] Scale disconnects during active espresso → dialog still shows immediately (not deferred, not affected by setting since it's post-Ready)
- [ ] `noScaleAbortDialog` still fires when starting a shot with weight-dependent profile and no scale
- [ ] Queued non-scale popups (update, bleError, refill) still show promptly after wake

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -357,11 +357,8 @@ ApplicationWindow {
         })
     }
 
-    Timer {
-        id: pendingPopupTimer
-        interval: 500  // Brief delay after wake to let UI settle
-        onTriggered: root.showNextPendingPopup()
-    }
+    // No timer needed — page transitions are instant (empty Transition{}),
+    // so Qt.callLater suffices to let the event loop finish the replace().
 
     // Periodic timer to keep steam heater on when idle
     // The DE1 may have an internal timeout that reduces steam heater power after some idle time.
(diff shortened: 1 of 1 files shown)


skialpine:
Fixed the Idle phase issue in 25e08e2 — `MachineStateType.Phase.Idle` is now in the deferral-clearing condition. Also replaced the 500ms `pendingPopupTimer` guard with `Qt.callLater` in f8b32f9.

PR #309, opened 2026-02-28
Add Show button to Auto-Favorites page

skialpine:
## Summary
- Adds a **Show** button (between Info and Load) on each Auto-Favorites card that opens Shot History filtered to exactly that group's shots
- Filter fields are derived from the current `groupBy` setting (bean, profile, grinder combinations)
- Search field is populated with the filter terms so users can **edit** or **save** the search
- A filter banner indicates exact-match mode is active; editing the search text switches to normal FTS search, and "Clear" resets both

## Test plan
- [ ] Open Favorites, pick a group showing "N shots"
- [ ] Tap "Show" — Shot History opens showing exactly N shots
- [ ] Verify search field contains the filter terms (e.g., bean brand, profile name)
- [ ] Tap "Save" to save the search — verify it appears in saved searches
- [ ] Edit the search text — filter banner disappears, switches to FTS search
- [ ] Tap "Clear" on the filter banner — both filter and search field reset, full history shown
- [ ] Test with different groupBy settings (bean only, profile only, bean+profile, all)

### DIFF
--- a/qml/pages/ShotHistoryPage.qml
+++ b/qml/pages/ShotHistoryPage.qml
@@ -184,7 +184,7 @@ Page {
             // Strip any remaining keyword tokens (e.g. duplicate dose:18 dose:20)
             searchText = searchText.replace(/\b(rating|dose|yield|time|tds|ey):\d+(?:\.\d+)?(?:-\d+(?:\.\d+)?|\+)?/g, "")
 
-            // Pass remaining text (after stripping keywords) as FTS search
+            // Pass remaining text as FTS search (skipped when exact initialFilter is active)
             searchText = searchText.trim().replace(/\s+/g, " ")
             if (searchText.length > 0 && !initialFilter) {
                 filter.searchText = searchText


skialpine:
### Code review

Found 3 issues:

1. `buildFilter()` applies both FTS search text and exact-match `initialFilter` fields simultaneously on first load. When the "Show" button navigates to ShotHistoryPage, the search field is populated with filter terms (e.g., "Ethiopia Decaf MyProfile") as a user convenience, but `buildFilter()` passes this text as `filter.searchText` for FTS *and* separately merges `initialFilter.beanBrand`, `initialFilter.profileName`, etc. as exact-match SQL conditions. This over-constrains the query — FTS tokenization may not align with exact-match results, potentially returning fewer shots than intended. The code comment says "Populate search field with filter terms so user can edit/save", implying the text is a convenience copy, not an active FTS filter. Fix: skip `filter.searchText` when `initialFilter` is active.

https://github.com/Kulitorum/Decenza/blob/5fbce697cad0ef0872a9ee77cd20b2a969f5437e/qml/pages/ShotHistoryPage.qml#L188-L201

2. `_populatingSearch` flag guards clearing `initialFilter` in `onTextChanged` but does not guard `searchTimer.restart()`. When `StackView.onActivated` programmatically sets `searchField.text`, `onTextChanged` fires synchronously. The `initialFilter = null` is correctly skipped, but `searchTimer.restart()` still executes, causing a second `loadShots()` call 300ms after the synchronous one at line 75. This produces two back-to-back database queries on every activation with a filter. Fix: also guard `searchTimer.restart()` with `if (!_populatingSearch)`.

https://github.com/Kulitorum/Decenza/blob/5fbce697cad0ef0872a9ee77cd20b2a969f5437e/qml/pages/ShotHistoryPage.qml#L340-L345

3. `showButton` sets `Accessible.ignored: true` on the parent Rectangle while putting accessibility attributes on the child `AccessibleMouseArea`. CLAUDE.md explicitly lists this as an anti-pattern: "Never set `Accessible.ignored: true` on a parent and put `Accessible.role` on a child occupying the same bounds. TalkBack can't reliably route activation to the child." The recommended alternative is `AccessibleButton`. Note: the existing `infoButton` and `loadButton` use the same pattern, so this is a pre-existing issue being replicated — but CLAUDE.md guidance applies to new code. (CLAUDE.md says "Use instead `AccessibleButton`")

https://github.com/Kulitorum/Decenza/blob/5fbce697cad0ef0872a9ee77cd20b2a969f5437e/qml/pages/AutoFavoritesPage.qml#L303-L349

skialpine:
### Code review

Found 3 issues:

1. Filter banner Rectangle has `Accessible.ignored: true` while containing an interactive `AccessibleButton` child. This matches the anti-pattern explicitly listed in CLAUDE.md: "Never set `Accessible.ignored: true` on a parent and put `Accessible.role` on a child occupying the same bounds. TalkBack can't reliably route activation to the child." The fix is to remove `Accessible.ignored: true` from the banner Rectangle -- the child Text already has its own `Accessible.ignored: true`, and the AccessibleButton needs to remain discoverable. (CLAUDE.md says "Never set `Accessible.ignored: true` on a parent and put `Accessible.role` on a child occupying the same bounds")

https://github.com/Kulitorum/Decenza/blob/01de7b15f2cc0c5de32acdef2b11b9aa1cda250a/qml/pages/ShotHistoryPage.qml#L419-L427

2. Empty `initialFilter` object `{}` is truthy in JavaScript. When a favorites group has all empty/null fields for the current groupBy (e.g., groupBy="bean" but shots have no bean metadata), `filter` remains `{}`. This is passed as `initialFilter: {}` to ShotHistoryPage, where `if (initialFilter)` evaluates to `true`. This causes: (a) an empty "Filtered:" banner with no terms, (b) FTS search is permanently suppressed (`!initialFilter` is false), (c) the exact-match merge loop finds no fields so no filter is applied. The user sees all shots with a confusing empty banner and broken search until they manually edit the field. Fix: guard with `if (Object.keys(filter).length > 0)` before passing the filter.

https://github.com/Kulitorum/Decenza/blob/01de7b15f2cc0c5de32acdef2b11b9aa1cda250a/qml/pages/AutoFavoritesPage.qml#L338-L356

3. `showButton` uses a raw `MouseArea` instead of `AccessibleMouseArea`, which every other button in the same delegate uses (`infoButton`, `loadButton`). `AccessibleMouseArea` provides announce-first-then-activate behavior in TalkBack mode (first tap announces, second tap activates). The raw `MouseArea` with `Accessible.onPressAction: showMouseArea.clicked(null)` bypasses this, so the Show button activates immediately while adjacent buttons require two taps. This creates inconsistent TalkBack behavior within the same card.

https://github.com/Kulitorum/Decenza/blob/01de7b15f2cc0c5de32acdef2b11b9aa1cda250a/qml/pages/AutoFavoritesPage.qml#L313-L358

PR #310, opened 2026-03-01
Fix CLAUDE.md compliance: async DB/IO and timer guards

skialpine:
## Summary

- Remove 3s timer guard from MQTT auto-connect (TCP broker, not BLE — no delay needed)
- Replace BatteryManager 1s timer with `applicationStateChanged` event-based flag (matching `UpdateChecker` pattern)
- Make `saveShot()` async: extract QObject data on main thread into `ShotSaveData` POD struct, run DB inserts on background thread via `saveShotStatic()`
- Make `handleBackupRestore()` and `DataMigrationClient::onShotsReply()` use `importDatabaseStatic()` on background threads
- Make `handleBackupFull()` async: collect QObject data on main thread, build archive on background thread

## Test plan

- [ ] Pull a shot — verify it saves without UI freeze and appears in history
- [ ] Web UI Remote Access: download full backup (`handleBackupFull`)
- [ ] Web UI Remote Access: restore from backup (`handleBackupRestore`)
- [ ] Data Migration between two devices (or same device via loopback)
- [ ] On Android: Settings > Battery > test battery optimization prompt — verify signal fires when returning to app
- [ ] Verify MQTT auto-connects on startup (if broker configured)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -901,9 +901,16 @@ int main(int argc, char *argv[])
         QObject::connect(&simulatedScale, &ScaleDevice::weightChanged,
                          &mainController, &MainController::onScaleWeightChanged);
 
-        // Set SimulatedScale as the active scale for MachineState
+        // Set SimulatedScale as the active scale (matching physical scale pattern)
         machineState.setScale(&simulatedScale);
+        timingController.setScale(&simulatedScale);
         context->setContextProperty("ScaleDevice", &simulatedScale);
+
+        // Reconnect WeightProcessor from FlowScale to SimulatedScale for espresso SOW
+        QObject::disconnect(&flowScale, &ScaleDevice::weightChanged,
+                            &weightProcessor, &WeightProcessor::processWeight);
+        QObject::connect(&simulatedScale, &ScaleDevice::weightChanged,
+                         &weightProcessor, &WeightProcessor::processWeight);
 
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

Found 2 issues:

1. Missing `PRAGMA busy_timeout` before `PRAGMA wal_checkpoint(FULL)` in `handleBackupFull()` background thread. Without it, the FULL checkpoint returns immediately with `busy=1` if a concurrent `saveShotStatic()` call holds an active write transaction, leaving WAL frames uncheckpointed and producing an incomplete backup. The `saveShotStatic()` function in the same PR correctly sets `busy_timeout = 5000` before its operations -- the backup connection should do the same.

https://github.com/Kulitorum/Decenza/blob/39a5cd43f2cdddb024032d2c3af8a413dbf709b1/src/network/shotserver_backup.cpp#L468-L474

2. Stale section numbering in `handleBackupFull()` main-thread phase. Sections 2-4 were moved to the background thread, but the comments still say `// 5. AI conversations` and `// 6. Extra QSettings data`. The numbering jumps from 1 to 5 to 6, which is confusing since there are only 3 items in the main-thread phase.

https://github.com/Kulitorum/Decenza/blob/39a5cd43f2cdddb024032d2c3af8a413dbf709b1/src/network/shotserver_backup.cpp#L400-L411

PR #403, opened 2026-03-12
feat: add custom button actions for navigation and commands

### LINKED ISSUE #401: FR: custom button actions
Request: more custom button actions.

In particular, the scale button actions include tare (single tap) and brew settings (double tap). I use this on every shot, to measure dose weight and compute espresso weight (via brew ratio). But I often end up with another single tap instead of a double tap, which screws up the calculations. I tried to add another button with single tap for brew settings (or add a long tap for brew settings), but this action (brew settings) is not available. (Unless I'm missing something, a distinct possibility. ;-)

Please add more actions (especially the brew settings popup) to buttons. TIA

skialpine:
## Summary
- Adds 9 new navigate actions to custom button picker: steam, hot water, flush, bean info, espresso, community, flow calibration, profile import, and shot review (with last-shot parameter)
- Adds 5 new command actions: open brew settings, toggle charging mode, upload last shot to visualizer, connect DE1, disconnect DE1
- Exposes `idleBrewDialog` property alias on IdlePage for the brew settings command

Closes #401

## Test plan
- [ ] Open layout editor → add custom button → verify all 14 new actions appear in picker
- [ ] Test navigate actions open correct pages
- [ ] Test `brewSettings` opens BrewDialog on idle page
- [ ] Test `toggleCharging` cycles charging modes (0→1→2→0)
- [ ] Test `connectDE1` starts BLE scan
- [ ] Test `disconnectDE1` disconnects
- [ ] Test `uploadVisualizer` uploads last shot
- [ ] Test `shotReview` opens PostShotReviewPage with last shot ID
- [ ] Verify existing actions still work


skialpine:
### Code review

Found 4 issues:

1. **`arguments.callee` signal leak in `uploadVisualizer`** — The anonymous handler connected to `shotReady` uses `arguments.callee` to disconnect itself. `arguments.callee` is deprecated and forbidden in strict mode JS. If the disconnect fails, every button press accumulates a permanent signal handler that fires on all future `shotReady` emissions, causing duplicate uploads. Additionally, when `shotId !== lastId`, the handler returns without disconnecting, leaking it permanently. Fix: assign the closure to a named variable before connecting, then reference that variable in the disconnect call.

https://github.com/Kulitorum/Decenza/blob/0b27a44bbd51367ce6c8ae8696340096695bc2de/qml/components/layout/items/CustomItem.qml#L306-L315

2. **`connectDE1` is functionally identical to `scanDE1`** — Both `command:connectDE1` (line 317) and `command:scanDE1` (line 286) call `BLEManager.startScan()`. Users selecting "Connect DE1" would expect it to reconnect to a known device, not re-scan. Either remove the duplicate or implement a distinct connect-to-last-device behavior.

https://github.com/Kulitorum/Decenza/blob/0b27a44bbd51367ce6c8ae8696340096695bc2de/qml/components/layout/items/CustomItem.qml#L316-L320

3. **Operation page navigations use `push()` instead of `replace()`** — `navigate:espresso` (new in this PR) pushes EspressoPage onto the stack. However, `main.qml`'s `onPhaseChanged` handler uses `pageStack.replace(null, ...)` for operation pages. Using `push` creates a broken stack state where pressing Back during a shot navigates away from the operation page. Pre-existing steam/hotwater/flush entries have the same issue, but adding `espresso` is highest risk since users could press Back mid-extraction.

https://github.com/Kulitorum/Decenza/blob/0b27a44bbd51367ce6c8ae8696340096695bc2de/qml/components/layout/items/CustomItem.qml#L236-L249

4. **Operation page navigate actions should not have `"all"` context** — `navigate:espresso`, `navigate:steam`, `navigate:hotwater`, `navigate:flush` are set to `contexts: ["idle", "all"]`, meaning they can be triggered during active operations. Pushing SteamPage while already steaming, or EspressoPage during extraction, creates duplicate stack entries. These should be `contexts: ["idle"]` only, matching the existing `command:startEspresso`/`command:startSteam` pattern.

https://github.com/Kulitorum/Decenza/blob/0b27a44bbd51367ce6c8ae8696340096695bc2de/qml/components/layout/CustomEditorPopup.qml#L1436-L1440

PR #433, opened 2026-03-14
fix: increase scale-only BLE drain timeout from 500ms to 1500ms

### LINKED ISSUE #429: BLE queue drain timeout too short on app exit
## Problem
When the app exits, the scale sleep command may not be delivered because the 500ms BLE queue drain timeout expires first. The Skale may stay awake after app close.

## Log evidence
```
WARN BLE queue drain timed out after 500 ms — sleep command may not have been delivered.
```
Observed in 2 out of ~6 sessions.

## Suggestion
Increase the 500ms timeout or make the scale sleep command fire-and-forget so app exit isn't blocked waiting for confirmation.

skialpine:
## Summary
- Increases the BLE queue drain timeout from 500ms to 1500ms when only a scale is connected (no DE1)
- Scale writes have no completion signal (`queueDrained` only exists on `DE1Transport`), so the app waits the full timeout — 500ms was too short for BLE writes that can take 1-2 connection intervals
- When DE1 is connected, the existing 2000ms timeout with `queueDrained` signal feedback is unchanged

Fixes #429

## Test plan
- [ ] Connect only a Skale (no DE1), close the app, verify scale goes to sleep
- [ ] Connect DE1 + scale, close the app, verify both go to sleep
- [ ] Verify app exit doesn't feel noticeably slower

### DIFF
--- a/src/ble/scaledevice.h
+++ b/src/ble/scaledevice.h
@@ -39,7 +39,7 @@ public slots:
     virtual void startTimer() {}
     virtual void stopTimer() {}
     virtual void resetTimer() {}
-    virtual void sleep() {}  // Put scale to sleep (battery power saving - full power off)
+    virtual void sleep() { emit sleepCompleted(); }  // Put scale to sleep (battery power saving - full power off)
     virtual void wake() {}   // Wake scale from sleep (enable LCD)
     virtual void disableLcd() {}  // Turn off LCD but keep scale powered (for screensaver)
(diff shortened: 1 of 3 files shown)


skialpine (commented):
### Code review

Verified that the three issues from the previous review are properly addressed:
- Heartbeat race: `stopHeartbeat()` now called before one-shot connection
- Sync emit ordering: `sleepCompleted` connection established before `sleep()` call
- Base class gap: default `sleep()` now emits `sleepCompleted()`

PR #437, opened 2026-03-14
Enable HTTPS web server support on iOS

skialpine:
## Summary

Enable HTTPS web server support on iOS using Apple's native Security.framework for certificate generation. The TLS transport already used Apple's SecureTransport via Qt's QSslServer — this PR adds the missing piece: self-signed certificate generation, without requiring OpenSSL on iOS.

- Generate RSA-2048 keys and self-signed X509v3 certificates with SANs using `SecKeyCreateRandomKey` and `SecKeyCreateSignature` (Security.framework)
- Manual ASN.1/DER encoding for X509v3 with RFC 5280-compliant date handling (UTCTime through 2049, GeneralizedTime for 2050+)
- No OpenSSL dependency on iOS — avoids export compliance concerns and reduces build time
- Desktop/Android continue using OpenSSL (EC P-256) unchanged

### Other changes
- Make iOS bundle ID and development team configurable via CMake variables (`-DIOS_BUNDLE_ID=...`, `-DIOS_DEVELOPMENT_TEAM=...`)
- Use `$(PRODUCT_BUNDLE_IDENTIFIER)` in Info.plist instead of hardcoding
- Set `PAHO_WITH_SSL OFF` on iOS (no OpenSSL = no SHA1 symbol conflict)

Previously, enabling "Enable Security" in Settings > Data on iOS would fail. Now iOS gets full HTTPS with self-signed certificates and TOTP authentication, identical to desktop and Android.

## Test plan
- [ ] Enable "Enable Security" toggle in Settings > Data on iOS — server starts with HTTPS
- [ ] Access `https://<device-ip>:8888` from a browser — cert is accepted (after trust)
- [ ] TOTP authenticator setup works on iOS
- [ ] Desktop/macOS build still works (OpenSSL path unchanged)
- [ ] Verify no OpenSSL symbols in iOS binary (`nm` or `otool`)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/network/shotserver.cpp
+++ b/src/network/shotserver.cpp
@@ -2115,7 +2115,6 @@ static QByteArray derTag(unsigned char tag, const QByteArray& content)
 static QByteArray derSequence(const QByteArray& content) { return derTag(0x30, content); }
 static QByteArray derSet(const QByteArray& content) { return derTag(0x31, content); }
 static QByteArray derOctetString(const QByteArray& content) { return derTag(0x04, content); }
-static QByteArray derBool(bool val) { QByteArray v; v.append(static_cast<char>(val ? 0xFF : 0x00)); return derTag(0x01, v); }
 
 static QByteArray derInteger(const QByteArray& val)
 {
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 2 issues:

1. PEM base64 output not line-wrapped at 64 characters. `QByteArray::toBase64(QByteArray::Base64Encoding)` produces a single unbroken line (~1600 chars for RSA-2048 private key). PEM format (RFC 7468) requires 64-char line wrapping. On iOS, Qt uses the SecureTransport backend — if it strictly validates PEM format, `QSslCertificate`/`QSslKey` will return null on reload, causing cert regeneration on every launch. The OpenSSL codepath does not have this problem because `PEM_write_X509()`/`PEM_write_PrivateKey()` wrap automatically. Fix: insert line breaks every 64 chars after `toBase64()`.

https://github.com/skialpine/Decenza/blob/605535946ac565c98065eeb709ce4b61ccae05af/src/network/shotserver.cpp#L2303-L2312

2. `derUtcTime` always uses ASN.1 UTCTime (tag `0x17`, 2-digit year). RFC 5280 section 4.1.2.5 requires GeneralizedTime (tag `0x18`, 4-digit year) for dates from 2050 onward. Since `notAfter = notBefore + 10 years`, any certificate generated after 2039-12-31 will have a malformed `notAfter` date that strict TLS validators will reject. Fix: use GeneralizedTime when year >= 2050.

https://github.com/skialpine/Decenza/blob/605535946ac565c98065eeb709ce4b61ccae05af/src/network/shotserver.cpp#L2143-L2150

Issue #395, opened 2026-03-11
[FR] Management of multiple scales


jtr00:
For users with multiple scales, I would appreciate a refinement. 

I use several scales that are visible via BT. It would be great to have a direct option in the settings to

1. prioritise the connection when several saved scales are available and
 
2. select the scale depending on the situation.

Currently, we only have the option of a generic scan.

PR #448, opened 2026-03-15
feat: multi-scale management (#440, #395)

### LINKED ISSUE #440: Phantom Bookoo scale causes repeated connect/disconnect loop
## Problem

When the Acaia Lunar scale disconnects after ~10 minutes of idle, the app discovers a nearby Bookoo scale and automatically switches to it. The Bookoo connects but never sends weight data and disconnects with "Unknown error" after ~5 minutes. Reconnect retries (3 attempts) all fail. This cycle repeats every time the Acaia disconnects.

Timeline (repeats 3x in a single session):
1. Acaia disconnects: "The connection has timed out unexpectedly."
2. App switches: "Scale type changed from 'acaia' to 'Bookoo'"
3. Bookoo connects but disconnects: "Unknown error."
4. 3 reconnect retries all timeout
5. Scale stays disconnected until manually reconnected

## Impact

Users lose scale connectivity during idle periods. The app chases a phantom/unreliable Bookoo device instead of reconnecting to the known-good Acaia.

## Possible Fix

- Prefer reconnecting to the previously-connected scale before switching to a newly-discovered one
- Track scale reliability (e.g., if a scale disconnects with error N times without ever sending data, deprioritize it)

skialpine:
## Summary
- **Fixes #440**: During auto-reconnect, only connect to the primary scale — nearby non-primary scales are ignored, preventing the connect/disconnect loop
- **Addresses #395**: Users can now manage multiple known scales from Settings > Connections — tap to set primary, forget individual scales
- One-time migration from single saved scale to known scales list on startup

## Test plan
- [ ] Connect to Scale A → appears in Known Scales as primary (★)
- [ ] Scan and connect Scale B → both appear, B is now primary
- [ ] Tap Scale A in list → A becomes primary, B demoted
- [ ] Disconnect Scale A → auto-reconnect targets A only, ignores nearby Scale B
- [ ] Forget Scale B → removed from list
- [ ] Forget primary while connected → disconnects, falls back to FlowScale
- [ ] Fresh install with old single-scale settings → migrates correctly as primary

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -1077,7 +1077,7 @@ ApplicationWindow {
         function onFlowScaleFallback() {
             // Only show "No Scale Found" if user has a saved scale.
             // Users without a saved scale expect FlowScale — no need to nag them.
-            if (!Settings.primaryScaleAddress !== "") return
+            if (Settings.primaryScaleAddress === "") return
             if (!Settings.showScaleDialogs) return
             // Don't nag if a USB scale is connected — it satisfies the requirement (not available on iOS)
             if (Qt.platform.os !== "ios" && UsbScaleManager.scaleConnected) return
--- a/qml/pages/settings/SettingsConnectionsTab.qml
+++ b/qml/pages/settings/SettingsConnectionsTab.qml
@@ -774,7 +774,6 @@ Item {
                             model: Settings.knownScales
 
                             Rectangle {
-                                id: scaleDelegate
                                 Layout.fillWidth: true
                                 height: knownScaleRow.implicitHeight + Theme.scaled(12)
                                 radius: Theme.scaled(6)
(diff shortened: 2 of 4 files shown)


skialpine:
### Code review

Found 4 issues:

1. **BLEManager auto-reconnect filter is dead code — core fix for #440 does not work** (logic bug)

The new filter at lines 278-285 checks `if (!m_scanningForScales && ...)`, but this condition is unreachable. The pre-existing guard at line 252 (`if (!m_scanningForScales) { return; }`) already exits the function when the flag is false. Furthermore, `tryDirectConnectToScale()` sets `m_scanningForScales = true` before scanning, so during auto-reconnect the flag is always true — making the new filter condition always false. Non-primary scales will still be emitted via `scaleDiscovered`, and the #440 bug remains unfixed. A separate flag (e.g. `m_userInitiatedScan`) is needed to distinguish user scans from auto-reconnect.

https://github.com/Kulitorum/Decenza/blob/4f0bf201b74efe037d0c50a49d1c3d34ce114d21/src/ble/blemanager.cpp#L251-L290

2. **Case-sensitive address comparison for `isPrimary` flag** (bug)

`knownScales()` uses `==` for the `isPrimary` comparison while all other address comparisons in the same file use `Qt::CaseInsensitive`. On Android, BLE MAC addresses can differ in case, causing the primary scale to not be highlighted in the UI.

https://github.com/Kulitorum/Decenza/blob/4f0bf201b74efe037d0c50a49d1c3d34ce114d21/src/core/settings.cpp#L286-L290

3. **Unicode star symbols used as icons** (CLAUDE.md says "Never use Unicode symbols as icons in text... Use SVG icons from `qrc:/icons/` with `Image` instead. Safe Unicode characters (deg, middot, mdash, rarr, times) that are in standard fonts are OK.")

The filled/outline stars (U+2605, U+2606) are not in the safe list and may render as tofu squares on devices without the right font glyphs.

https://github.com/Kulitorum/Decenza/blob/4f0bf201b74efe037d0c50a49d1c3d34ce114d21/qml/pages/settings/SettingsConnectionsTab.qml#L806-L811

4. **Overlapping accessible elements — AccessibleButton inside accessible Rectangle** (CLAUDE.md says "Never position accessible buttons inside another accessible element's bounds... TalkBack will only discover one element.")

The "Forget" `AccessibleButton` is nested inside a Rectangle that has `Accessible.role: Accessible.Button`. TalkBack may only discover one of the two interactive elements. Additionally, the `MouseArea` at z:-1 lacks `Accessible.ignored: true`.

https://github.com/Kulitorum/Decenza/blob/4f0bf201b74efe037d0c50a49d1c3d34ce114d21/qml/pages/settings/SettingsConnectionsTab.qml#L776-L854

PR #467, opened 2026-03-17
fix: hide keyboard button not working in expanded text dialog

skialpine:
## Summary
- The global hide keyboard button (top-right, main.qml) is untappable when a modal dialog is open because Qt's `Overlay` intercepts all touches behind it
- Added a local hide keyboard button inside `ExpandableTextArea`'s expanded dialog header, visible on mobile when the TextArea has focus
- Suppressed the global button when focus is inside a modal dialog to avoid showing two buttons

## Test plan
- [ ] Open PostShotReviewPage, tap notes, expand to full editor, verify hide keyboard button appears in dialog header on mobile
- [ ] Tap the button — keyboard should dismiss
- [ ] Verify the global hide keyboard button does NOT appear behind the dialog dimmer
- [ ] Check other ExpandableTextArea usages (RecipeEditor, ProfileEditor, ShotDetail, AutoFavoriteInfo) behave the same
- [ ] Verify global hide keyboard button still works normally outside of dialogs

### DIFF
--- a/qml/components/HideKeyboardButton.qml
+++ b/qml/components/HideKeyboardButton.qml
@@ -1,6 +1,5 @@
 import QtQuick
 import QtQuick.Controls
-import QtQuick.Effects
 import Decenza
 
 // Circular button to dismiss the soft keyboard inside modal dialogs.


skialpine:
### Code review

Found 6 issues:

1. **`instanceof Dialog` parent chain walk never matches** -- In QML, `item.parent` walks the visual parent chain, but Dialog content items are reparented to a `QQuickPopupItem` under the Overlay. The `Dialog` QObject is never in this chain, so the suppression logic always returns `true` and the global button remains visible (but untappable) behind modal dialogs.

https://github.com/Kulitorum/Decenza/blob/1527ebdea7da44aa319b30f1edfc24aa15c2d43e/qml/main.qml#L2879-L2884

2. **`Window.window.activeFocusItem` binding may not be reactive** -- `HideKeyboardButton` reads `activeFocusItem` through the `Window` attached property chain, unlike `main.qml` which reads it directly from the `ApplicationWindow`. The QML binding engine may not track changes through this attached property chain, causing the button to evaluate once and never update on focus changes.

https://github.com/Kulitorum/Decenza/blob/1527ebdea7da44aa319b30f1edfc24aa15c2d43e/qml/components/HideKeyboardButton.qml#L20-L26

3. **Uses raw `Rectangle+MouseArea` instead of `AccessibleMouseArea`** -- CLAUDE.md lists the raw pattern as "last resort, loses announce-first behavior." The codebase has an established `AccessibleMouseArea` component for custom-styled buttons.

https://github.com/Kulitorum/Decenza/blob/1527ebdea7da44aa319b30f1edfc24aa15c2d43e/qml/components/HideKeyboardButton.qml#L12-L50

4. **Missing "clear focus BEFORE hiding keyboard" ordering comment** -- The `onClicked` handler correctly clears focus before calling `Qt.inputMethod.hide()`, but lacks the comment explaining this constraint. The identical code in `main.qml` has it: "Must clear focus BEFORE hiding keyboard, otherwise KeyboardAwareContainer sees focus + no keyboard and reopens it."

https://github.com/Kulitorum/Decenza/blob/1527ebdea7da44aa319b30f1edfc24aa15c2d43e/qml/components/HideKeyboardButton.qml#L43-L49

5. **Header comment says "place inside a dialog header"** but `CustomEditorPopup` places it in `contentItem` anchored to top. Comment should reflect both valid placements.

https://github.com/Kulitorum/Decenza/blob/1527ebdea7da44aa319b30f1edfc24aa15c2d43e/qml/components/HideKeyboardButton.qml#L9-L11

6. **Cross-reference says "see ExpandableTextArea"** but the canonical definition is now `HideKeyboardButton.qml`. ExpandableTextArea is a consumer, not the definition.

https://github.com/Kulitorum/Decenza/blob/1527ebdea7da44aa319b30f1edfc24aa15c2d43e/qml/main.qml#L2878-L2880

---

PR #473, opened 2026-03-17
fix: add EMA smoothing to auto flow calibration

skialpine:
## Summary

- Adds exponential moving average smoothing (alpha=0.3) to per-profile flow calibration updates
- Instead of jumping straight to the per-shot ideal multiplier, blends 30% toward it each shot
- Converges in ~5-7 shots while preventing a single noisy shot from dominating

## Motivation

Recent shots showed the multiplier oscillating between 0.78–0.97 shot-to-shot:
```
Shot 780: 1.000 → 0.799  (ratio 1.30)
Shot 781: 0.799 → 0.880  (ratio 0.94)
Shot 782: 0.880 → 0.969  (ratio 0.94)
Shot 783: 0.969 → 0.824  (ratio 1.22)
```
With EMA at alpha=0.3, the same sequence converges smoothly to ~0.88 with no large swings.

## Test plan

- Pull a few shots with a profile that has auto flow calibration enabled
- Verify the multiplier converges smoothly rather than bouncing each shot
- Check logcat for `Auto flow cal: updated ... (ideal: ... EMA alpha: 0.3 ...)`

### DIFF
--- a/src/controllers/maincontroller.cpp
+++ b/src/controllers/maincontroller.cpp
@@ -2588,18 +2588,22 @@ void MainController::computeAutoFlowCalibration() {
 
     ideal = qBound(kCalibrationMin, ideal, kCalibrationMax);
 
-    // EMA smoothing: blend ideal toward current to dampen shot-to-shot oscillation.
-    // alpha=0.3 moves 30% toward the ideal each shot, converging in ~5-7 shots while
-    // preventing a single noisy shot from dominating (vs. jumping straight to ideal).
-    constexpr double kEmaAlpha = 0.3;
-    double computed = kEmaAlpha * ideal + (1.0 - kEmaAlpha) * currentEffective;
-
-    // Only update if relative change > 2%. The > 0.01 guard avoids division by zero
-    // on first use (before any calibration is set).
-    if (currentEffective > 0.01 && qAbs(computed - currentEffective) / currentEffective < kChangeThreshold) {
-        qDebug() << "Auto flow cal: computed" << computed << "≈ current" << currentEffective << "(< 2% change, skipping)";
+    // Only update if the ideal itself differs enough from current. Checking ideal (not the
+    // EMA output) preserves the original 2% deadband regardless of alpha. The > 0.01 guard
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 3 issues:

1. **EMA + 2% threshold creates a ~6.7% deadband where calibration never converges.** The threshold guard compares `computed` (already EMA-damped) against `currentEffective`. Because `computed - currentEffective = 0.3 * (ideal - currentEffective)`, the guard only triggers when `|ideal - currentEffective|` exceeds `0.02 / 0.3 = 6.7%`. Any steady drift smaller than 6.7% — e.g., ideal consistently at 5% from stored — is silently skipped on every shot, permanently. Fix: move the threshold check above the EMA line, comparing `ideal` vs `currentEffective` instead of `computed` vs `currentEffective`.

https://github.com/Kulitorum/Decenza/blob/2654897d2dfcb3299aace01770dde1af166214f5/src/controllers/maincontroller.cpp#L2594-L2601

2. **First-shot bias when no per-profile calibration exists yet.** On a brand-new profile, `effectiveFlowCalibration()` returns the global fallback (1.0). The EMA then computes `0.3 * ideal + 0.7 * 1.0`, anchoring the first stored per-profile value 70% toward the global. Pre-PR behavior was to use `ideal` directly on the first shot. Fix: gate the EMA on `m_settings->hasProfileFlowCalibration(m_baseProfileName)` — use `ideal` directly when no prior exists.

https://github.com/Kulitorum/Decenza/blob/2654897d2dfcb3299aace01770dde1af166214f5/src/controllers/maincontroller.cpp#L2578-L2596

3. **Comment direction is backwards.** Line 2591 says "blend ideal toward current" but the formula moves `computed` from `currentEffective` toward `ideal`, not the other way. Should read "blend current toward ideal" or "step 30% of the way from stored calibration toward the per-shot ideal."

https://github.com/Kulitorum/Decenza/blob/2654897d2dfcb3299aace01770dde1af166214f5/src/controllers/maincontroller.cpp#L2590-L2593

---

PR #474, opened 2026-03-17
ci: add sccache to Windows build, switch to Ninja

### LINKED ISSUE #385: Add sccache to Windows CI build
## Context

All 5 CI platforms now use compiler caching except Windows (18+ min builds). The other 4 platforms hit 88-95% ccache hit rates.

## Approach

Use `hendrikmuhs/ccache-action@v1` with `variant: sccache`, which handles install + cache storage automatically.

### Changes needed in `windows-release.yml`:

1. **Add sccache step** (after MSVC setup):
   ```yaml
   - name: Set up sccache
     uses: hendrikmuhs/ccache-action@v1
     with:
       variant: sccache
       key: windows-x64
       max-size: 500M
   ```

2. **Switch generator from Visual Studio to Ninja** (sccache works better with Ninja, and it's faster):
   ```yaml
   - name: Configure CMake
     run: |
       mkdir build && cd build
       cmake .. -G Ninja -DCMAKE_PREFIX_PATH="$env:QT_ROOT_DIR" -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache
   ```

3. **Install Ninja** — either via `choco install ninja` or use Qt's bundled copy (`find $QT_ROOT_DIR/../../Tools -name ninja.exe`)

4. **Update build step** — Ninja doesn't need `--config Release` (it's set at configure time)

### Considerations

- Switching to Ninja changes the build system — test that the installer still works (output paths may differ)
- MSVC compiler launcher support with sccache is well-established
- Expected improvement: 18min → ~3-5min on warm cache

## Current platform cache stats (for reference)

| Platform | Hit Rate | Tool |
|----------|----------|------|
| macOS | 88.4% | ccache |
| iOS | 87.8% | ccache |
| Linux x64 | 88.3% | ccache |
| Linux ARM64 | 88.2% | ccache |
| Android | 95.3% | ccache (wrapper) |
| **Windows** | **none** | — |

skialpine:
Closes #385

## Changes

- Find Qt's bundled `ninja.exe` dynamically and add to PATH (no `choco install ninja`)
- Add `hendrikmuhs/ccache-action@v1` with `variant: sccache`, 500MB cache, keyed on Qt version
- Switch CMake generator from `Visual Studio 17 2022` to `Ninja`
- Pass `-DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache`
- Remove `--config Release` from build step (set at configure time with Ninja)
- Fix `AppBuildDir` in installer config from `build\Release` → `build` (Ninja outputs flat)

## Expected impact

| | Before | After (warm cache) |
|---|---|---|
| Build time | ~18-20 min | ~5-7 min |
| Savings | — | ~65-70% |

Cold cache (first run, or after 7-day eviction) remains ~18-20 min.

## Test plan

- Trigger a manual `workflow_dispatch` build and confirm it succeeds end-to-end
- Check that the installer is produced and the sccache stats show a miss rate (expected on first run)
- Trigger a second build and confirm the sccache hit rate is high and build time is reduced

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -17,6 +17,7 @@ on:
 
 permissions:
   contents: write
+  actions: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/ios-release.yml
+++ b/.github/workflows/ios-release.yml
@@ -15,6 +15,10 @@ on:
     tags:
       - 'v*'
 
+permissions:
+  contents: write
+  actions: write
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/linux-arm64-release.yml
+++ b/.github/workflows/linux-arm64-release.yml
@@ -17,6 +17,7 @@ on:
 
 permissions:
   contents: write
+  actions: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/linux-release.yml
+++ b/.github/workflows/linux-release.yml
@@ -17,6 +17,7 @@ on:
 
 permissions:
   contents: write
+  actions: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/macos-release.yml
+++ b/.github/workflows/macos-release.yml
@@ -17,6 +17,7 @@ on:
 
 permissions:
   contents: write
+  actions: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/windows-release.yml
+++ b/.github/workflows/windows-release.yml
@@ -17,6 +17,7 @@ on:
 
 permissions:
   contents: write
+  actions: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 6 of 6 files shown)


skialpine:
### Code review

Found 2 issues:

1. **Windows build will fail: Ninja not found.** `install-qt-action@v4` does not install the Qt Tools directory (including Ninja) unless `tools: 'tools_ninja'` is explicitly specified. The "Find Qt bundled Ninja" step searches `$QT_ROOT_DIR\..\..\Tools` which will be empty, causing `exit 1` on every run. Fix: add `tools: 'tools_ninja'` to the Install Qt step.

https://github.com/Kulitorum/Decenza/blob/6027d913d30122e06afaa6a53330fdd0671ee2f7/.github/workflows/windows-release.yml#L54-L62

2. **Cache-warm dispatch silently fails: missing `actions: write` permission.** `gh workflow run` requires `actions: write` on the `GITHUB_TOKEN`. Five workflows only declare `contents: write` (actions defaults to `none`), and `ios-release.yml` has no permissions block at all. The `|| echo` guard swallows the 403 error, making the entire cache-warming feature dead code. Fix: add `actions: write` to the `permissions:` block in all 6 workflows.

https://github.com/Kulitorum/Decenza/blob/6027d913d30122e06afaa6a53330fdd0671ee2f7/.github/workflows/windows-release.yml#L18-L19

---

<details>

- [Worth a look] `$GITHUB_WORKFLOW` resolves to the display name (e.g., "Windows Build") not the filename -- works today with quoting but fragile if the name field is ever renamed; using the `.yml` filename directly would be more robust (score: 50)

</details>

PR #503, opened 2026-03-20
refactor: migrate duplicated selection dialogs to SelectionDialog

### LINKED ISSUE #499: Refactor: migrate StyledComboBox and SuggestionField to use SelectionDialog
## Summary

We now have a reusable `SelectionDialog` component (`qml/components/SelectionDialog.qml`) that provides a themed modal selection dialog with checkmark on the current item, fade gradients, scroll support, and a Cancel button.

The same dialog pattern is currently duplicated in:
- `qml/components/StyledComboBox.qml` (lines 100-313)
- `qml/components/SuggestionField.qml` (lines 358-584)

Both should be refactored to use `SelectionDialog` internally, reducing ~400 lines of duplicated code.

### SelectionDialog API
```qml
SelectionDialog {
    title: "Pick one"
    options: ["A", "B", "C"]
    currentIndex: 1
    onSelected: function(index) { /* handle */ }
}
```

### Notes
- `StyledComboBox` has extra features (`emptyItemText`, `textFunction`) that may need to be added to `SelectionDialog` or handled as overrides
- `SuggestionField` uses string matching (`_text === root.text`) instead of index-based selection
- Both use the exact same visual structure: header, ListView with Rectangle delegates, fade gradients, Cancel button

skialpine:
## Summary
- Enhance `SelectionDialog` with `emptyItemText`, `currentValue`, `emptyStateText` properties and `selected(int index, string value)` signal
- Migrate `StyledComboBox` and `SuggestionField` to use `SelectionDialog` internally, removing ~450 lines of duplicated dialog code
- Migrate three `CustomEditorPopup` dialogs (action picker, profile picker, variables) to `SelectionDialog`, removing ~400 more lines
- Fix suggestion fields not showing the current typed value in BeanInfoPage and PostShotReviewPage (Roaster, Coffee, Setting, Barista)

Closes #499

## Test plan
- [ ] Open any Settings dropdown (StyledComboBox) — verify dialog opens, selection works, checkmark shows
- [ ] Open Bean Info page — verify SuggestionField arrow button opens dialog, typing filters popup, selecting works
- [ ] Open Layout Editor → tap a text widget → tap an action field — verify action picker dialog works
- [ ] Select "Load Profile..." from action picker — verify profile picker sub-dialog opens
- [ ] In Layout Editor text widget, tap "Insert..." button — verify variable picker dialog opens and inserts token
- [ ] Type a new grinder setting in Bean Info → reopen dropdown — verify the typed value appears in the list
- [ ] Verify all dialogs with TalkBack: focus trapping, element navigation, double-tap activation

### DIFF
--- a/qml/components/SelectionDialog.qml
+++ b/qml/components/SelectionDialog.qml
@@ -15,7 +15,7 @@ import Decenza
 // For string-based matching (e.g. SuggestionField):
 //   SelectionDialog {
 //       currentValue: root.text     // highlights matching item
-//       onSelected: function(index, value) { root.setText(value) }
+//       onSelected: function(index, value) { root.selectSuggestion(value) }
 //   }
 Dialog {
     id: root
--- a/qml/components/layout/CustomEditorPopup.qml
+++ b/qml/components/layout/CustomEditorPopup.qml
@@ -1219,8 +1219,8 @@ Dialog {
         { token: "%CONNECTED%", label: TranslationManager.translate("customeditor.var.connected", "Online/Offline") },
         { token: "%CONNECTED_COLOR%", label: TranslationManager.translate("customeditor.var.statusColor", "Status Color") },
         { token: "%DEVICES%", label: TranslationManager.translate("customeditor.var.devices", "Devices") },
-        { token: "%MACHINE_CONNECTED%", label: TranslationManager.translate("customeditor.var.machineConnected", "Machine \u2713/\u2717") },
-        { token: "%SCALE_CONNECTED%", label: TranslationManager.translate("customeditor.var.scaleConnected", "Scale \u2713/\u2717") }
+        { token: "%MACHINE_CONNECTED%", label: TranslationManager.translate("customeditor.var.machineConnected", "Machine On/Off") },
+        { token: "%SCALE_CONNECTED%", label: TranslationManager.translate("customeditor.var.scaleConnected", "Scale On/Off") }
     ]
 
     // === Variable picker dialog ===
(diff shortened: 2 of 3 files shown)


skialpine:
### Code review

Found 3 issues:

1. **Misleading usage example in SelectionDialog comment** -- The "string-based matching" example calls `root.setText(value)` which does not exist. The actual method used by SuggestionField is `root.selectSuggestion(value)`. A developer copying this example would get a silent no-op.

https://github.com/Kulitorum/Decenza/blob/0c88810c8964f327387ca8708964b6d96c0e5b29/qml/components/SelectionDialog.qml#L14-L20

2. **Accessible.name missing fallback for empty scale name** -- The primary scale button's `Accessible.name` concatenates `scales[i].name + " " + scales[i].type`, but `name` can be an empty string (producing a leading space in TalkBack announcements). The display Text on line 861 correctly uses `scales[i].name || scales[i].type` but the `Accessible.name` does not apply the same fallback.

https://github.com/Kulitorum/Decenza/blob/0c88810c8964f327387ca8708964b6d96c0e5b29/qml/pages/settings/SettingsConnectionsTab.qml#L825-L836

3. **"Tap to change" hint announced when only 1 scale exists** -- `Accessible.name` always appends "Tap to change" regardless of scale count, but the `MouseArea` only opens the dialog when `knownScales.length > 1`. TalkBack users with a single scale hear "Tap to change" but nothing happens on activation. The dropdown arrow is correctly hidden with `visible: Settings.knownScales.length > 1` but the accessible hint is not conditioned similarly.

https://github.com/Kulitorum/Decenza/blob/0c88810c8964f327387ca8708964b6d96c0e5b29/qml/pages/settings/SettingsConnectionsTab.qml#L825-L888

---

<details>

- [OK to skip] Unicode `\u25BC` (BLACK DOWN-POINTING TRIANGLE) used as dropdown indicator in scale picker button -- pre-existing pattern used across StyledComboBox and SuggestionField; fixing requires codebase-wide refactor (score: 60)
- [OK to skip] Unicode `\u2713`/`\u2717` in variable picker fallback strings -- pre-existing content moved from inline list to property; not introduced by this PR (score: 68)

</details>

PR #524, opened 2026-03-20
feat: MCP profile editor tools + fix frame regeneration

### LINKED ISSUE #522: MCP dialing tools don't trigger profile frame regeneration
## Problem

The MCP `dialing_apply_change` and `settings_set` tools only write to `QSettings` — they never call `MainController::uploadRecipeProfile()`, so profile frames aren't regenerated and changes aren't uploaded to the machine.

For example, calling `dialing_apply_change` with `espressoTemperature: 92` returns `{"success": true}` but the profile's frame temperatures remain unchanged at the original values.

## Root Cause

The QML Recipe Editor calls `MainController.uploadRecipeProfile(recipeParams)` which:
1. Detects profile type (simple `settings_2a`/`2b` vs recipe vs D-Flow/A-Flow)
2. For simple profiles: calls `applyRecipeToScalarFields()` + `regenerateSimpleFrames()`
3. For recipe profiles: calls `setRecipeMode(true)` + `RecipeGenerator::generate()`
4. Uploads the regenerated frames to the machine

The MCP tools bypass all of this — they poke `QSettings` directly and nothing downstream reacts.

## What's Missing

1. **Frame regeneration** — after changing any profile parameter via MCP, `uploadRecipeProfile()` must be called to regenerate frames and upload to machine
2. **Recipe param editing** — simple profiles have params (`preinfusionTime`, `espressoPressure`, `holdTime`, `declineTime`, `limiterValue`, flow equivalents) that are invisible to MCP
3. **Editor type awareness** — `profiles_get_active` returns `editorType` but no write tool uses it to validate which parameters are applicable
4. **`profiles_set_active` fails to find valid profiles** — calling `profiles_set_active` with filenames like `blooming_espresso` (which appear in `profiles_list`) returns "Profile not found". Profile switching via MCP is broken.

## Proposed Solution

Add a new MCP tool `profiles_edit_params` that:
- Accepts RecipeParams fields (same ones the QML editor sends to `uploadRecipeProfile`)
- Calls `MainController::uploadRecipeProfile()` internally — reusing the existing code path that handles simple vs recipe vs D-Flow/A-Flow branching
- Returns the updated profile state (temperature, frames, etc.)

Update `dialing_apply_change` temperature/weight changes to also route through `uploadRecipeProfile()` instead of just setting QSettings values.

Fix `profiles_set_active` profile lookup to match the filenames returned by `profiles_list`.

## Affected Tools

- `dialing_apply_change` — `espressoTemperature` and `targetWeight` don't affect the active profile
- `settings_set` — same issue for `espressoTemperature` and `targetWeight`
- `profiles_set_active` — cannot find profiles by filename

## Relevant Code

- `MainController::uploadRecipeProfile()` — `src/controllers/maincontroller.cpp:1656`
- `MainController::applyRecipeToScalarFields()` — `src/controllers/maincontroller.cpp:1742`
- MCP dialing tool — `src/mcp/mcptools_write.cpp:237`
- MCP settings tool — `src/mcp/mcptools_write.cpp:123`
- MCP profiles tool — `src/mcp/mcptools_profiles.cpp`

skialpine:
## Summary

- Add 6 new MCP profile tools: `profiles_get_params`, `profiles_edit_params`, `profiles_save`, `profiles_delete`, plus fix `profiles_set_active` and `profiles_get_active`
- **Editor-type aware**: `profiles_get_params` returns only the fields the app's editor shows for the current profile type (pressure, flow, dflow, aflow, advanced)
- **Same code paths as the app**: `profiles_edit_params` uses `uploadRecipeProfile()` for recipe profiles and `uploadProfile()` for advanced profiles — no separate MCP-specific paths that could get out of sync
- Fix `dialing_apply_change` and `settings_set` to route temperature/weight changes through `uploadRecipeProfile()` instead of just writing QSettings (frames were silently unchanged)
- Fix `profileExists()` to check all profile sources (built-in, storage, downloaded, user) — `profiles_set_active` now finds built-in profiles
- Fix `profiles_get_active` filename field to use `baseProfileName()` (no `*` modified prefix)
- Use `currentEditorType()` for authoritative editor type detection (prevents advanced profiles like Adaptive v2 from incorrectly showing as dflow)
- Add `editorType` to `ProfileInfo` and `profiles_list` so MCP clients know which editor each profile uses

Fixes #522

## Test plan

- [x] `profiles_get_params` returns correct fields per editor type:
  - [x] pressure: preinfusion, hold, decline, per-step temps, espressoPressure, pressureEnd, limiter
  - [x] flow: preinfusion, hold, decline, per-step temps, holdFlow, flowEnd, limiter
  - [x] dflow: fill, infuse, pour phases
  - [x] aflow: fill, infuse, pour + rampTime, rampDownEnabled, flowExtractionUp, secondFillEnabled
  - [x] advanced: full profile with individual frames, notes, all metadata
- [x] `profiles_edit_params` changes temperature, marks profile modified
- [x] `profiles_save` persists to disk, clears modified flag
- [x] `profiles_delete` removes user override, reverts to built-in version
- [x] `profiles_get_active` returns clean filename (no `*` prefix)
- [x] `profiles_set_active` finds built-in profiles
- [x] `profiles_list` includes editorType for each profile
- [x] No app freeze (removed BlockingQueuedConnection deadlock)
- [x] No fire-and-forget (uploadRecipeProfile called directly on main thread)

### DIFF
--- a/docs/MCP_TEST_PLAN.md
+++ b/docs/MCP_TEST_PLAN.md
@@ -1,251 +1,431 @@
 # MCP Tool Test Plan
 
-Verifies all MCP tools work correctly as a second UI to the QML app. Each test records the tool called, expected behavior, and pass/fail result.
+Sequential runbook to verify all MCP tools work correctly as a second UI to the QML app. Each test has explicit tool calls, expected results, and cleanup steps. Tests must be run in order — later tests depend on state from earlier ones.
 
-Last run: 2026-03-20 (simulator mode, 37 tools registered)
+## Prerequisites
 
-## 1. Machine State (read)
+- App running (simulator or production)
+- MCP connected (`/mcp` in Claude Code)
+- Machine in Ready or Sleep state (not mid-operation)
+
+## State Capture (run before tests)
+
+Save these values — they will be restored at the end:
+```
+Call: profiles_get_active
+Save: ORIGINAL_PROFILE = filename, ORIGINAL_MODIFIED = modified
+
+Call: profiles_get_params
+Save: ORIGINAL_TEMP (tempStart for recipe, espresso_temperature for advanced)
+
+Call: settings_get
+Save: ORIGINAL_BRAND = dyeBeanBrand, ORIGINAL_GRIND = dyeGrinderSetting
+Save: ORIGINAL_BEAN_TYPE = dyeBeanType, ORIGINAL_ROAST = dyeRoastLevel
+
+Call: machine_get_state
+Save: ORIGINAL_PHASE = phase (if Sleep, will wake for tests then re-sleep at end)
+```
+
+If ORIGINAL_MODIFIED is true, warn the user — tests will modify the profile and the unsaved changes will be lost.
+
+## Run Log
+
+| Date | Tools | Passed | Skipped | Failed | Notes |
+|------|-------|--------|---------|--------|-------|
+| 2026-03-20 | 37 | 41 | 1 | 0 | Initial run, simulator mode |
+
+---
+
+## 1. Machine State (read-only, no state changes)
 
 ### 1.1 machine_get_state
-- **Action**: Read machine state
-- **Expected**: Returns phase, connection, water level, firmware
-- **Result**: PASS — returned Ready phase, waterLevelMl=872, firmwareVersion=SIM-1.0
+```
+Call: machine_get_state
+Expect: phase exists, connected=true, waterLevelMl > 0, firmwareVersion non-empty
+```
 
 ### 1.2 machine_get_telemetry
-- **Action**: Read live telemetry
-- **Expected**: Returns pressure, flow, temp, weight values
-- **Result**: PASS — returned pressure=0, flow=0, temperature=93, mixTemperature=92.5
+```
+Call: machine_get_telemetry
+Expect: temperature > 0, pressure/flow/weight are numbers
+```
 
 ## 2. Machine Control
 
-### 2.1 machine_wake
-- **Action**: Wake machine from sleep
-- **Expected**: Machine transitions to Idle then Ready
-- **Result**: PASS — "Wake command sent"
+### 2.1 machine_sleep
+```
+Call: machine_sleep (confirmed: true)
+Expect: success=true
+```
 
-### 2.2 machine_sleep
-- **Action**: Put machine to sleep
-- **Expected**: Machine transitions to Sleep
-- **Result**: PASS — "Sleep command sent"
+### 2.2 machine_wake
+```
+Call: machine_wake (confirmed: true)
+Expect: success=true
+Verify: machine_get_state → phase is "Ready" or "Idle" (not "Sleep")
+Note: machine must be awake for remaining tests. Will restore ORIGINAL_PHASE at end.
+```
 
-### 2.3 machine_stop
-- **Action**: Stop current operation (only valid during operation)
-- **Expected**: Returns success or appropriate error if not in operation
-- **Result**: PASS — returned error "No operation in progress" (correct)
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 5 issues:

1. **`settings_set` and `dialing_apply_change` corrupt advanced profiles by forcing them into recipe mode.** Both tools unconditionally call `uploadRecipeProfile()` when `espressoTemperature` or `targetWeight` is provided, without checking `currentEditorType()`. For advanced profiles, `getOrConvertRecipeParams()` returns default empty params, then `uploadRecipeProfile()` sets `isRecipeMode(true)` and calls `regenerateFromRecipe()`, destroying all existing custom frames. The `profiles_edit_params` tool in the same PR correctly gates on `currentEditorType() == "advanced"` and routes through `uploadProfile()` -- the same pattern should be applied here.

https://github.com/Kulitorum/Decenza/blob/51f7fae871a0ddbc82b78e7dc53f4bcff294ba61/src/mcp/mcptools_write.cpp#L151-L175

2. **New tools (`profiles_edit_params`, `profiles_save`, `profiles_delete`) bypass the chat confirmation gate.** `docs/MCP_SERVER.md` lists these tools as requiring confirmation, but `McpServer::needsChatConfirmation()` in `src/mcp/mcpserver.cpp` was not updated to include them. The `confirmed` parameter in their schemas is dead -- these settings-level operations execute immediately with no confirmation prompt.

https://github.com/Kulitorum/Decenza/blob/51f7fae871a0ddbc82b78e7dc53f4bcff294ba61/src/mcp/mcptools_profiles.cpp#L183-L271

3. **`profiles_get_active` always returns empty `editorType`.** Line 53 reads `editorType` from `getCurrentProfile()`, but that function never sets an `editorType` key. Should use `mainController->currentEditorType()` instead, matching the pattern already used in `profiles_get_params` and `profiles_edit_params`.

https://github.com/Kulitorum/Decenza/blob/51f7fae871a0ddbc82b78e7dc53f4bcff294ba61/src/mcp/mcptools_profiles.cpp#L51-L55

4. **`int` used with `QList::size()` in modified code block** (CLAUDE.md says "Use `qsizetype` (not `int`) for container sizes"). The loop at line 1336 uses `int i` with `m_allProfiles.size()` which returns `qsizetype`. The PR modifies code in this block (adding `editorType` extraction).

https://github.com/Kulitorum/Decenza/blob/51f7fae871a0ddbc82b78e7dc53f4bcff294ba61/src/controllers/maincontroller.cpp#L1334-L1340

5. **`infuseVolume` missing from `profiles_edit_params` input schema.** `profiles_get_params` returns `infuseVolume` for dflow/aflow profiles, but `profiles_edit_params` schema does not declare it as a parameter. MCP clients will see the field but not know they can set it.

https://github.com/Kulitorum/Decenza/blob/51f7fae871a0ddbc82b78e7dc53f4bcff294ba61/src/mcp/mcptools_profiles.cpp#L139-L150

---

<details>

- [OK to skip] `profileExists()` fallback disk check only covers `profilesPath()`, not QRC or subdirectories -- dead code since primary `m_availableProfiles` check covers all sources (score: 50)
- [Worth a look] `profileExists()` fallback comment says "for profiles loaded after initial scan" but only checks one directory -- misleading for future maintainers (score: 65)
- [Worth a look] `profiles_save` and `profiles_delete` perform synchronous disk I/O on main thread without a TODO acknowledging the CLAUDE.md constraint, unlike `mcptools_write.cpp` which has one (score: 75)

</details>

<sub>If this code review was useful, please react with :+1:. Otherwise, react with :-1:.</sub>

PR #561, opened 2026-03-24
fix: defer BLE profile upload to editor exit

### LINKED ISSUE #557: profile editor generates too many DE1 bluetooth updates
In 1.6.0 on Android, the profile editor pushes every change to the DE1 as the change is made. Parameters can no longer be changed using the on screen keyboard. The result is that when making changes to a parameter, you have to use the swipe gesture or +/-, and _each_ increment or decrement gets pushed via bluetooth to the DE1. Changing a single parameter by 1.0 in increments of 0.1 will generate 10 bluetooth profile updates pushed to the DE1.

This happens in the quick override dialog on the main screen as well.

(I see that there is a swipe to change the increment from 0.1 to 1, which would help, but not resolve the issue.)

Best solution would be to only push profile updates to DE1 after finishing profile edit and after closing the override dialog.

Workaround would be to re-enable keyboard entry of parameters, so a single change can be registered and pushed to DE1 for even a large change in parameter value.

skialpine:
## Summary
- Remove per-parameter BLE uploads from profile/recipe editors
- Single upload when user exits via Try, Save, or Discard
- Eliminates hundreds of unnecessary BLE writes during slider drags

## Problem
Every parameter change in the editors triggered an immediate BLE upload to the DE1 (1 header + ~12 frame writes). Dragging a temperature slider from 70 to 100 at 0.1 increments generated ~390 BLE packets. This caused BLE congestion and prevented keyboard entry of values.

## Fix
`uploadProfile()` and `uploadRecipeProfile()` now only update in-memory state. The BLE upload happens once on editor exit:

| Exit action | Upload via |
|---|---|
| Try | `uploadCurrentProfile()` added to QML `onTryClicked` |
| Save | `saveProfile()` already calls `uploadCurrentProfile()` |
| Save As | `saveProfileAs()` already calls `uploadCurrentProfile()` |
| Discard | `loadProfile(original)` already calls `uploadCurrentProfile()` |

No changes to BrewDialog (its "Update Profile" buttons save immediately, which triggers upload).

## Test plan
- [ ] Open ProfileEditorPage, drag temperature slider -- no BLE traffic during drag
- [ ] Click Done > Try -- profile uploads once to DE1
- [ ] Open RecipeEditorPage, adjust pour flow -- no BLE traffic during edit
- [ ] Click Done > Save -- profile saves and uploads once
- [ ] Click Done > Discard -- original profile reloads and uploads
- [ ] BrewDialog "Update Profile" button still uploads immediately
- [ ] Verify graph preview still updates in real-time during editing

Closes #557

### DIFF
--- a/src/mcp/mcptools_profiles.cpp
+++ b/src/mcp/mcptools_profiles.cpp
@@ -258,6 +258,7 @@ void registerProfileTools(McpToolRegistry* registry, MainController* mainControl
                     profileData[it.key()] = it.value().toVariant();
                 }
                 mainController->uploadProfile(profileData);
+                mainController->uploadCurrentProfile();  // MCP is one-shot, upload immediately
             } else {
                 // Recipe path: use uploadRecipeProfile() — same as RecipeEditorPage/SimpleProfileEditorPage
                 QVariantMap currentParams = mainController->getOrConvertRecipeParams();
@@ -266,6 +267,7 @@ void registerProfileTools(McpToolRegistry* registry, MainController* mainControl
                     currentParams[it.key()] = it.value().toVariant();
                 }
                 mainController->uploadRecipeProfile(currentParams);
+                mainController->uploadCurrentProfile();  // MCP is one-shot, upload immediately
             }
 
             result["success"] = true;
--- a/src/mcp/mcptools_write.cpp
+++ b/src/mcp/mcptools_write.cpp
@@ -415,6 +415,7 @@ void registerWriteTools(McpToolRegistry* registry, MainController* mainControlle
                     }
                     mainController->uploadRecipeProfile(currentParams);
                 }
+                mainController->uploadCurrentProfile();  // MCP is one-shot, upload immediately
 
                 // Sync QSettings so settings_get reads back the updated values.
                 // uploadRecipeProfile/uploadProfile update the profile object but


skialpine:
### Code review

Found 1 issue (already fixed in 4dd0533):

1. MCP tool handlers (`profiles_edit_params`, `settings_set`) call `uploadProfile()`/`uploadRecipeProfile()` as one-shot operations. After removing `uploadCurrentProfile()` from inside those methods, MCP calls silently fail to upload the profile to the DE1 -- the AI assistant gets a success response but the machine runs with stale parameters.

https://github.com/Kulitorum/Decenza/blob/a3550b9ddd80e9cb9836043433dba827ce79d22e/src/mcp/mcptools_profiles.cpp#L259-L269

Fixed in 4dd0533 by adding explicit `uploadCurrentProfile()` calls after the MCP tool uploadProfile/uploadRecipeProfile calls. MCP operations are one-shot (not slider drags), so immediate upload is correct.

---

<details>

- [OK to skip] Comments say "deferred to editor exit" but upload only triggers via Try/Save/Discard, not unmodified-exit -- technically accurate since no upload needed when nothing changed, but wording could be more precise (score: 50)

</details>

<sub>If this code review was useful, please react with :+1:. Otherwise, react with :-1:.</sub>

Issue #527, opened 2026-03-21
MCP write tools return success before QueuedConnection setters execute


skialpine:
## Problem

All MCP write tools (`settings_set`, screensaver/accessibility/translation setters) dispatch property changes via `QMetaObject::invokeMethod(..., Qt::QueuedConnection)` but return `{"success": true}` synchronously before any setter actually runs.

This means an MCP client receiving `success: true` has no guarantee the setting was written yet. If the app exits or crashes immediately after (possible on mobile), the change can be lost silently.

The profile update path (`mainController->uploadProfile()`) runs synchronously in contrast, making behavior inconsistent within the same tool.

## Root Cause

The MCP tool handler architecture requires synchronous `QJsonObject` return values. The `Qt::QueuedConnection` dispatch is necessary because `QSettings::setValue()` from a non-owner thread is unsafe when signals are connected to QML bindings. These two constraints conflict.

Acknowledged in the file-top TODO in `mcptools_write.cpp`:
```
// TODO: Move SQL queries to background thread per CLAUDE.md design principle.
// Current tool handler architecture (synchronous QJsonObject return) prevents this.
// Requires refactoring McpToolHandler to support async responses.
```

## Proposed Fix

Refactor `McpToolHandler` to support async responses (similar to the existing `PendingConfirmation` pattern that holds the HTTP response). The tool handler would return a deferred marker, dispatch the setter, and send the response after the setter completes on the main thread.

This would also unblock the `shots_update` SQL query moving to a background thread.

## Affected Tools

- `settings_set` — all 100+ writable fields
- Any future write tools using the same pattern

## Impact

Low in practice — settings writes are fast and the event loop processes them almost immediately. But the API contract is technically incorrect.

PR #562, opened 2026-03-24
refactor: extract ProfileManager from MainController

skialpine:
## Summary

- **Phase 1**: Mechanically extract ~40 profile methods (~1,500 lines) from MainController into a new `ProfileManager` class with lightweight dependencies (Settings, DE1Device, MachineState — no MQTT, OpenSSL, AI, network)
- **Phase 2**: Rewire MCP tool registration to take `ProfileManager*` directly, add `tst_mcptools_profiles` and `tst_mcptools_write` test suites (16 tests)
- **Phase 3**: Register `ProfileManager` as QML context property, migrate all 21 QML files, remove ~40 forwarding stubs + 20 Q_PROPERTY declarations + 5 profile signals from MainController, rewire all C++ consumers
- **Bug fix**: Fix pre-existing crash in `ShotHistoryStorage::requestDistinctCache()` — background threads dereferenced a destroyed `this` via `QMetaObject::invokeMethod`; added atomic destroyed-flag guards before all 19 `invokeMethod` calls

MainController shrinks from ~4,100 lines to ~2,600 lines. ProfileManager can now be compiled and tested in isolation without the full app dependency graph.

## Test plan

- [x] All 17 unit tests pass (0 failures)
- [x] `tst_mcptools_profiles` (10 tests) — profile CRUD, edit params, BLE upload
- [x] `tst_mcptools_write` (6 tests) — settings write triggers BLE upload
- [x] `tst_dbmigration` (17 tests) — previously crashed, now passes with threading fix
- [x] App compiles and runs on macOS via Qt Creator
- [ ] Manual: verify profile load/save/edit in QML editors
- [ ] Manual: verify MCP tools still work via Claude Code

### DIFF
--- a/docs/CLAUDE_MD/TESTING.md
+++ b/docs/CLAUDE_MD/TESTING.md
@@ -159,6 +159,31 @@ Located in `tests/tst_saw.cpp`. Tests `WeightProcessor::processWeight()`.
 | 2 | SAW blocked until frame 2 |
 | 3 | SAW blocked until frame 3 |
 
+## Known Coverage Gaps
+
+Areas where bugs have shipped undetected due to missing test coverage:
+
+### QML binding correctness (highest priority)
+
+No tests verify that QML files resolve property names and method calls to the expected C++ objects. During the ProfileManager extraction (PR #562), three QML bugs shipped past the full test suite:
+- `MainController.previousProfileName()` — method removed from MainController, QML silently returned `undefined`
+- `MainController.currentProfile` — never was a QML property (should be `currentProfileName`), always `undefined`
+- `typeof MainController` guards checking wrong object after data source moved to ProfileManager
+
+A QML binding smoke test that instantiates pages and verifies key property bindings resolve to non-undefined values would catch this class of bug.
+
+### MCP resource responses
+
+`tst_mcptools_profiles` and `tst_mcptools_write` test MCP *tools* but not MCP *resources* (`decenza://profiles/active`, `decenza://profiles/list`, etc.). A bug where the `"filename"` field returned a display title instead of a filename was undetectable.
+
+### ShotHistoryStorage async methods
+
+`tst_dbmigration` exercises schema migration and some query paths, but does not cover all async methods. `requestShotsFiltered()` had a missing destroyed-flag guard (use-after-free risk) that was only found by code review, not tests.
+
+### Flow calibration delegation
+
+`applyFlowCalibration()` existed as duplicate implementations in both MainController and ProfileManager. Since both were identical, no test could detect the divergence risk — but if either were updated independently, behavior would silently differ.
+
 ## Adding New Tests
 
 1. Create `tests/tst_yourtest.cpp` with `QTEST_GUILESS_MAIN(tst_YourTest)` and `#include "tst_yourtest.moc"`


skialpine:
### Code review

Found 10 issues:

1. **Missed migration: `MainController.previousProfileName()` will silently fail** (method removed from MainController)

`previousProfileName()` was moved to ProfileManager but this call was not migrated. In QML, calling a non-existent method returns `undefined`, so the "previousProfile" custom action does nothing.

https://github.com/Kulitorum/Decenza/blob/47f8d6b3af57a2aea5fd783c95a462e0450ad641/qml/components/layout/items/CustomItem.qml#L358-L362

2. **Missed migration: `MainController.currentProfile` is undefined -- current profile never highlighted** (pre-existing bug in modified file, CLAUDE.md says "fix pre-existing bugs and violations in the file you're modifying")

`MainController` never had a QML property called `currentProfile` (the property was `currentProfileName`). `isCurrentProfile` is always false, so the loaded profile row is never visually highlighted. Other lines in the same file correctly use `ProfileManager.currentProfileName`.

https://github.com/Kulitorum/Decenza/blob/47f8d6b3af57a2aea5fd783c95a462e0450ad641/qml/pages/ProfileSelectorPage.qml#L228-L230

3. **Stale `typeof MainController` guards check wrong object** (data reads from ProfileManager but availability guard checks MainController)

Lines 112 and 177-181 guard with `typeof MainController !== "undefined"` but read from `ProfileManager`. Works today since both are registered, but breaks if MainController is ever removed from QML context.

https://github.com/Kulitorum/Decenza/blob/47f8d6b3af57a2aea5fd783c95a462e0450ad641/qml/components/layout/items/CustomItem.qml#L112-L116

https://github.com/Kulitorum/Decenza/blob/47f8d6b3af57a2aea5fd783c95a462e0450ad641/qml/components/layout/items/CustomItem.qml#L176-L182

4. **Missing destroyed-flag pre-check in `requestShotsFiltered`** (every other async method has this guard)

The PR added `if (*destroyed) return;` before all 19 other `invokeMethod(this,...)` calls but missed this one. Same use-after-free risk the PR fixes elsewhere.

https://github.com/Kulitorum/Decenza/blob/47f8d6b3af57a2aea5fd783c95a462e0450ad641/src/history/shothistorystorage.cpp#L1446-L1450

5. **Dead `flowCalibrationAutoUpdated` signal on ProfileManager** (declared but never emitted)

The signal is only emitted from `MainController::computeAutoFlowCalibration()`. The copy on ProfileManager is dead code. Remove it or move the calibration logic.

https://github.com/Kulitorum/Decenza/blob/47f8d6b3af57a2aea5fd783c95a462e0450ad641/src/controllers/profilemanager.h#L154-L156

6. **Dead `friend class tst_McpTools` and `tst_ProfileManager`** (neither class exists)

Actual test classes are `tst_McpToolsProfiles` and `tst_McpToolsWrite`. Neither uses private member access.

https://github.com/Kulitorum/Decenza/blob/47f8d6b3af57a2aea5fd783c95a462e0450ad641/src/controllers/profilemanager.h#L189-L192

7. **`qsizetype` vs `int` in frame operations** (CLAUDE.md says "Use `qsizetype` (not `int`) for container sizes")

Frame methods compare `int index` against `.size()` (returns `qsizetype`). The `Q_INVOKABLE` params must stay `int`, but the comparison should cast explicitly: `static_cast<qsizetype>(index) >= steps.size()`.

https://github.com/Kulitorum/Decenza/blob/47f8d6b3af57a2aea5fd783c95a462e0450ad641/src/controllers/profilemanager.h#L128-L136

8. **Stale comments: "Load profile data from MainController"** (code now calls ProfileManager)

https://github.com/Kulitorum/Decenza/blob/47f8d6b3af57a2aea5fd783c95a462e0450ad641/qml/pages/ProfileEditorPage.qml#L1585-L1588

https://github.com/Kulitorum/Decenza/blob/47f8d6b3af57a2aea5fd783c95a462e0450ad641/qml/pages/RecipeEditorPage.qml#L147-L150

9. **Duplicate `applyFlowCalibration()` in both ProfileManager and MainController** (divergence risk)

Both classes have independent implementations. Changes to one won't propagate to the other. Consider consolidating into ProfileManager.

10. **MCP resource "filename" field contains display name** (CLAUDE.md MCP conventions: field names must be accurate for LLM consumers)

`result["filename"] = profileManager->currentProfileName()` returns the display title (e.g. `"*Decent DE1 v1.4"`), not a filename. Should use `baseProfileName()`.

https://github.com/Kulitorum/Decenza/blob/47f8d6b3af57a2aea5fd783c95a462e0450ad641/src/mcp/mcpresources.cpp#L78-L86

---

<details>

- [OK to skip] McpTestFixture `settings.setValue("profile/path", ...)` is a no-op since ProfileManager uses QStandardPaths directly -- currently harmless since tests use in-memory loading (score: 50)
- [OK to skip] `profiles_set_active` wraps synchronous `loadProfile()` in QueuedConnection -- verified as intentional design for correct MCP response ordering per issue #527 (score: 75 but not a bug)

</details>

PR #566, opened 2026-03-25
feat: add read_only profile protection matching de1app

skialpine:
## Summary

- Add de1app-compatible `read_only` field (integer 0/1/2) to Profile with JSON/TCL serialization
- Built-in profiles can never be overwritten — Save is disabled, Save As is required with a non-built-in name
- Modified read-only profiles get "(modified)" suffix in UI and Visualizer uploads to avoid confusion
- Migration compares user copies against built-ins: identical copies deleted (built-in takes over), truly modified copies renamed with "(user)" suffix
- Detects broken D-Flow/A-Flow profiles (wrong frame count → renamed with "(broken)", forced to advanced editor)
- Migration updates currentProfile setting and favorites after renames
- All 93 built-in profile JSON files now include `"read_only": 1`
- MCP tools expose `readOnly` in output and guard `profiles_save`
- Unified duplicate detection: `ProfileSaveHelper::compareProfiles()` now static, used by all 3 import paths (de1app import, device migration, backup restore) and the upgrade migration
- Device migration and backup restore now skip profiles identical to built-ins and avoid shadowing built-in filenames
- 22 new test cases (96 passed, 0 failed, 1 skipped)

## Test plan

- [x] Compile and run on macOS
- [x] Run test suite: 96 passed, 0 failed, 1 skipped
- [ ] Load a built-in D-Flow profile → verify "Read-Only" label in bottom bar, "Save" disabled in unsaved changes dialog
- [ ] Save As with new name → verify copy is editable (read_only cleared)
- [ ] Try Save As with a built-in name → verify "Reserved Name" dialog appears
- [ ] Import a TCL profile with `read_only 1` → verify flag is parsed and enforced
- [ ] Verify Visualizer upload shows "(modified)" suffix for edited read-only profiles
- [ ] Device migration: verify identical profiles are skipped, different ones get `_imported` suffix
- [ ] Backup restore (merge): verify identical profiles skipped, different ones deduplicated

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -149,6 +149,7 @@ set(PROFILEMANAGER_SOURCES
     ${CMAKE_SOURCE_DIR}/src/models/shotdatamodel.cpp
     ${CMAKE_SOURCE_DIR}/src/rendering/fastlinerenderer.cpp
     ${CMAKE_SOURCE_DIR}/src/network/visualizeruploader.cpp
+    ${CMAKE_SOURCE_DIR}/src/profile/profilesavehelper.cpp
 )
 
 # History sources (ShotHistoryStorage + transitive deps)
@@ -244,7 +245,7 @@ add_decenza_test(tst_profilemanager
     ${SIMULATOR_SOURCES}
     ${CMAKE_BINARY_DIR}/version_code.cpp
 )
-target_link_libraries(tst_profilemanager PRIVATE Qt6::Charts Qt6::Quick Qt6::Qml)
+target_link_libraries(tst_profilemanager PRIVATE Qt6::Charts Qt6::Quick Qt6::Qml paho-mqtt3a-static)
 target_include_directories(tst_profilemanager PRIVATE ${CMAKE_BINARY_DIR})
 target_compile_definitions(tst_profilemanager PRIVATE SRCDIR="${CMAKE_SOURCE_DIR}/tests")
 
@@ -265,7 +266,7 @@ add_decenza_test(tst_mcptools_profiles
     ${SIMULATOR_SOURCES}
     ${CMAKE_BINARY_DIR}/version_code.cpp
 )
-target_link_libraries(tst_mcptools_profiles PRIVATE Qt6::Charts Qt6::Quick)
+target_link_libraries(tst_mcptools_profiles PRIVATE Qt6::Charts Qt6::Quick paho-mqtt3a-static)
 target_include_directories(tst_mcptools_profiles PRIVATE ${CMAKE_BINARY_DIR})
 
 # --- tst_mcptools_write: Write MCP tools (settings_set temp/weight BLE upload) ---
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 5 issues:

1. **Migration does not update `settings.currentProfile`, causing silent profile switch on upgrade** (score: 95)

When `migrateReadOnlyProfiles()` renames a user file that shadows a built-in (e.g., `d_flow_default.json` to `d_flow_default_user_.json`), it never updates the `currentProfile` QSettings entry. On next launch, `loadProfile()` resolves the old filename to the stock built-in instead of the user's customized copy. The user silently loses their active profile selection.

Fix: After rename, check `if (m_settings && m_settings->currentProfile() == filename) m_settings->setCurrentProfile(newFilename);`

https://github.com/Kulitorum/Decenza/blob/2e8783e548b1e62a0e6a5c3ee117b0b66d58157e/src/controllers/profilemanager.cpp#L2413-L2430

2. **`profiles_list` MCP tool always returns `readOnly: false`** (score: 90)

The handler reads from `availableProfiles()` which only populates `name`, `title`, and `editorType` — it does not include `readOnly`. The `pm["readOnly"].toBool()` call returns false for a missing key. Every other list method (`selectedProfiles`, `allProfilesList`, etc.) correctly includes the field.

Fix: Add `profile["readOnly"] = info.readOnly;` to `availableProfiles()` alongside the other fields.

https://github.com/Kulitorum/Decenza/blob/2e8783e548b1e62a0e6a5c3ee117b0b66d58157e/src/controllers/profilemanager.cpp#L287-L304

3. **Favorites corruption when both shadow-rename and broken-detection trigger on the same profile** (score: 85)

When a user profile both shadows a built-in AND has the wrong frame count, the shadow-rename updates favorites to the intermediate `(user)` filename at line 2423, then the broken-detection overwrites `newFilename` to a `(broken)` name. Only the final filename is saved to disk, but favorites point to a file that never existed.

Fix: Move the favorites update to after both checks, just before the save, using the final `newFilename`.

https://github.com/Kulitorum/Decenza/blob/2e8783e548b1e62a0e6a5c3ee117b0b66d58157e/src/controllers/profilemanager.cpp#L2413-L2470

4. **`migrateReadOnlyProfiles()` runs disk I/O on the main thread** (CLAUDE.md: "Never run database or disk I/O on the main thread") (score: 85)

The migration reads, writes, and deletes profile files from the `ProfileManager` constructor on the main thread. The existing migrations have the same issue, but CLAUDE.md requires fixing pre-existing violations when modifying a file.

https://github.com/Kulitorum/Decenza/blob/2e8783e548b1e62a0e6a5c3ee117b0b66d58157e/src/controllers/profilemanager.cpp#L2384-L2535

5. **`SimpleProfileEditorPage.qml` does not close `saveAsDialog` before opening secondary dialogs** (score: 75)

In `doSave()`, both the built-in filename check and the overwrite check open their dialogs without first closing `saveAsDialog`, resulting in stacked modals. The other two editors correctly call `saveAsDialog.close()` first.

https://github.com/Kulitorum/Decenza/blob/2e8783e548b1e62a0e6a5c3ee117b0b66d58157e/qml/pages/SimpleProfileEditorPage.qml#L1043-L1061

---

<details>

- [OK to skip] Dead comment at `saveProfile()` line 1411 says "If saving a built-in profile" but the new read-only guard prevents built-in profiles from reaching that code. Comment is misleading but the code functions correctly for its actual use case (re-saving user copies that originated from built-ins). (score: 50)

</details>

PR #613, opened 2026-04-01
fix: prevent BLE writes before DE1 characteristics are discovered

skialpine:
## Summary

- `BleTransport::isConnected()` previously returned `true` as soon as the service object was created, but `m_characteristics` wasn't populated until characteristic discovery finished (~1s later)
- Commands arriving in that window (e.g. MQTT `steam_on`) passed the `isConnected()` guard in `startSteamHeating()` and similar methods, then were silently dropped with `"writeCharacteristic skipped - unknown characteristic"`
- Adds a `m_characteristicsReady` flag set only after `setupService()` completes, and includes it in `isConnected()` so all existing guards work correctly

## Root cause (from live log)

```
[3.867] DE1 service object created → isConnected() = true (but m_characteristics empty)
[4.074] MQTT steam_on → startSteamHeating() passes isConnected() guard
[4.140] writeCharacteristic(0000a00b) skipped - unknown characteristic  ← BUG
[4.290] writeCharacteristic(0000a006) skipped - unknown characteristic
[4.898] Characteristics ready: 18 registered  ← should have waited for this
```

## Test plan

- [ ] Connect DE1 via BLE — verify no "unknown characteristic" warnings in log at startup
- [ ] Trigger MQTT `steam_on` command immediately after app launch (before DE1 fully connects) — verify it is cleanly rejected by the `isConnected()` guard rather than silently dropped
- [ ] Normal espresso/steam/hot water flows unaffected

### DIFF
--- a/src/ble/bletransport.cpp
+++ b/src/ble/bletransport.cpp
@@ -326,6 +326,7 @@ void BleTransport::onControllerDisconnected() {
     m_writePending = false;
     m_writeTimeoutTimer.stop();
     m_commandTimer.stop();
+    m_characteristicsReady = false;
 
     emit disconnected();
 }
--- a/src/ble/bletransport.h
+++ b/src/ble/bletransport.h
@@ -21,7 +21,8 @@
  *   2. Call connectToDevice(deviceInfo)
  *   3. BleTransport handles service discovery, subscribes to notifications,
  *      reads initial values, and requests Idle state
- *   4. Emits connected() when ready for I/O
+ *   4. Emits connected() when ready for I/O (isConnected() returns false until
+ *      characteristic discovery completes, even if the BLE link is up)
  *   5. Call disconnect() or delete to tear down
  */
 class BleTransport : public DE1Transport {


skialpine:
### Code review

Found 2 issues:

1. `m_characteristicsReady` is not reset in `onControllerDisconnected()`. When the DE1 drops the BLE link unexpectedly (out of range, sleep, OS reset), this path fires instead of `disconnect()`. After that, `m_service` remains non-null and `m_characteristicsReady` stays `true`. `isConnected()` still returns `false` correctly because the controller state becomes `UnconnectedState`, but the inconsistent flag state means a future code path that checks `m_characteristicsReady` directly, or a reconnect race where the controller state briefly lags, could produce a false-positive `isConnected()` and attempt writes against a dead service object. The fix is one line: add `m_characteristicsReady = false;` alongside the existing `m_writePending = false` reset at line 326.

https://github.com/Kulitorum/Decenza/blob/b75dd2e37420ac97f22216a52d143104e5090c24/src/ble/bletransport.cpp#L316-L330

2. The class-level Lifecycle comment in the header still says `"Emits connected() when ready for I/O"` without documenting the tighter `isConnected()` semantics introduced by this PR — specifically that `isConnected()` returns `false` until characteristic discovery completes, not just when the BLE link is up. The [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/b75dd2e37420ac97f22216a52d143104e5090c24/CLAUDE.md) says "When touching existing code, fix pre-existing bugs and violations in the file you're modifying." A caller reading the header comment would not know that there is a ~1s window after service creation where `isConnected()` returns `false`.

https://github.com/Kulitorum/Decenza/blob/b75dd2e37420ac97f22216a52d143104e5090c24/src/ble/bletransport.h#L17-L27

---

PR #656, opened 2026-04-05
fix: validate favorites/currentProfile at startup, remove stale profile references

### LINKED ISSUE #654: Favorites show stale titles + load into wrong editor after profile rename
## Problem

Two related bugs surfaced while testing #653:

**1. Favorite name is frozen at the time of starring**
A favorite starred before profiles were renamed to the `A-Flow / <variant>` / `D-Flow / <variant>` scheme still shows the old short name in the favorites list — e.g. "A-Flow Medium" instead of "A-Flow / default-medium", "D-Flow Default" instead of "D-Flow / default".

Seen in:
- \`qml/pages/ProfileSelectorPage.qml\` favorites list (see screenshot)

**2. Tapping the favorite opens the wrong editor**
Clicking the "A-Flow Medium" favorite (filename \`a_flow_default_medium\`) opens the **advanced** editor instead of the **A-Flow** editor. Same happens with D-Flow favorites → advanced.

Verified current active profile state:
```
filename: d_flow_default.json
title: "Default"         <- should be "D-Flow / default"
editorType: advanced     <- should be "dflow"
```

Per CLAUDE.md:
> editor type is now derived at runtime from title + \`legacy_profile_type\`

The active profile holds a stale cached title that no longer matches the \`A-Flow /\` / \`D-Flow /\` pattern, so the editor-type derivation falls through to \`advanced\`. The on-disk profile has the correct current title (e.g. "A-Flow / default-medium"), so this is a staleness bug in either the favorite record, the current-profile cache, or the editor-type derivation when loading.

## Repro

1. Open Profiles page
2. Observe a favorite from before the naming rework (e.g. "A-Flow Medium", "D-Flow Default")
3. Tap the pencil-edit button on it
4. Wrong editor opens (advanced instead of A-Flow / D-Flow)

## Suggested fixes

- On profile load, refresh the active profile's cached title from disk so derivation sees the current title.
- **Or** make editor-type derivation also consult \`filename\` / \`legacy_profile_type\` when the title doesn't match the \`A-Flow /\` / \`D-Flow /\` pattern.
- Additionally: refresh stored favorite \`name\` fields on load if they've drifted from the on-disk title.

skialpine:
## Summary

- **Root cause of #654**: Settings stored `currentProfile = "a_flow_medium.json"` (a filename that no longer exists). `loadProfile()` fell back to a hardcoded in-memory `loadDefaultProfile()` that created a profile titled "Default" with no type — opening the advanced editor instead of the correct recipe editor. The stale name was re-persisted to settings every restart.
- `refreshProfiles()` now validates all favorites and `currentProfile` against the known profile set at startup, removing stale entries and resetting `currentProfile` to the first valid favorite (or `"default"`)
- `loadProfile()` strips `.json` extension to prevent double-extension lookups when settings stored the filename with extension
- `loadProfile()` `!found` branch now emits a `profileLoadFailed` signal and resolves to `"default"` so the stale filename is not re-written to settings
- `loadDefaultProfile()` loads the real `:/profiles/default.json` instead of building a hardcoded in-memory profile that shadowed it
- Removed `recipe.editorType` from `RecipeParams::toJson()` — editor type is derived at runtime from filename/title; existing profiles lose the field on next save
- Removed `recipe.editorType` from all 7 built-in `d_flow_*` / `a_flow_*` JSON files

## Test plan

- [ ] Build and run on Mac
- [ ] Open Profiles page — stale "A-Flow Medium" / "D-Flow Default" favorites are gone after startup
- [ ] `mcp__decenza__settings_get` — `currentProfile` is no longer `"a_flow_medium.json"`
- [ ] Add a D-Flow or A-Flow profile as favorite, restart — it survives validation and opens the recipe editor
- [ ] Verify `d_flow_default.json` no longer contains `recipe.editorType` in its `recipe` block

Closes #654

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -621,6 +621,7 @@ set(QML_FILES
 	qml/components/BrewDialog.qml
     qml/components/CircularGauge.qml
     qml/components/ConnectionIndicator.qml
+    qml/components/FavoritesListView.qml
     qml/components/PresetPillRow.qml
     qml/components/TouchSlider.qml
     qml/components/StepSlider.qml
--- a/qml/components/FavoritesListView.qml
+++ b/qml/components/FavoritesListView.qml
@@ -16,9 +16,14 @@ import Decenza
 //  - On release, emits rowMoved(startIndex, endIndex) once; consumer persists to its
 //    own store (Settings) and the resulting NOTIFY rebuilds the DelegateModel to match.
 //
-// Consumer API (identical to previous version):
-//   model, selectedIndex, displayTextFn, accessibleNameFn, deleteAccessibleNameFn,
-//   removeConfirmFn, trailingActionDelegate, showDeleteButton
+// Consumer API:
+//   model: QVariantList of row objects
+//   selectedIndex: external index of the selected row (or -1)
+//   displayTextFn(row, index), accessibleNameFn(row, index), deleteAccessibleNameFn(row, index)
+//   removeConfirmFn(row, index): return false to cancel delete
+//   trailingActionDelegate: Component whose loaded root sees `parent.row`, `parent.rowIndex`, `parent.selected`
+//   showDeleteButton: toggles the X button column
+//   rowAccessibleDescription: TalkBack/VoiceOver hint describing the long-press/double-tap action
 //
 // Signals: rowSelected(int), rowLongPressed(int), rowMoved(int from, int to), rowDeleted(int)
 Item {
@@ -36,6 +41,12 @@ Item {
                (row && row.name ? row.name : "")
     }
     property var removeConfirmFn: function(row, index) { return true }
+
+    // TalkBack/VoiceOver hint for the row's secondary (long-press/double-tap) action.
+    // Consumers should override with context-specific text (e.g. "rename preset", "edit profile").
+    property string rowAccessibleDescription: TranslationManager.translate(
+        "favorites.accessible.secondary_hint",
+        "Double-tap or long-press for more options.")
 
     signal rowSelected(int index)
     signal rowLongPressed(int index)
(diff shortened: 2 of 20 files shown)


skialpine:
### Code review

Found 2 issues:

1. `int` used for loop variable instead of `qsizetype` (CLAUDE.md says "Use `qsizetype` (not `int`) for container sizes — `QVector::size()`, `QList::size()`, `QString::size()` etc. return `qsizetype` (64-bit on iOS/macOS). Assigning to `int` causes `-Wshorten-64-to-32` warnings.")

https://github.com/Kulitorum/Decenza/blob/3e26ba738ab6f1496c4a27b0dfc483fa721cb4d9/src/controllers/profilemanager.cpp#L1169-L1171

Fix: `for (qsizetype i = favorites.size() - 1; i >= 0; --i)` — `qsizetype` is signed so the `>= 0` termination condition works correctly.

2. "No favorites yet" empty-state hint will never show after this PR's startup validation removes stale favorites. After pruning, `favoriteProfiles.length === 0` AND `selectedFavoriteProfile === -1` (set by `loadProfile()` at line 898 when the fallback profile isn't in favorites). The `!== -1` clause makes the condition logically impossible in this scenario.

https://github.com/Kulitorum/Decenza/blob/3e26ba738ab6f1496c4a27b0dfc483fa721cb4d9/qml/pages/ProfileSelectorPage.qml#L637-L649

Fix: `visible: Settings.favoriteProfiles.length === 0` (the non-favorite green pill at line 656 independently handles the "non-favorite profile loaded" case, so removing the `!== -1` guard is safe).

---

<details>

- [OK to skip] `loadDefaultProfile()` does not check if `Profile::loadFromFile(":/profiles/default.json")` succeeded — would silently write `targetWeight=0` and `temperature=0` as setting overrides if the resource failed to load. Risk is negligible since it's a compiled-in QRC resource. (score: 50)
- [OK to skip] `refreshProfiles()` validation can mutate `m_settings->currentProfile()` mid-session without updating `m_currentProfile` in memory, causing settings/memory divergence until the user manually selects another profile. Pre-existing gap also present in `deleteProfile()`. (score: 50)

</details>

PR #714, opened 2026-04-10
refactor: remove knowledge_base_id — rely on title/alias matching

skialpine:
## Summary

- Removes the `knowledge_base_id` field from all 23 built-in profile JSONs and the `Profile` C++ class — every case it handled is already covered by `Also matches:` aliases or prefix/fuzzy title matching in `matchProfileKey()`
- Removes `updateProfileKnowledgeBaseId()` from `ProfileManager` and simplifies KB ID resolution to always compute from title + editor type
- User-saved derivatives (e.g. "Turbo Shot - Mine") continue to match via prefix matching

## Test plan

- [ ] Load a built-in profile (e.g. D-Flow, Rao Allongé, Filter 2.1) — AI shot analysis should still inject the correct KB section
- [ ] Load a Damian variant (LM Leva, LRv2, LRv3) — should match D-Flow KB section via "Also matches" alias
- [ ] Save As a profile with a custom name like "Turbo Shot - Mine" — should still match Turbo Shot KB via prefix
- [ ] Profile selector KB viewer should still show KB text for all profiles


skialpine:
### Code review

Found 5 issues:

1. **Save-As regression: renamed profiles lose KB linkage and analysis flags** (bug introduced by PR)

When a user saves a renamed copy of a built-in profile (e.g. "Blooming Espresso" → "My Morning Bloom"), the shot-save path now calls `computeProfileKbId("My Morning Bloom", "advanced")` which returns empty — no KB section match. Consequence: the AI gets no curated knowledge for that profile, and `channeling_expected` / `grind_check_skip` flags are not checked, so false-positive channeling or grind badges may be written permanently to the DB. Previously, `profile->knowledgeBaseId()` returned "blooming espresso" (carried through Save As via the JSON field).

https://github.com/Kulitorum/Decenza/blob/7c2d0f33c4a5610d26e8e8dc670dc187a50185c8/src/history/shothistorystorage.cpp#L847-L851

2. **Stale comment in `shotsummarizer.cpp` still describes removed two-tier KB architecture** (file was modified by this PR)

The comment "Try direct KB ID first (from database), then fall back to fuzzy title/editorType matching" no longer describes what the code does — `profileKbId` is now always the result of `matchProfileKey()`, not a direct DB cache that bypasses matching.

https://github.com/Kulitorum/Decenza/blob/7c2d0f33c4a5610d26e8e8dc670dc187a50185c8/src/ai/shotsummarizer.cpp#L807-L812

3. **Orphaned comment in `profile.cpp` `fromJson()` with no associated code** (file was modified by this PR)

`// AI knowledge base ID (Decenza extension)` at line 541 was left behind after removing `profile.m_knowledgeBaseId = obj["knowledge_base_id"].toString()`.

https://github.com/Kulitorum/Decenza/blob/7c2d0f33c4a5610d26e8e8dc670dc187a50185c8/src/profile/profile.cpp#L540-L543

4. **`docs/AI_ADVISOR.md` still describes `knowledge_base_id` as active infrastructure**

Lines 55 and 59 say "Direct KB ID: stored in profile JSON… Survives Save As" and "31 stock profiles ship with knowledge_base_id baked into their JSON files" — both now false. The PR updated `PROFILE_KNOWLEDGE_BASE.md` and `UNIVERSAL_GRIND_SETTING.md` but missed this file.

https://github.com/Kulitorum/Decenza/blob/7c2d0f33c4a5610d26e8e8dc670dc187a50185c8/docs/AI_ADVISOR.md#L53-L61

5. **Double blank lines left in `profilemanager.cpp` after removing `updateProfileKnowledgeBaseId()` call sites**

Two locations: around line 873 (inside `loadProfile()`) and around line 2102 (after the deleted function body).

https://github.com/Kulitorum/Decenza/blob/7c2d0f33c4a5610d26e8e8dc670dc187a50185c8/src/controllers/profilemanager.cpp#L871-L876

---

skialpine:
Follow-up: issue #1 (Save-As regression) is a false positive for two reasons: (1) built-in profiles are read-only, so users cannot Save As a built-in to get a renamed copy; (2) even if they could, the old behavior was incorrect — `knowledge_base_id` would be copied verbatim to any Save As derivative regardless of how much the profile was changed, so a completely different custom profile would still receive the wrong KB section from the AI. The new title-matching approach is more correct: a substantially different custom profile correctly gets no KB match rather than wrong KB context.

PR #715, opened 2026-04-10
fix: suppress BLE writes when adjusting settings during active operation

skialpine:
## Summary

Fixes UI freeze on iPadOS 26.4 when pressing ±5s during steaming. Calling `startSteamHeating()` mid-steam was sending `setShotSettings` + `writeMMR` BLE commands to the DE1 while it was already running, causing the app to become unresponsive.

- **Steam ±5s buttons**: guard with `!isSteaming` — display updates immediately, BLE send suppressed during active steam
- **Steam pitcher pills**: same guard added (was an identical unguarded path in the same steaming view)
- **Hot water / flush live preset pills**: these are designed for mid-operation switching and are only reachable during the active operation, so the BLE send is kept unconditional (original behavior restored)
- Live steam flow control (`setSteamFlowImmediate`) is unaffected — it uses a separate single-MMR path
- Fix `waterVolumeMode || "weight"` → `?? "weight"` (empty string was being coerced to `"weight"`)

## Test plan

- [ ] Steam: press ±5s during steaming — display updates, no freeze
- [ ] Steam: press ±5s during heat-up (before steaming) — `startSteamHeating()` fires, updated timeout sent to machine
- [ ] Steam: switch pitcher preset during steaming — display updates, no freeze
- [ ] Hot water: tap a different vessel preset during dispensing — updates machine settings live
- [ ] Flush: tap a different preset during flushing — updates machine settings live
- [ ] Steam: adjust flow slider during steaming — still works (live flow path unchanged)

### DIFF
--- a/qml/pages/FlushPage.qml
+++ b/qml/pages/FlushPage.qml
@@ -105,8 +105,7 @@ Page {
                                 Settings.selectedFlushPreset = index
                                 Settings.flushFlow = modelData.flow
                                 Settings.flushSeconds = modelData.seconds
-                                if (!isFlushing)
-                                    MainController.applyFlushSettings()
+                                MainController.applyFlushSettings()
                             }
                         }
                     }
--- a/qml/pages/FlushPage.qml
+++ b/qml/pages/FlushPage.qml
@@ -105,7 +105,8 @@ Page {
                                 Settings.selectedFlushPreset = index
                                 Settings.flushFlow = modelData.flow
                                 Settings.flushSeconds = modelData.seconds
-                                MainController.applyFlushSettings()
+                                if (!isFlushing)
+                                    MainController.applyFlushSettings()
                             }
                         }
                     }
--- a/qml/pages/HotWaterPage.qml
+++ b/qml/pages/HotWaterPage.qml
@@ -113,10 +113,9 @@ Page {
                             onClicked: {
                                 Settings.selectedWaterVessel = index
                                 Settings.waterVolume = modelData.volume
-                                Settings.waterVolumeMode = (modelData.mode || "weight")
+                                Settings.waterVolumeMode = (modelData.mode ?? "weight")
                                 Settings.hotWaterFlowRate = (modelData.flowRate !== undefined) ? modelData.flowRate : 40
-                                if (!isDispensing)
-                                    MainController.applyHotWaterSettings()
+                                MainController.applyHotWaterSettings()
                             }
                         }
                     }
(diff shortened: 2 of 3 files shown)


skialpine:
### Code review

Found 4 issues:

1. **`root.isSteaming` is undefined — the ±5s guard is a no-op and the steam freeze is not fixed** (bug)

`SteamPage.qml` has no `id: root` on its Page element. `root` resolves to the `ApplicationWindow` in `main.qml`, which has no `isSteaming` property. `root.isSteaming` evaluates to `undefined`, so `!root.isSteaming` is always `true` and `startSteamHeating()` is never suppressed. Every other use of `isSteaming` in this file uses the unqualified name (lines 34, 38, 48, 63, 214, etc.). The guard should be `if (!isSteaming)`.

https://github.com/Kulitorum/Decenza/blob/1686eca81030341f259b7bc8c4dc4d4dd475783e/qml/pages/SteamPage.qml#L380-L388

https://github.com/Kulitorum/Decenza/blob/1686eca81030341f259b7bc8c4dc4d4dd475783e/qml/pages/SteamPage.qml#L433-L441

2. **Live pitcher-switching handler left unguarded** (incomplete fix)

The preset pill `livePitcherMa` at line 263 calls `MainController.startSteamHeating()` unconditionally. It sits inside the same "STEAMING VIEW" block (`visible: isSteaming || steamSoftStopped || ...`) as the ±5s buttons. Switching a pitcher during active steaming triggers the same BLE burst the PR was written to prevent.

https://github.com/Kulitorum/Decenza/blob/1686eca81030341f259b7bc8c4dc4d4dd475783e/qml/pages/SteamPage.qml#L255-L265

3. **`if (!isDispensing)` makes `applyHotWaterSettings()` dead code** (logic inversion)

The vessel preset pills are inside the "DISPENSING VIEW" `ColumnLayout` which has `visible: isDispensing` (line 74). The guard `!isDispensing` is therefore never true when the pills are reachable — `applyHotWaterSettings()` can never be called from this handler. The comment on line 79 reads "Preset pills for quick switching during dispensing"; the guard silently removes that capability.

https://github.com/Kulitorum/Decenza/blob/1686eca81030341f259b7bc8c4dc4d4dd475783e/qml/pages/HotWaterPage.qml#L112-L121

4. **`if (!isFlushing)` makes `applyFlushSettings()` dead code** (logic inversion)

Same pattern: the flush preset pills are inside `visible: isFlushing` (line 65). The guard `!isFlushing` is never true when the pills are reachable. The comment on line 70 reads "Preset pills for quick switching during flushing."

https://github.com/Kulitorum/Decenza/blob/1686eca81030341f259b7bc8c4dc4d4dd475783e/qml/pages/FlushPage.qml#L103-L111

---

<details>

- [Worth a look] `Settings.waterVolumeMode = (modelData.mode || "weight")` in the modified handler — CLAUDE.md says to use `??` not `||` where falsy values are valid; empty string `""` mode would be silently replaced with `"weight"`. One-character fix in code already being changed. (score: 70)

</details>

PR #716, opened 2026-04-10
fix: improve Visualizer upload parity with de1app

skialpine:
## Summary

- **Fix `water_dispensed` 10× scaling bug**: de1app stores `espresso_water_dispensed` at 0.1× scale (tenths of ml); apply same scaling before upload so Visualizer displays the correct volume
- **Add `water_dispensed` to history re-uploads**: was previously missing from `buildHistoryShotJson`
- **Add `resistance.by_weight`**: P/flow_weight² using scale-derived flow rate (g/s), matching de1app's `resistance_weight` field — both live and history paths
- **Add `scale` object**: raw weight series at native scale sample times (`weight`, `weight_arrival`, `weight_flow`, `espresso_start`), matching de1app format — both paths
- **Add `meta.out` firmware volume fallback**: when scale weight is 0/missing (e.g. scale disconnects mid-shot), fall back to `waterDispensed` integration volume — both paths
- **Embed `debug_log` in `app.data`**: shot debug log is now uploaded with every shot so support investigations don't require manually retrieving logs — both paths
- **Store `m_pendingDebugLog`** in `MainController` so the debug log is available for deferred (post-metadata) re-uploads

## Test plan
- [ ] Pull down branch and do a live shot — confirm Visualizer shows `resistance.by_weight` curve and correct `water_dispensed` volume
- [ ] Re-upload a shot from history — confirm same fields present and `debugLog` from DB is included
- [ ] Shot with scale disconnect mid-shot — confirm `meta.out` falls back to firmware volume instead of 0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ logcat_*.txt
 -.zip
 .claude/
 .DS_Store
+/.codex_tmp
--- a/src/controllers/maincontroller.cpp
+++ b/src/controllers/maincontroller.cpp
@@ -1261,6 +1261,9 @@ void MainController::onShotEnded() {
     // Must run before stopCapture() so its debug output is included in the shot log.
     computeAutoFlowCalibration();
 
+    // Capture shot-end epoch now so uploads (including deferred pending uploads) use consistent time
+    m_pendingShotEpoch = QDateTime::currentSecsSinceEpoch();
+
     // Stop debug logging and get the captured log
     QString debugLog;
     if (m_shotDebugLogger) {
@@ -1390,7 +1393,7 @@ void MainController::onShotEnded() {
     // Auto-upload if enabled (do this first, before showing metadata page)
     if (m_settings->visualizerAutoUpload() && m_visualizer) {
         qDebug() << "  -> Auto-uploading to visualizer";
-        m_visualizer->uploadShot(m_shotDataModel, m_profileManager->currentProfilePtr(), duration, finalWeight, doseWeight, metadata, debugLog);
+        m_visualizer->uploadShot(m_shotDataModel, m_profileManager->currentProfilePtr(), duration, finalWeight, doseWeight, metadata, debugLog, m_pendingShotEpoch);
     }
 
     // Store pending shot data for later upload (user can re-upload with updated metadata)
@@ -1457,9 +1460,10 @@ void MainController::uploadPendingShot() {
 
     m_visualizer->uploadShot(m_shotDataModel, m_profileManager->currentProfilePtr(),
                              m_pendingShotDuration, m_pendingShotFinalWeight,
-                             m_pendingShotDoseWeight, metadata, m_pendingDebugLog);
+                             m_pendingShotDoseWeight, metadata, m_pendingDebugLog, m_pendingShotEpoch);
 
     m_hasPendingShot = false;
+    m_pendingDebugLog.clear();
 }
 
 void MainController::generateFakeShotData() {
--- a/src/controllers/maincontroller.cpp
+++ b/src/controllers/maincontroller.cpp
@@ -1267,6 +1267,7 @@ void MainController::onShotEnded() {
         m_shotDebugLogger->stopCapture();
         debugLog = m_shotDebugLogger->getCapturedLog();
     }
+    m_pendingDebugLog = debugLog;
 
     // Build metadata for history
     ShotMetadata metadata;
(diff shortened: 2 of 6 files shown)


skialpine:
### Code review

Found 5 issues:

1. **Compile error: `weightFlowRateData` declared twice at function scope in `buildShotJson()`** — the live upload path will not build. Line 396 declares `const auto& weightFlowRateData = shotData->weightFlowRateData();` and line 482 redeclares it at the same function-body scope (the `{` on line 483 opens a block *after* the declaration, not around it). Fix: delete line 482 and rely on the existing declaration at line 396, which is already used for `flow["by_weight"]` at line 436.

https://github.com/Kulitorum/Decenza/blob/172b5f377660d562371db22801ad0cac8e578426/src/network/visualizeruploader.cpp#L394-L397

https://github.com/Kulitorum/Decenza/blob/172b5f377660d562371db22801ad0cac8e578426/src/network/visualizeruploader.cpp#L480-L485

2. **`resistance.resistance` uploads P/flow (linear) but Visualizer expects P/flow² (Darcy)** — the new comment at line 480 claims "P/flow² (DE1 flow)" but `shotData->resistanceData()` is computed in `shotdatamodel.cpp:263` as `pressure / flow` (capped at 15.0). The Darcy/quadratic formula that de1app sends for `espresso_resistance` is `pressure / (flow * flow)`, which Decenza computes separately as `darcyResistanceData()` (capped at 19.0, matching Visualizer's convention). The fix is to call `shotData->darcyResistanceData()` for this field. The same issue applies to the history path at line 1057 (`shotData["resistance"]` is the linear P/flow stored at save time — would also need remapping, or the DB field to be fixed at source). CLAUDE.md requires fixing pre-existing violations in touched files.

https://github.com/Kulitorum/Decenza/blob/172b5f377660d562371db22801ad0cac8e578426/src/network/visualizeruploader.cpp#L480-L487

3. **`scale.espresso_start` in the live path is the upload time, not the shot start time** — `root["clock"]` is set at line 403–404 as `QDateTime::currentSecsSinceEpoch()` when `buildShotJson()` runs. For `uploadPendingShot()` (called after user enters metadata), this can be minutes after extraction ended. The history path correctly uses `shotData["timestamp"]`. `weight_arrival` values are shot-relative seconds, so Visualizer uses `espresso_start + weight_arrival` for wall-clock display; an incorrect `espresso_start` offsets all scale timestamps. Fix: capture the shot-start epoch in `MainController::onEspressoCycleStarted()` and thread it through alongside the other pending-shot fields.

https://github.com/Kulitorum/Decenza/blob/172b5f377660d562371db22801ad0cac8e578426/src/network/visualizeruploader.cpp#L402-L406

https://github.com/Kulitorum/Decenza/blob/172b5f377660d562371db22801ad0cac8e578426/src/network/visualizeruploader.cpp#L528-L532

4. **de1app-imported shots get `water_dispensed` double-scaled in history re-uploads** — when a shot is imported from a de1app `.shot` file via `shotfileparser.cpp`, `espresso_water_dispensed` is stored verbatim at de1app's 0.1× convention (e.g. 4.0 for a 40 ml shot). The new code at lines 1046–1050 applies `* 0.1` unconditionally, producing 0.01× values (100× too small) for those shots. Natively-captured Decenza shots are unaffected. The same caveat applies to the `meta["out"]` fallback at line 1166 (de1app-imported shots would report ~0.4 g instead of ~40 g when scale weight is absent). The fix is to normalize at import time in `shotfileparser.cpp` (multiply by 10 when reading `espresso_water_dispensed`) so the DB always holds raw ml.

https://github.com/Kulitorum/Decenza/blob/172b5f377660d562371db22801ad0cac8e578426/src/network/visualizeruploader.cpp#L1043-L1052

5. **`m_pendingDebugLog` not cleared after `uploadPendingShot()` — stale log uploaded after `generateFakeShotData()`** — `uploadPendingShot()` resets `m_hasPendingShot` but leaves `m_pendingDebugLog` set. `generateFakeShotData()` sets `m_hasPendingShot = true` (line 1550) without writing `m_pendingDebugLog`, so a subsequent `uploadPendingShot()` call in dev/test mode would attach the previous real shot's debug log to the fake shot upload. Fix: clear `m_pendingDebugLog` alongside `m_hasPendingShot = false`.

https://github.com/Kulitorum/Decenza/blob/172b5f377660d562371db22801ad0cac8e578426/src/controllers/maincontroller.cpp#L1459-L1463

---

PR #717, opened 2026-04-10
refactor: unify BLE scan into single Scan for Devices button

skialpine:
## Summary

- Removes the separate \"Scan for DE1\" button from Connections settings — \"Scan for Devices\" now covers DE1, scales, and refractometers in one scan
- Renames \`scanForScales()\` → \`scanForDevices()\` to reflect unified scope
- Stops disconnecting the connected scale when a scan starts — discovered devices appear in the list and the user switches by tapping one
- Fixes a bug where a DE1-only scan wiped the discovered scale list, causing the scale to disappear from the UI

**Follow-up fixes from code review:**
- Use case-insensitive address comparison (\`address.compare(m_savedScaleAddress, Qt::CaseInsensitive)\`) in \`connectToScale()\` to avoid false disconnects when MAC address casing differs
- Remove duplicate \`scanDE1\` action from layout editor (now identical to \`scanScale\` after unification); rename \`scanScale\` label to \"Scan for Devices\"
- Disconnect flowScale before reconnecting in \`disconnectScaleRequested\` handler to prevent duplicate signal connections
- Re-enable scale/refractometer scanning in simulator mode (only DE1 auto-connect remains suppressed)
- Treat \`SimulatedScale\` as a configured scale for SAV/basic-profile skip logic so the simulated DE1 doesn't stop shots early at the volume limit
- Fix simulator yield model: lower \`PUCK_FILL_VOLUME\` (8→5ml), raise \`DRIP_START_EFFICIENCY\` (0.40→0.82), lower \`EFFICIENCY_RAMP_ML\` (25→8ml) so 36ml SAV produces ~30g in cup instead of ~15g

## Test plan

- [ ] Open Settings → Connections — confirm only one scan button exists (\"Scan for Devices\" in the scale section)
- [ ] With a scale connected, tap \"Scan for Devices\" — scale stays connected, discovered DE1/scale/refractometer devices populate their lists
- [ ] Tap a discovered scale in the list — switches to that scale
- [ ] With DE1 disconnected, run scan — DE1 appears in the discovered list for reconnection
- [ ] In simulator mode: tap \"Scan for Devices\" — BLE scan runs and real scales/refractometers appear
- [ ] In simulator mode with Default profile (settings_2a): run a shot to weight target — shot is not cut short by SAV at 36ml


skialpine:
### Code review

Found 6 issues:

1. **Tap-to-switch scale is silently broken** — removing `disconnectScaleRequested()` from `scanForDevices()` means the old scale stays live. The `scaleDiscovered` handler in `main.cpp` has an early return that fires immediately when a scale is already connected, so tapping a new scale from the discovered list is a no-op. Users cannot switch scales without hitting Forget first.

https://github.com/Kulitorum/Decenza/blob/dac40e956b77c54e75122ef040c733babcfbf094/src/main.cpp#L1103-L1110

2. **Tap-to-switch refractometer is silently broken** — same pattern: `refractometerDiscovered` has an identical early return at line 1277. Tapping a different R2 while one is connected does nothing.

https://github.com/Kulitorum/Decenza/blob/dac40e956b77c54e75122ef040c733babcfbf094/src/main.cpp#L1275-L1282

3. **Stale `accessibleName` on the scan button** — still reads `"Scan for Bluetooth scales and refractometers"` but the button now also discovers DE1 machines. Screen reader users are given an inaccurate description of the button's action.

https://github.com/Kulitorum/Decenza/blob/dac40e956b77c54e75122ef040c733babcfbf094/qml/pages/settings/SettingsConnectionsTab.qml#L754-L758

4. **Stale signal comment on `disconnectScaleRequested`** — the comment reads `// Emitted when starting scan, scale should disconnect` but `scanForDevices()` no longer emits it. A future developer reading this could re-add the emission to "fix" what looks like a missing call.

https://github.com/Kulitorum/Decenza/blob/dac40e956b77c54e75122ef040c733babcfbf094/src/ble/blemanager.h#L123-L125

5. **Stale user-visible log message in `scanForDevices()`** — `appendScaleLog("Starting scale scan...")` is shown in the Settings scale log panel; it should say "Starting device scan..." to match the unified scope.

https://github.com/Kulitorum/Decenza/blob/dac40e956b77c54e75122ef040c733babcfbf094/src/ble/blemanager.cpp#L657-L661

6. **`compactWarning` Rectangle is missing `Accessible.ignored: true`** — the root `Item` correctly handles accessibility, but the child `compactWarning` Rectangle (which contains a raw `MouseArea`) is not marked `Accessible.ignored: true`. TalkBack may traverse into it as a separate focusable element. CLAUDE.md: "Never put `Accessible.role`/`name`/`focusable` on a raw `MouseArea` child" and child elements of an accessible parent must be marked ignored.

https://github.com/Kulitorum/Decenza/blob/dac40e956b77c54e75122ef040c733babcfbf094/qml/components/layout/items/ScaleWeightItem.qml#L76-L117

---

<details>

- [Worth a look] `command:scanDE1` in `CustomItem.qml` still calls `BLEManager.startScan()` directly (DE1 only, no scales), while `command:scanScale` now calls `scanForDevices()` (everything). The two custom actions now have inconsistent scope under the same unified-scan goal — but this is pre-existing behavior not introduced by this PR (score: 50)

</details>

skialpine:
All 6 issues + the `command:scanDE1` item addressed in 622cf4cb:

- **Tap-to-switch scale**: `connectToScale()` now emits `disconnectScaleRequested()` when the requested address differs from the currently connected scale, allowing the `scaleDiscovered` handler to proceed
- **Tap-to-switch refractometer**: `refractometerDiscovered` handler now checks device identity before the early return — same device still short-circuits, different device falls through to cleanup + create
- **Stale `accessibleName`**: updated to "Scan for Bluetooth DE1, scales, and refractometers"
- **Stale `disconnectScaleRequested` comment**: updated to reflect actual emit sites
- **Stale log message**: "Starting scale scan..." → "Starting device scan..."
- **`compactWarning` accessibility**: added `Accessible.ignored: true` to the Rectangle
- **`command:scanDE1`**: now calls `scanForDevices()` instead of `startScan()`

skialpine:
### Code review

Found 2 issues:

1. `address != m_savedScaleAddress` uses raw string comparison instead of `deviceIdentifiersMatch()`. On non-iOS platforms, BLE MAC addresses may have different casing (e.g., \`AA:BB:CC\` vs \`aa:bb:cc\`), so the check could fail to recognise the same device, triggering an unnecessary disconnect when the user taps the already-connected scale. Line 150 in the same function already correctly uses `deviceIdentifiersMatch()`.

https://github.com/Kulitorum/Decenza/blob/622cf4cb4a6f41add43d1b53e020ade995f1401f/src/ble/blemanager.cpp#L148-L162

2. After unifying `scanDE1` and `scanScale` to both call `BLEManager.scanForDevices()`, the two entries remain as distinct selectable actions in the custom layout editor. Users who add both buttons get two buttons that do exactly the same thing, with no visible indication they are redundant.

https://github.com/Kulitorum/Decenza/blob/622cf4cb4a6f41add43d1b53e020ade995f1401f/qml/components/layout/CustomEditorPopup.qml#L1368-L1371

---

<details>

- [OK to skip] `// Different device selected — fall through to cleanup + create` uses switch-case "fall through" terminology for normal sequential flow — comment-only, no functional impact (score: 50)

</details>

PR #719, opened 2026-04-10
fix: prevent SIGSEGV during QML incubation on startup (#718)

skialpine:
## Summary

Fixes a crash (SIGSEGV) at app startup on Windows reported in #718.

**Root cause (best current theory):** GPS position update at ~2s after launch fired \`locationChanged\` synchronously during QML incubation, which triggered \`fetchWeather()\` → \`setLoading()\` → \`emit loadingChanged()\`. When \`loadingChanged\` fired while \`QQmlConnections::connectSignalsToMethods\` was setting up a binding, the V4 SharedArrayBuffer was not yet initialized → SIGSEGV. The Windows GUI event dispatcher's event ordering makes this timing reproducible there but not on other platforms.

**Changes:**

1. **QML-ready guard** (`WeatherManager::setQmlReady()`): any `fetchWeather()` call before `engine.load()` returns is deferred and replayed once QML is ready. This eliminates `loadingChanged()` firing during incubation entirely, regardless of signal ordering.

2. **`Qt::QueuedConnection`** on the `locationChanged` → `onLocationChanged` connection: breaks the synchronous chain as a secondary defense.

3. **Log the silent `locationChanged` path** in `onPositionUpdated`: the non-geocoding branch (city already known, position unchanged) emits the signal with no log — now it logs `"emitting locationChanged directly (synchronous)"` so future crash reports can confirm this path was taken.

4. **Log `setLoading()` transitions**: `"WeatherManager: loading = true/false"` — correlates fetch lifecycle with crash timing in future reports.

## Test plan

- [ ] Launch app with a cached GPS position — no crash on startup
- [ ] Crash log should show `"WeatherManager: QML not ready, deferring fetch"` if the fix is active during a near-miss
- [ ] Weather still loads after startup (`"WeatherManager: QML ready, running deferred fetch"` in log)
- [ ] GPS position updates still trigger weather fetches correctly


skialpine:
### Code review

Found 5 issues:

1. **`setQmlReady()` calls `fetchWeather()` synchronously — same race as #718** (bug introduced by this PR)

`setQmlReady()` is called synchronously in `main.cpp` right after `engine.load()` returns. It immediately calls `fetchWeather()` → `setLoading(true)` → `emit loadingChanged()`. Qt's `engine.load()` creates the root QML object but asynchronous sub-components (Loaders, async incubation) remain in the event queue. If any binding to `WeatherManager.loading` is still incubating, this fires the same `connectSignalsToMethods` race the PR claims to fix. The fix in `setLocationProvider` explicitly uses `Qt::QueuedConnection` to avoid exactly this — `setQmlReady()` should do the same.

Fix: use `QMetaObject::invokeMethod(this, &WeatherManager::fetchWeather, Qt::QueuedConnection)` inside `setQmlReady()` instead of calling `fetchWeather()` directly.

https://github.com/Kulitorum/Decenza/blob/4047623d6f043183225b259a4f5c4ac4331762fe/src/weather/weathermanager.cpp#L176-L185

2. **GPS position-unchanged log fires on every GPS poll on a stationary device** (diagnostic not removed)

The log `"LocationProvider: Position unchanged, emitting locationChanged directly (synchronous)"` was added to trace the #718 crash path. On a stationary device (e.g., a DE1 machine sitting on a counter), GPS updates arrive every few seconds and position never changes — so this fires continuously for the entire app lifetime, flooding the device log.

https://github.com/Kulitorum/Decenza/blob/4047623d6f043183225b259a4f5c4ac4331762fe/src/network/locationprovider.cpp#L186-L191

3. **`setLoading()` log fires ~48 times/day in normal operation** (diagnostic not removed)

`qDebug() << "WeatherManager: loading =" << loading` was added as a temporary diagnostic. Unlike the other new logs (which are one-time startup events), this fires on every fetch start and completion — at minimum twice per hourly refresh cycle, indefinitely. The commit message itself says "to correlate with crash timing in future reports," implying it was intended as temporary.

https://github.com/Kulitorum/Decenza/blob/4047623d6f043183225b259a4f5c4ac4331762fe/src/weather/weathermanager.cpp#L219-L226

4. **`simulatedScaleEnabled` toggle does not disable the simulated scale** (misleading UX)

The "Simulated Scale" toggle in debug settings only affects the SAV skip logic in `machinestate.cpp`. The `SimulatedScale` is unconditionally created and wired up at startup — there is no connection to `simulatedScaleEnabledChanged` that disconnects it or stops weight reporting. A user toggling it OFF still sees the scale connected, reporting weight in the UI, and triggering SAW. Only the volume-stop skip behavior changes, which is invisible in the UI. The label implies the scale will be disabled entirely.

https://github.com/Kulitorum/Decenza/blob/4047623d6f043183225b259a4f5c4ac4331762fe/qml/pages/settings/SettingsDebugTab.qml#L184-L201

5. **`scaleConfigured` fourth condition uses `isConnected()`, contradicting the stated design principle** (logic bug)

The existing comment says _"Uses 'configured' (scaleAddress non-empty) not 'connected' for BLE scales so a momentary disconnect mid-shot doesn't re-enable SAV."_ But the newly added fourth condition `(!simulatedScaleActive && m_scale->isConnected() && !m_scale->isFlowScale())` uses `isConnected()` directly. A USB scale or a newly connected BLE scale without a saved address will flip this condition on a mid-shot disconnect, re-enabling SAV mid-shot — the exact scenario the design principle was intended to prevent. Duplicated identically in both `checkStopAtVolume()` and `checkStopAtVolumeHotWater()`.

https://github.com/Kulitorum/Decenza/blob/4047623d6f043183225b259a4f5c4ac4331762fe/src/machine/machinestate.cpp#L700-L710

---

skialpine:
### Code review

All 5 issues from the first review are correctly fixed in commit 79915b40:
- `setQmlReady()` now defers via `QMetaObject::invokeMethod(..., Qt::QueuedConnection)` — no longer synchronous
- GPS position-unchanged log removed from `locationprovider.cpp`
- `setLoading()` debug log removed from `weathermanager.cpp`
- `simulatedScaleEnabled` toggle now calls `simulateConnection()`/`simulateDisconnection()` and wires/unwires the weight signal chain
- `scaleConfigured` simplified to 3 conditions; `isConnected()` on SimulatedScale is now the authoritative source controlled by the toggle

---

**Side note on issues #683 and #684**: Both crash reports show the identical stack (`QQmlConnections::connectSignalsToMethods` → `QQmlObjectCreator::finalize` → `QQmlIncubatorPrivate::incubate`) — the same crash class as #718. However, their triggers appear different: #683 fires on `RelayClient: WebSocket connected` (~1.8s into startup), and #684 fires on settings-page navigation (app had been running, 289MB RSS). PR #719 specifically defers the `WeatherManager` signal chain; if those crashes come from different signal sources firing during QML incubation, they may persist after this fix. Worth watching post-release crash reports.

PR #721, opened 2026-04-10
feat: PostShotReview header improvements + skip-first-frame quality badge

skialpine:
## Summary

- **PostShotReviewPage header** now matches the layout introduced for ShotDetailPage in cdcf4de: date inline with title, QualityBadges + ShotAnalysisDialog ("Shot Summary") in the header row, sparkle KB button, and profile name/date in the bottom bar while scrolling
- **Read TDS button** moved from the rating row to the header — visible whenever an R2 refractometer is configured, regardless of advanced mode
- **New "First step skipped" quality badge** (red) detects the known DE1 firmware bug where the machine skips profile frame 0, and the secondary case where the first step is under 2 seconds
- **Several R2/EY bug fixes** discovered during review and testing

## New badge details

Ported from the `skip_first_step_notice` de1app plugin by Damian. Detection uses phase markers stored in `shot_phases`:
- **FW bug**: first phase marker has `frameNumber != 0` → machine never executed frame 0, requires power-cycle to fix
- **Profile issue**: frame 0 present but next frame arrived in < 2 s → first step too short

Fully integrated into the quality badge pipeline:
- `ShotAnalysis::detectSkipFirstFrame()` — detection logic
- DB migration 12 adds `skip_first_frame_detected` column
- `requestReanalyzeBadges()` recomputes on shot open
- `shotBadgesUpdated` signal gains 4th bool param
- Filterable: `ShotFilter::filterSkipFirstFrame`
- Keyword search: `skipframe:yes` in history search bar (including help popover entry)
- `loadShotRecordStatic` always recomputes from phase markers (covers pre-migration-12 shots)
- ShotHistoryPage indicator dot + accessible label

## R2 / EY bug fixes

- **R2 reconnect button used wrong BLE call**: was calling `tryDirectConnectToRefractometer()` which silently does nothing if a scan is already in progress (from `onCompleted`). Now calls `scanForDevices()` (introduced in #717) which stops any current scan and restarts fresh.
- **EY not computed when TDS arrived before shot data loaded**: `onTdsChanged` could fire before `onShotReady` completed (async DB load), then `onShotReady` would overwrite `editDrinkTds` back to 0. Fixed by preserving the live R2 reading and calling `calculateEy()` at the end of `onShotReady`.
- **`dyeBeanWeight` silently zeroed by saving without a dose**: `saveCurrentValues()` unconditionally wrote `editDoseWeight` (potentially 0) back to `Settings.dyeBeanWeight`, causing every subsequent shot to be saved with `doseWeight=0`. Fixed with `> 0` guards on all four numeric DYE fields (dose, drink weight, TDS, EY).
- **Dose pre-fill**: when the shot has no stored dose, `editDoseWeight` now defaults to `Settings.dyeBeanWeight` so EY can be computed immediately when the TDS reading arrives.

## Test plan

- [ ] Open a shot in PostShotReview — confirm date is inline, QualityBadges appear in header when `profileKbId` is set, sparkle button opens KB dialog
- [ ] With R2 configured: confirm Read TDS button appears at top in both basic and advanced mode
- [ ] With R2 disconnected: tap "R2 Off" — confirm it triggers a fresh scan and reconnects
- [ ] Read TDS with R2 — confirm EY is computed immediately (even if dose wasn't previously stored)
- [ ] Confirm rating bar now fills full width
- [ ] Confirm profile name + date visible in bottom bar while scrolling
- [ ] On a shot from a machine with the skip-first-frame FW bug: confirm red "First step skipped" badge appears
- [ ] Type `skipframe:yes` in shot history search — confirm only flagged shots appear; confirm `skipframe:yes` is visible in the search help popover

### DIFF
--- a/qml/components/QualityBadges.qml
+++ b/qml/components/QualityBadges.qml
@@ -13,6 +13,7 @@ Item {
     required property bool channelingDetected
     required property bool temperatureUnstable
     required property bool grindIssueDetected
+    required property bool skipFirstFrameDetected
 
     signal summaryRequested()
 
@@ -129,9 +130,44 @@ Item {
             }
         }
 
+        // Skip-first-frame badge (red) — DE1 firmware bug or very short first step
+        Rectangle {
+            visible: root.skipFirstFrameDetected
+            width: skipFrameRow.width + Theme.spacingMedium * 2
+            height: Theme.scaled(28)
+            radius: Theme.scaled(14)
+            color: Qt.rgba(Theme.errorColor.r, Theme.errorColor.g, Theme.errorColor.b, 0.15)
+            border.color: Theme.errorColor
+            border.width: Theme.scaled(1)
+
+            Accessible.role: Accessible.StaticText
+            Accessible.name: skipFrameText.text
+            Accessible.focusable: true
+
+            Row {
+                id: skipFrameRow
+                anchors.centerIn: parent
+                spacing: Theme.scaled(4)
+                Rectangle {
+                    width: Theme.scaled(8); height: Theme.scaled(8); radius: Theme.scaled(4)
+                    color: Theme.errorColor; anchors.verticalCenter: parent.verticalCenter
+                    Accessible.ignored: true
+                }
+                Tr {
+                    id: skipFrameText
+                    key: "badges.skipFirstFrame"
+                    fallback: "First step skipped"
+                    font: Theme.captionFont
+                    color: Theme.errorColor
+                    anchors.verticalCenter: parent.verticalCenter
+                    Accessible.ignored: true
+                }
+            }
+        }
+
         // Clean extraction badge (green) — only shown when no flags are set
         Rectangle {
-            visible: !root.channelingDetected && !root.temperatureUnstable && !root.grindIssueDetected
+            visible: !root.channelingDetected && !root.temperatureUnstable && !root.grindIssueDetected && !root.skipFirstFrameDetected
             width: cleanRow.width + Theme.spacingMedium * 2
             height: Theme.scaled(28)
             radius: Theme.scaled(14)
--- a/qml/pages/PostShotReviewPage.qml
+++ b/qml/pages/PostShotReviewPage.qml
@@ -227,7 +227,7 @@ Page {
         editGrinderBurrs !== (editShotData.grinderBurrs || "") ||
         editGrinderSetting !== (editShotData.grinderSetting || "") ||
         editBarista !== (editShotData.barista || "") ||
-        editDoseWeight !== (editShotData.doseWeight ?? 0) ||
+        editDoseWeight !== ((editShotData.doseWeight > 0) ? editShotData.doseWeight : Settings.dyeBeanWeight) ||
         editDrinkWeight !== (editShotData.finalWeight ?? 0) ||
         editDrinkTds !== (editShotData.drinkTds ?? 0) ||
         editDrinkEy !== (editShotData.drinkEy ?? 0) ||
--- a/qml/pages/PostShotReviewPage.qml
+++ b/qml/pages/PostShotReviewPage.qml
@@ -270,10 +270,10 @@ Page {
         Settings.dyeGrinderBurrs = editGrinderBurrs
         Settings.dyeGrinderSetting = editGrinderSetting
         Settings.dyeBarista = editBarista
-        Settings.dyeBeanWeight = editDoseWeight
-        Settings.dyeDrinkWeight = editDrinkWeight
-        Settings.dyeDrinkTds = editDrinkTds
-        Settings.dyeDrinkEy = editDrinkEy
+        if (editDoseWeight > 0) Settings.dyeBeanWeight = editDoseWeight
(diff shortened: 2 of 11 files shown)


skialpine:
### Code review

Found 3 issues:

1. **Dead variable `sawRealFrameZero` never read** — the variable is set when `phase.frameNumber == 0` is encountered but is never checked in any conditional; the function returns `true` unconditionally for any non-zero frame before 2 s, making the variable misleading dead code (and a likely compiler warning). Either remove it and simplify the `frameNumber == 0` branch to a plain `continue`, or use it to gate the return if distinguishing the two cases is ever needed.

https://github.com/Kulitorum/Decenza/blob/5c1c091be59657a8cc779d67f8fa6277524c6833/src/ai/shotanalysis.cpp#L157-L172

2. **`skipframe:yes` missing from the search help popover** — the keyword is wired into the parser (line 202) and the strip regex (line 214), but no clickable button or description was added to the help popover alongside `channeling:yes`, `temp:yes`, and `grind:yes` (lines 1292–1334). The syntax hint text on line 1340 also omits it. Users who encounter the "First step skipped" badge have no way to discover the filter keyword through the UI.

https://github.com/Kulitorum/Decenza/blob/5c1c091be59657a8cc779d67f8fa6277524c6833/qml/pages/ShotHistoryPage.qml#L1325-L1341

3. **`skipFirstFrameDetected` not recomputed on-the-fly in `loadShotRecordStatic`** — `grindIssueDetected` (migration 11) has an explicit always-recompute block at lines 2262–2283 with the comment "cover all shot eras … predate migration 11 and thus have grind_issue_detected = 0 (DEFAULT) in the DB." `skipFirstFrameDetected` has the same characteristic (needs only `record.phases`, no derived curves) but the PR only reads it from the DB. Pre-migration-12 shots will show the badge as `false` in the history list until `requestReanalyzeBadges` is triggered by opening the shot. Adding a parallel block after the grind recompute block would fix this.

https://github.com/Kulitorum/Decenza/blob/5c1c091be59657a8cc779d67f8fa6277524c6833/src/history/shothistorystorage.cpp#L2261-L2284

---

skialpine:
### Code review

Found 4 issues:

1. `hasUnsavedChanges` false-positive when shot has no stored dose — `onShotReady` pre-fills `editDoseWeight` from `Settings.dyeBeanWeight` (line 125) when `doseWeight == 0`, but `hasUnsavedChanges` compares against `editShotData.doseWeight ?? 0` (line 230). For any shot with no stored dose, the save button lights up and navigation may prompt a discard-changes dialog immediately after opening the shot — before the user has touched anything. Fix: mirror the same fallback logic in the comparison: `editDoseWeight !== ((editShotData.doseWeight > 0) ? editShotData.doseWeight : Settings.dyeBeanWeight)`.

https://github.com/Kulitorum/Decenza/blob/e2f99c47960c5d249e25c4abda7c5eb599eb6b57/qml/pages/PostShotReviewPage.qml#L123-L131

2. `loadShotRecordStatic` calls `detectSkipFirstFrame` without `expectedFrameCount`, causing false-positive red badges for 1-frame profiles — the other two call sites (save-time and `requestReanalyzeBadges`) correctly pass `expectedFrameCount`. Without it, the `expectedFrameCount < 2` suppression guard never fires in this path, so a single-step profile will always show "First step skipped" when loaded from history, even though `requestReanalyzeBadges` would correctly clear it. Fix: `ShotAnalysis::detectSkipFirstFrame(record.phases, expectedFrameCountFromProfileJson(record.profileJson))`.

https://github.com/Kulitorum/Decenza/blob/e2f99c47960c5d249e25c4abda7c5eb599eb6b57/src/history/shothistorystorage.cpp#L2284-L2291

3. New "First step skipped" red badge has no corresponding text in the Shot Summary dialog — `generateSummary()` analyzes channeling, flow, grind, and temperature but has no branch for skip-first-frame. A user who sees the red badge and taps the shot analysis button to understand why gets a dialog with no mention of the issue, potentially showing "Clean shot" verdict. Fix: thread `skipFirstFrameDetected` (or re-run `detectSkipFirstFrame`) into `generateSummary` and append an observation when it's true.

https://github.com/Kulitorum/Decenza/blob/e2f99c47960c5d249e25c4abda7c5eb599eb6b57/src/ai/shotanalysis.cpp#L183-L186

4. Test suite missing `expectedFrameCount == 0` boundary case — `detectSkipFirstFrame` suppresses detection when `expectedFrameCount >= 0 && expectedFrameCount < 2`, but `tst_shotanalysis.cpp` only tests `expectedFrameCount = 1` (suppressed) and `= 3` (active). The `expectedFrameCount == 0` case (no frames in profile at all) is never tested, leaving a gap in the boundary coverage for the suppression guard.

https://github.com/Kulitorum/Decenza/blob/e2f99c47960c5d249e25c4abda7c5eb599eb6b57/tests/tst_shotanalysis.cpp#L51-L58

---

<details>

- [Worth a look] Test missing the primary FW bug scenario — no test case where the first phase marker is `frameNumber = 1` with no frame-0 entry at all (the actual DE1 firmware bug requiring power-cycle); all existing tests include a "Start" marker at frame 0 (score: 50)

</details>

PR #725, opened 2026-04-12
fix: keep scale connected when app backgrounded, fix flow cal feedback loop

skialpine:
## Summary
- **Scale stays connected when backgrounding the app.** Previously `scale.sleep()` was called on app suspend, powering off the scale and dropping BLE. Now scale sleep is only tied to the machine sleeping. On Android the scale BLE stays alive while the DE1 foreground service is running (wake lock); on iOS the `bluetooth-central` background mode keeps CoreBluetooth alive. If the OS does drop the connection, the reconnect-on-resume path handles it.
- **Fixed auto flow calibration feedback loop for flow profiles.** The DE1's PID locks reported flow to the target regardless of calibration factor, so the old formula `ideal = factor × weightFlow / (reportedFlow × density)` made ideal proportional to the current factor — it could only decrease, never converge. Over 30 shots the factor drifted from 1.0 down to 0.59. Fixed by using the profile's target flow directly: `ideal = weightFlow / (targetFlow × density)`. Only extraction frames are checked (preinfusion frames are skipped since they are always flow-controlled even in pressure profiles). Pressure profiles keep the existing formula (which works correctly since the machine doesn't control flow).
- **One-time migration resets all calibrations** so users don't have to wait for the EMA to slowly recover from drifted values.
- **Updated `docs/AUTO_FLOW_CALIBRATION.md`** with the v3 algorithm: flow vs pressure profile distinction, both formulas, feedback loop explanation, and v3 migration section.

## Test plan
- [ ] Background the app on Android (switch to Claude), confirm scale stays connected when returning (DE1 must be connected for wake lock)
- [ ] Background the app on iOS, confirm scale stays connected
- [ ] Run an espresso with a flow profile (D-Flow) and check debug log for `Auto flow cal: flow profile — using target flow`
- [ ] Verify the ideal factor is reasonable (~0.9-1.1) and doesn't systematically decrease over multiple shots
- [ ] Run an espresso with a pressure profile (e.g., Classic Italian with flow preinfusion) and verify it is NOT classified as a flow profile (`mode: pressure` in log)
- [ ] Confirm calibration values are reset to 1.0 on first launch after upgrade (check for `v3 flow profile feedback loop fix` in log)

### DIFF
--- a/docs/AUTO_FLOW_CALIBRATION.md
+++ b/docs/AUTO_FLOW_CALIBRATION.md
@@ -19,10 +19,13 @@ Automatic per-profile flow calibration using scale data as ground truth. After e
    - Scale data is recent (nearest weight flow point within 1 second)
    - Per-sample machine/weight flow ratio is within [0.4, 2.5] (rejects scale data glitches)
    - Window lasts at least 1.5 seconds with at least 7 samples
-3. **Ratio guard**: Window-mean machine/weight flow ratio must be within [0.75, 1.35] — rejects windows where scale data is systematically unreliable
-4. **Compute ratio**: `current_multiplier * mean(weight_flow) / (mean(machine_flow) * 0.963)` over the steady window
-5. **Sanity check**: Clamp to [0.5, 1.8] — cap extreme values that likely indicate measurement errors
-6. **Store & apply**: If the computed value differs from current by > 2%, save it per-profile and send to the machine
+3. **Profile classification**: Checks whether the profile's extraction frames (post-preinfusion) use flow control or pressure control — this determines which formula is used
+4. **Ratio guard**: Rejects windows where flow and weight diverge too much ([0.75, 1.35]). For flow profiles, compares `(target_flow * 0.963) / weight_flow`. For pressure profiles, compares `machine_flow / weight_flow`.
+5. **Compute calibration** (formula depends on profile type):
+   - **Flow profiles**: `mean(weight_flow) / (target_flow * 0.963)` — uses the profile's known target flow, independent of current calibration
+   - **Pressure profiles**: `current_multiplier * mean(weight_flow) / (mean(machine_flow) * 0.963)` — divides out current calibration from reported flow
+6. **Sanity check**: Clamp to [0.5, 1.8] — cap extreme values that likely indicate measurement errors
+7. **Store & apply**: If the computed value differs from current by > 2%, EMA-smooth toward it (alpha=0.3) and send to the machine
 
 ## Algorithm Details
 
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Found 4 issues:

1. **v3 migration resets pressure-profile calibrations that were correctly converged.** The feedback loop bug only affected flow-controlled profiles, but the migration unconditionally wipes all per-profile calibrations including pressure profiles. Users running only pressure profiles will lose their correctly-converged values (~5-7 shots per profile to re-converge) for no benefit. A targeted reset that only clears flow-profile calibrations would preserve pressure-profile work, though the global median contamination makes the broad reset defensible.

https://github.com/Kulitorum/Decenza/blob/0eaa9b9c4db98b3ae019513a219679e65ed96a6e/src/core/settings.cpp#L193-L206

2. **`docs/AUTO_FLOW_CALIBRATION.md` is outdated.** Still describes the v2 formula with no mention of the flow-profile vs pressure-profile distinction. The algorithm now has two separate code paths depending on profile type, but the docs only show the old formula. Future developers will be misled.

3. **Stale outer comment in `main.cpp`.** Line 1857 says "manage scale when app is suspended/resumed" but the handler no longer manages the scale on suspend -- that was the whole point of this PR. Should be updated to reflect the new scope (DE1 reconnect, accessibility, charger, etc.).

https://github.com/Kulitorum/Decenza/blob/0eaa9b9c4db98b3ae019513a219679e65ed96a6e/src/main.cpp#L1856-L1858

4. **Android scale-connected comment overstates platform protection.** The comment implies the scale stays connected on Android because of the foreground service wake lock, but `BleConnectionService` only runs when the DE1 is connected. If only a scale is connected (no DE1), the event loop can freeze and heartbeats stop. The reconnect-on-resume code handles this gracefully, but the comment should note the DE1 dependency.

https://github.com/Kulitorum/Decenza/blob/0eaa9b9c4db98b3ae019513a219679e65ed96a6e/src/main.cpp#L1875-L1880

Note: Two issues found during review (preinfusion frame misclassification and dead `flowSum` variable) were already fixed in a follow-up commit.

---

<details>
<summary>2 below-threshold issues (50-72)</summary>

- [OK to skip] Per-sample ratio guard comment (line 705) says "the tighter window-level check below catches systematic issues" but for flow profiles the window check uses a different formula entirely -- minor comment clarity issue (score: 72)
- [OK to skip] Flow profile ratio guard comment doesn't explicitly state the ratio direction (target*density / weightFlow) -- readers can verify from line 803 and debug output (score: 50)

</details>

<sub>If this code review was useful, please react with :thumbsup:. Otherwise, react with :thumbsdown:.</sub>

PR #726, opened 2026-04-12
fix: keep Qt event loop running on Android when backgrounded

skialpine:
## Summary

- Adds `android.app.background_running=true` meta-data to the Decenza activity so the Qt event loop continues processing while the app is in the background
- Fixes the MCP server (and all QTimers, BLE heartbeats) freezing when the user switches away from Decenza — e.g., to the Claude app for remote DE1 control
- Eliminates the session-storm on resume (log showed 5+ new MCP sessions created in 43ms when clients reconnected after a 5-minute background period)

## Background

Diagnosing the on-device log showed a 310-second gap with zero event loop activity between suspend and resume — no memory snapshots, no BLE heartbeats, no HTTP requests served. Qt for Android's default is to freeze the main event loop on `ApplicationSuspended` because the EGL surface is gone and painting would crash. The `background_running` meta-data opts out of that freeze; the render thread's missing-EGL-surface warnings are already non-fatal in our log.

The BLE foreground service + wake lock keeps the *process* alive (so DE1/scale connections survive), but without this flag the Qt event loop itself is paused, making the HTTP/MCP server unreachable.

## iOS

iOS has no equivalent flag. `bluetooth-central` (already present) only wakes the app for BLE events, not arbitrary TCP. The realistic workarounds there are split view or an in-app web view — out of scope for this PR.

## Test plan

- [ ] Build Android release, connect DE1
- [ ] Connect Claude Code MCP client, issue a few tool calls
- [ ] Switch to Claude app (or any other app) for >60 seconds
- [ ] Verify log shows continued memory snapshots, BLE heartbeats, and MCP requests being served while backgrounded
- [ ] Verify returning to Decenza does not produce a burst of "Session not found, auto-creating new session" entries
- [ ] Sanity: no new crashes or EGL-related SIGABRTs

### DIFF
--- a/android/AndroidManifest.xml.in
+++ b/android/AndroidManifest.xml.in
@@ -55,6 +55,12 @@
             </intent-filter>
 
             <meta-data android:name="android.app.lib_name" android:value="Decenza" />
+            <!-- Keep Qt event loop running when app is backgrounded. Without this,
+                 Qt freezes the event loop when the activity is suspended (no EGL surface),
+                 which kills the HTTP/MCP server. Users switch to Claude for remote control
+                 and need the server to stay responsive. The render thread safely handles
+                 the missing EGL surface (non-fatal "makeCurrent: no EGLSurface" warnings). -->
+            <meta-data android:name="android.app.background_running" android:value="true"/>
             <meta-data
                 android:name="android.hardware.usb.action.USB_DEVICE_ATTACHED"
                 android:resource="@xml/device_filter" />
--- a/android/AndroidManifest.xml.in
+++ b/android/AndroidManifest.xml.in
@@ -58,8 +58,14 @@
             <!-- Keep Qt event loop running when app is backgrounded. Without this,
                  Qt freezes the event loop when the activity is suspended (no EGL surface),
                  which kills the HTTP/MCP server. Users switch to Claude for remote control
-                 and need the server to stay responsive. The render thread safely handles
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

Found 1 issue:

1. The added comment claims "The render thread safely handles the missing EGL surface (non-fatal 'makeCurrent: no EGLSurface' warnings)", but the safety is not passive — `src/main.cpp` at `Qt::ApplicationSuspended` actively calls `QAccessible::setActive(false)` to prevent a SIGABRT deadlock between the accessibility bridge and the render thread's `eglSurface()` swap. Without that defensive call the render thread would crash, not "safely handle" the missing surface. A reader who later considers removing either the manifest flag or the accessibility disable would be misled into thinking they are independent. Suggest rewording to mention the `QAccessible::setActive(false)` guard.

https://github.com/Kulitorum/Decenza/blob/314cdfd08610112d7ca44081866300e7c82f8af1/android/AndroidManifest.xml.in#L57-L64

<details>

- [OK to skip] Wake lock comment in `BleConnectionService.java` becomes slightly incomplete (doesn't mention activity backgrounding) — file is unmodified; comment is still factually correct, just less complete (score: 35)
- [OK to skip] Alleged wake lock reference leak in `BleConnectionService.java` from prior-PR review — false positive, the `if (m_wakeLock == null)` guard at line 64 already prevents double-acquire (score: 0)

</details>

PR #729, opened 2026-04-12
fix: scope the preset "modified" indicator and track beans-modified state

### LINKED ISSUE #728: (minor) UI issue: marking profiles as modified marks flush/steam/hot water, not just espresso
When I make a profile change (e.g. brew temp, which I do regularly), and do not save the updated profile, the profile gets marked as modified (with asterisk or "modified"). That's fine. But the current flush/steam/hot water settings also get marked as modified, which is misleading (these have not been modified).

Low priority request: in an upcoming release, please restrict marking profile as modified to the espresso button, and do not mark other operations (flush/steam/hot water). TIA!

skialpine:
## Summary
Two related fixes so that the preset pill \"modified\" indicator behaves correctly for every preset type:

1. **Stop the espresso dirty flag from leaking onto other rows.** `PresetPillRow` is shared across Espresso, Steam, HotWater, Flush, and Beans rows, but read `ProfileManager.profileModified` unconditionally — so editing an espresso profile (e.g. brew temp) tagged the selected steam/flush/hot-water preset as modified too.

2. **Actually track modified state for beans.** Previously there was no equivalent for beans. If you modified bean/grinder fields in the Beans window and came back without saving (or used \"Use as entered\"), the live DYE values could diverge from the selected preset but the pill never showed it. Adds `Settings::beansModified` that compares current DYE fields to the selected preset.

### API change on `PresetPillRow`
Generalized the opt-in so any preset row can expose its own dirty state:
- `property bool modified` (bindable)
- `property bool modifiedIsReadOnly` — prefix `*` vs. suffix `(modified)`

Callers:
- Espresso → `modified: ProfileManager.profileModified; modifiedIsReadOnly: ProfileManager.isCurrentProfileReadOnly`
- Beans → `modified: Settings.beansModified`
- Steam / Hot Water / Flush → unset (default `false`)

Fixes #728

## Test plan
- [ ] Change brew temp without saving → only selected espresso pill shows `*Name` / `(modified)` (read-only); Steam, HotWater, Flush, Beans pills are clean
- [ ] Select a bean preset, then open beans window, change grinder setting, back out without saving → selected beans pill shows `*Name`
- [ ] Apply beans preset fully (fields match) → beans pill clean
- [ ] \"Use as entered\" clears `selectedBeanPreset` to -1 → no beans pill is selected, so no asterisk (expected)
- [ ] TalkBack: swipe to the dirty pill announces \"unsaved changes\" (espresso and beans); other preset rows do not

### DIFF
--- a/qml/components/PresetPillRow.qml
+++ b/qml/components/PresetPillRow.qml
@@ -13,6 +13,9 @@ FocusScope {
     property var pillSuffixFn: null  // Optional: function(index) => string suffix appended to pill text (e.g. " (125g)")
     property int pillSuffixVersion: 0  // Increment from outside to force pill text refresh without full layout recalc
     property real pillSuffixMaxWidth: 0  // Reserve extra horizontal space per pill for the suffix
+    property bool showProfileModified: false  // Only the espresso row reflects ProfileManager.profileModified;
+                                               // steam/flush/hot-water/bean rows share this component but are
+                                               // unrelated to the espresso profile's dirty state.
 
     // Effective max width - ensures we never exceed parent width even if maxWidth is larger
     readonly property real effectiveMaxWidth: {
@@ -53,7 +56,7 @@ FocusScope {
     function announceCurrentPill() {
         if (typeof AccessibilityManager !== "undefined" && AccessibilityManager.enabled && presets.length > 0) {
             var name = presets[focusedIndex].name || ""
-            var modified = (focusedIndex === selectedIndex && ProfileManager.profileModified) ? ", " + TranslationManager.translate("presets.unsaved", "unsaved changes") : ""
+            var modified = (showProfileModified && focusedIndex === selectedIndex && ProfileManager.profileModified) ? ", " + TranslationManager.translate("presets.unsaved", "unsaved changes") : ""
             var status = focusedIndex === selectedIndex ? ", " + TranslationManager.translate("presets.selected", "selected") : ""
             AccessibilityManager.announce(name + modified + status)
         }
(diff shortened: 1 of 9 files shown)


skialpine:
### Code review

Found 2 issues:

1. Hardcoded user-visible `" (modified)"` string in the pill label is not internationalized. The accessibility string on line 63 correctly uses `TranslationManager.translate("presets.unsaved", "unsaved changes")`, but the visual path here does not. CLAUDE.md says "All user-visible text must be internationalized" and "Rules for Modifying Existing Components: fix pre-existing bugs and violations in the file you're modifying — do not dismiss them as pre-existing."

https://github.com/Kulitorum/Decenza/blob/9ee8725b18c74e47cba0652cf37e80a6725578c0/qml/components/PresetPillRow.qml#L71-L76

2. Pre-existing `int index < arr.size()` narrowing comparisons in the bean-preset helpers (also at `settings.cpp:1535`, `:1541`, `:1579`). CLAUDE.md C++ section: "Use `qsizetype` (not `int`) for container sizes — `QVector::size()`, `QList::size()`, `QString::size()` etc. return `qsizetype` (64-bit on iOS/macOS). Assigning to `int` causes `-Wshorten-64-to-32` warnings." The pattern was partly cleaned up in PR #677 for steam helpers but the bean helpers still use raw `int`.

https://github.com/Kulitorum/Decenza/blob/9ee8725b18c74e47cba0652cf37e80a6725578c0/src/core/settings.cpp#L1506-L1511

<details>
<summary>3 below-threshold issues (50)</summary>

- [OK to skip] Selecting a bean preset triggers ~10 synchronous `recomputeBeansModified` passes (QML sets `selectedBeanPreset` first, then `applyBeanPreset` writes 9 DYE fields one by one). Final state is correct and `PresetPillRow`'s 1 ms recalc timer coalesces the visual update — signal churn only. (score: 50)
- [Worth a look] `saveBeanPresetFromCurrent` doesn't update `selectedBeanPreset` to the new/updated index. Pre-existing, but the new `beansModified` indicator makes the gap more visible: after a "save as new" the pill selection can still point at the old preset. (score: 50)
- [Worth a look] `BeansItem.qml` compact-mode `presetPopup` opens without setting `activePresetFunction = "beans"` or calling `AccessibilityManager.announce(...)` — TalkBack users get no preset-list announcement. Same pattern was flagged for `SteamItem` in PR #677. (score: 50)

</details>

PR #731, opened 2026-04-12
feat: flow-cal cap bump + firmware-gated auto-cal ceiling

skialpine:
## Summary
Two coordinated changes motivated by DE1 firmware v1337 raising the machine-side flow-calibration cap from 2.0 to 3.0 (to match newer pump hardware).

### Manual cap
Bumped the user-controlled multiplier cap 2.0 → 3.0 in all three sites:
- `Settings::setFlowCalibrationMultiplier` `qBound` upper
- `FlowCalibrationModel::setMultiplier` `qBound` upper
- `FlowCalibrationPage` \"+0.01\" enable threshold + `Math.min` + slider `to`

### Auto-cal ceiling (firmware-gated)
Auto-cal keeps ~10% headroom below the firmware cap, mirroring the old 2.0→1.8 pair (0.9× ratio):
- Pre-v1337 firmware: `kCalibrationMax = 1.8` (unchanged)
- v1337+ firmware: `kCalibrationMax = 2.7`

`DE1Device::firmwareBuildNumber()` is now exposed so `MainController::computeAutoFlowCalibration` can branch at runtime.

### Warning (option 3 from the ceiling discussion)
- `qWarning` logged whenever the post-clamp computed value exceeds the classic 1.8 ceiling (scale-bias telemetry).
- `ProfileInfoPage` colors the multiplier in `Theme.warningColor` and sets `Accessible.description` to \"Unusually high — verify scale accuracy\" when > 1.8. No Unicode icon — per CLAUDE.md.

### Persistence loosening
`Settings::setProfileFlowCalibration` and `SettingsSerializer` both accept `[0.5, 2.7]` now. Persistence just prevents corrupt values; the firmware-aware clamp at compute time is the correctness gate. This also keeps values computed on v1337+ readable if a user later downgrades firmware.

The window-ratio guard `[0.75, 1.35]` stays as the primary protection against bad scale data.

## Test plan
- [ ] Old firmware (< v1337): auto-cal still clamps per-profile values at 1.8; slider/UI still allows 3.0 manually.
- [ ] New firmware (≥ v1337): auto-cal can converge up to 2.7; values > 1.8 appear amber in Profile Info with \"Unusually high — verify scale accuracy\" TalkBack hint; `qWarning` fires.
- [ ] Import a settings backup containing per-profile cal values > 1.8 and ≤ 2.7 — accepted; values > 2.7 rejected.
- [ ] No reference to the new-machine codename in any commit message, comment, or doc.

Identified via `/de1app_review` 60-day scan. Reference: de1app commit `d2b0180c`.

### DIFF
--- a/docs/AUTO_FLOW_CALIBRATION.md
+++ b/docs/AUTO_FLOW_CALIBRATION.md
@@ -24,7 +24,7 @@ Automatic per-profile flow calibration using scale data as ground truth. After e
 5. **Compute calibration** (formula depends on profile type):
    - **Flow profiles**: `mean(weight_flow) / (target_flow * 0.963)` — uses the profile's known target flow, independent of current calibration
    - **Pressure profiles**: `current_multiplier * mean(weight_flow) / (mean(machine_flow) * 0.963)` — divides out current calibration from reported flow
-6. **Sanity check**: Clamp to [0.5, 1.8] — cap extreme values that likely indicate measurement errors
+6. **Sanity check**: Clamp to `[0.5, kCalibrationMax]` — cap extreme values that likely indicate measurement errors. The upper bound tracks DE1 firmware (1.8 on pre-v1337 firmware, 2.7 on v1337+)
 7. **Store & apply**: If the computed value differs from current by > 2%, EMA-smooth toward it (alpha=0.3) and send to the machine
 
 ## Algorithm Details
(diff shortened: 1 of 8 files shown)


skialpine:
### Code review

Found 3 issues:

1. `Accessible.description` on the flow-cal `Text` element is silently dropped because the element has no `Accessible.role` / `Accessible.name` / `Accessible.focusable`. TalkBack and VoiceOver only surface `Accessible.description` as a hint on elements they can discover (via role + name) — so the "Unusually high — verify scale accuracy" warning is effectively color-only, which fails screen-reader and color-blind users. CLAUDE.md says static text needs `Accessible.role: Accessible.StaticText` + `name` to be discoverable. Fix: fold the warning into `Accessible.name` (pattern used by `SteamPage.qml`'s warning banner).

https://github.com/Kulitorum/Decenza/blob/d04db175007160a31ab9e7ac5e121848c576523c/qml/pages/ProfileInfoPage.qml#L211-L228

2. Pre-existing `int` loop / cursor variables compared against `QVector::size()` (returns `qsizetype`) throughout `computeAutoFlowCalibration` — `-Wshorten-64-to-32` on iOS/macOS. CLAUDE.md: "Use `qsizetype` (not `int`) for container sizes... Rules for Modifying Existing Components: fix pre-existing bugs and violations in the file you're modifying — do not dismiss them as pre-existing." The PR modifies this function (adding firmware-gated `kCalibrationMax`, added a warn), triggering the rule. Affected lines: 621, 625, 644-645, 648, 673, 691.

https://github.com/Kulitorum/Decenza/blob/d04db175007160a31ab9e7ac5e121848c576523c/src/controllers/maincontroller.cpp#L646-L692

3. EMA-smoothed `computed` is not re-clamped before `setProfileFlowCalibration`, so persistence silently rejects valid auto-cal outputs. On v1337+ firmware `kCalibrationMax = 2.7`; if a user manually sets the global to 3.0 and a profile has no per-profile value yet, `currentEffective = 3.0`. EMA: `0.3*2.7 + 0.7*3.0 = 2.91`. `setProfileFlowCalibration` rejects values > 2.7 and logs a `qWarning` — auto-cal writes nothing, silently. One-line fix: `computed = qBound(kCalibrationMin, computed, kCalibrationMax);` before line 883.

https://github.com/Kulitorum/Decenza/blob/d04db175007160a31ab9e7ac5e121848c576523c/src/controllers/maincontroller.cpp#L876-L884

<details>

- [Worth a look] Pre-existing v3 migration at [`settings.cpp:201-206`](https://github.com/Kulitorum/Decenza/blob/d04db175007160a31ab9e7ac5e121848c576523c/src/core/settings.cpp#L201-L206) unconditionally wipes all per-profile flow calibrations; the underlying bug only affected flow-control profiles, so pressure-profile values that had correctly converged are also lost. Gated one-time migration, small affected cohort now, but PR #731 touches this file. (score: 70)
- [Worth a look] `DE1Device::setFlowCalibrationMultiplier` at [`de1device.cpp:937-940`](https://github.com/Kulitorum/Decenza/blob/d04db175007160a31ab9e7ac5e121848c576523c/src/ble/de1device.cpp#L935-L941) writes the multiplier to MMR without a firmware-based clamp. A stored 2.5 from v1337+ applied to an older-firmware DE1 (hard cap 2.0) gets silently clamped by firmware while the UI still shows 2.5. Rare multi-machine case; UI/machine mismatch, no data loss. (score: 65)

</details>

PR #733, opened 2026-04-12
fix: Android fresh install stuck on 'Bluetooth powered off' (#732)

### LINKED ISSUE #732: Reinstalled Build: 3267 -No connection
I deleted Decenza for a fresh install of Build: 3267. Result: no devices found. 

Rebooted tablet (Samsung A9+), closed and opened Decenza several times. Still no connection. Stock app connects just fine. Log attached. 

[debug-1.zip](https://github.com/user-attachments/files/26662771/debug-1.zip)

skialpine:
## Summary
- Fixes Android 12+ fresh-install case where Decenza reports "Bluetooth is powered off" and never prompts for the runtime Bluetooth permission, blocking all device discovery.
- Regression introduced in v1.6.5 by #695, which added an `isBluetoothAvailable()` guard that runs *before* `requestBluetoothPermission()`. On Android, `QBluetoothLocalDevice::hostMode()` returns `HostPoweredOff` until `BLUETOOTH_CONNECT` is granted, so the guard silently short-circuited the permission prompt.
- On Android, treat the adapter as available while permission status is not yet `Granted` so the scan flow reaches `requestBluetoothPermission()` and prompts the user. Once granted the real `hostMode()` check runs normally.

## Evidence from the reporter's log (#732)
```
WARN  Permissions not authorized ... QBluetoothPermission::CommunicationMode(Access)
WARN  Local device initialize() failed due to missing permissions
DEBUG BLEManager: scanForDevices - Bluetooth is powered off, skipping
```
No "Requesting Bluetooth permission…" log ever appears on build 3267. Earlier sessions in the same log worked only because permissions were carried over from a prior install — after the uninstall/reinstall Android revoked them and there was no way to re-grant.

## Test plan
- [ ] `adb uninstall io.github.kulitorum.decenza_de1` on Android 12+, install this build, tap Scan for Devices → permission prompt appears, grant, devices discovered.
- [ ] Deny the Bluetooth permission → error surfaces, app doesn't hang.
- [ ] With Bluetooth actually turned off in system settings, tap scan → "Bluetooth is powered off" is reported (real adapter-off guard still works post-grant).
- [ ] Existing install with permission already granted still scans without re-prompting.
- [ ] iOS unaffected (Q_OS_IOS path unchanged).

Fixes #732

### DIFF
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -61,10 +61,13 @@ bool BLEManager::isBluetoothAvailable() const
     // install, permission is Undetermined — if we bail out here, the scan flow never
     // gets a chance to call requestBluetoothPermission(), and the user is stuck with
     // "Bluetooth is powered off" even though the adapter is fine. Report available
-    // while permission is unknown so the scan proceeds and prompts the user.
+    // only while the status is Undetermined so the scan proceeds and prompts the
+    // user. If the user has explicitly denied, fall through to the real hostMode()
+    // check (which will report PoweredOff) so bluetoothAvailable accurately reflects
+    // that BLE is unusable.
     QBluetoothPermission perm;
     perm.setCommunicationModes(QBluetoothPermission::Access);
-    if (qApp->checkPermission(perm) != Qt::PermissionStatus::Granted) {
+    if (qApp->checkPermission(perm) == Qt::PermissionStatus::Undetermined) {
         return true;
     }
 #endif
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 1 issue:

1. `isBluetoothAvailable()` returns `true` for `Denied` as well as `Undetermined`. The PR's rationale (and the comment) cover only the fresh-install/`Undetermined` case, but the condition `!= Granted` also catches `Denied`. For a user who has explicitly denied the permission, this makes the `bluetoothAvailable` Q_PROPERTY report available (bound in `qml/pages/settings/SettingsConnectionsTab.qml`), and `tryDirectConnectToDE1()` will pass the guard at line 631 and emit `de1Discovered` before `requestBluetoothPermission()` surfaces the Denied error. Suggest tightening to `== Undetermined`:

```cpp
if (qApp->checkPermission(perm) == Qt::PermissionStatus::Undetermined) {
    return true;
}
```

https://github.com/Kulitorum/Decenza/blob/c185ff15e93f5c368e81b60a61568ac398deb027/src/ble/blemanager.cpp#L58-L72

<details>

- [Worth a look] `bluetoothAvailableChanged` is never emitted when permission transitions Undetermined→Granted inside the `requestPermission` callback, so QML bindings on `bluetoothAvailable` can remain stale until the adapter fires `hostModeStateChanged` — one-line fix to emit the signal from the Granted callback (score: 72)
- [OK to skip] The new comment describes only the Undetermined motivation, not the Denied arm — stylistic; becomes moot if the condition is tightened to `== Undetermined` (score: 50)

</details>

Issue #739, opened 2026-04-14
Monitor: v3 auto flow calibration convergence (D-Flow/Q, target ~0.89)


skialpine:
## Background

The v3 auto flow calibration algorithm was deployed on **2026-04-13** (commit [089bae2d](https://github.com/Kulitorum/Decenza/commit/089bae2d8ace5a7736d62372746575b71086b1f5)). It fixes a feedback loop in v2 for flow-controlled profiles: the DE1's PID holds reported flow at target regardless of calibration, so the v2 formula (`current × weightFlow / (reportedFlow × density)`) was proportional to the current factor and could only drift down.

**v3 formula for flow profiles:**
```
ideal = mean(weight_flow) / (target_flow × 0.963)
```

A migration reset all per-profile calibrations and the global multiplier to 1.0 on first launch.

## Expected convergence

Based on prior analysis of 26 D-Flow shots on a GFC-calibrated machine, the correct calibration for this machine was expected to converge near **~0.89** (doc mentions ~0.95 as the reference value on that machine).

## Observed shots since v3 (as of 2026-04-14)

| Shot | Date | C_before | meanMachineFlow | meanWeightFlow | Window | Ideal | C_after | Notes |
|------|------|----------|----------------|----------------|--------|-------|---------|-------|
| 846 | Apr 13 09:11 | 1.000 (reset) | 1.827 ml/s | 1.392 g/s | 19.2s / 93 samples | 0.803 | 0.803 | First v3 shot, grinder @11 |
| 847 | Apr 13 09:56 | 0.803 | **1.439 ml/s** | **1.743 g/s** | 5.0s / 25 samples | **1.006** | 0.864 | ⚠ Anomalous — machine flow well below target 1.8, scale > machine flow; likely puck releasing stored water. Short window. |
| 848 | Apr 14 07:14 | 0.864 | 1.824 ml/s | 1.337 g/s | 21.2s / 103 samples | 0.771 | 0.836 | Grinder @10 (coarser than 846) |

**Current multiplier: 0.836** (as of 2026-04-14)

## Concern

The two non-anomalous shots compute ideals of **0.803** and **0.771**, suggesting the algorithm may be converging toward ~0.77–0.80 rather than the expected ~0.89. However:

- Only 3 shots of v3 data exist — far too few to determine convergence (EMA α=0.3 needs ~10–15 shots from a cold start)
[...]

PR #742, opened 2026-04-14
cal: batched median updates + pressure smoothing for auto flow cal

skialpine:
## Summary
- **3-sample MA pressure smoothing**: Filters PID jitter in dpdt computation, increasing average steady window from 10.7s to 16.3s across 13 D-Flow shots
- **Batched median accumulator**: Collects 5 ideals at constant C before updating, breaking the pump-behavior feedback loop and providing natural outlier rejection
- **3% change threshold**: Only updates calibration when the shift is meaningful, reducing unnecessary pump changes
- Moves `AUTO_FLOW_CALIBRATION.md` into `docs/CLAUDE_MD/` and references it from `CLAUDE.md`

## Context

Analysis in #739 found two issues with the v3 auto flow calibration:

1. **dpdt threshold too strict for flow profiles**: The DE1's PID causes 0.1-0.2 bar corrections every ~0.2s that exceed the 0.5 bar/s threshold, producing artificially short windows (2.5-5s instead of 18-21s). A 3-sample moving average on pressure filters this jitter while preserving genuine transitions.

2. **Per-shot EMA creates feedback loop**: Each C update changes pump behavior → changes puck dynamics → changes weight flow → changes the next ideal. Simulation across 11 same-bean/same-grind shots showed neither the current algorithm nor smoothing alone produces convergence. Batching 5 shots at constant C breaks the loop — the pump stays stable for 5 shots, producing truly comparable data, then the median update rejects outliers automatically.

Simulation results (Pumphouse Beach Entry @ grinder 11, 11 shots):
| Strategy | Updates | Final C | Distance from true mean (0.815) |
|----------|---------|---------|-------------------------------|
| Current (per-shot EMA α=0.3) | 11 | 0.839 | 0.024 |
| **Batched N=5, median, α=0.5** | **3** | **0.836** | **0.021** |

## Test plan
- [ ] Pull 5+ D-Flow shots with BLE scale — verify batch accumulation logs (`Auto flow cal: accumulated ideal ... (N/5)`)
- [ ] Verify C updates only after 5th shot with correct median/alpha in log
- [ ] Verify clearing calibration (MCP `clear_flow_calibration`) also clears pending batch
- [ ] Verify first calibration for a new profile uses median directly (no EMA blend)
- [ ] Monitor convergence per #739

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -144,6 +144,7 @@ Detailed documentation lives in `docs/CLAUDE_MD/`. Read these when working in th
 | `CUP_FILL_VIEW.md` | CupFillView layer stack, GPU shaders, updating cup images |
 | `EMOJI_SYSTEM.md` | Twemoji SVG rendering, adding/switching emoji sets |
 | `ACCESSIBILITY.md` | TalkBack/VoiceOver rules, focus order, anti-patterns, implementation plan |
+| `AUTO_FLOW_CALIBRATION.md` | Auto flow calibration algorithm, batched median updates, windowing, convergence |
 
 Also in `docs/`:
 - `MCP_SERVER.md` — full MCP tool list, access levels, architecture
--- a/docs/CLAUDE_MD/AUTO_FLOW_CALIBRATION.md
+++ b/docs/CLAUDE_MD/AUTO_FLOW_CALIBRATION.md
@@ -12,7 +12,7 @@ Automatic per-profile flow calibration using scale data as ground truth. After e
 
 1. **Shot completes** with a Bluetooth scale connected
 2. **Steady-state detection**: The algorithm scans the shot data for a window where:
-   - Pressure is stable (change <= 0.5 bar/sec)
+   - Pressure is stable (3-sample smoothed change <= 0.5 bar/sec)
    - Pressure is above 1.5 bar (rejects empty-portafilter shots)
    - Weight flow is meaningful (> 0.5 g/s)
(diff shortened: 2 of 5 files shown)


skialpine:
### Code review

Found 1 issue:

1. `kBatchSize` declared as `constexpr int` but compared against `qsizetype` (CLAUDE.md says "Use `qsizetype` (not `int`) for container sizes — `QVector::size()`, `QList::size()`, `QString::size()` etc. return `qsizetype` (64-bit on iOS/macOS). Assigning to `int` causes `-Wshorten-64-to-32` warnings.")

https://github.com/Kulitorum/Decenza/blob/c7c5dbf9ad9a42e45270be140915e13b23a37c67/src/controllers/maincontroller.cpp#L887-L889

`pending.size()` returns `qsizetype` and is compared against `kBatchSize` on line 900. Commit 94a06898 recently fixed this exact pattern in the same function. Should be `constexpr qsizetype kBatchSize = 5;`.

---

<details>

- [Worth a look] Dead code: alpha-scaling comment (lines 914-917) describes "remainder batches" but `n >= kBatchSize` always at that point, so `qMin` always returns 1.0. Simplify to `double alpha = kBatchEmaAlpha;` or document the actual invariant (score: 65)
- [Worth a look] Stale comment on line 931 says "> 2% change" but threshold was changed to 3% in this PR. The debug message on line 934 correctly says "< 3% change" (score: 50)

</details>

PR #744, opened 2026-04-15
ble: verify profile upload frames via FrameToWrite ACKs

skialpine:
## Summary
- `DE1Device::uploadProfile()` / `uploadProfileAndStartEspresso()` now verify each `FRAME_WRITE` ACK's leading byte against the expected `FrameToWrite` sequence, modelled on de1app's `confirm_de1_send_shot_frames_worked`
- On mismatch, timeout (10 s), disconnect mid-upload, or queue-clear mid-upload, the upload now surfaces as `profileUploaded(false)` with a `qWarning()` that includes expected vs seen frame numbers and the profile title
- Previous behaviour just counted `writeComplete` signals and trusted them — silently-dropped or reordered frames reported success anyway

## Why
Shot 850 ended after 6 s because the DE1 was running a stale single-frame 9-bar setpoint instead of the D-Flow / Q profile we'd uploaded 14 h earlier. The old count-only check said "upload succeeded" so the bug was invisible until the shot failed with no obvious cause. The new verification would have logged exactly what went wrong the moment it happened.

## What a mismatch looks like in the log
\`\`\`
DE1Device: profile upload FAILED — frame sequence mismatch (expected [0x00, 0x01, 0x02, 0x22, 0x03], got [0x00, 0x01, 0x02]). Profile "D-Flow / Q" was likely NOT correctly loaded on the DE1.
\`\`\`

A successful upload now logs:
\`\`\`
DE1Device: profile upload verified — 5 frame(s) ACKed in order for profile "D-Flow / Q"
\`\`\`

## Test plan
- [x] New \`tst_profileupload\` covers: happy path, dropped frame ACK, out-of-order ACKs, unrelated writes ignored, \`uploadProfileAndStartEspresso\` requiring the \`REQUESTED_STATE(Espresso)\` ACK, and disconnect mid-upload reports failure
- [ ] Build + \`ctest\` clean locally
- [ ] Pull a real shot on an actual DE1 and confirm the "upload verified" log line appears; confirm no regressions in normal shot flow

## Docs
- Updated \`docs/CLAUDE_MD/BLE_PROTOCOL.md\` with a new "Profile Upload Frame-ACK Verification" section (tracked state, failure paths, example log lines)

### DIFF
--- a/docs/CLAUDE_MD/BLE_PROTOCOL.md
+++ b/docs/CLAUDE_MD/BLE_PROTOCOL.md
@@ -48,6 +48,36 @@ Qt's `QLowEnergyController::disconnected()` signal only fires on a Connected→D
 
 Symptom if this is broken: DE1 reboot drops BLE, app attempts one reconnect, the attempt fails, then app stays silent forever until restarted. The Scan Devices button also appears dead because the `de1Discovered` handler's `!isConnecting()` guard never clears.
 
+### Profile Upload Frame-ACK Verification
+
+`DE1Device::uploadProfile()` and `uploadProfileAndStartEspresso()` don't just count write completions — they verify that each `FRAME_WRITE` ACK's leading byte (the `FrameToWrite` field) matches the sequence we queued, in order. Modeled on de1app's `confirm_de1_send_shot_frames_worked` in `de1_comms.tcl`.
+
+**Why:** Counting `characteristicWritten` signals alone can falsely report success when frames are silently dropped, reordered, or a stale profile remains loaded on the DE1 (e.g., if the original upload was never re-sent after a connection hiccup). Verifying the frame-number sequence surfaces these cases as real failures instead of mysterious early shot endings.
+
+**Tracked state (per in-flight upload):**
+- `m_uploadProfileTitle` — echoed into success/failure logs so operators can tell which profile the verdict refers to
+- `m_uploadExpectedFrameBytes` — leading byte of every frame we queued (regular: `0..N-1`, extension: `i+32`, tail: `N`)
+- `m_uploadSeenFrameBytes` — leading byte captured from each `FRAME_WRITE` ACK
+- `m_uploadHeaderAcked` / `m_uploadEspressoStartAcked` — one-shot flags
+- `m_uploadExpectEspressoStart` — set true by `uploadProfileAndStartEspresso()` so the tracker also waits for the `REQUESTED_STATE(Espresso)` ACK before calling the upload complete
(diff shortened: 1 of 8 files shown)


skialpine:
### Code review

Found 3 issues:

1. `int` loop variable indexing into a `QList` in the new test — CLAUDE.md says "Use qsizetype (not int) for container sizes — QVector::size(), QList::size(), QString::size() etc. return qsizetype (64-bit on iOS/macOS). Assigning to int causes -Wshorten-64-to-32 warnings." The loop bound is a literal `3` so there's no runtime truncation, but the pattern is exactly what the project rule targets.

https://github.com/Kulitorum/Decenza/blob/5372f5d61ad5b75088d98180742c1fb8b5114043/tests/tst_profileupload.cpp#L202-L207

2. Doc says "BLE disconnect **mid**-upload" but the code emits "BLE disconnect **during** upload" — anyone searching logs or writing log-analysis tooling against the doc-cited string won't match. The test and the `McpTestFixture` warning filter both use the code's exact wording.

https://github.com/Kulitorum/Decenza/blob/5372f5d61ad5b75088d98180742c1fb8b5114043/docs/CLAUDE_MD/BLE_PROTOCOL.md#L63-L68

3. Success-log example in the doc reads `... for profile "D-Flow / Q"` (with quotes) but the emitter uses `.noquote()` with a format string of `for profile %2`, so real log lines read `... for profile D-Flow / Q` (no quotes). Failure case is correct because those quotes are baked into the reason string.

https://github.com/Kulitorum/Decenza/blob/5372f5d61ad5b75088d98180742c1fb8b5114043/src/ble/de1device.cpp#L937-L944

<details>

- [Worth a look] `Tracked state` bullet list in BLE_PROTOCOL.md omits `m_uploadProfileTitle`, `m_uploadExpectEspressoStart`, `m_uploadConnection` (and describes the timer in prose instead of as `m_uploadTimeoutTimer`). Doc completeness only — easy one-line fix while you're in the file (score: 65)
- [Worth a look] `DE1Device::profileUploaded(bool)` signal is declared/emitted but has no production listener; only tests observe it via `QSignalSpy`. A `qWarning` still surfaces the failure in the debug log, but no UI/ProfileManager reaction is wired. If that's intentional for now, fine; otherwise a follow-up PR should connect it (score: 65)
- [OK to skip] Opening comment on `uploadFailsWhenFrameAckIsDropped` describes the technique as a "phantom ACK" but the code actually duplicates the first-frame ACK three times. The clarifying inline comment two lines below is accurate, so test intent is preserved (score: 50)

</details>

PR #745, opened 2026-04-15
ble: recover reconnect loop after failed connect attempt

### LINKED ISSUE #1: AI
How does one initiate AI analysis? I’ve entered an API key for Claude but in my UI of v1.1.22 I find no way to start the analysis and I don’t see it in the visualizer shots

### LINKED ISSUE #2: Decent scale sleep and wakeup
I hope the sleep/wake function of the decent scale also works.

skialpine:
## Summary
Two related fixes for _"DE1 reboot → app never reconnects until restarted, Scan Devices button looks dead"_:

1. **`BleTransport` synthesizes `disconnected()` on Connecting→Unconnected failures.** Qt's `QLowEnergyController::disconnected()` only fires on a Connected→Disconnected transition — a failed connect attempt (e.g., DE1 still booting) leaves the controller in Unconnected without ever firing `disconnected()`. `DE1Device::onTransportDisconnected()` was therefore never called, `m_connecting` stayed `true` forever, and every subsequent reconnect retry plus the `de1Discovered` auto-connect handler bailed out. A per-attempt flag (`m_disconnectedEmittedForAttempt`) prevents double-emission when Qt does fire the native signal.
2. **`main.cpp`'s `connectedChanged` handler edge-tracks via `s_de1WasConnected`** instead of re-arming the retry timer on every spurious same-state emission. `DE1Device::disconnect()` emits `connectedChanged()` unconditionally and the reconnect path calls `disconnect()` on the old transport before creating a new one, so spurious emits were scrambling the backoff schedule (resetting `de1ReconnectAttempt=0` mid-attempt, converting the intended 30 s retry into 60 s).

Together: after a DE1 reboot, the app now cleanly sees the failed connect, schedules the next retry, and the Scan Devices button's discovery→connect path also works because the \`isConnecting()\` guard at main.cpp:967 no longer blocks.

## Evidence from the shot-850 session log
\`\`\`
[51762.385] CONTROLLER ERROR: ConnectionError   ← DE1 rebooted
[51762.399] Controller disconnected              ← live link dropped, Qt fires disconnected()
[51762.443] DE1 reconnect: scheduled first retry in 5000 ms
[51767.212] DE1 reconnect: attempt 1 of 12
[51767.231] DE1 reconnect: scheduled first retry in 5000 ms   ← spurious reset (fix #2)
[51767.265] Controller state: Connecting
[51777.272] CONTROLLER ERROR: ConnectionError   ← 2nd attempt fails
[51777.276] Controller state: Unconnected       ← Qt does NOT emit disconnected() (fix #1)
[51827.195] DE1 reconnect: already connected/connecting, stopping retries   ← m_connecting stuck true
... 25 min of silence until user force-killed the app ...
\`\`\`

## What you'll see in the log after the fix
On a connect-attempt failure:
\`\`\`
[BLE DE1] Controller state: Unconnected
[BLE DE1] Connection attempt failed — synthesizing disconnected()
\`\`\`
…and the next scheduled retry actually fires instead of bailing out.

## Test plan
- [ ] Build + \`ctest\` clean locally
- [ ] Reboot the DE1 mid-session and confirm the app reconnects on its own within the backoff schedule (5 s, then 30 s, then 60 s × 10)
- [ ] Verify the Scan Devices button finds and connects to the DE1 in that same post-reboot window
- [ ] Verify a clean Connected→Disconnected (e.g., turning off the DE1 via its switch) still drives the reconnect loop as before
- [ ] Verify USB-connected DE1 path is untouched (reconnect loop is only armed when BLE was the active transport)

## Docs
- Updated \`docs/CLAUDE_MD/BLE_PROTOCOL.md\` with a new \"Connection-Failure Handling\" section explaining Qt's signal gap and the synthesized-disconnect workaround


skialpine:
### Code review

Found 1 issue:

1. The service-discovery retry path doesn't reset `m_disconnectedEmittedForAttempt`, so a second failed connect on the retry silently skips the synthesized `disconnected()` — re-creating the exact stuck-state this PR fixes, just on a different path. The retry timer handler cleans up the old controller and calls `setupController() + m_controller->connectToDevice()` but never re-arms the flag, so if the previous attempt set it to `true` (either via `onControllerDisconnected` or the synthesized path) and the retry then goes Connecting→Unconnected, the guard at the `stateChanged` lambda drops the emit. The header comment and the new `BLE_PROTOCOL.md` section both describe this as a "per-attempt flag reset on `connectToDevice()`" — which is only true for the outer public `BleTransport::connectToDevice()`, not the internal retry. Fix: add `m_disconnectedEmittedForAttempt = false;` inside the retry timer lambda just before `m_controller->connectToDevice();`, and tighten the comment/doc wording.

https://github.com/Kulitorum/Decenza/blob/ff760d89566bfbb32dad15a3deac53ed780ff0b1/src/ble/bletransport.cpp#L122-L142

<details>

- [Worth a look] `BleTransport::disconnect()` emits `disconnected()` unconditionally at L245 without consulting `m_disconnectedEmittedForAttempt`, so a manual `disconnect()` call after the `stateChanged` synthesis would double-emit. Mitigated by the new `s_de1WasConnected` edge-tracking in `main.cpp`, but incomplete-guard: either wrap the emit in the same flag check, or reset the flag at the top of `disconnect()` so the next attempt starts clean regardless (score: 62).

</details>

PR #747, opened 2026-04-15
profile: auto-retry failed uploads, show dialog on exhaustion

skialpine:
## Summary

Stacks on top of #744. When a BLE profile upload fails with a transient reason (frame-ACK mismatch, write-ACK timeout), ProfileManager now auto-retries with exponential backoff (1s, 2s, 4s, 8s) up to 5 total attempts. If all 5 fail, a new \`de1CommunicationFailure\` Q_PROPERTY flips true and QML shows a dialog asking the user to power-cycle the DE1.

Motivation: most users trigger shots via the GHC hardware button, which bypasses the app entirely. If the initial BLE profile upload silently failed at startup, every subsequent shot would use a stale/missing profile — wasted coffee until the user noticed. Before #744, those failures were invisible (the old counter-based check lied about success); #744 made them visible in the debug log; this PR makes the app self-heal for transient issues and surface the rest to the user.

## Design highlights

- **Retryable reasons**: \`frame sequence mismatch\`, \`timeout waiting for write ACKs\`
- **Non-retryable reasons** (and why):
  - \`superseded by a new upload\` — newer upload in flight
  - \`command queue cleared during upload\` — intentional teardown (shot/steam started)
  - \`BLE disconnect during upload\` — the \`initialSettingsComplete\` → \`applyAllSettings\` → \`uploadCurrentProfile\` chain already re-uploads on reconnect; the timer would race with it
- **Reset points**: success, BLE disconnect, \`loadProfile\` / \`loadProfileFromJson\` (user profile switch), \`acknowledgeDe1CommunicationFailure\` (user OK'd dialog). **Not** at the top of \`uploadCurrentProfile\` — the retry timer calls that function, so resetting there would make the 5-attempt cap unreachable.
- Signal signature change: \`DE1Device::profileUploaded(bool)\` → \`profileUploaded(bool success, const QString& reason)\`. Only listeners today are QSignalSpy in tests; confirmed safe.

## Files

| File | What changed |
|------|-------------|
| \`src/ble/de1device.h\` / \`.cpp\` | Signal gets \`reason\` param; emit passes \`reason\` through |
| \`src/controllers/profilemanager.h\` | New \`QTimer m_profileUploadRetryTimer\`, attempt counter, \`de1CommunicationFailure\` Q_PROPERTY + signal + invokable acknowledgement |
| \`src/controllers/profilemanager.cpp\` | Constructor wires the \`profileUploaded\` / \`connectedChanged\` / timer handlers; \`loadProfile\` / \`loadProfileFromJson\` / \`acknowledgeDe1CommunicationFailure\` reset retry state |
| \`qml/components/De1CommunicationErrorDialog.qml\` (NEW) | Modelled on \`McpConfirmDialog.qml\`; single OK button |
| \`qml/main.qml\` | Instantiate dialog, bind open/close to \`ProfileManager.de1CommunicationFailure\` |
| \`CMakeLists.txt\` | Register new QML file |
| \`tests/mocks/MockTransport.h\` | New \`setConnectedSim()\` test helper — needed to drive transport disconnect in the retry-reset test |
| \`tests/tst_profilemanager.cpp\` | 9 new cases covering the full retry state machine |
| \`tests/tst_profileupload.cpp\` | Verify the reason string on failure paths |

## Test plan

- [x] \`ctest -R tst_profilemanager\` — 9 new cases
- [x] \`ctest -R tst_profileupload\` — extended signal signature
- [ ] Manual: force a bad upload (yank BLE mid-header); confirm retry logs show attempt counter and backoff intervals; confirm 5 failures trigger the dialog
- [ ] Manual GHC workflow: upload fails, ~15s silence, dialog appears, user power-cycles, next shot uses correct profile
- [ ] Verify dialog i18n wiring with \`TranslationManager.translate()\` and English fallbacks

## Explicit non-goals (follow-up work)

- Spinner/banner during the 15-second retry window (failures are silent to the user; only exhaustion is surfaced)
- Routing app-initiated \`startEspresso\` through \`uploadProfileAndStartEspresso\` — touches every start-espresso call site, deserves its own PR
- Translation keys — add with the rest of the i18n pass

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -692,6 +692,7 @@ set(QML_FILES
     qml/components/SwipeableArea.qml
     qml/components/CrashReportDialog.qml
     qml/components/McpConfirmDialog.qml
+    qml/components/De1CommunicationErrorDialog.qml
     qml/components/ColoredIcon.qml
     qml/components/CrtOverlay.qml
     qml/components/CrtShaderEffect.qml
--- a/qml/components/De1CommunicationErrorDialog.qml
+++ b/qml/components/De1CommunicationErrorDialog.qml
@@ -1,0 +1,137 @@
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+import Decenza
+
+// Shown when ProfileManager has exhausted its retry budget for BLE profile
+// uploads. Means a transient BLE issue (frame-ACK mismatch, write-ACK
+// timeout) didn't resolve on its own after 5 attempts spanning ~15 seconds,
+// and the DE1 almost certainly needs to be power-cycled.
+//
+// Bound to ProfileManager.de1CommunicationFailure. The single OK button
+// calls acknowledgeDe1CommunicationFailure() to clear the flag.
+Dialog {
+    id: root
+    anchors.centerIn: parent
+    width: Theme.dialogWidth + 2 * padding
+    modal: true
+    dim: true
+    padding: Theme.dialogPadding
+    closePolicy: Dialog.NoAutoClose
+
+    background: Rectangle {
+        color: Theme.surfaceColor
+        radius: Theme.cardRadius
+        border.width: 2
+        border.color: Theme.errorColor
+    }
+
+    onOpened: {
+        if (typeof AccessibilityManager !== "undefined" && AccessibilityManager.enabled) {
+            AccessibilityManager.announce(
+                TranslationManager.translate("de1CommError.announce",
+                    "DE1 communication issue. The current profile could not be loaded. Please power-cycle the DE1 and reconnect."))
+        }
+    }
+
+    contentItem: ColumnLayout {
+        spacing: 0
+
+        // Header (icon + title)
+        Item {
+            Layout.fillWidth: true
+            Layout.preferredHeight: Theme.scaled(50)
+            Layout.topMargin: Theme.scaled(10)
+
+            RowLayout {
+                anchors.left: parent.left
+                anchors.leftMargin: Theme.scaled(20)
+                anchors.verticalCenter: parent.verticalCenter
+                spacing: Theme.scaled(12)
+
+                Rectangle {
+                    width: Theme.scaled(32)
+                    height: Theme.scaled(32)
+                    radius: Theme.scaled(16)
+                    color: Theme.errorColor
+
+                    Text {
+                        anchors.centerIn: parent
+                        text: "!"
+                        font.pixelSize: Theme.scaled(18)
+                        font.bold: true
+                        color: Theme.primaryContrastColor
+                        Accessible.ignored: true
+                    }
+                }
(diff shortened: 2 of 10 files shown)


skialpine:
### Code review

Found 2 issues:

1. The `Q_PROPERTY` doc comment on `de1CommunicationFailure` says QML binds the dialog's `visible` to this property — it doesn't. `qml/main.qml` uses an imperative `Connections { onDe1CommunicationFailureChanged }` handler that calls `open()`/`close()`. Someone searching for a `visible:` binding will find nothing.

https://github.com/Kulitorum/Decenza/blob/0a731a1f2862f47485e73f5a8ffa66f94fc03b7a/src/controllers/profilemanager.h#L61-L66

2. Stale upload-timeout timer can poison the retry counter across an unexpected BLE drop. `DE1Device::onTransportDisconnected()` doesn't touch `m_uploadTimeoutTimer` or the in-flight upload tracker, only the app-initiated `DE1Device::disconnect()` does. Sequence: BLE drops mid-upload → `connectedChanged` fires → ProfileManager resets its retry counter to 0 → ~10s later the upload timeout timer fires → `finishProfileUpload(false, "timeout waiting for write ACKs")` → ProfileManager sees a retryable failure and bumps the counter 0→1 → retry fires → `uploadCurrentProfile()` silently returns because `!isConnected()` → on eventual reconnect, the next real failure walks 1→2 instead of 0→1, so the dialog appears after 4 post-reconnect failures instead of 5. Narrow (only matters if the reconnect takes >10s) but real. Fix either in `onTransportDisconnected()` (mirror `disconnect()` and call `finishProfileUpload(false, \"BLE disconnect during upload\")` when an upload is in flight — that reason is non-retryable) or in ProfileManager's `connectedChanged` handler (also reset the counter on reconnect, not just disconnect).

https://github.com/Kulitorum/Decenza/blob/0a731a1f2862f47485e73f5a8ffa66f94fc03b7a/src/ble/de1device.cpp#L90-L105

<details>

- [Worth a look] `tests/mocks/MockTransport.h` top-of-file comment still says "Used by tst_shotsettings" — the file is now shared by tst_shotsettings, tst_profileupload, and tst_profilemanager (score: 65)
- [OK to skip] `isRetryableUploadFailure` uses `startsWith` on reason strings without documenting why (the variable-suffix rationale). Future readers adding a new reason might use exact-match (score: 55)
- [Worth a look] No explicit test for "retry timer fires during an active machine phase" — behaviour is correct (uploadCurrentProfile sets `m_profileUploadPending` and the phaseChanged handler resumes) but the integration isn't locked down (score: 50)
- [OK to skip] String-coupling between `isRetryableUploadFailure` prefix matches and `finishProfileUpload` reason strings — mitigated today by `tst_profileupload` locking down the exact reason text, but an enum would be more robust (score: 50)

</details>

Issue #746, opened 2026-04-15
BR: possible steam heater idle bug


jmweiss:
Recent releases (1.6.6,1.6.7) may have an issue with the steam idle setting. E.g. this morning Decenza upgraded to 1.6.7, and the steam boiler did not increase over 74C, despite setting steam heater on when idle: 
<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/84d1c6ca-b678-4271-a9c8-885afc1c3a80" />
<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/0ea49f11-fcf0-4067-8982-5ea8824fd055" />

Correct steam heater behavior resumed after stopping/restarting Decenza, or toggling the steam heater idle switch.

skialpine:
can you attached the system debug log?

jmweiss:
Here you go:

[debug.zip](https://github.com/user-attachments/files/26757225/debug.zip)

skialpine:
not sure what is going on here so adding debugging. if it happens again can you attach another debug log.

jmweiss:
Will do, thanks.

jmweiss:
No problem with new 1.6.7 releases.

PR #751, opened 2026-04-15
ble: verify DE1 stored our ShotSettings, auto-heal drift

skialpine:
## Summary
- Subscribe to `SHOT_SETTINGS` indications + read initial value on connect, so we can see what the DE1 actually has stored (steam & group targets).
- New `[ShotSettings] write:` / `[ShotSettings] reported:` trace lines on every write and every indication, so `grep ShotSettings` reconstructs the full timeline from a debug log.
- `MainController::onShotSettingsReported` does a 3-way comparison (expected-from-Settings / commanded-last-write / reported-from-DE1) and, on drift, logs a rich `[SettingsDrift]` line with a `[cause: …]` classifier (`awaiting-write`, `stale-indication-suspect`, `DE1-dropped-write`) and auto-heals via `sendMachineSettings()` with a 3-attempt budget spaced ≥2 s apart.
- Commanded tracking lives on `DE1Device` so every `setShotSettings` caller (MainController, ProfileManager, SteamCalibrator, steam/water/flush helpers) feeds the same detector. Disconnect clears all state so reconnects don't race against a stale command from the previous session.

## Why
Issue #746: user reports boiler stayed at 74 °C with "keep heater on when idle" enabled. The debug log shows we correctly commanded `steam=160 °C` over BLE, but we had no way to confirm the DE1 actually stored it — de1app does the same fire-and-forget write we do. This change adds verification on top so we can tell "DE1 dropped the write" from "user toggled it off" from "stale indication crossed our new write," and self-heals when possible.

## Test plan
- [x] 7 new unit tests in `tst_shotsettings.cpp` covering parser, commanded tracking on write, disconnect reset, heater-off payload, and short-frame rejection
- [x] All 25 tests in the mac-test suite pass (`ctest`)
- [ ] Smoke test on real DE1: toggle steam-heater-when-idle switch — expect `[ShotSettings] write: steam=0C …` → `[ShotSettings] reported: steam=0.0C …` → no drift. Toggle back on → `write: steam=160C` → `reported: steam=160.0C` → no drift.
- [ ] Confirm no spurious `[SettingsDrift]` on reconnect (subscribe-time read of DE1 power-on state should be ignored before first write).

Ref: #746

### DIFF
--- a/src/ble/bletransport.cpp
+++ b/src/ble/bletransport.cpp
@@ -211,11 +211,15 @@ void BleTransport::subscribeAll() {
     subscribe(DE1::Characteristic::WATER_LEVELS);
     subscribe(DE1::Characteristic::READ_FROM_MMR);
     subscribe(DE1::Characteristic::TEMPERATURES);
+    // SHOT_SETTINGS is indicate-capable — subscribing lets us observe the
+    // DE1's stored steam/group targets and verify that our writes stuck.
+    subscribe(DE1::Characteristic::SHOT_SETTINGS);
 
     // Read initial values
     read(DE1::Characteristic::VERSION);
     read(DE1::Characteristic::STATE_INFO);
     read(DE1::Characteristic::WATER_LEVELS);
+    read(DE1::Characteristic::SHOT_SETTINGS);
 }
 
 void BleTransport::disconnect() {
--- a/src/ble/de1device.cpp
+++ b/src/ble/de1device.cpp
@@ -95,6 +95,15 @@ void DE1Device::onTransportDisconnected() {
     m_lastSawTriggerMs = 0;
     m_lastSawWriteMs = 0;
 
+    // Clear ShotSettings tracking so a reconnect doesn't compare the DE1's
+    // post-reconnect indication against a stale commanded value from the
+    // previous session (which would log a spurious drift).
+    m_commandedSteamTargetC = -1.0;
+    m_commandedGroupTargetC = -1.0;
+    m_lastShotSettingsWriteMs = 0;
+    m_deviceSteamTargetC = -1.0;
+    m_deviceGroupTargetC = -1.0;
+
     // If an upload was in flight when the transport dropped, surface it as
     // a non-retryable "BLE disconnect during upload" failure *now* rather
     // than letting the 10 s m_uploadTimeoutTimer eventually fire with a
@@ -120,6 +129,8 @@ void DE1Device::onTransportDataReceived(const QBluetoothUuid& uuid, const QByteA
         parseStateInfo(data);
     } else if (uuid == DE1::Characteristic::SHOT_SAMPLE) {
         parseShotSample(data);
+    } else if (uuid == DE1::Characteristic::SHOT_SETTINGS) {
+        parseShotSettings(data);
     } else if (uuid == DE1::Characteristic::WATER_LEVELS) {
         parseWaterLevel(data);
     } else if (uuid == DE1::Characteristic::VERSION) {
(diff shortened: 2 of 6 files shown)


skialpine:
### Code review

Found 5 issues:

1. **Initial-settings race produces a spurious `[SettingsDrift]` on every connect** (score: 90). `sendInitialSettings()` writes `steamTemp=0`/`groupTemp=93` and records those as commanded, then synchronously emits `initialSettingsComplete` → `applyAllSettings` → `sendMachineSettings()` overwrites commanded with user values **before** the DE1's indication for the initial write arrives. When that indication lands (reported=0/93), we compare it against the now-user-valued commanded and log `DE1-dropped-write` + trigger an extra resend, burning 1 of 3 retry attempts every reconnect. Fix: skip the commanded tracker for the initial defaults write, or reset the drift counter at the top of `applyAllSettings()`.

https://github.com/Kulitorum/Decenza/blob/f11ed7321fbd6315068407cba179a7eda9423f8f/src/ble/de1device.cpp#L1183-L1197

2. **Retry counter bumps even when the resend is a silent no-op** (score: 88). `onShotSettingsReported` increments `m_shotSettingsDriftResendCount` and timestamps before calling `sendMachineSettings()`. If the device disconnected between the entry-check at line 450 and line 602 (signals emitted along the way can drive disconnects), `sendMachineSettings()` bails at line 606 and the counter is bumped with zero writes issued — exhausting the 3-attempt budget without ever trying. Guard with a bool return from `sendMachineSettings()`, or re-check `isConnected()` and increment only after the write is actually dispatched.

https://github.com/Kulitorum/Decenza/blob/f11ed7321fbd6315068407cba179a7eda9423f8f/src/controllers/maincontroller.cpp#L595-L604

3. **`startSteamHeating`/`setSteamTimeoutImmediate`/`softStopSteam` bypass `expectedSteamTargetC()` gating** (score: 85). Those call sites send `m_settings->steamTemperature()` directly regardless of `keepSteamHeaterOn`. When `keepSteamHeaterOn=false` (a real user setting), commanded becomes e.g. 135 while `expectedSteamTargetC()` returns 0, and the drift handler will "fix" the drift by calling `sendMachineSettings()` which writes steam=0 — overriding the heating we just enabled, worst in `softStopSteam()` (the intended 1-second graceful stop gets clobbered by the full `steamTimeout`). Either gate those writes by the same rule `sendMachineSettings()` uses, or have the drift check compare against the last commanded value rather than re-deriving "expected" from Settings.

https://github.com/Kulitorum/Decenza/blob/f11ed7321fbd6315068407cba179a7eda9423f8f/src/controllers/maincontroller.cpp#L1282-L1305

4. **Retry bookkeeping not cleared on disconnect** (score: 75). `DE1Device::onTransportDisconnected()` resets all the device-side tracking, but `m_shotSettingsDriftResendCount` / `m_lastShotSettingsResendMs` on `MainController` are only reset inside `onShotSettingsReported()` when drift resolves cleanly. If a session exhausts the 3-attempt budget and the user then disconnects and reconnects, the counter stays at 3 — the `kMaxResendAttempts` guard returns immediately on every subsequent drift, so the auto-heal is dead for the rest of the app lifetime until a clean match happens to arrive first.

https://github.com/Kulitorum/Decenza/blob/f11ed7321fbd6315068407cba179a7eda9423f8f/src/controllers/maincontroller.cpp#L583-L595

5. **Comment claims `setShotSettings()` issues a read-back after every write — it doesn't** (score: 75). An earlier draft had an explicit read-after-write; it was removed because `BleTransport::read()` bypasses the command queue and would race with the queued write. The `setShotSettings()` body even documents this removal, but the slot-connection comment in `MainController` still says the opposite. Suggest: "Fires on every SHOT_SETTINGS indication from the DE1 (after each write the DE1 indicates its stored value) and on the initial read issued by `subscribeAll()` at connect time."

https://github.com/Kulitorum/Decenza/blob/f11ed7321fbd6315068407cba179a7eda9423f8f/src/controllers/maincontroller.cpp#L71-L77

---

<details>

- [Worth a look] Q_PROPERTY reset on disconnect doesn't emit `shotSettingsReported(-1,-1)` — no current QML bindings read these, but it violates Qt's NOTIFY contract and is a one-line fix (score: 65)
- [Worth a look] `stale-indication-suspect` classifier still falls through to `sendMachineSettings()` — the comment says "will self-correct" but the code resends anyway; may cause log churn during normal operation if the DE1 emits both pre- and post-write indications (score: 50)
- [OK to skip] `kMinResendIntervalMs` uses wall-clock elapsed time as a resend rate limiter, which nominally conflicts with CLAUDE.md's "never use timers as guards" rule, but the alternative (event-based flag) assumes the DE1 will always respond — which is exactly the failure mode this PR is guarding against (score: 50)

</details>

PR #755, opened 2026-04-15
cal: classify auto flow cal by window frames, not whole profile

skialpine:
## Summary

- Replaces profile-level flow/pressure classification in auto flow cal with a window-level classifier that looks at which frames the steady window actually touched (via `PhaseMarker`).
- Fixes false `"extraction anomaly"` rejections and spurious multiplier jumps on hybrid profiles like ASL9-3 (pressure declines + flow-controlled tail).
- Pure classifier lives in `src/controllers/autoflowcalclassifier.{h,cpp}` with full unit-test coverage.

## Why

From analysis in #739 of a user's `debug-4.log` on profile ASL9-3 (TCL import from Visualizer):

| # | Name | Pump | Target |
|---|------|------|--------|
| 2 | rise and hold | pressure | 9 bar |
| 3–8 | decline1…6 | pressure | 8→3 bar |
| 9 | maintain flow | **flow** | **1.5 ml/s** |

The classifier at `maincontroller.cpp:966` flagged the profile as \"flow\" because frame 9 is flow-controlled. But every observed steady window on this profile landed in the pressure declines (frames 3–8); frame 9 was never reached. Anchoring to 1.5 ml/s produced:
- **3 of 6 shots rejected** as `flow profile ratio … outside [0.75, 1.35]`
- **1 shot slipped through**, jumping C from 1.0 → 1.2762 on a 5s window where v2 formula would have given 1.01

## What changes

1. **New pure helper** `classifyAutoFlowCalWindow()` — takes profile frames + a list of `FrameTransition{time, frameNumber}` + window bounds + mean machine flow, returns a classification with modes: `isFlowProfile` / `mixedMode` / `fallbackToProfileScan`.
2. **MainController** converts `PhaseMarker → FrameTransition` at the call site and branches on the result. Mixed-mode windows are skipped with a log line; missing markers fall back to the historical profile-level scan so nothing regresses for legacy shots.
3. **New log lines** so field telemetry can observe the fix:
   - `Auto flow cal: window mode=pressure frames=[3..5]`
   - `Auto flow cal: window spans mixed flow/pressure frames — skipping`
4. **Docs**: `docs/CLAUDE_MD/AUTO_FLOW_CALIBRATION.md` updated with a new \"Window-level Classification\" section.

Preinfusion is no longer a special case — `windowStart >= 10s` passes through the classifier the same way as extraction frames, and any frame's actual mode is used.

## Test plan

- [x] `tst_autoflowcal` — 11 new cases covering:
  - Flow-only profile (D-Flow / Q shape) — regression guard
  - Pressure-only profile (Spring Lever shape) — regression guard
  - **Hybrid profile, window in pressure declines (ASL9-3 shot 175 shape) — primary regression**
  - Hybrid profile, window in flow tail
  - Mixed-mode window (skipped)
  - Multi-frame same-mode window (pressure declines)
  - Missing transitions (falls back)
  - Out-of-range frame index (falls back)
  - Multi-target flow picks closest
- [x] Full ctest run: 26/26 pass
- [x] Main app builds clean on macOS
- [ ] Monitor field logs post-merge for new `window mode=` lines; confirm hybrid-profile users (spring lever, adaptive, ASL) see a drop in `flow profile ratio … skipping` rejections

## Related

- Resolves the hybrid-profile half of #739
- Complements #742 (batched median + pressure smoothing)
- For users whose \`asl9_3\` cal walked up due to this bug, the batched median will gradually pull it back. A targeted reset migration is out of scope here; can follow up if field logs show multiple affected users.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,7 @@ set(SOURCES
     src/machine/steamhealthtracker.cpp
     src/machine/steamcalibrator.cpp
     src/controllers/maincontroller.cpp
+    src/controllers/autoflowcalclassifier.cpp
     src/controllers/profilemanager.cpp
     src/controllers/directcontroller.cpp
     src/controllers/shottimingcontroller.cpp
--- a/docs/CLAUDE_MD/AUTO_FLOW_CALIBRATION.md
+++ b/docs/CLAUDE_MD/AUTO_FLOW_CALIBRATION.md
@@ -19,7 +19,7 @@ Automatic per-profile flow calibration using scale data as ground truth. After e
    - Scale data is recent (nearest weight flow point within 1 second)
    - Per-sample machine/weight flow ratio is within [0.4, 2.5] (rejects scale data glitches)
    - Window lasts at least 1.5 seconds with at least 7 samples
-3. **Profile classification**: Checks whether the profile's extraction frames (post-preinfusion) use flow control or pressure control — this determines which formula is used
+3. **Window classification**: Checks which profile frames were actually active during the steady window (using the shot's `PhaseMarker` stream) to decide whether to use the flow or pressure formula — see "Window-level Classification" below
 4. **Ratio guard**: Rejects windows where flow and weight diverge too much ([0.75, 1.35]). For flow profiles, compares `(target_flow * 0.963) / weight_flow`. For pressure profiles, compares `machine_flow / weight_flow`.
 5. **Compute calibration** (formula depends on profile type):
    - **Flow profiles**: `mean(weight_flow) / (target_flow * 0.963)` — uses the profile's known target flow, independent of current calibration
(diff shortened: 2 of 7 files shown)


skialpine:
### Code review

Found 1 issue:

1. Inline transition comments in `flowOnlyProfile_windowInFlowFrame` don't match the frame names built by `buildFlowOnlyProfile()` — they're labelled `// 2s infuse` / `// infuse` (copied from the ASL9-3 builder) but the profile at L47-L50 uses `"Filling"`, `"Infusing"`, `"Pouring"`. Harmless to the test but will confuse future readers comparing against the builder.

https://github.com/Kulitorum/Decenza/blob/7de430e26af4191a5bfc4e8d6d7f2782960b05f1/tests/tst_autoflowcal.cpp#L85-L91

<details>

- [Worth a look] `int idx` compared against `steps.size()` (qsizetype) at autoflowcalclassifier.cpp:74 — CLAUDE.md rule targets narrowing assignments, so strictly speaking not a `-Wshorten-64-to-32` violation, but prior auto-cal PRs (#742, #731) consistently moved similar patterns to `qsizetype`. Easy one-liner if applying the style rule broadly. (score: 50)

</details>

Also reviewed but below threshold: missing v4 pending-batch reset migration (score 42 — author explicitly called out as an out-of-scope tradeoff given only 1 known affected user), `preinfuseCount` widening in the fallback loop (25 — widening to `qsizetype` doesn't trigger the warning the rule targets), `firstFrameInWindow`/`lastFrameInWindow` as `int` (10 — used only for logging), and a minor imprecision in the header comment saying `shotdatamodel.h` pulls in Qt Charts / Quick (25 — Qt Quick is only transitive via `FastLineRenderer`).

PR #765, opened 2026-04-16
fix: navigate directly to shot review when SAW settling outlasts overlay

skialpine:
## Summary

- After a SAW-triggered shot end, the app briefly shows the idle page before navigating to the shot review page. This happens because SAW weight settling (4–8s on the last two shots) outlasts the 3-second stop overlay timer.
- Root cause: `stopOverlayTimer` fires at 3s, sees `pendingMetadataNavigation == false` (shot not saved yet — still settling), runs `goToIdle()`. When the shot finally saves and `shotEndedShowMetadata` fires, it just restarts the timer for another 3 seconds on idle.
- Fix: in `onShotEndedShowMetadata`, check `stopOverlayVisible`. If the overlay is still showing, use the existing deferred path. If it already expired (settling took too long), navigate directly to shot review — no timer restart, no idle flash.

Evidence from app log (shot 854):
- `[7558.487]` stop overlay shown, 3s timer starts
- `[7561.5]` timer fires, `pendingMetadataNavigation == false` → `goToIdle()`
- `[7562.463]` shot saved, `shotEndedShowMetadata` fires, restarts timer for 3 more seconds
- `[7565.5]` timer fires again → finally navigates to shot review

Shot 853 was even worse — settling took 8 seconds.

## Test plan

- [ ] Pull a SAW-triggered shot → should go directly from espresso (with overlay) to shot review, no idle flash
- [ ] If settling is fast (<3s), the overlay path should work unchanged
- [ ] With Edit After Shot OFF, should still go to idle normally after overlay
- [ ] With postShotReviewTimeout = 0 (Instant), should stay on idle even after save

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -1802,8 +1802,10 @@ ApplicationWindow {
                     goToIdle()
                 }
             } else {
-                // pendingMetadataNavigation is only set by onShotEndedShowMetadata (Edit After
-                // Shot ON). If false here, post-shot review is disabled — go straight to idle.
+                // pendingMetadataNavigation is set by onShotEndedShowMetadata only when
+                // the overlay was still visible at signal time. False here means either
+                // Edit After Shot is OFF, or the shot save arrived after the overlay
+                // expired (SAW settling outlasted 3s) and was handled directly.
                 goToIdle()
             }
         }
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 3 issues (all fixed in f63c5e7f):

1. Stale comment in `stopOverlayTimer.onTriggered` else-branch claims `pendingMetadataNavigation` is false only when Edit After Shot is OFF. After this PR, it can also be false when SAW settling outlasted the overlay timer and navigation was handled directly by `onShotEndedShowMetadata`. Updated the comment to reflect both cases.

https://github.com/Kulitorum/Decenza/blob/df377fbd452db11bfccfc612a14d5a4869c65661/qml/main.qml#L1804-L1807

2. Missing active-shot guard in the overlay-expired branch. If a user starts a new shot while the old shot's save is still pending, `onShotEndedShowMetadata` from the old shot would fire with `stopOverlayVisible == false` (cleared by `onShotStarted`) and immediately navigate to the old shot's review page, interrupting the active shot. The old code deferred through the timer, giving `onShotStarted` time to clean up. Added a `currentPage === "espressoPage"` guard to skip stale navigation.

https://github.com/Kulitorum/Decenza/blob/df377fbd452db11bfccfc612a14d5a4869c65661/qml/main.qml#L2714-L2726

3. Missing `goToIdle()` in the `timeout === 0` ("Instant") branch of the overlay-expired path. The code logged and did nothing, relying on the timer's `goToIdle()` having already succeeded. If that call was blocked by `pageStack.busy`, the user would be stranded. Added the explicit `goToIdle()` call to match the `stopOverlayTimer.onTriggered` behavior.

https://github.com/Kulitorum/Decenza/blob/df377fbd452db11bfccfc612a14d5a4869c65661/qml/main.qml#L2719-L2720

---

<details>
<summary>3 below-threshold issues (scored 25)</summary>

- [OK to skip] `stopOverlayTimer.restart()` in overlay-visible branch extends overlay to ~6s on fast settling — pre-existing behavior, timer is allowed for UI auto-dismiss per CLAUDE.md (score: 25)
- [OK to skip] `Number(Settings.value("postShotReviewTimeout", 31))` without NaN guard — `??` vs `||` rule doesn't apply to function default parameters; duplicates existing pattern at line 1792 (score: 25)
- [OK to skip] `startNavigation()` race between timer's `goToIdle()` and direct `goToShotMetadata()` — Qt.callLater clears `navigationInProgress` within ms; signals arrive seconds apart (score: 25)

</details>

skialpine:
### Code review (addendum — settling fix)

No issues found in commit 8f1d2892. Checked the `SETTLING_SILENCE_OVERRIDE_MS` change for premature stability, wrong final weights, and interaction with existing fast-path and rolling-average paths.

The 2s silence override is strictly more conservative than the existing 1s fast-path (which has no avg-margin guard at all). Without new BLE samples, the rolling average can never catch up — waiting longer serves no purpose.

PR #768, opened 2026-04-16
ble: extend ShotSettings drift detection to all fields + steam diagnostics

skialpine:
## Summary
- Extends ShotSettings drift detection from 2 fields (steam temp, group temp) to all 5 variable fields: adds steam duration, hot water temp, and hot water volume verification
- The indication-pending flag now requires ALL 5 fields to match before clearing — a partial match (e.g. temp OK but duration byte dropped) stays pending and triggers auto-resend
- Adds `[Steam]` substate transition logging (`Steaming -> Puffing -> Ending`) and flow start/stop logging for diagnosing "steam didn't stop" reports
- All `[SettingsDrift]` log lines expanded to include all 5 fields in both reported and commanded sections

## Context
PR #751 added drift detection but only covered steam temp and group temp. The 9-byte ShotSettings payload has 3 more variable fields (steam duration, hot water temp, hot water volume) that were written but never verified from the DE1's indication. This means a dropped steam timeout byte would go undetected — the DE1 might heat to the right temperature but run with a wrong or zero timeout.

PR #766 fixed a steam timeout bug, and this follow-up ensures we have the logging to diagnose any recurrence without BLE sniffing.

## Test plan
- [ ] Build with `-DBUILD_TESTS=ON` and run `ctest -R tst_shotsettings` — all tests pass including 2 new partial-match tests
- [ ] Connect to DE1, observe `[ShotSettings] write:` and `[ShotSettings] reported:` logs include all 5 fields
- [ ] Start steam, observe `[Steam] substate:` transition logs
- [ ] Verify `[Steam] flow started` / `[Steam] flow stopped` appear in logs
- [ ] If drift occurs, verify `[SettingsDrift]` includes duration + hot water fields

### DIFF
--- a/src/ble/de1device.cpp
+++ b/src/ble/de1device.cpp
@@ -1312,6 +1312,7 @@ void DE1Device::parseShotSettings(const QByteArray& data) {
         && std::abs(hotWaterTempC - m_commandedHotWaterTempC) <= kTempTolerance
         && m_commandedHotWaterVolMl >= 0
         && hotWaterVolMl == m_commandedHotWaterVolMl
+        && m_commandedGroupTargetC >= 0.0
         && std::abs(groupTargetC - m_commandedGroupTargetC) <= kTempTolerance) {
         m_shotSettingsIndicationPending = false;
     }
--- a/src/ble/de1device.cpp
+++ b/src/ble/de1device.cpp
@@ -1300,8 +1300,9 @@ void DE1Device::parseShotSettings(const QByteArray& data) {
     // indication (leave pending set — the real post-write one will arrive
     // next) or a genuine dropped-write (MainController will detect it and
     // trigger a resend, which itself sets pending). Either way we wait.
-    // All 5 variable fields must match — a partial match (e.g. temp OK but
-    // duration wrong) still counts as unacknowledged.
+    // All 5 tracked fields must match — bytes 5-6 (TargetHotWaterLength,
+    // TargetEspressoVol) are hardcoded in every write and excluded. A partial
+    // match (e.g. temp OK but duration wrong) still counts as unacknowledged.
     constexpr double kTempTolerance = 0.5;  // u8p0 encoding rounding
     if (m_commandedSteamTargetC >= 0.0
         && std::abs(steamTargetC - m_commandedSteamTargetC) <= kTempTolerance
--- a/src/ble/de1device.cpp
+++ b/src/ble/de1device.cpp
@@ -100,16 +100,22 @@ void DE1Device::onTransportDisconnected() {
     // post-reconnect indication against a stale commanded value from the
     // previous session (which would log a spurious drift).
     m_commandedSteamTargetC = -1.0;
+    m_commandedSteamDurationSec = -1;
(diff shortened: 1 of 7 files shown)


skialpine:
### Code review

Found 4 issues:

1. Stale comment on `m_lastShotSettingsPayload` — says duration/hotWater fields "aren't covered by the commanded-target pair above", but this PR adds `m_commandedSteamDurationSec`, `m_commandedHotWaterTempC`, and `m_commandedHotWaterVolMl` which now track those fields. The comment should explain the actual remaining purpose: replaying hardcoded bytes 5-6 (TargetHotWaterLength, TargetEspressoVol) that have no commanded members.

https://github.com/Kulitorum/Decenza/blob/c631a4742b58c3a824abadbd2ae0f0bddd4111f9/src/ble/de1device.h#L329-L332

2. Missing `deviceHotWaterTempC` in the sentinel guard in `onShotSettingsReported()`. The guard checks 4 of 5 parameters (`deviceSteamTargetC`, `deviceGroupTargetC`, `deviceSteamDurationSec`, `deviceHotWaterVolMl`) but omits `deviceHotWaterTempC`. Not a functional bug today (steam temp catches the disconnect sentinel first), but inconsistent with the pattern this PR establishes for the other 3 new fields.

https://github.com/Kulitorum/Decenza/blob/c631a4742b58c3a824abadbd2ae0f0bddd4111f9/src/controllers/maincontroller.cpp#L455-L459

3. Comment in `parseShotSettings()` says "All 5 variable fields must match" without explaining that bytes 5-6 (TargetHotWaterLength=60, TargetEspressoVol=200) are hardcoded in every write and intentionally excluded from the match. A reader seeing the 9-byte wire format comment above would wonder where those 2 bytes went.

https://github.com/Kulitorum/Decenza/blob/c631a4742b58c3a824abadbd2ae0f0bddd4111f9/src/ble/de1device.cpp#L1303-L1304

4. Comment references "#766 follow-up" which is an ephemeral issue reference. The next line already explains the purpose ("Reconstructs the full Steaming->Puffing->Ending->Idle sequence..."), so the issue reference adds no value and will become opaque once #766 is closed.

https://github.com/Kulitorum/Decenza/blob/c631a4742b58c3a824abadbd2ae0f0bddd4111f9/src/machine/machinestate.cpp#L200-L202

---

<details>

- [Worth a look] `m_previousSubState` initialized to `Ready` causes spurious `[Steam] substate: Ready -> Steaming` on first steam session -- misleading for diagnostics (score: 60)
- [OK to skip] Duplicate "flow stopped" logs during GHC purge -- the two messages describe different lifecycle events (substate change vs phase exit), arguably useful (score: 62)
- [OK to skip] Hot water temp summary uses `'f', 0` while main log uses `'f', 1` -- follows pre-existing steam temp pattern, u8p0 encoding guarantees distinguishability (score: 50)
- [OK to skip] Comment says "target temperatures" but new tracked fields include duration and volume -- comment scoped to the 2 Q_PROPERTY lines below it, technically correct (score: 50)
- [OK to skip] Redundant second sentence in steam substate logging comment repeats the first (score: 50)

</details>

PR #770, opened 2026-04-16
fix: limit SAW spike filter to active extraction only

skialpine:
## Summary
- Gates the SAW spike filter on `m_active` so it only rejects 100g+ jumps during extraction
- Between shots, cup placement/removal and tare drift legitimately exceed 100g — the filter was rejecting these, freezing the flow rate display and filling the debug log with warnings
- `startExtraction()` already resets all spike filter state before each shot, so extraction safety is unchanged

## Context
Found by analyzing DE1 debug logs: a -361g baseline from the previous session's cup removal persisted across app restart. Every near-zero reading was rejected as a "spike" for 67 minutes until the 3-rejection auto-reset kicked in. The filter was running outside extraction where it serves no purpose — SAW logic is already gated on `m_active` at line 134.

## Test plan
- [ ] Pull shot — SAW spike filter still protects during extraction (inject a 200g spike in simulator if possible)
- [ ] Between shots, place/remove cup — no spike rejection warnings in log, flow rate display updates immediately
- [ ] Run `ctest -R tst_` to verify no test regressions

### DIFF
--- a/src/machine/weightprocessor.cpp
+++ b/src/machine/weightprocessor.cpp
@@ -25,7 +25,12 @@ void WeightProcessor::processWeight(double weight)
     // false SAW stop. Any reading that jumps more than 100g from the previous
     // sample is rejected. Auto-resets after 3 consecutive rejections to handle
     // legitimate shifts (cup removal, tare, scale reconnect at different offset).
-    if (m_hasLastWeight && qAbs(weight - m_lastRawWeight) > 100.0) {
+    //
+    // Only active during extraction — between shots, cup placement/removal and
+    // tare drift produce legitimate 100g+ swings that the filter would reject,
+    // freezing the flow rate display and polluting the debug log. The filter is
+    // reset by startExtraction() before each shot.
+    if (m_active && m_hasLastWeight && qAbs(weight - m_lastRawWeight) > 100.0) {
         if (++m_consecutiveRejections < 3) {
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 1 issue:

1. Two existing tests (`consecutiveRejectionsAutoReset`, `spikeDoesNotCorruptFlowRate`) exercise the spike filter without calling `startExtraction()`. With the `m_active` gate, the filter is bypassed, `QTest::ignoreMessage` expectations go unmatched, and signal count assertions fail.

https://github.com/Kulitorum/Decenza/blob/45240acd8a6fe822d49157b689a57d2970b8c402/tests/tst_weightprocessor.cpp#L330-L351

Fixed in follow-up commit: added `startExtraction()` to both tests and a new `spikeBypassedWhenInactive` test covering the new behavior.

---

<details>

- [OK to skip] `resetForRetare()` does not reset `m_consecutiveRejections` -- pre-existing, only matters if 2 rejections happen before a mid-shot retare (unlikely), and `m_hasLastWeight=false` already prevents the stale counter from triggering on the next reading (score: 55)
- [OK to skip] `resetForRetare()` does not reset `m_uncalibratedBatchWarned` -- pre-existing, the warning would just be suppressed for the rest of the shot after a retare, no functional impact (score: 50)

</details>

PR #772, opened 2026-04-16
fix: measure SteamHealth against actual flow time, not state time

skialpine:
## Summary
- Gate steam sample recording on `m_machineState->isFlowing()` so only true flow time (Steaming/Pouring substates) counts toward SteamHealth metrics
- Excludes Puffing (GHC purge puff) and Ending substates which previously inflated `rawTime()` by ~10s per session
- Restores the `MIN_DURATION_FOR_ANALYSIS=20s` filter so brief steaming tests (under 20s of actual flow) are correctly skipped

## Context
A 10-second steam test session was recording ~30s of samples (17s actual flow + 13s purge/wind-down), so the 20s minimum filter never triggered. Worse, the post-flow samples (during Puffing/Ending where pressure and temperature behave differently) were polluting the avg pressure/peak temperature metrics that drive SteamHealth's drift detection and auto-reset.

`MachineState::isFlowing()` already correctly distinguishes flow from purge for steam:
```cpp
if (m_phase == Phase::Steaming && m_device) {
    DE1::SubState subState = m_device->subState();
    return subState == DE1::SubState::Steaming || subState == DE1::SubState::Pouring;
}
```

## Test plan
- [ ] Steam for less than 20s — verify the session is skipped with `SteamHealth [skip] N samples Ms` log
- [ ] Steam for over 20s of actual flow — verify session is recorded and `duration` matches actual flow time
- [ ] Verify `SteamHealth [session] ... duration: Ns` no longer includes purge/ending time
- [ ] Verify session-end logic still triggers (`onSessionComplete` called when phase leaves Steaming)

### DIFF
--- a/src/controllers/maincontroller.cpp
+++ b/src/controllers/maincontroller.cpp
@@ -138,7 +138,10 @@ MainController::MainController(QNetworkAccessManager* networkManager,
                 m_settings->setSteamDisabled(false);
             }
 
-            // Steam session ended — run post-session analysis
+            // Steam session ended — run post-session analysis. m_steamStartTime
+            // is only set when isFlowing() was true (Steaming/Pouring substates),
+            // so this fires after all flowing samples have been collected even
+            // though the Steaming phase persists through Puffing/Ending substates.
             if (phase != MachineState::Phase::Steaming && m_steamStartTime > 0) {
                 if (m_steamHealthTracker && m_steamDataModel) {
                     m_steamHealthTracker->onSessionComplete(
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

No issues scoring 75+ found. Checked for bugs, CLAUDE.md compliance, comment accuracy, and historical context. The PR cleanly delegates to `MachineState::isFlowing()` which is the canonical "actually-flowing" predicate already used by the shot timer and scale timer.

---

<details>
<summary>5 below-threshold issues (all addressed in follow-up commit)</summary>

- [Worth a look] Comment framed as blacklist ("not Puffing/Ending") instead of whitelist (Steaming/Pouring) — could mislead if future substates added (score: 50)
- [OK to skip] "~10s" duration claim is machine/firmware-dependent (score: 25)
- [OK to skip] Redundant `m_machineState &&` null check — already guaranteed by early return (score: 35)
- [Worth a look] Session-end handler comment doesn't explain interaction with new gate (score: 30)
- [Worth a look] Pre-existing `isFlowing()` comment in machinestate.cpp uses same misleading blacklist framing (score: 15, fixed for consistency)

</details>

PR #773, opened 2026-04-16
fix: verify ShotSettings via read-after-write, dedupe duplicates

skialpine:
## Summary
Three fixes that work together to make ShotSettings drift detection actually function:

1. **Read-after-write verification** — adds `read(SHOT_SETTINGS)` after every `setShotSettings()` write so the drift detector receives the DE1's stored value
2. **Queue-aware reads** — `BleTransport::read()` now goes through the command queue instead of bypassing it, so reads after writes don't race ahead and return stale values
3. **Dedupe identical writes** — skip writes whose payload exactly matches the last one sent, eliminating the 4-5 duplicate BLE writes per steam session
4. **Remove unused subscription** — drop `subscribe(SHOT_SETTINGS)` since the DE1 firmware never pushes notifications on this characteristic

## Context
The drift detector added in PR #768 was built on a false premise: that the DE1 firmware pushes notifications on the `SHOT_SETTINGS` characteristic when written. Inspection of de1app's `de1_comms.tcl` confirms it does NOT — de1app doesn't subscribe to `SHOT_SETTINGS`, only reads it on demand.

Production logs from build 3278 show this clearly: 12 ShotSettings writes in a 20-minute session, but only **one** `[ShotSettings] reported:` line (from the explicit startup read). Every steam timeout change, every group temp change — none of them got verified.

The duplicate writes came from multiple QML signals (state change, phase change, page open, isSteaming change) all firing `startSteamHeating()` with the same values within milliseconds. With read-after-write added, these duplicates would also produce duplicate reads — so deduping at the write site eliminates both at once.

## Test plan
- [ ] Connect to DE1 — verify a single `[ShotSettings] reported:` from startup read
- [ ] Change steam timeout in Settings — verify a `[ShotSettings] reported:` follows the `write:` line within ~50-100ms with the new value
- [ ] Run a steam session — verify ONE write per actual change (no `[ShotSettings] write skipped:` spam, but a few skips per session is expected from the QML signal cascade)
- [ ] Run `ctest -R tst_shotsettings` to verify existing tests pass + 3 new dedup tests

### DIFF
--- a/src/ble/bletransport.cpp
+++ b/src/ble/bletransport.cpp
@@ -216,12 +216,10 @@ void BleTransport::subscribeAll() {
     subscribe(DE1::Characteristic::WATER_LEVELS);
     subscribe(DE1::Characteristic::READ_FROM_MMR);
     subscribe(DE1::Characteristic::TEMPERATURES);
-    // SHOT_SETTINGS: the DE1 firmware does not push notifications on this
-    // characteristic when we write to it (de1app confirms this — it doesn't
-    // subscribe). Verification happens via explicit read() after each write
-    // in DE1Device::setShotSettings(). We still subscribe in case the firmware
-    // ever starts pushing — costs nothing extra.
-    subscribe(DE1::Characteristic::SHOT_SETTINGS);
+    // SHOT_SETTINGS is intentionally NOT subscribed: the DE1 firmware does
+    // not push notifications on writes (confirmed in de1app's de1_comms.tcl).
+    // Verification happens via explicit read() after each write in
+    // DE1Device::setShotSettings().
 
     // Read initial values
     read(DE1::Characteristic::VERSION);
--- a/src/ble/bletransport.cpp
+++ b/src/ble/bletransport.cpp
@@ -180,11 +180,16 @@ void BleTransport::writeUrgent(const QBluetoothUuid& uuid, const QByteArray& dat
 }
 
 void BleTransport::read(const QBluetoothUuid& uuid) {
-    if (!m_service || !m_characteristics.contains(uuid)) {
-        log(QString("read(%1) skipped - %2").arg(uuid.toString().mid(1, 8), !m_service ? "no service" : "unknown characteristic"));
-        return;
-    }
-    m_service->readCharacteristic(m_characteristics[uuid]);
+    // Queue the read so it runs after any pending writes complete. Without
+    // queueing, a read issued right after a write executes immediately and
+    // returns the pre-write value, defeating any read-after-write verification.
+    queueCommand([this, uuid]() {
+        if (!m_service || !m_characteristics.contains(uuid)) {
+            log(QString("read(%1) skipped - %2").arg(uuid.toString().mid(1, 8), !m_service ? "no service" : "unknown characteristic"));
+            return;
+        }
+        m_service->readCharacteristic(m_characteristics[uuid]);
+    });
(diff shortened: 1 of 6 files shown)


skialpine:
### Code review

Found 3 issues — all fixed in follow-up commit:

1. **Queue stall after queued read.** `processCommandQueue()` only re-fires from `onCharacteristicWritten` (writes only) or new `queueCommand()` calls. After dispatching a read, the queue stalled until something else queued. With back-to-back writes (e.g. `applyAllSettings()` plus a steam preset change), the second write would sit idle.

   https://github.com/Kulitorum/Decenza/blob/7687c403/src/ble/bletransport.cpp#L619-L626

2. **Stuck `m_shotSettingsIndicationPending` defeats drift detection.** If DE1 dropped a write, the queued read returned drift, but the pending flag stayed set (cleared only on match). MainController's stale-indication guard then suppressed the drift, silently disabling auto-heal — the exact failure case PR Kulitorum/Decenza#768 was built to catch.

   https://github.com/Kulitorum/Decenza/blob/7687c403/src/controllers/maincontroller.cpp#L527-L544

3. **Dedup interaction with stuck pending flag.** With the previous "clear only on match" semantics, an identical follow-up write would skip both the write and its read, leaving the flag stuck forever. Resolved by the same "clear on every indication" change as #2.

Fixed in commit d356e84c: re-arm the queue timer after dispatching a read; clear the pending flag on every indication; remove the now-unnecessary stale-indication guard in MainController; update tests for the new semantics.

---

skialpine:
### Code review

Re-reviewing commit d356e84c (the follow-up that addressed prior findings). Found 3 issues — all comment/dead-state cleanup, no behavioral bugs:

1. **Stale doc comment on `shotSettingsIndicationPending()` accessor.** Still describes the OLD semantics: "True between issuing a setShotSettings() write and receiving an indication that **matches** the commanded value... any mismatch indication is presumed to be a stale pre-write value still in flight, and the drift check ignores it." Both halves are now wrong — the flag clears on any indication, and the drift-check guard was removed in the same commit.

   https://github.com/Kulitorum/Decenza/blob/d356e84cd3a47b6c14fa66e9b0f7e9b8e8e8c5b1/src/ble/de1device.h#L124-L131

2. **Stale "matching value" comment in `setShotSettings()`.** Reads "An indication for this write is now outstanding. Cleared in parseShotSettings() when the DE1 reports a matching value." The "matching value" qualifier is now wrong — clears on any indication.

   https://github.com/Kulitorum/Decenza/blob/d356e84cd3a47b6c14fa66e9b0f7e9b8e8e8c5b1/src/ble/de1device.cpp#L1245-L1248

3. **`m_shotSettingsIndicationPending` has no production consumers.** Searched src/, qml/, src/network/ — only tests read `shotSettingsIndicationPending()`. The flag is set on every write and cleared on every indication, but nothing acts on it (the MainController guard that used it was removed in this commit). Either remove the flag entirely, or leave it but update the docs (Finding 1) to match.

   https://github.com/Kulitorum/Decenza/blob/d356e84cd3a47b6c14fa66e9b0f7e9b8e8e8c5b1/src/ble/de1device.h#L334-L338

---

<details>

- [OK to skip] Historical "(Previously this branch had a stale-indication filter...)" parenthetical in MainController — useful for explaining why the guard is absent, though arguably belongs in commit message (score: 35)

</details>

PR #774, opened 2026-04-16
fix: remove QML page preload + defer page side effects to onActivated

skialpine:
## Summary
Two related fixes:

1. **Remove the QML page preload Loaders** in `main.qml`. Qt 6.5+'s `qt_add_qml_module` enables `qmlcachegen` by default, which compiles QML to bytecode at build time. The hidden-Loader preload pattern predated that and no longer warms anything useful — it just instantiated 11 page object trees, ran every `Component.onCompleted` (firing real BLE/scale side effects), then destroyed them.

2. **Defer SteamPage/HotWaterPage/FlushPage side effects to `StackView.onActivated`** so they only run when the user actually opens the page, not whenever the QML object is instantiated. Matches the pattern EspressoPage already uses.

## Context
Production logs (build 3278) show SteamPage's `MainController.startSteamHeating()` firing at app startup, producing a `[ShotSettings] write: steam=160.0C ...` BLE write before the user did anything. It also overwrote `Settings.steamTimeout` and `Settings.steamFlow` with the pitcher preset defaults.

HotWaterPage was sending `MachineState.tareScale()` at startup. FlushPage was overwriting `Settings.flushFlow`/`flushSeconds`.

Tracing back, both issues had the same root cause: `Component.onCompleted` ran during the preload Loader's instantiation. Removing the preload eliminates the entire class of "side effects fire before user interaction" bugs, plus it speeds up cold start by skipping 11 unnecessary instantiations.

## Test plan
- [ ] Cold launch app — verify NO `[ShotSettings] write: steam=160.0C ...` from `startSteamHeating()` before opening Steam page
- [ ] Open Steam page — verify settings sync, heater turns on, focus moves to slider
- [ ] Open Hot Water page — verify tare fires (when not dispensing), volume/flow reset to preset
- [ ] Open Flush page — verify flow/seconds reset to preset
- [ ] Navigate between pages — verify navigation doesn't feel noticeably slower than before
- [ ] Cold launch should be faster (one fewer pass of QML instantiation)

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -596,97 +596,13 @@ ApplicationWindow {
         }
     }
 
-    // Pre-compile pages at startup by loading them once (warms QML cache)
-    // These are NOT used for navigation - StackView uses Components below
-    // The Loaders just ensure QML is parsed/compiled before first navigation
-    Loader {
-        id: preloadIdle
-        active: true
-        asynchronous: true
-        visible: false
-        sourceComponent: Component { IdlePage {} }
-        onLoaded: active = false  // Unload after compilation (Loader owns the item)
-    }
-    Loader {
-        id: preloadEspresso
-        active: true
-        asynchronous: true
-        visible: false
-        sourceComponent: Component { EspressoPage {} }
-        onLoaded: active = false
-    }
-    Loader {
-        id: preloadSteam
-        active: true
-        asynchronous: true
-        visible: false
-        sourceComponent: Component { SteamPage {} }
-        onLoaded: active = false
-    }
-    Loader {
-        id: preloadHotWater
-        active: true
-        asynchronous: true
-        visible: false
-        sourceComponent: Component { HotWaterPage {} }
-        onLoaded: active = false
-    }
-    Loader {
-        id: preloadFlush
-        active: true
-        asynchronous: true
-        visible: false
-        sourceComponent: Component { FlushPage {} }
-        onLoaded: active = false
-    }
-    Loader {
-        id: preloadSettings
-        active: true
-        asynchronous: true
-        visible: false
-        sourceComponent: Component { SettingsPage {} }
-        onLoaded: active = false
-    }
-    Loader {
-        id: preloadProfileSelector
-        active: true
-        asynchronous: true
-        visible: false
-        sourceComponent: Component { ProfileSelectorPage {} }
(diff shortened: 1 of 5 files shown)


skialpine:
### Code review

Found 3 issues — all fixed in follow-up commit:

1. **Stale "fires during preload Loader" comments** in SteamPage, HotWaterPage, FlushPage. The comments described the OLD `Component.onCompleted` but now sit on `StackView.onActivated`. Replaced with a positive description.

2. **EspressoPage dead-code guard.** With preload removed, `if (StackView.status === StackView.Active)` inside `Component.onCompleted` is never true (status is `Inactive` during construction, before activation). Removed the entire `onCompleted` block; `StackView.onActivated` already does the same setup.

   https://github.com/Kulitorum/Decenza/blob/48175029/qml/pages/EspressoPage.qml#L29-L39

3. **SteamPage/HotWaterPage/FlushPage pop-back risk** (latent). `StackView.onActivated` fires on pop-back too. If a future feature pushes a popup or sub-page from one of these operation pages, popping back would silently reset the user's in-progress settings and re-issue BLE writes. Gated the preset-reset and BLE push on `!isSteaming`/`!isDispensing`/`!isFlushing`.

   https://github.com/Kulitorum/Decenza/blob/48175029/qml/pages/SteamPage.qml#L15-L27

Fixed in commit c0c8f1b4.

---

skialpine:
### Code review

Re-reviewing commit c0c8f1b4 (the follow-up that addressed prior findings). No issues scoring 75+. The dead-code removal in EspressoPage is verified correct, and the new !isSteaming/!isDispensing/!isFlushing guards correctly defer GHC-initiated flows to main.qml's authoritative phase-change handlers.

---

<details>

- [Worth a look] HotWaterPage lost the BLE-contention rationale comment ("C++ flow-start handler tares with 200ms delay to avoid BLE command contention; a duplicate tare here would overwrite the baseline and risk BLE packet drops") when restructuring. The `!isDispensing` guard is still there but the "why" is now undocumented (score: 60)
- [Worth a look] SteamPage's `onIsSteamingChanged` and `StackView.onActivated` both reset preset values, but at different lifecycle moments. Future maintainer reading one in isolation might not understand why both exist. A brief cross-reference comment would help (score: 50)

</details>

PR #775, opened 2026-04-16
Unify steam two-tap stop setting

skialpine:
## Summary

The Steam Heater section now has a single **Two-tap to stop steaming** toggle that controls both the GHC firmware behavior (via MMR) and the headless on-screen stop button. Removes the old duplicated setup:

- The **Headless Machine → Single press to stop & purge** toggle (app-side, headless-only) and the **Heater Calibration → Steam two-tap stop** toggle (firmware MMR, GHC-only) were two implementations of the same UX feature surfaced in two different places. They never interacted because each only affected one machine type.
- Folded into a single \`steamTwoTapStop\` setting, moved to the Steam Heater section so it's discoverable.
- New-install default flips from \`true\` → \`false\` to match de1app's firmware default (\`steam_two_tap_stop = 0\`).
- **Existing installs keep their previous two-tap behavior.** A one-shot seed in the \`Settings\` constructor locks in \`steamTwoTapStop = true\` on upgrade for any user who didn't have the key explicitly set, so their stop button doesn't change on launch. Fresh installs (detected via \`m_settings.allKeys().isEmpty()\` snapshot at the top of the constructor) skip the seed and get the new single-tap default.
- Headless users who had explicitly enabled the old \`Single press to stop & purge\` toggle have that preference preserved (legacy \`headless/skipPurgeConfirm = true\` migrates to \`steamTwoTapStop = false\`).

## Test plan

- [ ] Fresh install: \`steamTwoTapStop\` is unset, defaults to \`false\`, single-tap on stop button.
- [ ] Upgrade from prior version (no settings touched): \`steamTwoTapStopDefaultMigrated\` sentinel is written, \`steamTwoTapStop = true\` is seeded, button still requires two taps.
- [ ] Upgrade with legacy \`headless/skipPurgeConfirm = true\`: migrates to \`steamTwoTapStop = false\`, single-tap behavior preserved.
- [ ] Upgrade with explicit \`calibration/steamTwoTapStop\` set in the old Heater Calibration popup: value preserved as-is.
- [ ] On a GHC machine, toggling the new setting writes the firmware MMR (visible via DE1 button behavior on the group head).
- [ ] On a headless machine with two-tap on: first tap soft-stops (button goes blue and shows "PURGE"), second tap purges.
- [ ] Settings search: searching "headless", "two-tap", or "purge" surfaces the Steam Heater card.
- [ ] Heater Calibration popup tab order: heaterTestTimeout → Defaults for cafe → Done (no orphan KeyNavigation).


skialpine:
### Code review

Found 8 issues:

1. Description text under the new toggle reads "First tap goes to puffs, **second tap stops steam**" — but the second tap purges; steam already stopped on the first tap (which sends `softStopSteam()` and puts the firmware in Puffing). On both GHC and headless, second tap = purge, not stop.

https://github.com/Kulitorum/Decenza/blob/f0ada8e07d4fad4af218c786ba686f63d2c66368/qml/pages/settings/SettingsMachineTab.qml#L353-L362

2. The migration `qDebug` line fires unconditionally whenever the legacy key existed, even when its value was `false` (in which case nothing was migrated, only the key was removed). Either move the log inside the inner `if`, or split the message.

https://github.com/Kulitorum/Decenza/blob/f0ada8e07d4fad4af218c786ba686f63d2c66368/src/core/settings.cpp#L186-L196

3. Headless-migration comment ends "Otherwise drop the old key" without acknowledging that the seeding-migration block immediately below will then set `steamTwoTapStop = true` for those users. The two blocks are coupled; reading either one alone misrepresents the final state.

https://github.com/Kulitorum/Decenza/blob/f0ada8e07d4fad4af218c786ba686f63d2c66368/src/core/settings.cpp#L186-L209

4. Seeding-migration comment "seed the old default (true)" is confusing. The removed `headlessSkipPurgeConfirm()` getter defaulted to `false`; the comment is talking about behavior (two-tap), not the legacy field's stored value. Worth rewording to make the value-vs-behavior distinction explicit.

https://github.com/Kulitorum/Decenza/blob/f0ada8e07d4fad4af218c786ba686f63d2c66368/src/core/settings.cpp#L198-L208

5. The settings JSON importer no longer reads the legacy `headless.skipPurgeConfirm` field — the entire block was removed. A user re-importing an older JSON export silently loses their preference (constructor migration only fires for persisted QSettings, not JSON imports). Worth keeping a one-line back-compat read in `importFromJson`.

https://github.com/Kulitorum/Decenza/blob/f0ada8e07d4fad4af218c786ba686f63d2c66368/src/core/settingsserializer.cpp#L417-L428

6. The "Defaults for cafe" button no longer touches `steamTwoTapStop`. It was added intentionally in commit `b9a74f6b` for steam-workflow safety on cafe machines (\"Two-tap mode ensures the machine enters Puffing before stopping, which lets the app's auto-flush timer work reliably on GHC machines\"). Removing it is a semantic regression — cafe defaults should arguably still seed two-tap regardless of where the toggle UI lives.

https://github.com/Kulitorum/Decenza/blob/f0ada8e07d4fad4af218c786ba686f63d2c66368/qml/pages/settings/SettingsCalibrationTab.qml#L734-L748

7. MCP `get_settings` still categorizes `steamTwoTapStop` under `\"calibration\"` (`include(\"steamTwoTapStop\", \"calibration\")`), but the user-facing UI moved to the Machine tab in this PR. AI agents filtering by `category=\"machine\"` won't surface it.

https://github.com/Kulitorum/Decenza/blob/f0ada8e07d4fad4af218c786ba686f63d2c66368/src/mcp/mcptools_settings.cpp#L83-L89

8. Descriptive Text under the new Two-tap toggle is missing `Accessible.ignored: true`. The label inside the same `RowLayout` correctly sets it; the description below does not. Per [docs/CLAUDE_MD/ACCESSIBILITY.md](https://github.com/Kulitorum/Decenza/blob/f0ada8e07d4fad4af218c786ba686f63d2c66368/docs/CLAUDE_MD/ACCESSIBILITY.md) decorative text adjacent to an interactive element with its own `Accessible.name` should be ignored to prevent double-announcement.

https://github.com/Kulitorum/Decenza/blob/f0ada8e07d4fad4af218c786ba686f63d2c66368/qml/pages/settings/SettingsMachineTab.qml#L354-L361

<details>

- [Worth a look] Headless migration unconditionally overwrites `calibration/steamTwoTapStop` if `headless/skipPurgeConfirm = true`, even when the user had explicitly set the GHC-side value via the old calibration popup. Practically very rare (would require having toggled both old UIs on the same install) but easy to guard against (score: 75 — kept here because impact is low). https://github.com/Kulitorum/Decenza/blob/f0ada8e07d4fad4af218c786ba686f63d2c66368/src/core/settings.cpp#L188-L196
- [OK to skip] `SteamPage.qml` block-level comment says \"second tap purges\" while the inline comment says \"request Idle to trigger purge\" — pure phrasing inconsistency, both correct (score: 50).
- [OK to skip] `steamSoftStopped` is only reset when `isSteaming` becomes true; if the user toggles `steamTwoTapStop` mid-soft-stop the flag could remain `true`, but the `(steamSoftStopped && Settings.steamTwoTapStop)` binding short-circuits and the click handler also short-circuits, so no user-visible bad state (score: 50).

</details>

PR #779, opened 2026-04-17
a11y: make AI response TextAreas discoverable by TalkBack

### LINKED ISSUE #777: Accessibility Issue: AI Advice interface not readable with TalkBack
I hope you are doing well. I am reporting an accessibility regression in the "Get AI Advice" interface.
Previously, this section was compatible with screen readers, but in the current version, I am unable to access or read the AI-generated responses using TalkBack. The focus does not seem to reach the text area where the advice is displayed.
Could you please look into this so I can read the AI suggestions again?
Thank you for your continuous support for accessibility!
**Platform:** Android (Samsung Tablet)
**App Version:** 1.6.6 / 1.6.7 Beta

skialpine:
## Summary
- TalkBack/VoiceOver focus never landed on the read-only `TextArea` that shows the AI conversation transcript or dialing recommendation, so the generated advice could not be read aloud (reported in #777).
- Root cause: these `TextArea`s have never had explicit `Accessible.*` properties. They relied on Qt's default accessibility exposure, which tightened in the Qt 6.10.x upgrades — Android no longer treats a read-only `TextArea` as focusable by default, so TalkBack skips it.
- Adds the standard `EditableText` a11y pattern already used in `ExpandableTextArea.qml` and the ShotDetail debug-log `TextArea`: role + name + description + focusable. Applied to all three places AI responses render — `ConversationOverlay`, `DialingAssistantPage`, and `SettingsAITab` conversation panel.

Fixes #777.

## Test plan
- [ ] On Android with TalkBack on: tap **AI Advice** from PostShotReview or Shot Detail → swipe through the conversation overlay → focus lands on the transcript and reads the AI response.
- [ ] On Android with TalkBack on: open Dialing Assistant → request a recommendation → swipe to the recommendation area → focus lands and the text is read.
- [ ] On Android with TalkBack on: Settings → AI tab → open the conversation panel → swipe to the transcript → focus lands and the text is read.
- [ ] Verify keyboard/mouse text selection still works (no regression in the existing scroll-during-selection behavior).

### DIFF
--- a/qml/Theme.qml
+++ b/qml/Theme.qml
@@ -155,6 +155,43 @@ QtObject {
         return stripEmoji(html.replace(/<[^>]*>/g, "")).trim()
     }
 
+    // Strip Markdown syntax so screen readers don't read formatting
+    // characters (e.g. "star star bold star star" for **bold**). Also
+    // caps length so large transcripts don't flood the accessibility tree.
+    // Pass to Accessible.description on read-only TextAreas that render
+    // Text.MarkdownText. Default cap: 2000 characters.
+    function stripMarkdown(text, maxLen) {
+        if (!text) return ""
+        var cap = maxLen || 2000
+        var s = text
+            .replace(/```[\s\S]*?```/g, " ")
+            .replace(/`([^`]+)`/g, "$1")
+            .replace(/!\[[^\]]*\]\([^)]*\)/g, "")
+            .replace(/\[([^\]]+)\]\([^)]+\)/g, "$1")
+            .replace(/\*\*([^*]+)\*\*/g, "$1")
+            .replace(/__([^_]+)__/g, "$1")
+            .replace(/\*([^*]+)\*/g, "$1")
+            .replace(/_([^_]+)_/g, "$1")
+            .replace(/~~([^~]+)~~/g, "$1")
+            .replace(/^#{1,6}\s+/gm, "")
+            .replace(/^\s*[-*+]\s+/gm, "")
+            .replace(/^\s*\d+\.\s+/gm, "")
+            .replace(/^\s*>\s?/gm, "")
+            .replace(/^\s*-{3,}\s*$/gm, "")
+            .replace(/\n{3,}/g, "\n\n")
+            .trim()
+        return s.length > cap ? s.substring(0, cap) + "\u2026" : s
+    }
+
+    // Cap plain text length for Accessible.description so very large
+    // strings (log output, crash dumps) don't flood the accessibility
+    // tree. Default cap: 2000 characters.
+    function capAccessibleText(text, maxLen) {
+        if (!text) return ""
+        var cap = maxLen || 2000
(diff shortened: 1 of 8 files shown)


skialpine:
### Code review

Found 1 issue:

1. `Accessible.description: text` binds the entire multi-turn AI transcript as raw Markdown source. TalkBack/VoiceOver will read `**bold**` literally as "star star bold star star", `#` as "hash", etc., and the transcript can grow to several kilobytes after a few turns. The commit message cites the ShotDetail debug-log TextArea as the reference pattern, but that one caps the description at `text.substring(0, 200)` — see https://github.com/Kulitorum/Decenza/blob/3de4ad8be7af5db1466b1cd6c63c882fa3189320/qml/pages/ShotDetailPage.qml#L1034-L1040. Consider stripping Markdown syntax, capping the length, or dropping `Accessible.description` entirely (the TextArea's own text is already exposed to AT via its value). Applies to all three added blocks; one example:

https://github.com/Kulitorum/Decenza/blob/3de4ad8be7af5db1466b1cd6c63c882fa3189320/qml/components/ConversationOverlay.qml#L284-L293

<details>

- [OK to skip] `Accessible.focusable: true` without `activeFocusOnTab: true` — `docs/CLAUDE_MD/ACCESSIBILITY.md` says the two should be set together, but the same doc explicitly marks keyboard chains as "low priority" for this Android tablet app, and both canonical references (ShotDetail debug log, ExpandableTextArea) also omit it (score: 50)

</details>

skialpine:
### Code review (second pass on 13534e46)

Original review findings addressed: Markdown now stripped via `Theme.stripMarkdown`, description capped at 2000 chars, and `activeFocusOnTab: true` added. A11y extended to seven additional read-only surfaces (crash log, release notes dialog, four connection logs, profile knowledge base).

Found 1 issue:

1. `stripMarkdown` and `capAccessibleText` use `var cap = maxLen || 2000`. The root CLAUDE.md explicitly forbids this in the QML section: *"Use `??` not `||` for numeric defaults where 0 is valid. JavaScript `||` treats `0` as falsy."* A caller passing `maxLen = 0` (e.g. to mean "no cap") silently gets 2000. One-liner fix: `var cap = maxLen ?? 2000`.

https://github.com/Kulitorum/Decenza/blob/13534e467a3740bb0f58f46664326844c4cd7cbd/qml/Theme.qml#L163-L167

https://github.com/Kulitorum/Decenza/blob/13534e467a3740bb0f58f46664326844c4cd7cbd/qml/Theme.qml#L189-L193

<details>

- [Worth a look] `s.substring(0, cap)` in both helpers can split a UTF-16 surrogate pair if an emoji sits on the `cap` boundary, leaving a lone high surrogate before the ellipsis. AI conversations and release notes occasionally contain emoji so it is reachable at 2000 chars. One-line guard: back off one code unit when `s.charCodeAt(cap - 1)` is in `0xD800–0xDBFF` (score: 72)

</details>

PR #780, opened 2026-04-17
fix: unify ShotSettings volume-mode gate and attribute BLE writes

skialpine:
## Summary

Started as a fix for a false-positive ShotSettings drift warning on startup; grew into a small audit of every code path that writes the DE1's ShotSettings characteristic.

### The original bug
\`ProfileManager::uploadCurrentProfile()\` unconditionally passed \`Settings::waterVolume()\` as \`TargetHotWaterVol\`. \`MainController::sendMachineSettings()\` gates this on \`waterVolumeMode\` — sending 0 in weight mode so the app stops hot water via the scale instead of the DE1's flowmeter. \`applyAllSettings\` called both back-to-back; in weight mode the two produced different payloads (65 → 0), the DE1's echo notifications arrived out of order, and the drift detector from #773 falsely flagged \"DE1-dropped-write\" and resent.

Live log evidence:
\`\`\`
[5.347] [ShotSettings] write: ... vol=65ml ...    # ProfileManager (unconditional)
[5.356] [ShotSettings] write: ... vol=0ml ...     # MainController (weight mode)
[5.429] reported: ... vol=0ml ...                 # echo of 2nd arrives first
[6.254] reported: ... vol=65ml ...                # echo of 1st arrives late
[6.263] WARN DE1-dropped-write
[6.348] resolved after 1 resend(s)
\`\`\`

### What the PR does

**1. Single source of truth for the volume-mode gate.** Adds \`Settings::effectiveHotWaterVolume()\` that returns 0 in weight mode and \`qBound(0, waterVolume(), 255)\` in volume mode. All eight \`setShotSettings\` call sites now use it — one in \`ProfileManager\`, seven in \`MainController\` (\`sendMachineSettings\`, plus the six steam paths \`setSteamTemperatureImmediate\`, \`sendSteamTemperature\`, \`startSteamHeating\`, \`turnOffSteamHeater\`, \`setSteamTimeoutImmediate\`, \`softStopSteam\`). Those six were passing \`waterVolume()\` raw before, which would have kept the same race latent on every user steam-action in weight mode.

**2. Lower-bound clamp.** The previous \`qMin(..., 255)\` let a negative \`waterVolume()\` (corrupted \`QSettings\`) slip through and wrap to 255 via the downstream \`uint8_t\` cast. \`qBound(0, v, 255)\` in the helper fixes it at the source. \`BinaryCodec::encodeU8P0\` also clamps, so the DE1 payload was already safe — this is defense at the right layer.

**3. Attribute ShotSettings BLE writes to their originator.** The existing session log showed four \`startSteamHeating\` calls per steam session (all elided by #773's dedup) without identifying which QML signal fired each one. Added an optional \`reason\` tag that flows from every \`startSteamHeating\` / \`sendMachineSettings\` / \`uploadCurrentProfile\` callsite into \`DE1Device::setShotSettings\`, where it's appended to both \`[ShotSettings] write:\` and \`write skipped: payload unchanged\` log lines. Every QML and C++ entry point now passes a tag (\`de1-state-steam\`, \`phase-steaming\`, \`steampage-activated\`, \`is-steaming-changed\`, \`live-pitcher-click\`, \`decrease-5s\`, \`increase-5s\`, \`pitcher-a11y/return/space/click\`, \`custom-widget-toggle\`, \`mqtt-steam-on\`, \`applySteamSettings\`, \`applyHotWaterSettings\`, \`applyFlushSettings\`, \`applyAllSettings\`, etc.). Future redundancy audits become self-service.

### Tests

- \`tst_profilemanager::uploadCurrentProfileRespectsWaterVolumeMode\` — byte-4 assertion in both modes.
- \`tst_settings::effectiveHotWaterVolumeRespectsMode\` — helper round-trip including negative lower-bound clamp and >255 upper-bound clamp.

### Test plan
- [x] \`ctest -R tst_profilemanager\` passes locally
- [x] \`ctest -R tst_settings\` passes locally
- [x] Two on-device steam sessions on the prior build confirmed steam control healthy (pre-fix; the race is latent outside the startup path)
- [ ] Post-merge: verify next session log shows \`[reason]\` suffix on every \`[ShotSettings]\` line and no \`DE1-dropped-write\` warning at startup

### DIFF
--- a/qml/components/layout/items/CustomItem.qml
+++ b/qml/components/layout/items/CustomItem.qml
@@ -334,7 +334,7 @@ Item {
                 case "tempToggleSteam":
                     if (typeof Settings !== "undefined" && typeof MainController !== "undefined") {
                         if (Settings.steamDisabled)
-                            MainController.startSteamHeating()
+                            MainController.startSteamHeating("custom-widget-toggle")
                         else
                             MainController.turnOffSteamHeater()
                     }
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -215,7 +215,7 @@ ApplicationWindow {
             // DE1::State::Steam = 5
             if (DE1Device.state === 5) {
                 console.log("DE1 entered Steam state - starting heater, navigating to SteamPage")
-                MainController.startSteamHeating()  // This clears steamDisabled flag
+                MainController.startSteamHeating("de1-state-steam")  // This clears steamDisabled flag
                 // Navigate to SteamPage immediately so user sees heating progress
                 var currentPage = pageStack.currentItem ? pageStack.currentItem.objectName : ""
                 if (currentPage !== "steamPage" && !pageStack.busy) {
@@ -2080,7 +2080,7 @@ ApplicationWindow {
             if (phase === MachineStateType.Phase.Steaming && wasIdle) {
                 // Start steam heating when entering steam (from GHC button)
                 // startSteamHeating clears steamDisabled flag and forces heater on regardless of keepSteamHeaterOn
-                MainController.startSteamHeating()
+                MainController.startSteamHeating("phase-steaming")
                 console.log("Started steam heating on phase change to Steaming")
                 // Stop any pending auto-flush timer when starting new steam
                 steamAutoFlushTimer.stop()
--- a/qml/pages/SteamPage.qml
+++ b/qml/pages/SteamPage.qml
@@ -26,7 +26,7 @@ Page {
             Settings.steamFlow = getCurrentPitcherFlow()
             // Start heating steam heater (ignores keepSteamHeaterOn - user wants to steam)
             // startSteamHeating clears steamDisabled flag automatically
-            MainController.startSteamHeating()
+            MainController.startSteamHeating("steampage-activated")
             durationSlider.forceActiveFocus()
         }
     }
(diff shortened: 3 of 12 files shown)


skialpine:
### Code review

Found 2 issues:

1. Same race lurks at six other `setShotSettings` call sites in `MainController` that still pass `m_settings->waterVolume()` unconditionally: `setSteamTemperatureImmediate`, `sendSteamTemperature`, `startSteamHeating`, `turnOffSteamHeater`, `setSteamTimeoutImmediate`, `softStopSteam`. In weight mode each of these emits `vol=<waterVolume>`, and the next `applyAllSettings`/`sendMachineSettings` will emit `vol=0` — the ShotSettings dedup doesn't elide because the payloads differ, and BLE echo reordering can trip the same drift detector this PR was written to silence. Extracting a private `MainController::effectiveHotWaterVolume()` helper (or `Settings::effectiveHotWaterVolume()`) and using it from all 8 call sites would close the gap and eliminate the duplicated gate.

https://github.com/Kulitorum/Decenza/blob/df27528a213094abf3a150c99380fd06c3849337/src/controllers/maincontroller.cpp#L1263-L1273

2. `qMin(m_settings->waterVolume(), 255)` only clamps the upper bound. If a future bug or corrupted QSettings returned a negative value, `qMin(-1, 255) = -1` flows forward. `BinaryCodec::encodeU8P0` downstream also clamps, so the DE1 payload is safe today — but the PR's own defense is weaker than the downstream one. Prefer `qBound(0, m_settings->waterVolume(), 255)` here and at `maincontroller.cpp:638` so both call sites stay safe independently.

https://github.com/Kulitorum/Decenza/blob/df27528a213094abf3a150c99380fd06c3849337/src/controllers/profilemanager.cpp#L1482-L1486

<details>

- [OK to skip] Literal `255` is duplicated across the two call sites; introducing a named `kHotWaterVolumeMaxMl` constant would be nicer but there's no explicit CLAUDE.md rule on magic numbers for BLE protocol encoding and the adjacent comment `// BLE uint8 max` already documents the intent (score: 50)

</details>

PR #782, opened 2026-04-17
feat(ValueInput): add valueCommitted signal, defer BLE writes to end of interaction

skialpine:
## Summary
The reason-tag logging from #780 exposed two classes of BLE over-writes driven by \`ValueInput\`:

1. **Flush page storm (observed in session log):** 30+ \`[applyFlushSettings]\` calls in 2.5 s while holding +/-. \`setShotSettings\` dedups but \`sendMachineSettings\` also writes three MMR registers that are **not deduped**, producing ~90 BLE writes in under 3 s.
2. **Profile editor (latent, worse):** 13 sliders each call \`uploadProfile()\` from \`onValueModified\`. Holding a slider for 2 s re-uploads the entire profile ~25 times — ~150 BLE writes per 2 s of adjustment.

CLAUDE.md forbids timer-based debounce (\"including debounce — should use event-based flags\"), so the fix is an event-based signal.

### ValueInput contract
Adds \`signal valueCommitted(real newValue)\` alongside the existing \`valueModified\`. Fires exactly once at the end of a user interaction (hold release, drag end, single tap, text commit). A private dirty flag ensures it only fires when the interaction actually modified the value — tapping + at max emits neither signal. The internal \`commitText()\` function was renamed to avoid shadowing the new root-level \`commitValue()\`.

| Interaction | valueModified | valueCommitted |
|-------------|---------------|----------------|
| Single tap on +/- (value changed) | 1× | 1× |
| Single tap on +/- (no change — e.g. pinned at max) | 0 | 0 |
| Hold +/- (N ticks) | N× | 1× on release |
| Drag scrubber (M pixels) | M× | 1× on release |
| Type a value + Enter | 1× | 1× |
| Type same value + Enter | 0 | 0 |
| User drags off the button (cancel) | varies | 0 |

### Migrations (all keep cheap bookkeeping in \`onValueModified\`, move BLE writes to \`onValueCommitted\`)

- **\`FlushPage.qml\`**: secondsInput, flowInput — \`applyFlushSettings()\`.
- **\`ProfileEditorPage.qml\`** (13 handlers): global temp, recommended dose, preheat tank, preinfusion ends, stop-at-volume, stop-at-weight, flow/pressure range limits, step temp, pressure/flow goal, max duration/volume/weight/flow-or-pressure, exit value — all now call \`uploadProfile()\` only on commit.
- **\`SteamPage.qml\`** (3 handlers): steamingFlowSlider, flowSlider — \`setSteamFlowImmediate\`; steamTempSlider — \`setSteamTemperatureImmediate\`.
- **\`HotWaterPage.qml\`** (3 handlers): live flowRateInput — \`setHotWaterFlowRateImmediate\`; settings volumeInput + temperatureInput — \`applyHotWaterSettings\`.

### Intentionally NOT migrated (no BLE cost per tick)
- \`SettingsMachineTab.qml\`: pure QSettings persistence.
- \`BrewDialog.qml\`: local state (\`doseValue\`, \`ratio\`, \`targetValue\`).
- \`AutoFavoritesPage\`, \`RecipeEditorPage\`, \`SimpleProfileEditorPage\`, \`PostShotReviewPage\`, other settings tabs: in-memory or QSettings only.

## Evidence (pre-fix, from post-#780 session log)
\`\`\`
[4141.996] [ShotSettings] write skipped ... [applyFlushSettings]
[4142.026] ... [applyFlushSettings]
...  (30 more, every ~25–80 ms for 2.5 s)
[4144.503] ... [applyFlushSettings]
\`\`\`
Each iteration also fired three un-deduped MMR writes (0x803828 / 0x803840 / 0x803848). Issue #783 tracks adding per-register dedup to \`writeMMR\` as belt-and-suspenders.

## Test plan
- [ ] On-device: hold duration +/- on FlushPage; log should show \`[applyFlushSettings]\` dedup-skips during hold, then the final real BLE payload once on release.
- [ ] On-device: hold a temperature or pressure slider on ProfileEditorPage; log should show no \`uploadProfile\` activity until release, then exactly one profile upload.
- [ ] On-device: hold steam temperature + on SteamPage; log should show one \`setSteamTemperatureImmediate\` on release, not ~12 Hz.
- [ ] On-device: hold hot-water volume in HotWaterPage settings view; log should show one \`applyHotWaterSettings\` on release.
- [ ] Single taps, popup drags, and text entry still commit correctly.
- [ ] Tapping + when pinned at max does nothing (no valueModified, no valueCommitted — verify by placing a log line in either handler).

### DIFF
--- a/qml/components/ValueInput.qml
+++ b/qml/components/ValueInput.qml
@@ -80,7 +80,11 @@ Item {
     Accessible.description: TranslationManager.translate("valueinput.accessibility.description", "Use plus and minus buttons to adjust. Tap center for full-screen editor. Double-tap value to type a number.")
     Accessible.focusable: true
 
-    // Keyboard handling
+    // Keyboard handling. Press fires adjustValue per key-press (including
+    // auto-repeat while held); release fires commitValue once when the
+    // physical key lifts — same contract as the +/- button MouseAreas, so
+    // BLE-bound consumers that migrated to onValueCommitted don't miss
+    // hardware-keyboard adjustments.
     Keys.onUpPressed: adjustValue(1)
     Keys.onDownPressed: adjustValue(-1)
     Keys.onLeftPressed: adjustValue(-1)
@@ -98,6 +102,17 @@ Item {
         } else if (event.key === Qt.Key_PageDown) {
             adjustValue(-10)
             event.accepted = true
+        }
+    }
+
+    Keys.onReleased: function(event) {
+        if (event.isAutoRepeat) return
+        switch (event.key) {
+            case Qt.Key_Up: case Qt.Key_Down: case Qt.Key_Left: case Qt.Key_Right:
+            case Qt.Key_PageUp: case Qt.Key_PageDown:
+                commitValue()
+                event.accepted = true
+                break
         }
     }
 
@@ -532,13 +547,15 @@ Item {
 
             Accessible.name: TranslationManager.translate("valueinput.editor.title", "Value editor")
 
-            // Tap outside to close (exit edit mode first)
+            // Tap outside to close (exit edit mode first). When the TextInput
+            // is in edit mode, commit the typed value instead of silently
+            // discarding it — tapping elsewhere is a common tablet "done"
+            // gesture. Escape in the TextInput remains an explicit discard.
             MouseArea {
                 anchors.fill: parent
                 onClicked: {
                     if (popupContent.editMode) {
-                        popupContent.editMode = false
-                        popupValueContainer.forceActiveFocus()
+                        popupTextInput.commitText()
                     } else {
                         scrubberPopup.close()
                     }
@@ -616,7 +633,9 @@ Item {
                         Accessible.description: TranslationManager.translate("valueinput.popup.hint", "Double-tap to type a number.")
                         Accessible.focusable: true
 
-                        // Keyboard navigation — honors the selected gear
+                        // Keyboard navigation — honors the selected gear.
+                        // Release fires commitValue once when the key lifts,
+                        // mirroring the +/- button contract.
                         Keys.onEscapePressed: scrubberPopup.close()
                         Keys.onUpPressed: popupAdjust(1)
                         Keys.onDownPressed: popupAdjust(-1)
@@ -624,6 +643,15 @@ Item {
                         Keys.onRightPressed: popupAdjust(1)
                         Keys.onReturnPressed: scrubberPopup.close()
                         Keys.onEnterPressed: scrubberPopup.close()
+                        Keys.onReleased: function(event) {
+                            if (event.isAutoRepeat) return
+                            switch (event.key) {
+                                case Qt.Key_Up: case Qt.Key_Down: case Qt.Key_Left: case Qt.Key_Right:
+                                    root.commitValue()
+                                    event.accepted = true
+                                    break
+                            }
+                        }
 
                         Text {
                             anchors.centerIn: parent
--- a/qml/components/ValueInput.qml
+++ b/qml/components/ValueInput.qml
@@ -45,11 +45,28 @@ Item {
     signal valueModified(real newValue)
     signal valueCommitted(real newValue)
 
-    // Helper: emit valueCommitted with the current value. Used by all
-    // release/click handlers so a single tap, a hold release, a drag release,
-    // and a keyboard arrow all produce one commit.
+    // Internal dirty flag: set when valueModified is emitted, cleared by
+    // commitValue(). Ensures valueCommitted only fires when the interaction
+    // actually changed the value (e.g. tapping + when already at max emits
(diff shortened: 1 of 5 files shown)


skialpine:
### Code review

Found 3 issues:

1. **HotWaterPage live flow-rate slider is broken after this PR.** `liveFlowRateInput` previously had `onValueModified: setHotWaterFlowRateImmediate(...)` — setHotWaterFlowRateImmediate writes back to `Settings.hotWaterFlowRate`, which is also the slider's `value:` binding, so each tick updated the displayed value. The migration dropped `onValueModified` entirely. `ValueInput.adjustValueWithStep` only *emits* `valueModified` — it doesn't self-mutate `root.value`, so during a hold the displayed number stays pinned at the starting value, every tick recomputes `newVal` from the same old `root.value`, and on release exactly one step is committed regardless of how long the user held. The live-pour UX goes from "continuous adjustment during water draw" to "single step on release." Fix: add `onValueModified: function(v) { Settings.hotWaterFlowRate = v }` (cheap Settings write — no BLE) to keep the display live, and keep `onValueCommitted` for the actual BLE write. Same pattern to verify at every migrated live-view slider.

https://github.com/Kulitorum/Decenza/blob/743b359d2a9fd32c5eed3e96bbbde4b2e546aec9/qml/pages/HotWaterPage.qml#L217-L240

2. **Pressing Escape (or tapping outside) in the popup number-entry silently discards the typed value.** `Keys.onEscapePressed` only sets `editMode = false` and gives focus back, and the outer `MouseArea.onClicked` at line 538 does the same — neither calls `commitText()`. Users who type a number and then tap outside the text field lose the edit with no indication. Pre-existing, but the PR is reorganizing this exact commit logic (renaming the function, wiring Return/Enter), so it's the natural place to decide between commit-on-blur (common tablet UX) or a visible cancel. Root CLAUDE.md: *"fix pre-existing violations in any file you touch."*

https://github.com/Kulitorum/Decenza/blob/743b359d2a9fd32c5eed3e96bbbde4b2e546aec9/qml/components/ValueInput.qml#L670-L680

3. **`||` for numeric defaults where 0 is valid — pre-existing in ProfileEditorPage, touched by this PR.** Root CLAUDE.md: *"Use `??` not `||` for numeric defaults where 0 is valid."* Three sites in the file the PR modifies:
   - `profile.recommended_dose || 18` — a dose of 0 g becomes 18 g
   - `profile.tank_desired_water_temperature || 0` — semantically wrong operator (happens to be benign because both sides are 0)
   - `profile.preinfuse_frame_count || 0` — same

   https://github.com/Kulitorum/Decenza/blob/743b359d2a9fd32c5eed3e96bbbde4b2e546aec9/qml/pages/ProfileEditorPage.qml#L527-L530

   https://github.com/Kulitorum/Decenza/blob/743b359d2a9fd32c5eed3e96bbbde4b2e546aec9/qml/pages/ProfileEditorPage.qml#L593-L596

   https://github.com/Kulitorum/Decenza/blob/743b359d2a9fd32c5eed3e96bbbde4b2e546aec9/qml/pages/ProfileEditorPage.qml#L612-L615

<details>

- [Worth a look] Keyboard arrow keys and PageUp/Down on `ValueInput` call `adjustValue()` but never `commitValue()` — since every BLE write moved to `onValueCommitted`, hardware-keyboard and switch-access users can change the displayed value but the change never propagates to the DE1. `docs/CLAUDE_MD/ACCESSIBILITY.md` says keyboard chains are "low priority" for this tablet app, so scoring below threshold, but one-line additions to each `Keys.on*Pressed` handler in `ValueInput.qml` would close the gap (score: 72)

</details>

skialpine:
### Code review (second pass on `a70cc663`)

No issues found. The four prior-review findings are all addressed:
- HotWaterPage live flow slider: `onValueModified: Settings.hotWaterFlowRate = v` restored so display tracks during hold.
- Popup tap-outside: now calls `popupTextInput.commitText()` instead of silent discard.
- Keyboard arrows: new `Keys.onReleased` handlers at both scopes fire `commitValue()` once on physical release.
- `||` → `??` applied at all three ProfileEditorPage sites.

<details>

- [Worth a look] `Keys.onReleased` handlers guard with `if (event.isAutoRepeat) return` and comment it as suppressing "Qt's synthetic release between autorepeat pairs." The guard is actually functional on platforms that DO emit autorepeat-flagged releases (qutebrowser PR #4936 relied on this), so the code is correct — but the comment's specific claim about "synthetic release between autorepeat pairs" is fuzzy. Consider tightening to something like "skip autorepeat-synthesized release events so a held key commits exactly once on physical release" (score: 50)

  https://github.com/Kulitorum/Decenza/blob/a70cc66348e5076d02a6f3dc1ba1a027fa1867d9/qml/components/ValueInput.qml#L103-L114

</details>

PR #784, opened 2026-04-17
feat(BLE): add per-register dedup to DE1Device::writeMMR (#783)

### LINKED ISSUE #783: BLE: add per-register dedup to DE1Device::writeMMR (matching ShotSettings pattern)
## Background

\`DE1Device::setShotSettings()\` dedups identical payloads to elide redundant BLE writes. Introduced in #773 and it works well: the reason-tag instrumentation from #780 showed ~8 \"write skipped: payload unchanged\" events across a typical session, all legitimately elided.

\`DE1Device::writeMMR()\` (the MMR register-write primitive used for steam flow, flush flow, flush timeout, hot water flow rate, heater voltage, group voltage, etc.) has no equivalent dedup. Every call sends a fresh BLE packet regardless of whether the value changed.

## Why this matters

The session log captured after #780 showed a tight burst on FlushPage:

\`\`\`
[4141.996] → [4144.503]: 30+ [applyFlushSettings] calls in 2.5 s
\`\`\`

Each \`applyFlushSettings\` calls \`sendMachineSettings\` which writes:
1. \`setShotSettings\` — **deduped** via the #773 mechanism → 0 BLE writes when payload unchanged
2. \`writeMMR(0x803828, ...)\` for steam flow — **not deduped** → 1 BLE write
3. \`writeMMR(0x803840, ...)\` for flush flow — **not deduped** → 1 BLE write
4. \`writeMMR(0x803848, ...)\` for flush timeout — **not deduped** → 1 BLE write

So the 2.5-second flush burst produced ~90 real BLE MMR writes even though the values didn't change.

#782 fixes the top-level signal storm at the QML layer (via \`valueCommitted\`), which addresses this particular surface. But MMR dedup would be cheap defense-in-depth for any other current or future caller that legitimately fires multiple identical MMR writes in rapid succession (phase transitions, settings apply on connect, remote-control echoes, etc.).

## Proposal

Add a simple per-register \`QHash<int, int> m_lastMMRValues\` to \`DE1Device\`. In \`writeMMR(int address, int value)\`:

- If \`m_lastMMRValues.value(address, INT_MIN) == value\`, log \`[MMR] write skipped: 0x%1 unchanged (%2) [reason]\` and return.
- Otherwise update the cache, log as usual, and proceed with the BLE write.

Extend the optional \`reason\` parameter pattern already in \`setShotSettings\` to \`writeMMR\` so dedup log lines show who was blocked.

Invalidate the cache on BLE disconnect (clear the map in the device-disconnect handler) so a fresh connect re-writes the real values rather than trusting stale cached ones.

## Acceptance criteria

- [ ] \`writeMMR\` skips BLE write when address + value match the last successful write.
- [ ] Log line for both accepted and skipped writes, with optional reason tag.
- [ ] Cache cleared on disconnect.
- [ ] Unit test with a mock transport asserting dedup-skip count vs. actual write count.
- [ ] On-device: repeated \`applyFlushSettings\` calls with unchanged values emit only the first real BLE packet; subsequent calls show as \`write skipped\`.

## Related

- #773 — original \`setShotSettings\` dedup
- #780 — reason-tag instrumentation that made this problem visible
- #782 — fixes the QML signal storm that surfaced this (different layer; MMR dedup is still worthwhile belt-and-suspenders)

skialpine:
## Summary
- Per-register `QHash m_lastMMRValues` dedup inside `DE1Device::writeMMR`, matching the `setShotSettings` dedup from #773. `applyFlushSettings` fans out through convergent QML signals and was producing ~30 identical MMR writes in a 2.5 s burst even when the underlying value hadn't changed (visible in the session log after #780).
- Optional `reason` tag plumbed through `writeMMR` / `writeMMRUrgent` so the `[MMR] write` and `[MMR] write skipped` log lines attribute redundant writes to their origin. MainController call sites (sendMachineSettings, applyHotWaterSettings, applyHeaterTweaks, startSteamHeating, set…Immediate) tag themselves.
- `force=true` opt-out on `writeMMR` for callers with refresh semantics — `setUsbChargerOn(on, force=true)` forwards it through so the DE1's 10-minute USB-charger auto-enable timeout keepalive still reaches the wire. `writeMMRUrgent` always bypasses dedup (urgent implies must-reach) but still updates the cache so subsequent non-urgent writes dedup correctly.
- Cache cleared on transport disconnect so a reconnect re-writes real values rather than trusting stale cached ones.

Closes #783.

## Test plan
- [x] `tst_mmrwrite` (14 new cases): first-write-fires, distinct-values-fire, per-address keying (not collapsed), duplicate-write-skipped, changed-write-fires-after-duplicate, 30-call burst→1-write, force bypasses dedup, `setUsbChargerOn` force reaches wire, urgent always fires, urgent updates cache, disconnect clears cache, reason tag in skip log.
- [x] `tst_shotsettings`, `tst_machinestate`, `tst_profileupload`, `tst_sav` continue to pass (writeMMR signature change is source-compatible via default args).
- [x] Main Decenza app links clean on macOS.
- [ ] On-device: repeated `applyFlushSettings` calls with unchanged values emit only the first real BLE packet; subsequent calls appear as `[MMR] write skipped` in the debug log.

### DIFF
--- a/src/ble/de1device.cpp
+++ b/src/ble/de1device.cpp
@@ -106,6 +106,10 @@ void DE1Device::onTransportDisconnected() {
     m_commandedGroupTargetC = -1.0;
     m_lastShotSettingsWriteMs = 0;
     m_lastShotSettingsPayload.clear();
+    // Clear the MMR dedup cache so a reconnect re-writes real values rather
+    // than trusting cached values from the previous session (the DE1 may have
+    // power-cycled or had its firmware state reset between sessions).
+    m_lastMMRValues.clear();
     m_deviceSteamTargetC = -1.0;
     m_deviceSteamDurationSec = -1;
     m_deviceHotWaterTempC = -1.0;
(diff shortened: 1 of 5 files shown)


skialpine:
### Code review

Found 2 issues:

1. `writeMMR` updates `m_lastMMRValues` before the transport write is acknowledged, so anything `BleTransport::clearQueue()` silently drops leaves the cache recording a value the DE1 never received. Callers of `DE1Device::clearCommandQueue()` include `MainController::onShotStarted` (every espresso start), `DE1Device::stopOperationUrgent` (SAW trigger), and `MachineState` preinfusion entry; a flush/steam slider change landing ~50–150 ms before one of those paths would strand the register on its old value until the user picks a different value or the BLE link drops.

https://github.com/Kulitorum/Decenza/blob/cac428eb704ae3996c7aea38a9c8b3d82eed559e/src/ble/de1device.cpp#L803-L815

2. `burstDeduplication` registers one `QTest::ignoreMessage` before a 30-call loop, but `ignoreMessage` matches only a single emission per registration — the other 28 skip log lines print as unexpected-debug noise.

https://github.com/Kulitorum/Decenza/blob/cac428eb704ae3996c7aea38a9c8b3d82eed559e/tests/tst_mmrwrite.cpp#L87-L100

Both are fixed in ecfc922f1d46e69a59372b613e61fea19f43665a: the cache is now cleared in `clearCommandQueue()` and in `goToSleep()` (which also drops the transport queue directly), a new `clearCommandQueueClearsCache` test covers it, and the burst test loops `ignoreMessage` 29 times.

PR #785, opened 2026-04-17
fix(BLE): only invalidate MMR cache when clearQueue actually drops a write

skialpine:
## Summary
Follow-up to #784. Every non-espresso flow-begin (steam Pouring, hot water Pouring) defensively calls `DE1Device::clearCommandQueue()` via [machinestate.cpp:350](https://github.com/Kulitorum/Decenza/blob/977e159b/src/machine/machinestate.cpp#L350) to guard against stale queued profile-upload frames. In practice nothing is queued, but #784 wired the MMR-cache clear into that same path, so every steam/hot-water cycle ended with three spurious re-sends of `STEAM_FLOW` / `FLUSH_FLOW` / `FLUSH_TIMEOUT` as `applySteamSettings` repopulated the cache. Visible in the tablet debug log at every steam-stop.

Make `DE1Transport::clearQueue()` return the number of commands it actually dropped. `DE1Device::clearCommandQueue()` and `goToSleep()` only invalidate `m_lastMMRValues` when the count is >0. Preserves the safety fix for the real race (BLE write queued → `clearQueue` drops it → cache claims DE1 has the value) while eliminating the nuisance burst on every steam stop.

## Test plan
- [x] `tst_mmrwrite::clearCommandQueueDropsInvalidateCache` — bumps `MockTransport::pendingQueueSize` before the call; confirms cache is invalidated (1 BLE write on same-value retry).
- [x] `tst_mmrwrite::clearCommandQueueWithEmptyQueuePreservesCache` (new) — leaves `pendingQueueSize=0`; confirms cache survives (0 BLE writes on same-value retry).
- [x] `tst_shotsettings`, `tst_machinestate`, `tst_profileupload`, `tst_sav`, `tst_saw` still pass after the `clearQueue()` signature change.
- [x] Main Decenza app links clean on macOS.
- [ ] On-device: verify steam-end no longer shows the 3-MMR burst in the debug log.

### DIFF
--- a/src/ble/bletransport.cpp
+++ b/src/ble/bletransport.cpp
@@ -271,7 +271,11 @@ void BleTransport::disconnect() {
 }
 
 qsizetype BleTransport::clearQueue() {
-    qsizetype cleared = m_commandQueue.size();
+    // processCommandQueue dequeues a command before dispatching it, so the
+    // currently-in-flight write is no longer in m_commandQueue but is still
+    // live (m_writePending=true). Count it too — otherwise an aborted MMR
+    // write would leave m_lastMMRValues claiming the DE1 has the value.
+    qsizetype cleared = m_commandQueue.size() + (m_writePending ? 1 : 0);
     m_commandQueue.clear();
     m_writePending = false;
     m_writeTimeoutTimer.stop();
--- a/src/ble/bletransport.cpp
+++ b/src/ble/bletransport.cpp
@@ -270,7 +270,7 @@ void BleTransport::disconnect() {
     m_disconnectedEmittedForAttempt = false;
 }
 
-void BleTransport::clearQueue() {
+qsizetype BleTransport::clearQueue() {
     qsizetype cleared = m_commandQueue.size();
     m_commandQueue.clear();
     m_writePending = false;
@@ -279,7 +279,7 @@ void BleTransport::clearQueue() {
     m_writeRetryCount = 0;
     m_lastWriteUuid.clear();
     m_lastWriteData.clear();
-    Q_UNUSED(cleared);
+    return cleared;
 }
 
 bool BleTransport::isConnected() const {
--- a/src/ble/bletransport.h
+++ b/src/ble/bletransport.h
@@ -39,7 +39,7 @@ public:
     void subscribe(const QBluetoothUuid& uuid) override;
     void subscribeAll() override;
     void disconnect() override;
-    void clearQueue() override;
+    qsizetype clearQueue() override;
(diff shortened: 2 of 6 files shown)


skialpine:
### Code review

Found 2 issues:

1. `BleTransport::clearQueue()` counts only `m_commandQueue.size()`, but a write that has already been dequeued by `processCommandQueue()` (with `m_writePending=true`, `m_lastCommand` set, bytes handed to Qt) is aborted by the same `clearQueue()` call without being counted. An in-flight MMR write can therefore return 0 → `DE1Device` skips invalidating `m_lastMMRValues` → the cache claims the DE1 has a value the wire never delivered → future identical `writeMMR` calls get deduped and the register is wedged until the user changes the value or BLE reconnects. Tight window (a few ms) but real. Fix: add `(m_writePending ? 1 : 0)` to the returned count.

https://github.com/Kulitorum/Decenza/blob/5a23c6a40039f475077ee90b0a037ae575c3c46b/src/ble/bletransport.cpp#L272-L284

2. Two comments reference `MainController::onShotStarted`, which does not exist — the actual method is `MainController::onEspressoCycleStarted` ([maincontroller.cpp:1445](https://github.com/Kulitorum/Decenza/blob/5a23c6a40039f475077ee90b0a037ae575c3c46b/src/controllers/maincontroller.cpp#L1445-L1448)). CLAUDE.md requires comments to be "correct — no claims that don't match the code."

https://github.com/Kulitorum/Decenza/blob/5a23c6a40039f475077ee90b0a037ae575c3c46b/src/ble/de1device.cpp#L819-L824

https://github.com/Kulitorum/Decenza/blob/5a23c6a40039f475077ee90b0a037ae575c3c46b/tests/tst_mmrwrite.cpp#L172-L181

Both fixed in e7afeb38: clearQueue now adds the in-flight count, comments corrected, and parallel `goToSleepDropsInvalidateCache` / `goToSleepWithEmptyQueuePreservesCache` tests added to cover the same logic in `goToSleep`. 18 MMR tests pass.

<details>

- [Worth a look] Test comment invented symbols `FLUSH_FLOW` / `FLUSH_TIMEOUT` — only `STEAM_FLOW` exists in `DE1::MMR`. Fixed in e7afeb38 (score: 58)
- [Worth a look] `goToSleep` had the same conditional-invalidation logic as `clearCommandQueue` but no test coverage. Added in e7afeb38 (score: 60)
- [OK to skip] `MockTransport::pendingQueueSize` is decoupled from the `writes` list; the existing comment on the field already documents the intentional behavior (score: 50)

</details>

PR #786, opened 2026-04-17
feat(history): add Export Shots to File toggle (#778)

### LINKED ISSUE #778: FR: history export
Decenza exports profiles in JSON format (Documents/Decenza), but there does not seem to be any way to export shot data. (You can _import_ shots from the DE1 app, but not _export_ shots.)

I'm keeping historic data for my own analysis, and it would be nice to maintain a record of all my extractions, not just the DE1 app shots.

Either format is fine (SHOT or JSON). TIA.

skialpine:
Closes #778.

## Summary
- New **Export Shots to File** toggle on Settings → History & Data (below the Share Data card). Off by default.
- When enabled, each shot in the DB is mirrored as a visualizer-format JSON file into `profiles/history/` alongside the user profiles folder. Files are for customer archive/analysis only — the app never reads them back.
- Reuses `VisualizerUploader::buildHistoryShotJson` (promoted to public static since it had no instance state) so the files are byte-identical to visualizer.coffee uploads.

## Stateless behaviour
| Action | Effect |
| --- | --- |
| Toggle off → on | Bulk-exports every shot, overwriting anything in the folder |
| Toggle on → off | Stops; leaves files alone |
| New shot saved (`shotSaved`) | Single JSON written |
| Metadata edited (`shotMetadataUpdated`, success=true) | JSON refreshed |
| Shot deleted (`shotDeleted` / `shotsDeleted`) | Matching `*_<id>.json` removed |
| Bulk export finished | Toast: `Exported N shots` (or `Exported N shots; M failed`) |

Nothing is persisted about what's been exported, so toggle-off + toggle-on = clean full re-export.

## Filename scheme
`YYYYMMDDTHHmmss_<shotId>.json` — timestamp matches the legacy `.shot` convention; shot ID guarantees uniqueness and enables glob-based deletion without a DB lookup.

## Threading
All DB reads and disk I/O run on background threads via `QThread::create` + `withTempDb`, following the existing `ShotHistoryStorage::requestCreateBackup` pattern. A `std::shared_ptr<std::atomic<bool>>` destroyed-flag guards callback delivery, and a `compare_exchange_strong` on `m_bulkRunning` prevents overlapping bulk passes.

## Files
- **New**: `src/history/shothistoryexporter.{h,cpp}` — exporter class
- **Modified**:
  - `src/core/settings.{h,cpp}` — new `exportShotsToFile` Q_PROPERTY (default false, QSettings key `export/shotsToFile`)
  - `src/core/profilestorage.{h,cpp}` — new `userHistoryPath()` helper
  - `src/network/visualizeruploader.{h,cpp}` — promoted `buildHistoryShotJson` to public `static`
  - `src/main.cpp` — constructs the exporter and registers it as a QML context property
  - `qml/pages/settings/SettingsHistoryDataTab.qml` — new Export Shots card below the Share Data card (i18n via new `settings.data.exportshots*` keys, `StyledSwitch` with `accessibleName`)
  - `qml/main.qml` — toast wired to `ShotHistoryExporter.bulkExportFinished` (new `main.toast.exportShots*` keys)
  - `CMakeLists.txt` — registered the new sources/headers

## Test plan
Verified end-to-end on macOS (fresh app, 1002 existing shots in the DB):
- [x] Toggle defaults to off
- [x] Enabling toggle creates `profiles/history/` and writes 992 files (matches DB, 10 gaps are deleted shots) — naming follows `YYYYMMDDTHHmmss_<id>.json`, no duplicates, average ~28 KB/shot
- [x] One file inspected: valid visualizer v2 JSON (`version: 2`), aligned time-series, full `meta`, embedded `profile`, populated `app.data.settings`
- [x] Pulling a new shot appended a fresh `*_1003.json` within seconds
- [x] Editing that shot's enjoyment/TDS/EY refreshed the file (mtime updated, new fields appear in `meta.shot` and `app.data.settings`)
- [x] Deleting that shot removed the file (folder back to 992)
- [ ] Android device: confirm files land under `Documents/Decenza/history/` with permission, internal fallback otherwise
- [ ] Windows device

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -277,6 +277,7 @@ set(SOURCES
     src/history/shotdebuglogger.cpp
     src/history/shotfileparser.cpp
     src/history/shotimporter.cpp
+    src/history/shothistoryexporter.cpp
     src/models/shotcomparisonmodel.cpp
     src/models/flowcalibrationmodel.cpp
     src/network/shotserver.cpp
@@ -424,6 +425,7 @@ set(HEADERS
     src/history/shotdebuglogger.h
     src/history/shotfileparser.h
     src/history/shotimporter.h
+    src/history/shothistoryexporter.h
     src/models/shotcomparisonmodel.h
     src/models/flowcalibrationmodel.h
     src/network/shotserver.h
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -2795,6 +2795,67 @@ ApplicationWindow {
         onTriggered: flowCalToast.opacity = 0
     }
 
+    // ============ SHOT EXPORT BULK COMPLETION TOAST ============
+    // Shown once the initial "export all shots" pass triggered by enabling
+    // Settings.exportShotsToFile has finished writing files to the user
+    // history folder. Silent-until-done so the toggle behaves like a plain
+    // boolean preference.
+    property string shotExportToastText: ""
+
+    Rectangle {
+        id: shotExportToast
+        anchors.bottom: parent.bottom
+        anchors.bottomMargin: Theme.scaled(40)
+        anchors.horizontalCenter: parent.horizontalCenter
+        width: shotExportToastLabel.implicitWidth + Theme.scaled(32)
+        height: shotExportToastLabel.implicitHeight + Theme.scaled(16)
+        radius: Theme.cardRadius
+        color: Theme.surfaceColor
+        opacity: 0
+        visible: opacity > 0
+        z: 600
+        Accessible.ignored: true
+
+        Behavior on opacity {
+            NumberAnimation { duration: 300 }
+        }
+
+        Text {
+            id: shotExportToastLabel
+            anchors.centerIn: parent
+            text: shotExportToastText
+            color: Theme.textColor
+            font.pixelSize: Theme.scaled(13)
+            Accessible.ignored: true
+        }
+    }
+
+    Timer {
+        id: shotExportToastTimer
+        interval: 4000
+        onTriggered: shotExportToast.opacity = 0
+    }
+
+    Connections {
+        target: ShotHistoryExporter
+        function onBulkExportFinished(ok, failed) {
+            if (failed > 0) {
+                shotExportToastText = TranslationManager.translate(
+                    "main.toast.exportShotsPartial",
+                    "Exported %1 shots; %2 failed").arg(ok).arg(failed)
+            } else {
+                shotExportToastText = TranslationManager.translate(
+                    "main.toast.exportShotsDone",
+                    "Exported %1 shots").arg(ok)
+            }
+            shotExportToast.opacity = 1
+            shotExportToastTimer.restart()
+            if (AccessibilityManager.enabled) {
+                AccessibilityManager.announce(shotExportToastText)
+            }
+        }
+    }
+
     // ============ ACCESSIBILITY: Machine State Announcements ============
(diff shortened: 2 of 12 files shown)


skialpine:
### Code review

Found 4 issues:

1. **Bulk-export race can silently drop shots**: `startBulkExport` runs `SELECT id FROM shots` once up front, and `onShotSaved` early-returns while `m_bulkRunning` is true. A shot saved between the SELECT and the end of the bulk loop (a window of seconds to tens of seconds for a large library) is neither in the enumerated list nor handled by the live hook, so its JSON mirror is missing until the next metadata edit or toggle-off→on cycle. Fix: re-query IDs at the end of the bulk loop and cover any gap, or buffer `shotSaved` IDs that arrive during the bulk.

https://github.com/Kulitorum/Decenza/blob/ebba4f1ede0bb5ef280f770a3ed8601e4efc112d/src/history/shothistoryexporter.cpp#L134-L158

2. **Batch deletes spawn N+1 racing threads and log spurious `remove failed` warnings**: `ShotHistoryStorage::deleteShots` emits `shotDeleted(id)` once per id **and** `shotsDeleted(shotIds)` for the batch ([shothistorystorage.cpp#L2461-L2463](https://github.com/Kulitorum/Decenza/blob/ebba4f1ede0bb5ef280f770a3ed8601e4efc112d/src/history/shothistorystorage.cpp#L2460-L2464)). The exporter connects to both, so each batch delete spawns N+1 background threads that race to remove the same files; all but one log a `qWarning` per file. Fix: connect to only one — `shotDeleted` covers every deletion path.

https://github.com/Kulitorum/Decenza/blob/ebba4f1ede0bb5ef280f770a3ed8601e4efc112d/src/history/shothistoryexporter.cpp#L37-L42

3. **No `shotId <= 0` guard in `onShotSaved` / `onShotMetadataUpdated`**: `ShotHistoryStorage::saveShot` emits `shotSaved(-1)` on precondition failure ([shothistorystorage.cpp#L843](https://github.com/Kulitorum/Decenza/blob/ebba4f1ede0bb5ef280f770a3ed8601e4efc112d/src/history/shothistorystorage.cpp#L842-L844)). The exporter spawns a background thread anyway, the load returns an empty record, and we log `"ShotHistoryExporter: failed to load shot -1"`. Add `if (shotId <= 0) return;` to both handlers.

https://github.com/Kulitorum/Decenza/blob/ebba4f1ede0bb5ef280f770a3ed8601e4efc112d/src/history/shothistoryexporter.cpp#L55-L71

4. **Class doc comment misleads about delete scope**: the "Toggle on" bullet lists deletion alongside shotSaved/metadata, but `onShotDeleted` / `onShotsDeleted` intentionally run regardless of toggle state (so files stay in sync even if the user turned the toggle off after files were written). Either move the delete bullet out of "Toggle on" and add a note that deletion always propagates, or drop it from the list.

https://github.com/Kulitorum/Decenza/blob/ebba4f1ede0bb5ef280f770a3ed8601e4efc112d/src/history/shothistoryexporter.h#L17-L24

<details>

- [Worth a look] `onShotDeleted` / `onShotsDeleted` call `userHistoryPath()` unconditionally, and that helper unconditionally `mkpath`s the folder — so every shot deletion creates an empty `history/` folder and spawns a thread even when the feature was never enabled. Cheap fix: skip the work if the folder doesn't already exist (score: 50)

</details>

PR #787, opened 2026-04-17
cleanup: drop redundant applySteamSettings() on Idle/Ready phase

skialpine:
## Summary

Removes `MainController.applySteamSettings()` from the `MachineState.phaseChanged` handler at `qml/main.qml:2206`. This call fired on every phase→Idle/Ready transition with unchanged Settings values — `applyAllSettings()` at connect already primes the registers, and user-initiated Settings changes propagate via QML `onValueCommitted` handlers.

This also eliminates the "50 ms after `applyAllSettings`" startup duplicate, where phase transitions to Ready immediately after device connect and re-fired `applySteamSettings` with the exact values `applyAllSettings` had just written.

## Context

PR #784 (per-register MMR dedup) and PR #785 (cache invalidation tightening) are doing work — the two dedup layers were absorbing 91 writes across two sessions today. Investigation traced most of that to one QML call site re-firing on every operation-end phase transition. This PR cuts the call at the source; the dedup safety net stays in place and continues catching legitimate convergent callers.

## Impact (from today's 2 DE1 sessions)

Removing this single call site eliminates an estimated **~24 apply calls/day**, or **~96 BLE write attempts**, split as:
- ~42 attempts that were already dedup-skipped (less log noise)
- ~30 attempts that were actually hitting the wire (real BLE reduction)
- ~24 \`setShotSettings\` payload-unchanged log entries

Remaining dedup skips (~44/session) are all legitimate convergent-caller cases (\`startSteamHeating\` during steam session startup, \`applyFlushSettings\` during multi-step user flush interaction) that the dedup layer is designed to handle.

## Mid-steam update safety

\`+5s\`/\`-5s\` buttons on SteamPage call \`startSteamHeating()\` directly, which flows through \`setShotSettings\` payload comparison — new duration ≠ old, so the write goes through. Settings-tab edits call \`applySteamSettings\` explicitly via \`onValueCommitted\`. Neither path is affected by this change.

## Test plan

- [ ] Startup: connect DE1, confirm \`[MMR] write: 0x803828 = N [applyAllSettings]\` fires at startup with no follow-up \`unchanged [applySteamSettings]\` ~50 ms later
- [ ] Mid-steam \`+5s\`: during active steaming, press +5s, confirm \`[ShotSettings] write\` (payload differs) and steaming extends
- [ ] Settings-tab steam-temp edit: change via \`SettingsMachineTab\`, confirm single \`[ShotSettings] write\` with new temp
- [ ] Operation end: trigger a flush, let it end, confirm phase→Ready transition produces no \`applySteamSettings\` entries
- [ ] GHC-initiated HotWater still works (via \`main.qml:2089\`)
- [ ] Dedup skip counts in a comparable session drop from ~42/session to ~15 or fewer


skialpine:
### Code review

<details>

- [Worth a look] Defensive re-prime gap: `DE1Device::clearCommandQueue()` at https://github.com/Kulitorum/Decenza/blob/4c9157e766b0de8d75ffd8fe0dc6f5d5aa581aad/src/ble/de1device.cpp#L808-L832 can drop a pending MMR write and invalidate `m_lastMMRValues`. The removed Idle/Ready handler would have re-primed steam flow / flush flow / flush timeout after every operation; now only the user touching the affected slider re-pushes. Very rare in practice (PR #785 narrowed cache-clear conditions, DE1 firmware retains the last-written value, any subsequent commit re-sends), so the gap is real but low-probability. (score: 50)
- [OK to skip] No comment in the Idle/Ready branch at https://github.com/Kulitorum/Decenza/blob/4c9157e766b0de8d75ffd8fe0dc6f5d5aa581aad/qml/main.qml#L2188-L2202 explains how machine settings stay current without the removed re-push. Root CLAUDE.md says "Default to writing no comments" so no action needed. (score: 50)
- [OK to skip] Comment at https://github.com/Kulitorum/Decenza/blob/4c9157e766b0de8d75ffd8fe0dc6f5d5aa581aad/src/ble/de1device.cpp#L1075-L1081 characterises MMR dedup as catching "slider change fans out" bursts; after #787 the highest-frequency source it historically absorbed (phase-transition replay) is gone, so the description is now narrower than complete. File not touched by this PR. (score: 50)

</details>

PR #788, opened 2026-04-17
fix(AI): force scene graph repaint when AI reply arrives on mobile

skialpine:
## Summary
- On iOS and Android, AI replies frequently don't appear until the user taps the screen. The signal flow and imperative text update work correctly — only the repaint is missing because the Qt Quick render thread stays asleep when a property changes from a network reply.
- Fix: call `update()` on the TextArea and enclosing Flickable inside the existing `Qt.callLater` scroll handlers, for `responseReceived` / `historyChanged` / `recommendationReceived` across the three AI surfaces.
- Item-level `update()` is a no-op on desktop (render loop already wakes on binding changes), so no behavior change there.

Files:
- `qml/components/ConversationOverlay.qml` — shot-detail + post-shot AI conversation
- `qml/pages/settings/SettingsAITab.qml` — settings AI conversation panel
- `qml/pages/DialingAssistantPage.qml` — dial-in recommendation page

## Test plan
- [ ] Android: open AI conversation from a shot, send a message, wait for reply — confirm it appears without touching the screen.
- [ ] iOS: same flow.
- [ ] Settings → AI → in-panel conversation: send a message, confirm reply appears automatically.
- [ ] Dialing Assistant page: trigger a recommendation, confirm text appears without tapping.
- [ ] Desktop (macOS/Windows): regression check — conversation still flows normally, scroll positions still behave, no flicker.

Confidence: moderate (~50%) — diagnosis of render-thread suspension matches the "appears on tap" symptom, but unreproduced locally. If the item-level `update()` isn't enough, next step is escalating to `Window.window.update()`.

### DIFF
--- a/qml/components/ConversationOverlay.qml
+++ b/qml/components/ConversationOverlay.qml
@@ -729,7 +729,7 @@ Rectangle {
             conversationText.text = MainController.aiManager.conversation.getConversationText()
             Qt.callLater(function() {
                 conversationFlickable.contentY = Math.max(0, overlay._preResponseHeight)
-                // Android's render thread can stay asleep when a property changes from
+                // Mobile render thread can stay asleep when a property changes from
                 // a network reply. Force a scene graph repaint so the reply appears
                 // without requiring a touch to wake the render loop.
                 conversationText.update()
--- a/qml/components/ConversationOverlay.qml
+++ b/qml/components/ConversationOverlay.qml
@@ -729,6 +729,11 @@ Rectangle {
             conversationText.text = MainController.aiManager.conversation.getConversationText()
             Qt.callLater(function() {
                 conversationFlickable.contentY = Math.max(0, overlay._preResponseHeight)
+                // Android's render thread can stay asleep when a property changes from
+                // a network reply. Force a scene graph repaint so the reply appears
+                // without requiring a touch to wake the render loop.
+                conversationText.update()
+                conversationFlickable.update()
             })
         }
         function onErrorOccurred(error) {
@@ -747,6 +752,8 @@ Rectangle {
             // Scroll to bottom to show the user's message / thinking indicator
             Qt.callLater(function() {
                 conversationFlickable.contentY = Math.max(0, conversationFlickable.contentHeight - conversationFlickable.height)
+                conversationText.update()
+                conversationFlickable.update()
             })
         }
     }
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 3 issues:

1. Comment in `ConversationOverlay` says "Android's render thread" while the other two files say "Mobile" — inconsistent, and the author has since noted the bug may affect iOS too, so "Mobile" is correct in all three.

https://github.com/Kulitorum/Decenza/blob/598863aa99e14629f94e275e8b8d0d141d534b4f/qml/components/ConversationOverlay.qml#L730-L737

2. `SettingsAITab.qml` imperative `conversationText.text = ...` at line 1193 permanently destroys the declarative binding set up at line 1064 (`text: MainController.aiManager?.conversation?.getConversationText() ?? ""`). Pre-existing, but CLAUDE.md says "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'." The PR touches this exact handler to add `update()` calls. After the first `onHistoryChanged` fires, the binding is gone; the TextArea relies solely on the imperative handler to keep itself current.

https://github.com/Kulitorum/Decenza/blob/598863aa99e14629f94e275e8b8d0d141d534b4f/qml/pages/settings/SettingsAITab.qml#L1186-L1200

3. `SettingsAITab.qml`'s `onHistoryChanged` is missing the `_waitingForResponse` guard that `ConversationOverlay.qml` has. `aiconversation.cpp` emits `historyChanged` both when the user sends AND when the response arrives (src/ai/aiconversation.cpp#L200). Without the guard, `_preResponseHeight` is overwritten with the post-response content height, so `onResponseReceived`'s scroll-to-top-of-new-response logic (line 1182) ends up scrolling to the bottom. Compare to the guarded version in `ConversationOverlay.qml` L743-L749. Pre-existing user-visible scroll bug in the settings AI panel.

https://github.com/Kulitorum/Decenza/blob/598863aa99e14629f94e275e8b8d0d141d534b4f/qml/pages/settings/SettingsAITab.qml#L1188-L1200

<details>

- [Worth a look] New `Connections` blocks added at `DialingAssistantPage.qml` L137-L154 duplicate pre-existing page-scope Connections at L341 and L355 listening to the same signals on the same targets. `update()` calls should go inside the existing blocks, not in new ones. Harmless (idempotent) but a maintainability trap, and the new blocks have an asymmetric lifetime (scoped inside `recommendationFlickable`). (score: 50)
- [OK to skip] The new `onResponseReceived` handler in `DialingAssistantPage` listens to the shared conversation singleton, so it fires for Settings-tab conversations too — the `update()` calls become no-ops when the page isn't visible, so no real impact. (score: 50)

</details>

PR #789, opened 2026-04-17
feat(steam): add "Add Off" preset option that disables the steam heater

skialpine:
## Summary
User request: allow creating a steam preset that turns the steam heater **off** rather than setting a temp/time. Useful for seasonal workflows (cold drinks, summer) where you want the boiler to stop heating until you're back to steaming.

Instead of a single built-in "Off" preset, the Add Preset dialog now has a third button **Add Off** alongside Cancel / Add — users name their own off pill ("Cold brew", "Summer", etc.) and can have multiple if they want.

## What changed
- **Data model** ([settings.cpp](../blob/feat/steam-preset-off/src/core/settings.cpp), [settings.h](../blob/feat/steam-preset-off/src/core/settings.h)): presets now support an optional `disabled: true` flag; new `Settings::addSteamPitcherPresetDisabled(name)` writes it. Existing `updateSteamPitcherPreset` already preserves unknown fields, so editing an Off preset keeps the flag.
- **Dialog** ([SteamPage.qml](../blob/feat/steam-preset-off/qml/pages/SteamPage.qml)): new "Add Off" button between Cancel and Add, tab-ordered.
- **Pill styling**: disabled pills use `Theme.textSecondaryColor` for the selected-background and text, so they read as "not a temp/time preset" without looking broken.
- **Apply path**: config-row pills go through one `applyPitcher()` helper that branches to `MainController.turnOffSteamHeater()` for disabled presets instead of `startSteamHeating`. Page activation does the same.
- **Live pill row**: Off pills are hidden while steaming — nothing useful to do mid-session.
- **Sliders**: `saveCurrentPitcher()` no-ops for disabled presets, and `getCurrentPitcher*` helpers fall back to current Settings values so sliders don't snap to undefined when an Off pill is selected.
- **A11y**: pill Accessible.name includes "turns steam heater off" when disabled.

## Test plan
- [ ] Add a new preset named "Off" via "Add Off" in the dialog — pill appears in the config row with muted styling, selected.
- [ ] Tap the Off pill from an idle steam page — steam heater turns off (no heating).
- [ ] Tap a normal pill afterwards — steam heater starts heating again, sliders update to preset values.
- [ ] Start a steam session from a normal preset, confirm the Off pill is **hidden** from the live pill row during steaming.
- [ ] Edit an Off preset via long-press: rename works, the `disabled: true` flag survives the edit.
- [ ] Move sliders while an Off preset is selected: changes do **not** get saved into the Off preset's JSON.
- [ ] Navigate to SteamPage with an Off preset selected: heater stays off, doesn't kick on.
- [ ] TalkBack/VoiceOver reads "turns steam heater off" for disabled pills.

### DIFF
--- a/qml/components/PresetPillRow.qml
+++ b/qml/components/PresetPillRow.qml
@@ -242,13 +242,19 @@ FocusScope {
 
                         property bool isSelected: modelData.index === root.selectedIndex
                         property bool isFocused: root.activeFocus && modelData.index === root.focusedIndex
+                        // A preset may mark itself as disabled (e.g. a steam "Off" preset
+                        // that stops heating instead of setting a temp/time). Render it
+                        // muted so it's visibly distinct from real time/temp presets.
+                        property bool isDisabled: modelData && modelData.preset && modelData.preset.disabled === true
 
                         width: pillText.implicitWidth + root.pillPadding
                         height: Theme.scaled(50)
                         radius: Theme.scaled(10)
 
-                        color: isSelected ? Theme.primaryColor : Theme.backgroundColor
-                        border.color: isSelected ? Theme.primaryColor : Theme.textSecondaryColor
+                        color: isSelected
+                            ? (isDisabled ? Theme.textSecondaryColor : Theme.primaryColor)
+                            : Theme.backgroundColor
+                        border.color: isSelected && !isDisabled ? Theme.primaryColor : Theme.textSecondaryColor
                         border.width: 1
 
                         // Accessibility: Let AccessibleTapHandler handle screen reader interaction
--- a/qml/components/layout/items/SteamItem.qml
+++ b/qml/components/layout/items/SteamItem.qml
@@ -216,6 +216,14 @@ Item {
                     var wasAlreadySelected = (index === Settings.selectedSteamPitcher)
                     Settings.selectedSteamPitcher = index
                     var preset = Settings.getSteamPitcherPreset(index)
+                    if (preset && preset.disabled) {
+                        // "Off" preset — disable the steam heater. Don't write
+                        // undefined preset.duration/flow into int Settings, and
+                        // don't start steam on re-tap.
+                        MainController.turnOffSteamHeater()
+                        presetPopup.close()
+                        return
+                    }
                     if (preset) {
                         Settings.steamTimeout = preset.duration
                         Settings.steamFlow = preset.flow !== undefined ? preset.flow : 150
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -214,8 +214,20 @@ ApplicationWindow {
         function onStateChanged() {
             // DE1::State::Steam = 5
             if (DE1Device.state === 5) {
-                console.log("DE1 entered Steam state - starting heater, navigating to SteamPage")
-                MainController.startSteamHeating("de1-state-steam")  // This clears steamDisabled flag
+                // Match de1app: if the currently selected steam preset is an "Off" pill,
+                // leave the heater target at 0 (already pushed via sendMachineSettings /
+                // turnOffSteamHeater). The DE1 still enters Steam state on GHC press, but
+                // with TargetSteamTemp=0 no steam is produced — consistent with the user's
+                // intent to keep the boiler off.
+                var currentPitcher = Settings.getSteamPitcherPreset(Settings.selectedSteamPitcher)
+                var currentPitcherDisabled = currentPitcher && currentPitcher.disabled === true
+                if (currentPitcherDisabled) {
+                    console.log("DE1 entered Steam state but Off preset selected — heater stays off")
+                    MainController.turnOffSteamHeater()
+                } else {
+                    console.log("DE1 entered Steam state - starting heater, navigating to SteamPage")
+                    MainController.startSteamHeating("de1-state-steam")  // This clears steamDisabled flag
+                }
                 // Navigate to SteamPage immediately so user sees heating progress
                 var currentPage = pageStack.currentItem ? pageStack.currentItem.objectName : ""
                 if (currentPage !== "steamPage" && !pageStack.busy) {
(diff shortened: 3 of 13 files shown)


skialpine:
### Code review

Found 5 issues. Four fixed in [72660047](https://github.com/Kulitorum/Decenza/commit/726600472c2215b05d9877a79bd4c7a6304509fb); one left as a follow-up design question.

**Fixed:**

1. IdlePage steam preset tap crashed with "Cannot assign [undefined] to int" when an Off preset was tapped, and the heater did not turn off (user-reported during review). `preset.duration` is undefined for disabled presets.

https://github.com/Kulitorum/Decenza/blob/726600472c2215b05d9877a79bd4c7a6304509fb/qml/pages/IdlePage.qml#L292-L319

2. Same bug in `SteamItem.qml` (layout-widget steam popup). Applied the same fix.

https://github.com/Kulitorum/Decenza/blob/726600472c2215b05d9877a79bd4c7a6304509fb/qml/components/layout/items/SteamItem.qml#L215-L243

3. `getCurrentPitcherDuration/Flow` used `||` for numeric fallbacks on `Settings.steamTimeout` and `Settings.steamFlow`, where `0` is a valid value (CLAUDE.md: "Use `??` not `||` for numeric defaults where 0 is valid.").

https://github.com/Kulitorum/Decenza/blob/726600472c2215b05d9877a79bd4c7a6304509fb/qml/pages/SteamPage.qml#L120-L133

4. IdlePage / layout-widget pill rows (via shared `PresetPillRow`) were not styling disabled pills, so Off pills looked identical to real temp/time presets. Added muted styling to the shared component — no-op for other callers (espresso / flush / hotwater) which don't set `disabled`.

https://github.com/Kulitorum/Decenza/blob/726600472c2215b05d9877a79bd4c7a6304509fb/qml/components/PresetPillRow.qml#L240-L258

**Not fixed — design question:**

5. `main.qml` `onStateChanged` unconditionally calls `startSteamHeating("de1-state-steam")` whenever the DE1 enters Steam state (e.g. a GHC button press). If the user has an Off preset selected, the heater turns on as soon as anything puts the DE1 in Steam state, overriding the Off intent. This may be correct semantics ("user physically pressed GHC → they want steam now") but contradicts the `StackView.onActivated` branch which explicitly honors Off. Needs a design decision before changing.

https://github.com/Kulitorum/Decenza/blob/726600472c2215b05d9877a79bd4c7a6304509fb/qml/main.qml#L212-L226

<details>

- [Worth a look] Keyboard nav in the live pill row can land on a hidden Off pill — `forceActiveFocus()` on an invisible item is a no-op, so arrow-key users get stuck. Not a correctness bug for the heater. (score: 70)
- [OK to skip] Edit-dialog rename of an Off preset passes `preset.duration`/`preset.flow` (undefined) into the int-typed `updateSteamPitcherPreset` which coerces to 0. The `disabled` flag is preserved because the C++ method reads the existing object first, so behavior is unchanged in practice. (score: 60)
- [OK to skip] Back button on SteamPage with `keepSteamHeaterOn=true` + fresh app restart + Off preset selected: `applySteamSettings()` sends the stored steamTemperature because the session-only `steamDisabled` flag has not been set yet. `StackView.onActivated` sets it on page entry, so the window is only "page not yet visited this session". (score: 55)

</details>

skialpine:
### Code review (round 2)

Found 3 issues in the new commits. All fixed in [6491890d](https://github.com/Kulitorum/Decenza/commit/6491890d32a88e0055f5e0d8c8c4049b83f1b6a2).

1. `turnOffSteamHeater()` sets `Settings.steamDisabled=true`, and only `startSteamHeating()` clears it. But IdlePage and SteamItem's preset-select paths call `applySteamSettings()` (not `startSteamHeating`), so switching from an Off preset to a normal preset via those surfaces left the session flag set. Next `sendMachineSettings` then commanded 0°C to the heater even though the user had a real preset selected. Fix: in `sendMachineSettings`, when the currently selected preset is not disabled, clear the stale flag — the preset is the authoritative intent.

https://github.com/Kulitorum/Decenza/blob/6491890d32a88e0055f5e0d8c8c4049b83f1b6a2/src/controllers/maincontroller.cpp#L617-L646

2. `DE1Simulator::setTargetSteamTemp` always emitted `idleSteamTempChanged` on target change, including mid-steam. The SteamPage +5s/-5s buttons call `setShotSettings` mid-session, which now forwards to the sim and fires the idle emit; `DE1Device::setSimulatedIdleSteamTemp` then built a minimal `ShotSample` with zero pressure/flow and `timer=0`, emitted `shotSampleReceived`, and `MainController::onShotSampleReceived` plotted it into the live steam graph at a large negative x-coordinate. Suppressed the emit when `m_state == Steam`.

https://github.com/Kulitorum/Decenza/blob/6491890d32a88e0055f5e0d8c8c4049b83f1b6a2/src/simulator/de1simulator.cpp#L218-L237

3. SteamPage `StackView.onActivated` unconditionally called `durationSlider.forceActiveFocus()`. When an Off preset is selected, the duration row has `visible: false` — focusing a hidden item silently kills keyboard navigation. Moved the `forceActiveFocus` into the non-disabled branch.

https://github.com/Kulitorum/Decenza/blob/6491890d32a88e0055f5e0d8c8c4049b83f1b6a2/qml/pages/SteamPage.qml#L21-L40

<details>

- [OK to skip] `m_lastShotSettingsPayload` is never updated in the sim-mode `setShotSettings` branch. If the app ever switched from sim mode to real BLE mid-session (not a supported workflow), the first post-sim write could produce misleading drift warnings. Comment now documents this is intentional. (score: 55)
- [OK to skip] Duplication: SteamPage has both a reactive `currentPitcherDisabled` property and an `isCurrentPitcherDisabled()` imperative function. Both exist because the function is used from inside `saveCurrentPitcher()` where reactivity isn't needed. Consolidating to the property alone would work. (score: 50)
- [OK to skip] MMR writes in `sendMachineSettings` (steamFlow, flushFlow, flushSeconds) are silently dropped in sim mode because `writeMMR` guards on `!m_transport`. Pre-existing behavior, not a regression from this PR. (score: 55)

</details>

PR #790, opened 2026-04-18
fix(steam): restore mid-session SteamFlow control + verify-and-retry MMR

Kulitorum:
## Summary

- Restores mid-session steam flow control — drag-on-value-text now actually reaches the DE1
- Adds `writeMMRVerified()` (write + read-back + retry) for confirmed-delivery MMR writes; `setSteamFlowImmediate` routes through it
- Live preset row now writes the new flow to the DE1 mid-session and syncs the slider
- Doc commit: refresh the WiFi debug IP that DHCP rotated and add a USB-recovery hint

## Root cause

PR #782 split BLE writes from `onValueModified` to `onValueCommitted` in `ValueInput`, but missed the `valueDragArea.onReleased` handler (`ValueInput.qml:315`). The +/- button MouseAreas got the new `commitValue()` call; the value-text drag MouseArea did not.

So horizontal drags fired `_emitValueModified` per ~20 px (setting `_dirtySinceCommit=true`) but `commitValue()` never ran — for every consumer that migrated to `onValueCommitted` (SteamPage, FlushPage, HotWaterPage, SettingsMachineTab) drag-to-write became a silent no-op.

For SteamPage's mid-session flow slider this was visible as "the steam intensity doesn't react when I drag the slider mid-session". The displayed value updated, `Settings.steamFlow` persisted, but no BLE packet ever reached the DE1.

## Why verify-and-retry too

While hunting the regression I built `writeMMRVerified()` thinking the firmware needed multiple writes to land — a wrong theory the on-device test debunked (zero retries needed across many drags). Kept it anyway as defensive engineering: writes the value, reads `0x803828` back via `READ_FROM_MMR` after ~50 ms, retries up to 5× with `force=true` on mismatch. Costs one extra read per change, makes future BLE drops or firmware quirks self-healing on this register, and self-documents that `SteamFlow` is a confirmed-delivery setting.

## Files

| File | Purpose |
|------|---------|
| `qml/components/ValueInput.qml` | **Root-cause fix** — `valueDragArea.onReleased` calls `commitValue()`. Affects every page that wired BLE to `onValueCommitted`. |
| `qml/pages/SteamPage.qml` | Live preset row: mid-session taps push `setSteamFlowImmediate(flow)` and imperatively sync `steamingFlowSlider.value` (binding gets broken by the first drag). |
| `src/ble/de1device.{h,cpp}` | New `writeMMRVerified()` + `PendingMMRVerify` state + `parseMMRResponse` hook + clear on disconnect. |
| `src/controllers/maincontroller.cpp` | `setSteamFlowImmediate()` routes through `writeMMRVerified()`. |
| `CLAUDE.md` | Refresh stale WiFi debug IP, add USB-fallback hint for next DHCP rotation. |

## Test plan

- [x] On-device (DE1Pro v1333, Galaxy Tab A9): drag steam flow 40 → 145 → 40 mid-Pouring; every commit shows `[MMR] verify ok: 0x803828 = N` on the first read-back; steam intensity visibly tracks.
- [x] No retries needed across multiple sessions — firmware honors mid-session writes when they actually arrive.
- [ ] Sanity-check that drag-to-write also works on FlushPage and HotWaterPage (same `ValueInput` change benefits them).
- [ ] Verify the +/- button paths still commit correctly (unchanged by this PR but co-tenant of the same component).
- [ ] Verify nothing broke when a drag is canceled (drag-off-screen) — should not commit.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -154,7 +154,7 @@ Also in `docs/`:
 
 - **ADB path**: `/c/Users/Micro/AppData/Local/Android/Sdk/platform-tools/adb.exe`
 - **Uninstall app**: `adb uninstall io.github.kulitorum.decenza_de1`
-- **WiFi debugging**: `192.168.1.208:5555` (reconnect: `adb connect 192.168.1.208:5555`)
+- **WiFi debugging**: `192.168.1.212:5555` (reconnect: `adb connect 192.168.1.212:5555`). The DHCP lease can rotate — if reconnect fails, plug in USB and run `adb shell ip route | grep wlan` to read the current IP, then `adb tcpip 5555` + `adb connect <ip>:5555`.
 - **Qt version**: 6.10.3
 - **Qt path**: `C:/Qt/6.10.3/msvc2022_64`
 - **C++ standard**: C++17
--- a/qml/components/ValueInput.qml
+++ b/qml/components/ValueInput.qml
@@ -320,12 +320,26 @@ Item {
                         if (!isDragging && !hasMoved) {
                             scrubberPopup.open()
                         }
+                        // Commit on drag release. PR #782 added the
+                        // valueCommitted contract for the +/- buttons but
+                        // missed this MouseArea, so drags were updating
+                        // _dirtySinceCommit during onPositionChanged
+                        // (via _emitValueModified) without ever firing
+                        // commitValue() — silently dropping the BLE write
+                        // for every consumer that migrated to onValueCommitted.
+                        if (isDragging) {
+                            root.commitValue()
+                        }
                         isDragging = false
                         dragReady = false
                         hasMoved = false
                     }
 
                     onCanceled: {
+                        // Drag canceled (e.g. finger left the screen edge,
+                        // or another gesture stole focus). Don't commit —
+                        // this matches the +/- button onCanceled semantics
+                        // (treat cancel as "user changed their mind").
                         isDragging = false
                         dragReady = false
                         hasMoved = false
--- a/qml/pages/SteamPage.qml
+++ b/qml/pages/SteamPage.qml
@@ -386,18 +386,19 @@ Page {
                                     if (!isSteaming) {
                                         MainController.startSteamHeating("live-pitcher-click")
                                     } else {
-                                        // Sync the live slider imperatively. Its binding to
-                                        // Settings.steamFlow gets broken the first time the
-                                        // user drags it (onValueModified writes value=newVal),
-                                        // so a Settings update from this preset tap won't
-                                        // reach the slider on its own.
-                                        steamingFlowSlider.value = flow
-                                        // Settings assignment alone only updates QSettings; the BLE
-                                        // MMR write to 0x803828 lives in MainController. Push the
-                                        // new flow immediately so the DE1 picks it up mid-session.
-                                        // Don't push timeout — the DE1's session timer is already
-                                        // running on the prior value and a mid-session timeout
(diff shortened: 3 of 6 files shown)


skialpine:
### Code review

Found 1 issue:

1. Code comments assert a confirmed firmware deficiency ("a single MMR write isn't reliably picked up by the firmware's sample-tick loop") as the rationale for `writeMMRVerified`, but the PR description explicitly says on-device testing debunked that theory — "zero retries needed across many drags." Future readers asking "why the verify-and-retry instead of plain `writeMMR`?" will be misled to believe it's a workaround for an observed firmware bug, when it's actually defensive forward-engineering. Reword to "defensive" or "insurance" framing (CLAUDE.md: "Only add [comments] when the WHY is non-obvious" — and the WHY should be accurate).

https://github.com/Kulitorum/Decenza/blob/fbaf2086729a7d1b24e776d0fb7c96d25942f384/src/ble/de1device.h#L228-L240

https://github.com/Kulitorum/Decenza/blob/fbaf2086729a7d1b24e776d0fb7c96d25942f384/src/controllers/maincontroller.cpp#L1413-L1423

<details>

- [Worth a look] `steamingFlowSlider.value = flow` is imperatively assigned from the mid-session pill handler while the slider has a declarative binding to `Settings.steamFlow`. QML permanently breaks the binding on first imperative assignment. Works today because the slider's own handlers are imperative-first, but any other code path that expects the binding to update the slider via `Settings.steamFlow` will silently stop working. Use `Qt.binding(() => Settings.steamFlow)` if you want to preserve reactivity. (score: 62)
- [OK to skip] `writeMMRVerified` uses a 50ms `QTimer::singleShot` to pace the read-back after a write. CLAUDE.md flags timers-as-guards, but this is waiting on hardware I/O — not a state guard. A fully event-driven alternative would hook `writeComplete` on `WRITE_TO_MMR`, but that signal doesn't distinguish which MMR address completed, so using it requires tracking pending writes in a separate queue (non-trivial refactor). (score: 65)
- [Worth a look] Stale retry timer after rapid disconnect/reconnect: a 50ms verify timer scheduled before disconnect can fire after reconnect and consume a fresh pending-verify entry with the old expected value. Realistic BLE reconnect takes seconds not milliseconds, so the window is narrow. Tag pending entries with a generation counter if you want full safety. (score: 50)
- [OK to skip] `parseMMRResponse` function-entry guard `data.size() < 5` is inconsistent with the new verify block's implicit `data.size() >= 8` requirement. Safe today because real DE1 responses are always ≥ 8 bytes, but fragile — a future handler added between existing ones could assume function-entry-guard alone and read out of bounds. Tighten the function-entry guard or document the per-block minimum. (score: 50)

</details>

Worth noting separately: profile uploads already verify writes, but via BLE **ACK sequence tracking** ([de1device.cpp#L970-L1009](https://github.com/Kulitorum/Decenza/blob/fbaf2086729a7d1b24e776d0fb7c96d25942f384/src/ble/de1device.cpp#L970-L1009)), not value-readback. ShotSettings verifies via **read-back with drift auto-heal** ([de1device.cpp#L1340-L1345](https://github.com/Kulitorum/Decenza/blob/fbaf2086729a7d1b24e776d0fb7c96d25942f384/src/ble/de1device.cpp#L1340-L1345)). `writeMMRVerified` introduces a third pattern (write + timer + readback + retry). If the real need is just "catch silent write drops mid-session," the existing ACK-sequence pattern would work and is cheaper; if you want "catch firmware-didn't-store-value", readback is the only option. Given the PR's own testing showed zero retries needed, the ACK-level pattern may be sufficient for steam flow.

PR #793, opened 2026-04-18
feat(scale): add keepScaleOn toggle for disconnect on DE1 sleep

skialpine:
## Summary

Adds a new **Settings.keepScaleOn** toggle on the Connections settings page (below the Scan for Devices button). When enabled (default), behaviour is unchanged: the scale stays BLE-connected when the DE1 sleeps and only its LCD is disabled. When disabled, the scale is powered off via its native BLE sleep command and the BLE link is dropped, matching de1app's default behaviour.

Motivated by a user report: a Decent HDS scale left on battery overnight drained fully because Decenza kept the BLE connection alive. This toggle lets users with battery-only scales avoid that drain while preserving the current fast-wake behaviour for everyone else.

## Default

**True** — preserves current Decenza behaviour on upgrade. Users with battery-only scales can flip the toggle in Settings → Connections.

## How it works

- On DE1 phase → Sleep, with \`keepScaleOn=false\`:
  1. Call \`physicalScale->sleep()\` — sends the scale's power-off command.
  2. On \`ScaleDevice::sleepCompleted\`, call \`physicalScale->disconnectFromScale()\` to drop the BLE link.
  3. Saved scale address stays in Settings; existing reconnect-on-resume and DE1-wake paths reconnect automatically.
- Event-driven (no wall-clock timer) per CLAUDE.md's no-timers-as-guards rule.

## Scale coverage

| Scale | Native sleep command | Benefit |
|-------|:-:|-------|
| Decent (HDS) | Yes (\`0A 02 00\`) | Explicit power-off + BLE disconnect |
| Skale | Yes | Explicit power-off + BLE disconnect |
| Eureka Precisa | Yes | Explicit power-off + BLE disconnect |
| Acaia / Bookoo / DiFluid / Felicita / Hiroia / SmartChef / SoloBarista / Timemore / VariaAku / AtomheartEclair | No (inherits no-op base) | BLE disconnect only — still lets most peripherals drop into low-power advertising |

## Test plan

- [ ] With \`keepScaleOn=true\` (default): DE1 → Sleep still sends LCD-off only; scale stays connected. No regression.
- [ ] With \`keepScaleOn=false\` on a Decent scale: DE1 → Sleep sends \`0A 02\`, BLE disconnects within ~1 s. Scale powers off.
- [ ] DE1 wake (Idle phase): existing reconnect path restores the scale connection.
- [ ] Non-Decent scale (e.g. Acaia): toggle still causes BLE disconnect on DE1 sleep; scale enters low-power mode on peripheral side.
- [ ] Toggle UI on Settings → Connections page, directly under Scan for Devices, reads/writes \`Settings.keepScaleOn\`.
- [ ] Setting persists across app restarts.

### DIFF
--- a/qml/pages/settings/SettingsConnectionsTab.qml
+++ b/qml/pages/settings/SettingsConnectionsTab.qml
@@ -777,6 +777,47 @@ Item {
                         }
                     }
 
+                    // Keep scale connected when DE1 sleeps
+                    ColumnLayout {
+                        Layout.fillWidth: true
+                        spacing: Theme.scaled(4)
+
+                        RowLayout {
+                            Layout.fillWidth: true
+
+                            Text {
+                                text: TranslationManager.translate("settings.bluetooth.keepScaleOn",
+                                                                   "Keep scale connected when DE1 sleeps")
+                                color: Theme.textColor
+                                font.family: Theme.bodyFont.family
+                                font.pixelSize: Theme.scaled(14)
+                                Accessible.ignored: true
+                            }
+
+                            Item { Layout.fillWidth: true }
+
+                            StyledSwitch {
+                                id: keepScaleOnSwitch
+                                checked: Settings.keepScaleOn
+                                accessibleName: TranslationManager.translate(
+                                    "settings.bluetooth.keepScaleOn",
+                                    "Keep scale connected when DE1 sleeps")
+                                onToggled: Settings.keepScaleOn = checked
+                            }
+                        }
+
+                        Text {
+                            Layout.fillWidth: true
+                            text: TranslationManager.translate(
+                                "settings.bluetooth.keepScaleOnDesc",
+                                "Turn off to power down and disconnect the scale when the DE1 sleeps. Recommended for battery-only scales; reconnects automatically when the DE1 wakes.")
+                            color: Theme.textSecondaryColor
+                            font.pixelSize: Theme.scaled(12)
+                            wrapMode: Text.WordWrap
+                            Accessible.ignored: true
+                        }
+                    }
+
                     // Connected BLE scale name + battery
                     RowLayout {
                         Layout.fillWidth: true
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Found 1 issue:

1. The `keepScaleOn=false` path disconnects the scale, but the existing `ScaleDevice::connectedChanged` else-branch immediately schedules `scaleReconnectTimer.start(reconnectDelays[0])` (5 s) because `Settings.scaleAddress` is still set — so 5 s after the DE1 sleeps, the app reconnects to the scale and the feature does nothing. This needs a "deliberate disconnect" flag that the reconnect handler checks, cleared on DE1 wake or app resume.

https://github.com/Kulitorum/Decenza/blob/e232ad6bbb5aa1b7d869393909d40fca8b80ec9e/src/main.cpp#L2004-L2015

See interaction with https://github.com/Kulitorum/Decenza/blob/e232ad6bbb5aa1b7d869393909d40fca8b80ec9e/src/main.cpp#L1240-L1248 (existing 5 s / 30 s / 60 s retry schedule from `reconnectDelays`).

<details>

- [Worth a look] Stale pre-PR comment at [`src/main.cpp:L1978-L1986`](https://github.com/Kulitorum/Decenza/blob/e232ad6bbb5aa1b7d869393909d40fca8b80ec9e/src/main.cpp#L1978-L1986) says "Uses disableLcd() instead of sleep() to keep BLE connected — no reconnection needed on wake". Accurate for the `keepScaleOn=true` arm only; should be updated to describe both branches (score: 70)
- [OK to skip] `sleepCompleted` may not fire if the BLE link drops between the sleep write and its ack — harmless because the parallel `connectedChanged` handler still performs the FlowScale-fallback cleanup (score: 50)
- [OK to skip] If phase transitions Sleep → Sleep without intervening state, a second `Qt::SingleShotConnection` can accumulate on `sleepCompleted` — benign because `disconnectFromScale()` is idempotent (score: 50)

</details>

PR #795, opened 2026-04-18
feat(espresso): +10g button to extend SAW mid-shot (#792)

### LINKED ISSUE #792: [FR} Salvage Shot button - and/or AI way to over-ride stop at weight on-the-fly
Many times I have ground too coarse and instead of wasting the shot, throwing it away, the shot could be turned into a "spontaneous allonge" 

it would work like this:

AUTO
1/ start your shot, target weight 36g
2/flow is much too high by mistake
3/ system recognises flow is much too high for a 36g stop, so adds to that weight target according to flow, ie a sliding scale from 36 to 90g

MANUAL
1/ ... 
2/ ...
3/ system recognises flow is too high, offers "add 10g" button to repeatedly press to reach desired stop at weight.

USING UNIVERSAL GRIND SETTING
From knowlege AI has gained knowing UGS units needed to make changes to flow for that profile, it would be able to offer an UGS grind setting correction, and if you're calibrated/has a lot of varied learnings for your grinder already, it would offer a grind setting for your grinder.

skialpine:
Closes #792 (manual variant — AI/auto-detect and UGS branches deferred).

## Summary
- Adds a **+10 g** button next to **Skip** on EspressoPage. Each press extends the live stop-at-weight target by 10 g, so a too-fast shot can be salvaged as a "spontaneous allonge" instead of binned.
- Wires `MachineState::targetWeightChanged` → `WeightProcessor` on its worker thread so mid-shot target changes actually move the SAW trigger. Previously `m_targetWeight` was set once in `configure()` and documented as read-only during extraction.
- `MainController::bumpTargetWeight()` updates `MachineState` only — the persisted profile/setting is **not** mutated, so the next shot starts at the user's normal target. Button is hidden when `targetWeight == 0` (e.g. basic profile without SAW).
- Display bindings on the weight column, brew-by-ratio label, CupFillView, and accessibility readouts switched from `ProfileManager.targetWeight` to `MachineState.targetWeight` so the bumped target shows live (progress bar, "/ 46 g", cup fill ceiling).

## Drive-by: Skip button fixed in simulator
While exploring I noticed Skip does nothing in simulator mode. `DE1Device::requestState()`'s simulator switch had no case for `DE1::State::SkipToNext` and was silently falling into `default: break;`. Added `DE1Simulator::skipFrame()` (calls existing `advanceToNextFrame()`, no-op outside Preinfusion/Pouring) and the missing switch case. Bundled here because it's tiny and was blocking simulator testing of +10 alongside Skip.

## Files
- `qml/pages/EspressoPage.qml` — new button + display bindings + tab order (Back → +10 → Skip → view-mode)
- `src/controllers/maincontroller.{h,cpp}` — `bumpTargetWeight(deltaG)`
- `src/machine/weightprocessor.{h,cpp}` — `setTargetWeight()` slot
- `src/main.cpp` — connect `targetWeightChanged` → worker via `Qt::QueuedConnection`
- `src/ble/de1device.cpp`, `src/simulator/de1simulator.{h,cpp}` — simulator Skip fix

## Test plan
- [ ] **Simulator — Skip regression**: start espresso, press Skip mid-shot, confirm logs show frame advance.
- [ ] **Simulator — +10 button**: start espresso with SAW set, press **+10** during Preinfusion/Pouring; confirm progress-bar label "/ X g" increments by 10 each press, CupFillView ceiling updates, qInfo log shows old → new target on both MainController and WeightProcessor.
- [ ] **Visibility gating**: button hidden in Idle/Heating/Ready and when targetWeight === 0 (basic profile with no SAW).
- [ ] **Real DE1**: SAW = 36 g, start shot, press **+10** twice → shot stops at ≈56 g (not 36).
- [ ] **No persistence**: subsequent shot still uses the persisted 36 g target (profile/Settings untouched).
- [ ] **Brew-by-ratio**: ratio label "1:2.0 (36g)" updates the parenthetical to reflect bumped target.
- [ ] **Accessibility**: TalkBack/VoiceOver — focus order Back → +10 → Skip → view-mode; +10 announces "Add 10 grams to weight target".
- [ ] **Existing tests**: ctest -R weight and ctest -R machinestate pass with -DBUILD_TESTS=ON.

## Out of scope (per #792, deferred)
- AUTO branch: AI/flow-detection auto-extends SAW.
- UGS grind correction suggestions.
- Configurable bump amount (+5/+10/+20).

### DIFF
--- a/qml/components/AccessibleButton.qml
+++ b/qml/components/AccessibleButton.qml
@@ -21,6 +21,10 @@ Button {
     // For external reference
     property Item accessibleItem: root
 
+    // Live press state — Button.down stays false because touchArea below accepts
+    // the press. Bind custom background visuals to this instead of `down`.
+    readonly property alias isPressed: touchArea.pressed
+
     // Optional font overrides (0/-1 = use defaults)
     property int _customFontSize: 0
     property int _customFontWeight: -1
--- a/qml/components/ShotGraph.qml
+++ b/qml/components/ShotGraph.qml
@@ -140,8 +140,13 @@ ChartView {
         id: weightAxis
         min: 0
         // Live shots may bump SAW past the configured target (#792 +10g button), so
-        // take the larger of profile target and current MachineState target.
-        max: Math.max(10, Math.max(ProfileManager.targetWeight || 0, MachineState.targetWeight || 0, 36) * 1.1)
+        // take the larger of profile target and current MachineState target. Each
+        // source uses an explicit > 0 check because targetWeight == 0 means SAW
+        // disabled, and JS `||` would conflate that with "no data".
+        max: Math.max(10, Math.max(
+            ProfileManager.targetWeight > 0 ? ProfileManager.targetWeight : 0,
+            MachineState.targetWeight > 0 ? MachineState.targetWeight : 0,
+            36) * 1.1)
         tickCount: 5
         visible: false
     }
--- a/qml/components/ShotGraph.qml
+++ b/qml/components/ShotGraph.qml
@@ -139,7 +139,9 @@ ChartView {
     ValueAxis {
         id: weightAxis
         min: 0
-        max: Math.max(10, (ProfileManager.targetWeight || 36) * 1.1)
+        // Live shots may bump SAW past the configured target (#792 +10g button), so
+        // take the larger of profile target and current MachineState target.
+        max: Math.max(10, Math.max(ProfileManager.targetWeight || 0, MachineState.targetWeight || 0, 36) * 1.1)
         tickCount: 5
         visible: false
     }
--- a/qml/pages/EspressoPage.qml
+++ b/qml/pages/EspressoPage.qml
@@ -1001,8 +1001,9 @@ Page {
                 onClicked: MainController.bumpTargetWeight(10.0)
 
                 background: Rectangle {
+                    implicitWidth: Theme.scaled(56)
                     implicitHeight: Theme.scaled(24)
-                    color: addTenButton.down ? Qt.darker(Theme.accentColor, 1.3) : Theme.accentColor
+                    color: addTenButton.isPressed ? Qt.darker(Theme.accentColor, 1.3) : Theme.accentColor
                     radius: Theme.scaled(12)
                 }
                 contentItem: Text {
@@ -1036,8 +1037,9 @@ Page {
                 onClicked: DE1Device.skipToNextFrame()
 
                 background: Rectangle {
+                    implicitWidth: Theme.scaled(56)
                     implicitHeight: Theme.scaled(24)
-                    color: skipFrameButton.down ? Qt.darker(Theme.accentColor, 1.3) : Theme.accentColor
+                    color: skipFrameButton.isPressed ? Qt.darker(Theme.accentColor, 1.3) : Theme.accentColor
                     radius: Theme.scaled(12)
                 }
                 contentItem: Text {
--- a/qml/pages/EspressoPage.qml
+++ b/qml/pages/EspressoPage.qml
@@ -981,13 +981,15 @@ Page {
             }
 
             // Bump SAW target by +10g (issue #792 — "salvage" a too-fast shot)
-            Rectangle {
+            AccessibleButton {
                 id: addTenButton
                 Layout.preferredWidth: Theme.scaled(56)
                 Layout.preferredHeight: Theme.scaled(24)
                 Layout.alignment: Qt.AlignVCenter
-                radius: Theme.scaled(12)
-                color: addTenTapHandler.pressed ? Qt.darker(Theme.accentColor, 1.3) : Theme.accentColor
+                leftPadding: 0
+                rightPadding: 0
+                text: TranslationManager.translate("espresso.button.add10", "+10 g")
+                accessibleName: TranslationManager.translate("espresso.accessible.add10", "Add 10 grams to weight target")
                 visible: MachineState.targetWeight > 0 &&
                          (MachineState.phase === MachineStateType.Phase.Preinfusion ||
                           MachineState.phase === MachineStateType.Phase.Pouring)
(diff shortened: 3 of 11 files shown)


skialpine:
### Code review

Found 4 issues:

1. **ShotGraph weight Y-axis still uses `ProfileManager.targetWeight`, won't expand on +10 bump** — The PR switched EspressoPage display bindings to `MachineState.targetWeight` so the bumped target shows live, but the chart Y-axis ceiling in ShotGraph is unchanged. With a 36 g target the axis caps at 39.6 g; a single +10 press makes the target 46 g and the weight line will plot off the top of the chart.

https://github.com/Kulitorum/Decenza/blob/a2326622a35ca549422e3300000bddb834936857/qml/components/ShotGraph.qml#L140-L144

2. **`KeyNavigation.backtab` from `skipFrameButton` lands on a possibly-invisible `addTenButton`** — `addTenButton` is gated on `MachineState.targetWeight > 0` AND phase, while `skipFrameButton` only gates on phase. For basic profiles without SAW the skip button is visible but +10 is hidden, so Shift+Tab from skip targets a hidden item. Match the pattern at line 408 (`viewModeMouseArea`'s backtab): `KeyNavigation.backtab: addTenButton.visible ? addTenButton : espressoBackButton`.

https://github.com/Kulitorum/Decenza/blob/a2326622a35ca549422e3300000bddb834936857/qml/pages/EspressoPage.qml#L1028-L1035

3. **New `addTenButton` uses raw `Rectangle + TapHandler` instead of `AccessibleButton`** (CLAUDE.md says "Prefer `AccessibleButton` or `AccessibleMouseArea` over raw `Rectangle+MouseArea`"). The accessibility properties are all present manually, but per CLAUDE.md "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'", both `addTenButton` and the sibling `skipFrameButton` should migrate to `AccessibleButton`.

https://github.com/Kulitorum/Decenza/blob/a2326622a35ca549422e3300000bddb834936857/qml/pages/EspressoPage.qml#L984-L1018

4. **Misleading \"no-op\" wording in main.cpp wiring comment** — The comment claims the pre-shot `targetWeightChanged` signals are a "no-op for them", but the queued `WeightProcessor::setTargetWeight()` call IS made and DOES mutate `m_targetWeight` on the worker thread; it's only harmless because `configure()` overwrites it moments later. Reword to reflect that the call happens but has no functional effect.

https://github.com/Kulitorum/Decenza/blob/a2326622a35ca549422e3300000bddb834936857/src/main.cpp#L681-L692

<details>

- [Worth a look] `weightprocessor.h` class header still says "configure(): called once at shot start with targets and learning data" — outdated now that `setTargetWeight()` also writes `m_targetWeight` (score: 50)
- [Worth a look] `bumpTargetWeight` and `setTargetWeight` doc comments include some WHAT lines (e.g., "No-op outside Preinfusion/Pouring") that the code already expresses; CLAUDE.md says don't explain WHAT (score: 62)

</details>

skialpine:
### Code review (commit e939a4bb)

Found 4 issues:

1. **Pressed-state visual is dead on both pill buttons** — both `addTenButton` and `skipFrameButton` bind their background color to `down`, but `AccessibleButton`'s internal `MouseArea touchArea` accepts the press (`mouse.accepted = true` at AccessibleButton.qml:138-140), so Qt Button's `pressed`/`down` never fires. The darken-on-press feedback that worked with the old `Rectangle + TapHandler` is now non-functional. Fix: either expose `touchArea.pressed` as a `readonly property alias` on AccessibleButton and bind to that, or add a `TapHandler` inside the background Rectangle for local press tracking.

https://github.com/Kulitorum/Decenza/blob/e939a4bbd6e50cff66627e937b114878e0aee793/qml/pages/EspressoPage.qml#L1003-L1008

2. **`bumpTargetWeight` doc dropped its precondition note** — the prior comment included "No-op outside Preinfusion/Pouring or when no SAW target is set." The reword removed that, leaving only the design-rationale line. The function does silently return on those guards (maincontroller.cpp:2049-2055), so callers reading just the header could be surprised. Add the precondition note back as a single line — guards on a public Q_INVOKABLE are part of the contract.

https://github.com/Kulitorum/Decenza/blob/e939a4bbd6e50cff66627e937b114878e0aee793/src/controllers/maincontroller.h#L168-L174

3. **`||` numeric default in ShotGraph instead of `??`** (CLAUDE.md says: "Use `??` not `||` for numeric defaults where 0 is valid. JavaScript `||` treats `0` as falsy. Only use `||` when `0` genuinely means 'no data'"). The 36 g floor inside `Math.max` makes the runtime impact harmless here, but every other `targetWeight` consumer in the codebase uses an explicit `> 0` check. Match that pattern or use `??`.

https://github.com/Kulitorum/Decenza/blob/e939a4bbd6e50cff66627e937b114878e0aee793/qml/components/ShotGraph.qml#L141-L146

4. **Custom `background` Rectangle on both pill buttons sets `implicitHeight` but no `implicitWidth`** — works today only because each button has `Layout.preferredWidth: Theme.scaled(56)`. PR #419 explicitly fixed AccessibleButton size-propagation issues; the current override sidesteps that. Buttons here are inline single-use, so practical risk is low, but adding `implicitWidth: Theme.scaled(56)` to the background makes the geometry self-contained.

https://github.com/Kulitorum/Decenza/blob/e939a4bbd6e50cff66627e937b114878e0aee793/qml/pages/EspressoPage.qml#L1003-L1011

<details>

- [OK to skip] Tap-to-announce in accessibility mode adds a two-tap activation flow vs. the prior single-tap. Intentional behavior of `AccessibleButton` and consistent with the rest of the codebase (score: 25, below 50 threshold but worth flagging for awareness)

</details>

skialpine:
### Code review (commit 77416a4b)

No issues found. All 4 findings from the prior review are addressed; the new `isPressed` alias on AccessibleButton (https://github.com/Kulitorum/Decenza/blob/77416a4b314cc1403099459548d2fd57e826eaf5/qml/components/AccessibleButton.qml#L23-L27) is safe (read-only alias, no binding loops, doesn't affect existing consumers, naming avoids Qt's `Button.pressed`), the ShotGraph ternaries produce correct output for all input cases, and both pill buttons' new `implicitWidth` makes their geometry self-contained.

PR #805, opened 2026-04-18
fix(ble): warn on Linux when CAP_NET_ADMIN is missing

### LINKED ISSUE #804: [Bug] RaspberryOs version stop getting De1+ online despite finding it
### What happened?

[RaspberryOs version stop getting De1+ online despite finding it after Scanning for Devices. 

Default App still works.
Not sure what happened but maybe after a routinely OS (and libraries) upgrade (not even the App itself) it stop working. 

I upgraded to the latest stable (V1.6.6) and took the attached log.

[TerminaLog.txt](https://github.com/user-attachments/files/26860390/TerminaLog.txt)

### Steps to reproduce

It never  changes to "Online" despite message stating "Found" after a Scan

### Decenza version

1.6.6

### Platform

Linux

### Device / OS version

Raspberry Pi 4

### Scale (if applicable)

Skale 2

[TerminaLog.txt](https://github.com/user-attachments/files/26860403/TerminaLog.txt)

### Screenshots or video

Image

### Anything else?

_No response_

skialpine:
## Summary
- Fixes [#804](https://github.com/Kulitorum/Decenza/issues/804): on Raspberry Pi OS (and other Linux), an OS update can strip the `CAP_NET_ADMIN` file capability from the Decenza binary. Without it, BlueZ can't tell random from public BLE addresses and rejects connects to the DE1 (random static address) with `UnknownRemoteDeviceError` — the DE1 is found by scans but never comes online.
- Detects a missing effective `CAP_NET_ADMIN` at `BLEManager` startup via `/proc/self/status` and exposes it to QML as a CONSTANT Q_PROPERTY along with the exact `sudo setcap 'cap_net_admin+eip' <binary>` command.
- Shows a prominent modal dialog (red border, non-dismissible-by-outside-click) at startup with the command in a selectable read-only field plus a Copy button. Dialog is deferred until after the first-run welcome on fresh installs.
- Adds a fallback actionable hint to the connection log when `UnknownRemoteDeviceError` fires (for DE1 and scale BLE transports), in case the user dismisses the dialog or the check somehow misses.

## Test plan
- [ ] On Linux without the capability: launch app → verify modal dialog appears with the exact command and binary path, Copy button puts it on the clipboard.
- [ ] Run the command, restart: dialog no longer appears, DE1 connects normally.
- [ ] On macOS/Windows/Android: no dialog, no regressions (the check is compiled out).
- [ ] On first-run (fresh install) Linux without cap: welcome dialog shows first, capability dialog appears after clicking Continue.
- [ ] Verify the log hint still fires on `UnknownRemoteDeviceError` (e.g. if the user dismissed the dialog) and surfaces in the connection log.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,6 +223,7 @@ set(SOURCES
     src/core/translationmanager.cpp
     src/core/updatechecker.cpp
     src/ble/protocol/binarycodec.cpp
+    src/ble/blecapability.cpp
     src/ble/blemanager.cpp
     src/ble/de1device.cpp
     src/ble/bletransport.cpp
@@ -371,6 +372,7 @@ set(HEADERS
     src/ble/protocol/binarycodec.h
     src/ble/protocol/de1characteristics.h
     src/ble/protocol/decentscaleprotocol.h
+    src/ble/blecapability.h
     src/ble/blemanager.h
     src/ble/de1transport.h
     src/ble/de1device.h
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -1925,6 +1925,7 @@ ApplicationWindow {
 
                 AccessibleButton {
                     text: trLinuxBleCapCopy.text
+                    accessibleName: trLinuxBleCapCopy.text
                     onClicked: {
                         cmdText.selectAll()
                         cmdText.copy()
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -466,6 +466,13 @@ ApplicationWindow {
             // On subsequent launches, still check if storage setup is needed
             // (e.g., after reinstall when QSettings was restored but SAF permission wasn't)
             checkStorageSetup()
+
+            // Linux-only: warn if the binary is missing CAP_NET_ADMIN, which
+            // causes BLE connections to the DE1 to fail silently. Only shown
+            // after first run so the welcome dialog isn't obscured.
+            if (BLEManager.linuxBleCapabilityMissing) {
+                Qt.callLater(function() { linuxBleCapabilityDialog.open() })
+            }
         }
 
         // Initialize sleep countdowns (fresh app start, not auto-woken)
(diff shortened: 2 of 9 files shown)


skialpine:
### Code review

Found 2 issues:

1. Missing `Accessible.*` on the read-only `TextEdit` that displays the setcap command (CLAUDE.md says "every interactive element needs `Accessible.role`, `Accessible.name`, `Accessible.focusable: true`, and `Accessible.onPressAction`" and ACCESSIBILITY.md requires `Accessible.role: Accessible.EditableText` for editable-text components). The field is keyboard-focusable and `selectByMouse: true`, so a screen reader needs role/name/focusable. Precedent exists in `ConversationOverlay.qml` and `SettingsAITab.qml`.

https://github.com/Kulitorum/Decenza/blob/db89553536a8c4909ccfc7da298dd8bed7d666e2/qml/main.qml#L1905-L1921

2. `checkLinuxBleCapability()` runs unconditionally in the `BLEManager` constructor, *before* `main.cpp` calls `bleManager.setDisabled(settings.simulationMode())`. A Linux developer running the simulator without CAP_NET_ADMIN will hit the blocking modal dialog even though BLE is fully disabled. The QML triggers at `main.qml:473` and `main.qml:1992` also don't gate on `BLEManager.disabled`. Fix: check `!BLEManager.disabled && BLEManager.linuxBleCapabilityMissing` at the two QML call sites, or defer the capability check until after `setDisabled`.

https://github.com/Kulitorum/Decenza/blob/db89553536a8c4909ccfc7da298dd8bed7d666e2/src/ble/blemanager.cpp#L46-L54

<details>

- [Worth a look] The Linux hint in both transports fires on every `UnknownRemoteDeviceError`, not only when CAP_NET_ADMIN is actually missing — could gate on `BLEManager::linuxBleCapabilityMissing()` so users who already have the cap aren't told to run setcap when the error is caused by something else (score: 70)
- [Worth a look] Cross-reference `// See BleTransport::onControllerError for context.` in `qtscalebletransport.cpp` is fragile if that function is renamed — the two-line explanation below it stands alone without it (score: 65)
- [OK to skip] On Linux with a crash log AND missing cap, `crashReportDialog` and `linuxBleCapabilityDialog` both open via `Qt.callLater` on the same frame and stack — both remain reachable, minor UX ordering (score: 50)
- [OK to skip] `sudo setcap 'cap_net_admin+eip' <binary>` is built in three places instead of routing through `BLEManager::linuxBleSetcapCommand()` — avoids coupling low-level transports to `BLEManager` (score: 50)

</details>

PR #809, opened 2026-04-19
fix(android): use PackageInstaller session API for self-updates

skialpine:
## Summary

- Replaces the legacy `Intent(ACTION_VIEW)` APK install path with the `PackageInstaller` session API, routed through an `IntentSender` + dynamic `BroadcastReceiver`. Fixes the Samsung / Android 16 bug where users had to tap "Install" twice because a 158 ms lifecycle flicker was dismissing the system install dialog on first tap.
- Removes the brittle `applicationStateChanged` guard (plus the two tracking bools) and the fsync-failure gate that were producing the two observed "two-tap" scenarios.
- Session write runs on a worker thread, so the Qt UI thread is no longer blocked during the ~146 MB APK → session copy.

## Details

The debug log for session -3 on my Samsung SM-X210 / Android 16 showed:

```
[3754.996] APK install intent launched
[3755.154] app resumed from background, install intent guard cleared  ← 158 ms later
[3757.858] user taps Install again → second intent fires
```

The app went inactive→active within 158 ms of the `startActivity` call — far too fast for a human. The heuristic guard cleared while the installer sheet was still transitioning, and something about the focus flip caused the first dialog to be dismissed/hidden before the user could act.

The new path uses `PackageInstaller.Session` + `IntentSender`, which is Android's officially-sanctioned flow and is not subject to the flicker race — the system hosts the dialog and delivers `STATUS_PENDING_USER_ACTION` to our receiver, which forwards `EXTRA_INTENT` to `startActivity`. On API 31+, `setRequireUserAction(USER_ACTION_NOT_REQUIRED)` is set so future updates *can* be silent once this app becomes the package's update owner (currently still requires `UPDATE_PACKAGES_WITHOUT_USER_ACTION` to fully take effect — left out of this PR).

Files:

- `android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java` (new) — validates the APK, registers a `RECEIVER_NOT_EXPORTED` broadcast receiver, dispatches the session create / write / commit to `Thread("ApkInstallerWorker")`.
- `src/core/updatechecker.cpp` — `installApk()` is now a small JNI call into `ApkInstaller.install(activity, apkPath)`. Removed the `FileProvider.getUriForFile` dance, the `applicationStateChanged` listener, the guard check in `downloadAndInstall`, and the fsync-failure return (proceeds with install and lets PackageInstaller reject a truncated APK if it actually happens).
- `src/core/updatechecker.h` — dropped `m_installIntentPending` / `m_installIntentLeftActive`.

No manifest changes. `REQUEST_INSTALL_PACKAGES` was already declared; FileProvider is still declared (used by BLE + ShotServer upload paths). Android-only changes — iOS / desktop paths are untouched.

## Test plan

- [ ] Build Android release, push to SM-X210 / Android 16 where the two-tap bug reproduces. Trigger an update.
- [ ] Verify logcat shows `ApkInstaller: install: session N committed` followed by `install: user confirmation dialog launched` (from the broadcast receiver).
- [ ] First tap should produce a stable install confirmation sheet; second tap should not be required.
- [ ] Confirm UI stays responsive during the APK→session copy (scroll / tap other controls while the worker streams).
- [ ] Check behavior when "Install Unknown Apps" permission is revoked — expect the system's permission redirect flow.
- [ ] Regress-test on a second Android device (ideally non-Samsung) to confirm no regression.

### DIFF
--- a/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java
+++ b/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java
@@ -16,6 +16,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Streams an APK into a PackageInstaller session and commits it. Replaces the
@@ -46,6 +47,11 @@ public class ApkInstaller {
     private static final int INTERNAL_STATUS_WRITE_FAILED    = -101;
     private static final int INTERNAL_STATUS_NO_CONFIRM_INTENT = -102;
 
+    // Guards against concurrent sessions. Both the UpdateChecker and ShotServer
+    // paths share a single BroadcastReceiver whose callbacks route to one global
+    // nativeOnInstallStatus — a second session would corrupt m_installInFlight.
+    private static final AtomicBoolean sInstallInFlight = new AtomicBoolean(false);
+
     /**
      * Implemented in C++. Forwards install status (Android PackageInstaller
      * STATUS_* or INTERNAL_STATUS_*) to {@code UpdateChecker} on the Qt main
@@ -74,6 +80,11 @@ public class ApkInstaller {
         final long apkLen = apk.length();
         if (!apk.exists() || apkLen <= 0) {
             Log.e(TAG, "install: APK missing or empty: " + apkPath);
+            return false;
+        }
+
+        if (!sInstallInFlight.compareAndSet(false, true)) {
+            Log.w(TAG, "install: session already in flight, ignoring duplicate request");
             return false;
         }
 
@@ -114,6 +125,7 @@ public class ApkInstaller {
             sessionId = installer.createSession(params);
         } catch (IOException e) {
             Log.e(TAG, "install: createSession failed: " + e);
+            sInstallInFlight.set(false);
             reportStatus(INTERNAL_STATUS_CREATE_FAILED, e.toString());
             return;
         }
@@ -150,6 +162,7 @@ public class ApkInstaller {
                     Log.w(TAG, "session.abandon() failed: " + e2);
                 }
             }
+            sInstallInFlight.set(false);
             reportStatus(INTERNAL_STATUS_WRITE_FAILED, e.toString());
         } finally {
             if (session != null) {
@@ -202,6 +215,7 @@ public class ApkInstaller {
                 }
                 if (confirm == null) {
                     Log.w(TAG, "install: STATUS_PENDING_USER_ACTION with no EXTRA_INTENT");
+                    sInstallInFlight.set(false);
                     reportStatus(INTERNAL_STATUS_NO_CONFIRM_INTENT,
                             "STATUS_PENDING_USER_ACTION delivered with no EXTRA_INTENT");
                     return;
@@ -212,12 +226,14 @@ public class ApkInstaller {
                     Log.i(TAG, "install: user confirmation dialog launched");
                 } catch (Exception e) {
                     Log.e(TAG, "install: startActivity for confirm failed: " + e);
-                    reportStatus(INTERNAL_STATUS_WRITE_FAILED, "failed to launch install dialog: " + e);
+                    sInstallInFlight.set(false);
+                    reportStatus(INTERNAL_STATUS_NO_CONFIRM_INTENT, "failed to launch install dialog: " + e);
                 }
                 return;
             }
 
             Log.i(TAG, "install: status=" + status + " msg=" + msg);
+            sInstallInFlight.set(false);
             reportStatus(status, msg);
         }
     };
--- a/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java
+++ b/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java
@@ -60,9 +60,13 @@ public class ApkInstaller {
 
     /**
      * Implemented in C++. Forwards install status (Android PackageInstaller
-     * STATUS_* or INTERNAL_STATUS_*) to {@code UpdateChecker} on the Qt main
-     * thread. Registered via {@code QJniEnvironment::registerNativeMethods}
-     * when the first {@code UpdateChecker} is constructed.
+     * STATUS_* or INTERNAL_STATUS_*) to the Qt main thread via a
+     * {@code QMetaObject::invokeMethod} call. Registered by
+     * {@code UpdateChecker} via {@code QJniEnvironment::registerNativeMethods}.
+     *
+     * Both {@code UpdateChecker}-triggered and {@code ShotServer}-triggered
+     * sessions share this callback. The C++ side uses {@code m_installInFlight}
+     * to ignore statuses that belong to sessions it did not dispatch.
      */
     static native void nativeOnInstallStatus(int status, String message);
 
@@ -80,6 +84,26 @@ public class ApkInstaller {
      * activity or path argument is null; the APK file is missing or empty; or a
      * session is already in flight ({@link #sInstallInFlight} is set).
      */
+    /** Returns true if a PackageInstaller session is currently in flight. */
+    public static boolean isInFlight() {
+        return sInstallInFlight.get();
+    }
+
+    private static volatile boolean sNativeRegistered = false;
+
+    /**
+     * Called by UpdateChecker via JNI after successful native method registration.
+     * ShotServer can query this to know whether status callbacks will reach C++.
+     */
+    static void onNativeRegistered() {
+        sNativeRegistered = true;
+    }
+
+    /** Returns true if the JNI status callback is registered and functional. */
+    public static boolean isNativeRegistered() {
+        return sNativeRegistered;
+    }
+
     public static boolean install(Activity activity, String apkPath) {
         if (activity == null || apkPath == null) {
             Log.e(TAG, "install: null activity or path");
@@ -121,6 +145,9 @@ public class ApkInstaller {
             // Catches Error subclasses (OutOfMemoryError, etc.) that escape the
             // narrower catches inside doSessionInstall, ensuring sInstallInFlight
             // is always reset and the C++ side always receives a terminal status.
+            // Reports INTERNAL_STATUS_WRITE_FAILED as a generic unexpected-error
+            // sentinel; the more specific INTERNAL_STATUS_CREATE_FAILED is only
+            // produced by the narrower catches inside doSessionInstall.
             Log.e(TAG, "install: unexpected error in worker: " + t);
             sInstallInFlight.set(false);
             reportStatus(INTERNAL_STATUS_WRITE_FAILED, t.toString());
--- a/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java
+++ b/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java
@@ -42,14 +42,20 @@ public class ApkInstaller {
 
     // Sentinel codes for errors that happen before PackageInstaller produces a
     // STATUS_*. Values are chosen outside the PackageInstaller.STATUS_* range
-    // (STATUS_PENDING_USER_ACTION = -1, STATUS_SUCCESS..STATUS_FAILURE_TIMEOUT = 0..8).
+    // (STATUS_PENDING_USER_ACTION = -1, STATUS_SUCCESS = 0, STATUS_FAILURE = 1,
+    //  STATUS_FAILURE_BLOCKED = 2, STATUS_FAILURE_ABORTED = 3,
+    //  STATUS_FAILURE_INVALID = 4, STATUS_FAILURE_CONFLICT = 5,
+    //  STATUS_FAILURE_STORAGE = 6, STATUS_FAILURE_INCOMPATIBLE = 7,
+    //  STATUS_FAILURE_TIMEOUT = 8 [API 34+]).
     private static final int INTERNAL_STATUS_CREATE_FAILED   = -100;
     private static final int INTERNAL_STATUS_WRITE_FAILED    = -101;
     private static final int INTERNAL_STATUS_NO_CONFIRM_INTENT = -102;
 
-    // Guards against concurrent sessions. Both the UpdateChecker and ShotServer
-    // paths share a single BroadcastReceiver whose callbacks route to one global
-    // nativeOnInstallStatus — a second session would corrupt m_installInFlight.
+    // Guards against concurrent sessions. Both UpdateChecker and ShotServer call
+    // install() but share a single BroadcastReceiver. A ShotServer-triggered
+    // session's terminal status is misdelivered to UpdateChecker.onInstallStatus()
+    // via nativeOnInstallStatus, potentially triggering spurious errors or APK
+    // cleanup. The AtomicBoolean ensures only one session exists at a time.
     private static final AtomicBoolean sInstallInFlight = new AtomicBoolean(false);
 
     /**
@@ -69,6 +75,10 @@ public class ApkInstaller {
      * Returns true on successful dispatch of the worker thread. All subsequent
      * failures (create / write / commit / PackageInstaller terminal statuses)
      * are reported via {@link #nativeOnInstallStatus}.
+     *
+     * Returns false without calling {@link #nativeOnInstallStatus} if: the
+     * activity or path argument is null; the APK file is missing or empty; or a
+     * session is already in flight ({@link #sInstallInFlight} is set).
      */
     public static boolean install(Activity activity, String apkPath) {
         if (activity == null || apkPath == null) {
@@ -105,6 +115,20 @@ public class ApkInstaller {
     }
 
     private static void runSessionInstall(Context appContext, File apk, long apkLen) {
+        try {
+            doSessionInstall(appContext, apk, apkLen);
+        } catch (Throwable t) {
+            // Catches Error subclasses (OutOfMemoryError, etc.) that escape the
+            // narrower catches inside doSessionInstall, ensuring sInstallInFlight
+            // is always reset and the C++ side always receives a terminal status.
+            Log.e(TAG, "install: unexpected error in worker: " + t);
+            sInstallInFlight.set(false);
+            reportStatus(INTERNAL_STATUS_WRITE_FAILED, t.toString());
+        }
+    }
+
+    private static void doSessionInstall(Context appContext, File apk, long apkLen)
+            throws Exception {
         PackageInstaller installer =
                 appContext.getPackageManager().getPackageInstaller();
 
@@ -123,17 +147,24 @@ public class ApkInstaller {
         int sessionId;
         try {
             sessionId = installer.createSession(params);
-        } catch (IOException e) {
+        } catch (IOException | SecurityException | IllegalArgumentException | IllegalStateException e) {
             Log.e(TAG, "install: createSession failed: " + e);
             sInstallInFlight.set(false);
             reportStatus(INTERNAL_STATUS_CREATE_FAILED, e.toString());
             return;
         }
 
-        PackageInstaller.Session session = null;
+        PackageInstaller.Session session;
         try {
             session = installer.openSession(sessionId);
+        } catch (IOException | SecurityException e) {
+            Log.e(TAG, "install: openSession failed: " + e);
+            sInstallInFlight.set(false);
+            reportStatus(INTERNAL_STATUS_CREATE_FAILED, e.toString());
+            return;
+        }
 
+        try {
             try (InputStream in = new FileInputStream(apk);
                  OutputStream out = session.openWrite("base.apk", 0, apkLen)) {
                 byte[] buf = new byte[1 << 16];
@@ -157,17 +188,13 @@ public class ApkInstaller {
             Log.i(TAG, "install: session " + sessionId + " committed (" + apkLen + " bytes)");
         } catch (Exception e) {
             Log.e(TAG, "install: write/commit failed: " + e);
-            if (session != null) {
-                try { session.abandon(); } catch (Exception e2) {
-                    Log.w(TAG, "session.abandon() failed: " + e2);
-                }
+            try { session.abandon(); } catch (Exception e2) {
+                Log.w(TAG, "session.abandon() failed: " + e2);
             }
             sInstallInFlight.set(false);
             reportStatus(INTERNAL_STATUS_WRITE_FAILED, e.toString());
         } finally {
-            if (session != null) {
-                session.close();
-            }
+            session.close();
         }
     }
 
@@ -184,6 +211,9 @@ public class ApkInstaller {
 
     private static volatile boolean sReceiverRegistered = false;
 
+    // Registers the receiver once for the application lifetime. It is
+    // deliberately not unregistered — sessions can outlive individual install()
+    // calls, and ACTION_INSTALL_STATUS is scoped to this package (not exported).
     private static synchronized void registerStatusReceiver(Context appContext) {
         if (sReceiverRegistered) return;
         IntentFilter filter = new IntentFilter(ACTION_INSTALL_STATUS);
@@ -224,6 +254,11 @@ public class ApkInstaller {
                 try {
                     context.getApplicationContext().startActivity(confirm);
                     Log.i(TAG, "install: user confirmation dialog launched");
+                    // sInstallInFlight stays true while the user interacts with
+                    // the confirmation dialog; the terminal STATUS_* broadcast
+                    // clears it below. Note: some OEM ROMs fail to deliver
+                    // STATUS_FAILURE_ABORTED on back-dismiss, which can leave the
+                    // flag stuck until the process restarts.
                 } catch (Exception e) {
                     Log.e(TAG, "install: startActivity for confirm failed: " + e);
                     sInstallInFlight.set(false);
--- a/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java
+++ b/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java
@@ -23,16 +23,34 @@ import java.io.OutputStream;
  * the confirmation dialog was being dismissed by a lifecycle flicker, forcing
  * the user to tap "Install" a second time.
  *
- * On API 31+ we request USER_ACTION_NOT_REQUIRED. When this app is already the
- * package's update owner, Android may apply the update silently; otherwise the
- * system fires STATUS_PENDING_USER_ACTION and we forward its EXTRA_INTENT to
- * startActivity, which is the officially-sanctioned way to surface the
- * confirmation sheet and is not subject to the flicker race.
+ * On API 31+ we request USER_ACTION_NOT_REQUIRED. If this app is the installer
+ * of record for its own package (common after a prior self-install), Android
+ * may apply the update silently; otherwise the system fires
+ * STATUS_PENDING_USER_ACTION and we forward its EXTRA_INTENT to startActivity,
+ * which is the officially-sanctioned way to surface the confirmation sheet and
+ * is not subject to the flicker race.
+ *
+ * Terminal statuses (SUCCESS / FAILURE*) and internal worker-thread errors are
+ * reported back to Qt via {@link #nativeOnInstallStatus}.
  */
 public class ApkInstaller {
     private static final String TAG = "ApkInstaller";
     private static final String ACTION_INSTALL_STATUS =
             "io.github.kulitorum.decenza_de1.APK_INSTALL_STATUS";
+
+    // Sentinel codes for errors that happen before PackageInstaller produces a
+    // STATUS_*. Values are chosen outside the PackageInstaller.STATUS_* range
+    // (STATUS_PENDING_USER_ACTION = -1, STATUS_* = 0..8) to avoid collision.
+    private static final int INTERNAL_STATUS_CREATE_FAILED = -100;
+    private static final int INTERNAL_STATUS_WRITE_FAILED = -101;
+
+    /**
+     * Implemented in C++. Forwards install status (Android PackageInstaller
+     * STATUS_* or INTERNAL_STATUS_*) to {@code UpdateChecker} on the Qt main
+     * thread. Registered via {@code QJniEnvironment::registerNativeMethods}
+     * when the first {@code UpdateChecker} is constructed.
+     */
+    static native void nativeOnInstallStatus(int status, String message);
 
     /**
      * Validates the APK, registers the status receiver, then streams the APK
@@ -40,9 +58,9 @@ public class ApkInstaller {
      * quickly so the Qt UI thread stays responsive during the (potentially
      * multi-second) session write for large APKs.
      *
-     * Returns true on successful dispatch of the worker thread. Write/commit
-     * failures are reported asynchronously via logcat; STATUS_PENDING_USER_ACTION
-     * and terminal install statuses arrive on the registered BroadcastReceiver.
+     * Returns true on successful dispatch of the worker thread. All subsequent
+     * failures (create / write / commit / PackageInstaller terminal statuses)
+     * are reported via {@link #nativeOnInstallStatus}.
      */
     public static boolean install(Activity activity, String apkPath) {
         if (activity == null || apkPath == null) {
@@ -94,6 +112,7 @@ public class ApkInstaller {
             sessionId = installer.createSession(params);
         } catch (IOException e) {
             Log.e(TAG, "install: createSession failed: " + e);
+            reportStatus(INTERNAL_STATUS_CREATE_FAILED, e.toString());
             return;
         }
 
@@ -122,15 +141,27 @@ public class ApkInstaller {
 
             session.commit(pi.getIntentSender());
             Log.i(TAG, "install: session " + sessionId + " committed (" + apkLen + " bytes)");
-        } catch (IOException e) {
+        } catch (Exception e) {
             Log.e(TAG, "install: write/commit failed: " + e);
             if (session != null) {
                 try { session.abandon(); } catch (Exception ignored) {}
             }
+            reportStatus(INTERNAL_STATUS_WRITE_FAILED, e.toString());
         } finally {
             if (session != null) {
                 session.close();
             }
+        }
+    }
+
+    private static void reportStatus(int status, String message) {
+        try {
+            nativeOnInstallStatus(status, message);
+        } catch (UnsatisfiedLinkError e) {
+            // Native side not yet registered (unit tests, or install()
+            // called before UpdateChecker construction). Logged status is
+            // the fallback signal.
+            Log.w(TAG, "reportStatus: native not registered, status=" + status);
         }
     }
 
@@ -158,7 +189,12 @@ public class ApkInstaller {
                     PackageInstaller.EXTRA_STATUS_MESSAGE);
 
             if (status == PackageInstaller.STATUS_PENDING_USER_ACTION) {
-                Intent confirm = intent.getParcelableExtra(Intent.EXTRA_INTENT);
+                Intent confirm;
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                    confirm = intent.getParcelableExtra(Intent.EXTRA_INTENT, Intent.class);
+                } else {
+                    confirm = intent.getParcelableExtra(Intent.EXTRA_INTENT);
+                }
                 if (confirm == null) {
                     Log.w(TAG, "install: STATUS_PENDING_USER_ACTION with no EXTRA_INTENT");
                     return;
@@ -169,11 +205,13 @@ public class ApkInstaller {
                     Log.i(TAG, "install: user confirmation dialog launched");
                 } catch (Exception e) {
                     Log.e(TAG, "install: startActivity for confirm failed: " + e);
+                    reportStatus(INTERNAL_STATUS_WRITE_FAILED, "failed to launch install dialog: " + e);
                 }
                 return;
             }
 
             Log.i(TAG, "install: status=" + status + " msg=" + msg);
+            reportStatus(status, msg);
         }
     };
 }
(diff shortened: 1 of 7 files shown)


skialpine:
### Code review

Found 5 issues (all resolved in 20f3682d):

1. Silent async install failures — worker-thread errors (createSession / openSession / write / commit IOException) and terminal PackageInstaller statuses were only logged to logcat, never surfaced to `m_errorMessage`. CLAUDE.md-aligned concern (silent failures leave the UI in a broken state).

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L110-L166

Resolved by adding a `nativeOnInstallStatus` JNI bridge that hops to the Qt main thread and dispatches to a new `UpdateChecker::onInstallStatus` slot that maps status codes to user-visible errors:

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/src/core/updatechecker.cpp#L27-L72

2. `dismissUpdate()` TOCTOU vs. the worker thread's `FileInputStream` open — dismiss would `QFile::remove` the cached APK even while the worker was about to stream from it, leaving no window for `setSize(apkLen)` mismatch or `FileNotFoundException`. The parallel reasoning in `parseReleaseInfo` already avoided deletion on version-change for the same reason.

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/src/core/updatechecker.cpp#L598-L617

Resolved by adding `m_installInFlight`, set in `installApk()` and cleared when `onInstallStatus` sees a terminal code; `dismissUpdate()` now skips the file delete while install is streaming.

3. `Intent.getParcelableExtra(String)` is deprecated on API 33+ (`targetSdkVersion` is 34). Deprecated form still works today but may silently return null on future versions.

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L191-L197

Resolved by guarding with a `Build.VERSION.SDK_INT >= TIRAMISU` check that picks the type-safe `getParcelableExtra(String, Class)` overload.

4. Stale \"install intent\" terminology + a comment claim the code no longer upholds — \"Without this, PackageInstaller can open a truncated APK\" remained above the fsync calls, but the PR now proceeds on fsync failure. CLAUDE.md calls out this exact kind of comment rot.

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/src/core/updatechecker.cpp#L475-L479

Resolved by softening the comment to \"best-effort ... Failures are non-fatal — if the file really is incomplete, PackageInstaller's verification will reject it.\"

5. Remaining stale comments (\"may still be held by PackageInstaller\" / \"leaves the file for PackageInstaller\" / \"update owner\" conflation with \"installer of record\" / \"reported asynchronously via logcat\") — all reference the pre-session-API model or conflate distinct Android concepts.

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/src/core/updatechecker.cpp#L356-L362

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L20-L35

Resolved across both files.

<details>

- [OK to skip] `setSize(apkLen)` captured up front could go stale if the APK file is mutated before the worker opens it — now fully protected by the new `m_installInFlight` guard, so this can't happen in practice (score: 50)
- [Worth a look, out of scope] `ShotServer::installApk()` in `src/network/shotserver_upload.cpp` still uses the legacy `Intent(ACTION_VIEW) + FileProvider` path and would suffer the same Samsung flicker bug for web-uploaded APKs. Separate code path, separate flow — worth a follow-up PR, not this one (score: 50)

</details>

Discarded below 50: BroadcastReceiver registered for process lifetime (32) — benign, receiver is on `applicationContext` with a static instance, and re-entrancy is gated by `m_downloading` in the C++ caller.

skialpine:
### Code review

Found 6 issues:

1. `m_installInFlight` permanently stuck when `EXTRA_INTENT` is null — when `STATUS_PENDING_USER_ACTION` arrives and `getParcelableExtra(EXTRA_INTENT)` returns `null`, the receiver logs and returns without calling `reportStatus()`. `m_installInFlight` stays `true` forever. `dismissUpdate()` checks `!m_installInFlight` before cleaning up the cached APK, so the user is stuck (stale APK on disk, dismiss blocked) until the app is restarted.

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L196-L202

2. `ShotServer::installApk()` still uses the legacy `Intent(ACTION_VIEW)` + FileProvider path — the web-UI APK upload path was not updated to use the new session API. It will reproduce the Samsung Android 16 double-tap bug that this PR was created to fix, for any user who uploads an APK via the web interface.

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/src/network/shotserver_upload.cpp#L357-L427

3. Stale cached APK leaks to disk after deferred-dismiss + install failure — `dismissUpdate()` intentionally skips file cleanup when `m_installInFlight` is true. But `onInstallStatus()` failure branches (e.g., `STATUS_FAILURE_BLOCKED`) clear `m_installInFlight` and emit an error without removing the file or clearing `m_downloadedApkPath`. If `dismissUpdate()` was called first, the orphaned APK accumulates on device storage with no cleanup trigger.

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/src/core/updatechecker.cpp#L607-L614

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/src/core/updatechecker.cpp#L760-L767

4. `downloadAndInstall()` does not guard against `m_installInFlight` — line 355 checks `m_downloading || m_checking` but not `m_installInFlight`. After the download finishes, `m_downloading` is false. A second tap on "Download & Install" while a session is in flight calls `installApk()` again, creating a duplicate `PackageInstaller` session. `m_installInFlight` is not a `Q_PROPERTY` so QML cannot disable the button to prevent this.

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/src/core/updatechecker.cpp#L353-L358

5. `onInstallStatus` declared `public` instead of in `private slots:` — CLAUDE.md says "Slots: `onEventName()`". The method is invoked via `QMetaObject::invokeMethod(..., Qt::QueuedConnection)` and every sibling `on*` callback in this class (`onReleaseInfoReceived`, `onDownloadProgress`, `onDownloadFinished`, `onPeriodicCheck`) is in `private slots:`. This one is not.

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/src/core/updatechecker.h#L60-L64

6. Class Javadoc claims silent updates may happen when they cannot — lines 26–31 say "Android may apply the update silently" for installer-of-record apps, but silent updates actually require `UPDATE_PACKAGES_WITHOUT_USER_ACTION`, a privileged permission unavailable to sideloaded apps and not declared in the manifest. The PR description itself acknowledges this ("currently still requires `UPDATE_PACKAGES_WITHOUT_USER_ACTION` to fully take effect — left out of this PR"). The Javadoc as written will mislead future maintainers into believing silent updates work today.

https://github.com/Kulitorum/Decenza/blob/20f3682d5842539fa7e759e0ae78210c865585c7/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L24-L33

---

<details>

- [Worth a look] `FLAG_MUTABLE` is guarded on `Build.VERSION_CODES.M` (API 23) instead of `Build.VERSION_CODES.S` (API 31) — functionally harmless (pre-31 PendingIntents are mutable by default, unknown flag bits are ignored), but the wrong constant misleads future readers about when the flag was introduced (score: 68)
- [Worth a look] `STATUS_FAILURE_ABORTED` (user cancel) clears `m_installInFlight` but emits no signal and does not clear `m_errorMessage`, so a stale prior-attempt error can remain visible in the UI even though the user just cancelled rather than encountering a new error (score: 68)

</details>

skialpine:
### Code review

Found 4 issues (all now fixed in eecd4464):

1. **Concurrent install race** — both `UpdateChecker` and `ShotServer` call `ApkInstaller.install()` but share a single `BroadcastReceiver` and JNI callback (`nativeOnInstallStatus → s_activeChecker`). A second session started while the first is in flight corrupts `m_installInFlight` and reintroduces the dismissed-APK cleanup race. Fixed by adding `AtomicBoolean sInstallInFlight` in `ApkInstaller`; `install()` now returns false if a session is already in progress.

https://github.com/Kulitorum/Decenza/blob/379c9666/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L73-L104

2. **Wrong sentinel for `startActivity` failure** — the catch block for `context.startActivity(confirm)` reported `INTERNAL_STATUS_WRITE_FAILED` ("Failed to write the update package") instead of `INTERNAL_STATUS_NO_CONFIRM_INTENT` ("Install dialog could not be launched"), giving the user a misleading error. Fixed.

https://github.com/Kulitorum/Decenza/blob/379c9666/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L227-L231

3. **`handleUpload` file write on main thread** — `file.write(body)` for a 30 MB APK was blocking the Qt main thread (violates CLAUDE.md: "Never run disk I/O on the main thread"). Fixed by moving the write to `QThread::create` with a `QMetaObject::invokeMethod` callback to call `installApk()` and send the response.

https://github.com/Kulitorum/Decenza/blob/379c9666/src/network/shotserver_upload.cpp#L348-L379

4. **Path traversal via `X-Filename` header** — `handleUpload` used the header value directly in the file path (`savePath + "/" + filename`) without stripping path components. A request with `X-Filename: ../../evil.apk` could write outside the intended directory. Fixed with `filename = QFileInfo(filename).fileName()`.

https://github.com/Kulitorum/Decenza/blob/379c9666/src/network/shotserver_upload.cpp#L329-L330

---

<details>

- [Worth a look] `QFile::remove(m_downloadedApkPath)` in `UpdateChecker` runs on the main thread (CLAUDE.md: no disk I/O on main thread) — score: 65. Fixed in the same commit by moving to `QThread::create`.
- [OK to skip] `onInstallStatus` in `public slots:` instead of `private slots:` per CLAUDE.md convention — score: 75 (but must be public; the JNI bridge lambda is a free function outside the class). Already documented with a comment.
- [OK to skip] `s_activeChecker` not explicitly cleared in `UpdateChecker` destructor — score: 25. `QPointer` auto-nulls on destruction; single-instance design prevents the hypothetical second-instance scenario.

</details>

skialpine:
### Code review

Found 6 issues:

1. `/upload` APK path materializes the entire APK as `QByteArray` on the **main thread** before the background worker starts, and bypasses `MAX_UPLOAD_SIZE` entirely (the check on `contentLength` only applies to the streaming path, but APKs arrive via `handleUpload` which reads from `request.mid(headerEndPos + 4)` synchronously)

https://github.com/Kulitorum/Decenza/blob/bd2928c4419d0c9486ee3448098f9a80027b02ae/src/network/shotserver_upload.cpp#L308-L319

2. OEM ROMs that fail to deliver `STATUS_FAILURE_ABORTED` on back-dismiss (noted in the code's own comment) will leave `sInstallInFlight` set permanently for the process lifetime. The next `install()` call returns `false` silently, which `UpdateChecker` maps to the generic install-dispatch error — currently that triggers a "Please enable Install Unknown Apps" message, which is misleading when the real cause is a stuck flag from a dismissed dialog on a previous boot.

https://github.com/Kulitorum/Decenza/blob/bd2928c4419d0c9486ee3448098f9a80027b02ae/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L255-L268

3. ShotServer-triggered installs share the same `BroadcastReceiver` and `nativeOnInstallStatus` → `UpdateChecker::onInstallStatus` path. A terminal `STATUS_*` from a ShotServer session is delivered to whatever `UpdateChecker` is currently registered via `s_activeChecker`, triggering its error/cleanup logic (APK file deletion, `m_installInFlight` reset, error message emission) spuriously. The `AtomicBoolean` only prevents concurrent sessions — it does not disambiguate which caller owns the result.

https://github.com/Kulitorum/Decenza/blob/bd2928c4419d0c9486ee3448098f9a80027b02ae/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L54-L59

4. `emit installationStarted()` fires **before** `installApk()` is called in both code paths. If `installApk` returns early (e.g. `s_nativeRegistrationFailed`, null activity, or `sInstallInFlight` already set), the signal has already been emitted with no corresponding terminal event, leaving the QML install UI in a permanently "started" state with no way to reset.

https://github.com/Kulitorum/Decenza/blob/bd2928c4419d0c9486ee3448098f9a80027b02ae/src/core/updatechecker.cpp#L595-L600

5. `ShotServer::installApk` does not check `s_nativeRegistrationFailed` before calling `ApkInstaller.install()`. If JNI registration failed, the Java method will call `nativeOnInstallStatus` which will throw `UnsatisfiedLinkError` (caught and logged in Java), but the C++ side never gets a status callback. The HTTP response is already sent as 200 OK (or 500 on `!ok`), so the web UI shows success/failure based solely on whether `install()` returned true — but the install silently never completes.

https://github.com/Kulitorum/Decenza/blob/bd2928c4419d0c9486ee3448098f9a80027b02ae/src/network/shotserver_upload.cpp#L383-L420

6. The Javadoc for `nativeOnInstallStatus` says it forwards status to `UpdateChecker` — but ShotServer-triggered sessions also deliver their terminal status through this same method. The Javadoc is misleading to future maintainers about the method's actual scope.

https://github.com/Kulitorum/Decenza/blob/bd2928c4419d0c9486ee3448098f9a80027b02ae/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L62-L67

---

<details>
<summary>1 below-threshold issue (50)</summary>

- [Worth a look] `runSessionInstall`'s `Throwable` catch comment says "any phase" but always reports `INTERNAL_STATUS_WRITE_FAILED` — if the error actually occurred during session creation, the wrong sentinel is used (score: 50)

</details>

skialpine:
### Code review

Found 3 issues:

1. The `install()` Javadoc block (lines 73–86) is orphaned because `isInFlight()` was inserted immediately after it with its own one-line Javadoc on line 87. Javadoc tools and IDE tooltips attach a comment block to the immediately following declaration, so the `isInFlight()` one-liner (line 87) takes precedence for `isInFlight()`, the big block is orphaned, and `install()` (line 107) has no Javadoc at all — the method with the most complex contract is invisible to tooling.

https://github.com/Kulitorum/Decenza/blob/64b4f6cdc4a194fdc13a6b75ee859b2f94f091fd/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L73-L107

2. `session.close()` is called unconditionally in the `finally` block even when the `catch` block has already called `session.abandon()`. On some Android versions, `close()` on an abandoned session throws `IllegalStateException`, which propagates through the outer `Throwable` catch in `runSessionInstall` and results in a second call to `reportStatus(INTERNAL_STATUS_WRITE_FAILED, ...)`. The duplicate is dropped on the C++ side (the `m_installInFlight` guard at `onInstallStatus` line 750 rejects it), but only by accident — on a future code path where `m_installInFlight` logic changes, this double-report would become a bug. Fix: wrap `session.close()` in its own try/catch in the finally block.

https://github.com/Kulitorum/Decenza/blob/64b4f6cdc4a194fdc13a6b75ee859b2f94f091fd/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L215-L226

3. `installationStarted` is emitted when `installApk()` succeeds, but there is no `onInstallationStarted` handler anywhere in QML. With the old `ACTION_VIEW` approach the system install dialog appeared immediately. With the new session API the worker thread may take several seconds to write the APK into the session before Android shows the confirmation dialog, leaving the user with no visual feedback that their tap did anything.

https://github.com/Kulitorum/Decenza/blob/64b4f6cdc4a194fdc13a6b75ee859b2f94f091fd/src/core/updatechecker.cpp#L599-L603

---

<details>
<summary>1 below-threshold issue (50)</summary>

- [Worth a look] `nativeOnInstallStatus` Javadoc says it forwards "PackageInstaller STATUS_*" but `STATUS_PENDING_USER_ACTION` is intercepted entirely in Java and never forwarded — a one-sentence clarification would prevent future confusion when debugging why -1 never appears on the C++ side (score: 50)

</details>

skialpine:
### Code review

Found 4 issues:

1. `sInstallInFlight` permanently stuck on OEM ROMs that skip `STATUS_FAILURE_ABORTED` on back-dismiss — no recovery path other than app restart. `downloadAndInstall()` returns early (line 364) before reaching `installApk()`, so the user sees no error and the "Installing..." spinner stays on-screen indefinitely. The Java comment at lines 292-295 documents the OEM bug but provides no mitigation on the C++ side.

https://github.com/Kulitorum/Decenza/blob/962339398e781101d16a5fbd4015bd23567ae692/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L289-L297

https://github.com/Kulitorum/Decenza/blob/962339398e781101d16a5fbd4015bd23567ae692/src/core/updatechecker.cpp#L362-L366

2. `installationStarted()` signal is dead code — emitted at lines 382 and 602 but nothing connects to it in QML or C++. The `installing` Q_PROPERTY added in this PR replaced its purpose for QML binding, leaving both emit sites unreachable in terms of observable effect.

https://github.com/Kulitorum/Decenza/blob/962339398e781101d16a5fbd4015bd23567ae692/src/core/updatechecker.h#L68-L72

https://github.com/Kulitorum/Decenza/blob/962339398e781101d16a5fbd4015bd23567ae692/src/core/updatechecker.cpp#L380-L384

3. ShotServer-triggered APK install failures are silently dropped — HTTP 200 is sent to the web UI before the async PackageInstaller session completes (line 363), and all terminal `STATUS_FAILURE_*` / `INTERNAL_STATUS_*` codes from Java are discarded at `UpdateChecker::onInstallStatus` with only `qDebug` because `m_installInFlight` is never set for ShotServer-initiated installs. The user believes the install succeeded.

https://github.com/Kulitorum/Decenza/blob/962339398e781101d16a5fbd4015bd23567ae692/src/network/shotserver_upload.cpp#L361-L365

https://github.com/Kulitorum/Decenza/blob/962339398e781101d16a5fbd4015bd23567ae692/src/core/updatechecker.cpp#L749-L755

4. `QFile::remove(outputPath)` in `handleMediaUpload` is disk I/O on the main thread — called directly in the `addPersonalMedia` failure branch. `shotserver_upload.cpp` was modified by this PR; CLAUDE.md says "Never run database or disk I/O on the main thread" and requires fixing pre-existing violations in touched files.

https://github.com/Kulitorum/Decenza/blob/962339398e781101d16a5fbd4015bd23567ae692/src/network/shotserver_upload.cpp#L1174-L1179

---

<details>

- [Worth a look] `handleUpload()` still materializes the full APK as a `QByteArray` on the main thread — only reachable for APK uploads if a request arrives without a `Content-Length` header (the `isApkUpload` flag normally forces the streaming path); score: 68
- [Worth a look] JNI bridge comment says "worker thread / BroadcastReceiver" but `BroadcastReceiver.onReceive()` runs on the Android main thread, not a worker — inaccurate threading context in the comment; score: 70
- [OK to skip] Web-upload filename collision with UpdateChecker's cached APK path — requires simultaneous auto-update download and a web upload with an identical filename; niche race; score: 68

</details>

skialpine:
### Code review

Found 5 issues:

1. `onDownloadFinished()` has 4 synchronous `QFile::remove()` calls on the main thread (CLAUDE.md: "Never run database or disk I/O on the main thread"). The PR correctly moved identical calls in `dismissUpdate()` and `onInstallStatus()` to `QThread::create()` background threads but missed these four error-path calls at lines 488, 531, 549, and 567.

https://github.com/Kulitorum/Decenza/blob/32de30dd756c51908441d77aa08ad0827dedb901/src/core/updatechecker.cpp#L486-L492

2. `pending.tempFile` (`QFile*` allocated with `new QFile(...)` at line 629) is set to `nullptr` at line 701 without `delete`, leaking the heap object on every large upload. The comment "Transfer ownership" is inaccurate — downstream handlers only receive a `QString` path, not the `QFile*`.

https://github.com/Kulitorum/Decenza/blob/32de30dd756c51908441d77aa08ad0827dedb901/src/network/shotserver.cpp#L699-L705

3. `handleMediaUpload()` runs blocking file I/O (QFile::rename/copy/remove, QImage resizing, QPainter) and external FFmpeg/FFprobe `QProcess::waitForFinished()` calls synchronously on the Qt main thread. `shotserver_upload.cpp` was modified by this PR, triggering CLAUDE.md: "Fix pre-existing violations in any file you touch."

https://github.com/Kulitorum/Decenza/blob/32de30dd756c51908441d77aa08ad0827dedb901/src/network/shotserver_upload.cpp#L1071-L1080

4. When `installApk(fullPath)` returns false in `handleUploadFromFile()`, the renamed APK at `fullPath` is left on disk — neither deleted nor tracked for later cleanup, leaving an orphaned file in the cache directory.

https://github.com/Kulitorum/Decenza/blob/32de30dd756c51908441d77aa08ad0827dedb901/src/network/shotserver_upload.cpp#L357-L364

5. The new "Installing update..." `RowLayout` (BusyIndicator + Tr text) added to `SettingsUpdateTab.qml` has no `Accessible.role` or `Accessible.name`, so TalkBack users receive no announcement that an install is in progress. CLAUDE.md: "Fix pre-existing violations in any file you touch."

https://github.com/Kulitorum/Decenza/blob/32de30dd756c51908441d77aa08ad0827dedb901/qml/pages/settings/SettingsUpdateTab.qml#L314-L334

---

<details>

- [OK to skip] `onInstallStatus` declared in `public slots:` — CLAUDE.md only mandates the `onEventName()` naming convention, which is met; slot access level is not specified (score: 50)
- [Worth a look] `sInstallInFlight` permanently stuck on OEM ROMs that skip `STATUS_FAILURE_ABORTED` on back-dismiss — documented in code (ApkInstaller.java:291-295); `downloadAndInstall()` surfaces "restart the app" message as a fallback (score: 50)

</details>

skialpine:
### Code review

Found 1 issue:

1. `QFile::remove(fullPath)` runs on the main thread when `installApk()` fails in `handleUploadFromFile` (CLAUDE.md says "Never run database or disk I/O on the main thread")

https://github.com/Kulitorum/Decenza/blob/3b9a5d2a896cb1f754e2ba90d94a2ee8bb33c66a/src/network/shotserver_upload.cpp#L359-L366

The `invokeMethod` callback at line 359 runs on the Qt main thread. The `QFile::remove(fullPath)` at line 362 (APK failure cleanup) is synchronous disk I/O on that thread. The rest of this PR correctly wraps removes in `QThread::create` lambdas; this path was missed. Fix: move the `QFile::remove` into its own `QThread::create` using the same pattern at lines 344–348.

---

<details>

- [Worth a look] Fast-path guard gap: if `installApk()` returns false on the cached-APK path in `downloadAndInstall()`, neither `m_downloading` nor `m_installInFlight` is set, so the button guard never fires and the user can trigger repeated JNI calls (each safely rejected by Java's `compareAndSet`, but each re-emits `errorMessageChanged`) (score: 65)
- [Worth a look] STATUS_SUCCESS comment doesn't mention the new APK cleanup added by this PR — a reader sees `// app is typically being killed for the upgrade` but then file removal and signal emission below with no explanation (score: 65)
- [Worth a look] `dismissUpdate()` when `m_installInFlight` is true does not emit `installingChanged`, leaving the "Installing update..." spinner visible until the terminal PackageInstaller status arrives (score: 55)

</details>

skialpine:
### Code review

Found 6 issues:

1. `↓` Unicode symbol (U+2193) used as icon in the scroll-indicator button — replace with an SVG `Image` (CLAUDE.md says "Never use Unicode symbols as icons in text... Use SVG icons from `qrc:/icons/` with `Image` instead" and "Fix pre-existing violations in any file you touch")

https://github.com/Kulitorum/Decenza/blob/7044cbe4b10df84d3dff5b1870aade3717b468a5/qml/pages/settings/SettingsUpdateTab.qml#L606-L613

2. `QFileInfo::exists()` and `QFileInfo(path).size()` called synchronously on the main thread in `downloadAndInstall()` — `updatechecker.cpp` was heavily modified by this PR, triggering the "fix pre-existing violations in files you touch" rule (CLAUDE.md says "Never run database or disk I/O on the main thread")

https://github.com/Kulitorum/Decenza/blob/7044cbe4b10df84d3dff5b1870aade3717b468a5/src/core/updatechecker.cpp#L379-L385

3. `dismissUpdate()` background-thread APK deletion has no `m_downloadGeneration` guard, unlike the 4 other `QFile::remove` paths in the same file (lines 490–597 all capture and recheck `capturedGen`). If a periodic auto-check re-fires `startDownload()` for the same version immediately after dismiss, the background thread can delete the file that `startDownload()` just opened for write+truncate.

https://github.com/Kulitorum/Decenza/blob/7044cbe4b10df84d3dff5b1870aade3717b468a5/src/core/updatechecker.cpp#L651-L658

4. ShotServer-triggered APK install failures are silently dropped — HTTP 200 "Upload complete" is sent before the async PackageInstaller session finishes, and all terminal `STATUS_FAILURE_*` / `INTERNAL_STATUS_*` callbacks arrive at `UpdateChecker::onInstallStatus` which returns early because `m_installInFlight` is false for ShotServer-initiated installs. The browser shows "Installing..." indefinitely on failure (write error, incompatible APK, user denial).

https://github.com/Kulitorum/Decenza/blob/7044cbe4b10df84d3dff5b1870aade3717b468a5/src/core/updatechecker.cpp#L782-L790

https://github.com/Kulitorum/Decenza/blob/7044cbe4b10df84d3dff5b1870aade3717b468a5/src/network/shotserver_upload.cpp#L360-L370

5. `ApkInstaller.java` Javadoc describes `m_installInFlight` as a "C++ side" shared mechanism, implying ShotServer has a symmetric filter. Only `UpdateChecker` has this guard; ShotServer makes no attempt to filter the resulting status callbacks. The comment should say "UpdateChecker uses its `m_installInFlight` flag" to make the asymmetry clear.

https://github.com/Kulitorum/Decenza/blob/7044cbe4b10df84d3dff5b1870aade3717b468a5/android/src/io/github/kulitorum/decenza_de1/ApkInstaller.java#L65-L74

6. `cleanupPendingRequest` calls `QFile::remove(pending.tempFilePath)` synchronously on the main thread — `shotserver.cpp` was modified by this PR to add `isApkUpload` tracking and update the guard condition, triggering the "fix pre-existing violations" rule. This can block the main thread while deleting a 30MB+ APK temp file on timed-out or aborted uploads (CLAUDE.md says "Never run database or disk I/O on the main thread")

https://github.com/Kulitorum/Decenza/blob/7044cbe4b10df84d3dff5b1870aade3717b468a5/src/network/shotserver.cpp#L821-L828

---

<details>

- [OK to skip] `onInstallStatus` comment lists `STATUS_PENDING_USER_ACTION=-1` as part of the arriving code range, but the immediately following comment (line 778–779) clarifies it is handled in Java and never forwarded — contradictory but no runtime impact (score: 50)
- [Worth a look] `sendResponse(200)` in `handleUploadFromFile` includes the full internal Android cache path in the body (e.g. `/data/.../cache/Decenza_uploaded.apk`), inconsistent with the media upload which returns a generic `"Media uploaded successfully"` string (score: 55)

</details>

skialpine:
### Code review

Found 3 issues:

1. `dismissUpdate()` clears `m_installInFlight` but the terminal status callback is then dropped by the early-return guard at line 791, leaving the cached APK (~146 MB) on disk forever and `m_downloadedApkPath` set to a stale path. `downloadReadyChanged()` is never emitted. On the next \"Install\" tap the fast-path re-enters with a path that may be stale. Fix: in the new `if (m_installInFlight)` block at lines 666–669, also clear `m_downloadedApkPath`/`m_expectedDownloadSize` and emit `downloadReadyChanged()` (the APK file can be removed at this point — the Java worker finished before the confirmation dialog appeared).

https://github.com/Kulitorum/Decenza/blob/7a71dadb21ef78f752bb368103a3de3aea4c4142/src/core/updatechecker.cpp#L644-L670

2. The fast-path in `downloadAndInstall()` calls `installApk(m_downloadedApkPath)` and discards the return value. When the cached APK was deleted (OS cache eviction), Java's `install()` detects `!apk.exists()` and returns `false` without calling `nativeOnInstallStatus()`. C++ receives `false` from `installApk()`, but `downloadAndInstall()` ignores it and returns silently — no error message, no re-download, UI frozen. Fix: check the return value; on `false`, clear `m_downloadedApkPath`/`m_expectedDownloadSize` and fall through to the fresh-download path.

https://github.com/Kulitorum/Decenza/blob/7a71dadb21ef78f752bb368103a3de3aea4c4142/src/core/updatechecker.cpp#L379-L388

3. `canDownloadUpdate` is declared `CONSTANT` in updatechecker.h, but on desktop (macOS/Windows/Linux) the implementation returns `!m_downloadUrl.isEmpty()`, which is `false` at startup and only becomes `true` after `parseReleaseInfo()` runs. QML evaluates CONSTANT properties once at bind time and never re-reads them, so the \"Download & Install\" button (SettingsUpdateTab.qml:411) is permanently hidden on desktop. This is pre-existing, but `updatechecker.h` was modified by this PR, and [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/7a71dadb21ef78f752bb368103a3de3aea4c4142/CLAUDE.md) requires fixing pre-existing violations in touched files. Fix: change to `NOTIFY updateAvailableChanged` and emit that signal (or a dedicated one) when `m_downloadUrl` is populated.

https://github.com/Kulitorum/Decenza/blob/7a71dadb21ef78f752bb368103a3de3aea4c4142/src/core/updatechecker.h#L24-L26

---

<details>

- [Worth a look] Five `QThread::create` lambdas in `onDownloadFinished`/`dismissUpdate` capture `this` to read `m_downloadGeneration`; the destructor does not join them, so rapid app exit during cleanup is a dangling-pointer UB site — easy to fix by capturing `m_downloadGeneration.loadAcquire()` by value instead of `this` (score: 72)
- [Worth a look] `downloadAndInstall()` checks C++ `m_installInFlight` but not Java-level `ApkInstaller.isInFlight()`; if a ShotServer-triggered session is active, a full ~146 MB download completes before `install()` returns `false` with \"already in progress\" (score: 65)
- [OK to skip] Comment at fast-path says \"it's the right size\" but the size is no longer compared — `m_expectedDownloadSize > 0` is only a sentinel for \"download was started\" (`updatechecker.cpp` ~line 378) (score: 60)
- [OK to skip] `cleanupPendingRequest` comment says \"small uploads < MAX_SMALL_BODY_SIZE don't increment\" but APK uploads now always increment regardless of size (`shotserver.cpp` ~line 832) (score: 55)

</details>

skialpine:
### Code review

Found 2 issues:

1. Partial-download remove in `dismissUpdate()` has no generation guard, unlike every other cleanup lambda in the file. If a new `startDownload()` runs at the same filename before the background thread fires, the thread's `QFile::remove(partialPath)` will delete the newly-opened download's file. The inline comment argues the UI-hidden update card prevents this, but nothing at the C++ API boundary enforces that — programmatic callers (automation, tests, ShotServer) can bypass the UI.

https://github.com/Kulitorum/Decenza/blob/77006624c7cc53a4f10c089261bc2f7a8a8dc0dd/src/core/updatechecker.cpp#L699-L712

2. Pre-existing main-thread `QFileInfo(pending.tempFilePath).size()` stat() syscall inside `onReadyRead` ([CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/77006624c7cc53a4f10c089261bc2f7a8a8dc0dd/CLAUDE.md): \"Never run database or disk I/O on the main thread.\" CLAUDE.md also says \"Fix pre-existing violations in any file you touch\"; `shotserver.cpp` has been modified by this PR). It's only used for a qDebug log — either drop the size from the log or capture it from the QFile position before closing.

https://github.com/Kulitorum/Decenza/blob/77006624c7cc53a4f10c089261bc2f7a8a8dc0dd/src/network/shotserver.cpp#L689-L694

---

<details>

- [Worth a look] `s_downloadGeneration` file-scope comment now describes a contract that only applies to the onDownloadFinished lambdas — dismissUpdate's lambda no longer re-checks, so \"cleanup lambdas compare against the captured value\" is misleading (score: 60)
- [OK to skip] Two safety-net comments in onInstallStatus say the status is \"dropped by the early-return above\" without naming the mechanism (dismissUpdate clears m_installInFlight) — requires the reader to hunt through dismissUpdate to verify (score: 50)
- [OK to skip] ShotServer-triggered install failures (STORAGE, INCOMPATIBLE, etc.) are still silent to the web UI — the HTTP 200 response fires before the async session completes, and onInstallStatus early-returns for !m_installInFlight. Android's native system UI does surface most errors, so user is not completely unaware. Architectural, persistent across passes (score: 50)

</details>

skialpine:
### Code review

Found 4 issues:

1. Main-thread synchronous file read — `QFile f(tempPath); f.open(...); f.readAll(); f.close()` inside `onReadyRead()` on the Qt main thread ([CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/71d64a92b43e7b63b2db2d80e5d56d5fe22f9b78/CLAUDE.md): \"Never run database or disk I/O on the main thread\"). This is the \"Large non-media request with temp file — reconstruct\" fallback, so it can pull an arbitrarily large body into memory while blocking the event loop.

https://github.com/Kulitorum/Decenza/blob/71d64a92b43e7b63b2db2d80e5d56d5fe22f9b78/src/network/shotserver.cpp#L718-L732

2. Main-thread `QFile::exists(pending.tempFilePath)` stat() syscall inside `onReadyRead()` ([CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/71d64a92b43e7b63b2db2d80e5d56d5fe22f9b78/CLAUDE.md): \"Never run database or disk I/O on the main thread\"). Same function as #1, guarding the reconstruct path.

https://github.com/Kulitorum/Decenza/blob/71d64a92b43e7b63b2db2d80e5d56d5fe22f9b78/src/network/shotserver.cpp#L718-L723

3. Main-thread `QFile::open/write/close` sequences writing the request body to a temp file on disk inside `handleRequest()` — two occurrences (small-body `/upload/media` and `/api/backup/restore` paths). Same CLAUDE.md rule as #1/#2.

https://github.com/Kulitorum/Decenza/blob/71d64a92b43e7b63b2db2d80e5d56d5fe22f9b78/src/network/shotserver.cpp#L2179-L2192

4. The file-scope comment on `s_downloadGeneration` asserts \"Every background `QFile::remove` in this file follows the pattern\" but three lambdas in `onInstallStatus` are unguarded: `QThread::create([path]() { QFile::remove(path); })` at the STATUS_SUCCESS branch, STATUS_FAILURE_ABORTED safety-net, and the generic failure safety-net. Either the claim needs to narrow to \"download-phase\" lambdas or the three should be guarded like the other five.

https://github.com/Kulitorum/Decenza/blob/71d64a92b43e7b63b2db2d80e5d56d5fe22f9b78/src/core/updatechecker.cpp#L891-L897

---

<details>

- [Worth a look] Three unguarded `QFile::remove` lambdas in onInstallStatus (SUCCESS, aborted safety-net, generic failure safety-net) — root cause of finding #4; guarding them would resolve the comment inconsistency (score: 65)
- [OK to skip] Two safety-net comments say the block runs \"only for a future code path that clears m_updateAvailable without going through dismissUpdate()\" — but `onCheckFinished` at lines 308-312 already clears m_updateAvailable when newer=false, so the path is present-day reachable (score: 50)

</details>

skialpine:
### Code review

Found 2 issues:

1. Temp-file leak in the new small-body background-thread dispatches. Both `/upload/media` and `/api/backup/restore` paths write the body to `tempPath` on a background thread, then queue a callback guarded by `if (safeThis && safeSocket)`. If the client disconnects (or ShotServer is torn down) between the write completing and the queued callback running, the handler is skipped and `tempPath` is never cleaned up — temp files accumulate in `TempLocation`. The fix is a simple `else QFile::remove(tempPath)` in the guarded branch; the reconstruct path at `onReadyRead` correctly removes the temp file unconditionally on the background thread before invoking the handler.

https://github.com/Kulitorum/Decenza/blob/9b42f4844641e5f2e186f1b09927b5a07ae88a66/src/network/shotserver.cpp#L2217-L2226

2. Two new safety-net comments in `onInstallStatus` reference `onCheckFinished` — a function that does not exist in the codebase. The actual functions that set `m_updateAvailable` are `onReleaseInfoReceived` (user-triggered checks) and the lambda in `onPeriodicCheck`, both of which call `parseReleaseInfo`. A maintainer grepping for `onCheckFinished` will find only these comments and conclude the comment is stale.

https://github.com/Kulitorum/Decenza/blob/9b42f4844641e5f2e186f1b09927b5a07ae88a66/src/core/updatechecker.cpp#L913-L919

---

skialpine:
### Code review

Found 1 issue:

1. Pass-17's fix for the temp-file leak added `else { QFile::remove(tempPath); }` inside a `QMetaObject::invokeMethod(safeThis, ..., Qt::QueuedConnection)` callback. That callback runs on `safeThis`'s thread — the Qt main thread — so the cleanup ended up being synchronous main-thread disk I/O, the very violation the earlier passes worked to eliminate ([CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/312867c22edc4a56d3eb020f980284faa1c1b987/CLAUDE.md): \"Never run database or disk I/O on the main thread\"). Both new occurrences are affected.

https://github.com/Kulitorum/Decenza/blob/312867c22edc4a56d3eb020f980284faa1c1b987/src/network/shotserver.cpp#L2219-L2227

https://github.com/Kulitorum/Decenza/blob/312867c22edc4a56d3eb020f980284faa1c1b987/src/network/shotserver.cpp#L2376-L2384

Fix: dispatch the cleanup to a `QThread::create` lambda (the same pattern used in `cleanupPendingRequest`).

Also noted (comment-only, no functional impact): the pass-17 fix for `onCheckFinished` caught only one of the two occurrences; the second at line 973 still references the phantom function name.

https://github.com/Kulitorum/Decenza/blob/312867c22edc4a56d3eb020f980284faa1c1b987/src/core/updatechecker.cpp#L971-L977

---

PR #810, opened 2026-04-19
fix(ble): follow-up Linux BLE diagnostics after #805

### LINKED ISSUE #804: [Bug] RaspberryOs version stop getting De1+ online despite finding it

skialpine:
## Summary

Follow-up to #805 covering the case where CAP_NET_ADMIN is effective but BLE still fails — the exact situation reported in #804, where Qt's misleading `"Missing CAP_NET_ADMIN permission"` warning sent the user down the setcap rabbit hole while the real cause was a stale BlueZ cache.

Three changes, all Linux-only (no-op on macOS/Windows/Android):

### 1. `BtLogFilter` — suppress Qt's false cap warning
New chained `qInstallMessageHandler` in `src/core/btlogfilter.{h,cpp}`. When `BleCapability::linuxMissing()` is false and the incoming message is Qt's `"Missing CAP_NET_ADMIN permission"` warning, drop it on the floor. Installed in `main.cpp` right after `WebDebugLogger` so it sees Bluetooth warnings before they propagate.

### 2. BlueZ-cache recovery dialog
When `UnknownRemoteDeviceError` fires on a DE1 or scale connect attempt **and** caps are effective, the host BlueZ state is almost always stale. A new modal surfaces the recovery commands:

```
MAC=$(bluetoothctl devices | awk '/DE1/ {print $2; exit}')
bluetoothctl remove "$MAC"
sudo systemctl restart bluetooth
```

with a Copy button. Plumbed via a new `BLEManager::linuxBlueZCacheHintNeeded()` signal, a public `requestBluezCacheHint()` method (de-dupes via `BleCapability::takeBluezCacheHintToken()`), and a new static `BLEManager::instance()` so the transport layer can reach it without a constructor change. Dialog modeled on the setcap dialog from #805 and gated on `!disabled && !linuxBleCapabilityMissing` so it never stacks with it.

### 3. One-shot BT diagnostics dump
First time any BLE transport error fires in a session, `BleCapability::logLinuxBtDiagnosticsOnce()` writes a block to the debug log via `qInfo()`:

- `CapEff`/`CapBnd`/`CapInh`/`CapAmb` from `/proc/self/status`
- `getcap` on the binary
- `bluetoothctl --version`
- `hciconfig -a` (if installed)

Guarded by `std::once_flag`. Since #802/#803 wired the issue template to attach the debug log, these diagnostics flow into every future Linux BLE bug report for free — no UI surface needed, no support round-trips.

## Test plan
- [ ] **Caps missing (existing #805 behavior unchanged):** `sudo setcap -r` → launch → confirm setcap dialog still opens
- [ ] **Caps present + Qt false-alarm suppressed:** `sudo setcap 'cap_net_admin+eip' Decenza_DE1` → launch with `QT_LOGGING_RULES="qt.bluetooth.warning=true"` → scan → confirm the `"Missing CAP_NET_ADMIN"` line does **not** appear in stdout/debug log
- [ ] **BlueZ-cache dialog:** caps present, power off DE1 to force `UnknownRemoteDeviceError` → confirm new dialog opens with recovery commands; Copy works; fires once per session; does not stack with setcap dialog
- [ ] **Diagnostics dump:** trigger any transport error → confirm single `BtDiagnostics:` block in debug log with all four sections; second error does **not** re-dump
- [ ] **Non-Linux no-op:** build/run on macOS — confirm no new code paths run, no new log lines, and the new QML connection is silent (signal never emitted)

Closes #804.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -246,6 +246,7 @@ set(SOURCES
     src/ble/scales/flowscale.cpp
     src/ble/transport/qtscalebletransport.cpp
     src/core/asynclogger.cpp
+    src/core/btlogfilter.cpp
     src/machine/machinestate.cpp
     src/machine/weightprocessor.cpp
     src/profile/profile.cpp
@@ -396,6 +397,7 @@ set(HEADERS
     src/ble/transport/scalebletransport.h
     src/ble/transport/qtscalebletransport.h
     src/core/asynclogger.h
+    src/core/btlogfilter.h
     src/machine/machinestate.h
     src/machine/weightprocessor.h
     src/profile/profile.h
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -1959,6 +1959,113 @@ ApplicationWindow {
         }
     }
 
+    // Linux BLE BlueZ-cache recovery hint — shown when UnknownRemoteDeviceError
+    // fires on a DE1 connection attempt while CAP_NET_ADMIN is effective. That
+    // combination almost always means the host-side BlueZ state is stale
+    // (cached address type, expired pairing) after an OS upgrade or similar.
+    // Gated to Linux + caps-OK; the setcap dialog above covers the caps-missing
+    // case.
+    Connections {
+        target: BLEManager
+        function onLinuxBlueZCacheHintNeeded() {
+            if (BLEManager.disabled) return
+            if (BLEManager.linuxBleCapabilityMissing) return  // setcap dialog takes precedence
+            Qt.callLater(function() { linuxBleBluezCacheDialog.open() })
+        }
+    }
+
+    Dialog {
+        id: linuxBleBluezCacheDialog
+        modal: true
+        dim: true
+        anchors.centerIn: parent
+        width: Theme.dialogWidth + 2 * padding
+        closePolicy: Dialog.NoAutoClose
+        padding: Theme.dialogPadding
+
+        background: Rectangle {
+            color: Theme.surfaceColor
+            radius: Theme.cardRadius
+            border.width: 2
+            border.color: Theme.errorColor
+        }
+
+        Tr { id: trBluezCacheTitle; key: "main.dialog.linuxBleBluezCache.title"; fallback: "Can't connect to DE1 — try clearing the Bluetooth cache"; visible: false }
+        Tr { id: trBluezCacheMessage; key: "main.dialog.linuxBleBluezCache.message"; fallback: "The DE1 was discovered but the Bluetooth stack rejected the connection (\"Remote device not found\"). This usually means BlueZ cached stale pairing or address information after an OS update.\n\nIn a terminal, remove the cached entry and restart the Bluetooth service, then power-cycle the DE1 and try again:"; visible: false }
+        Tr { id: trBluezCacheCopy; key: "main.dialog.linuxBleBluezCache.copy"; fallback: "Copy commands"; visible: false }
+        Tr { id: trBluezCacheCommandField; key: "main.dialog.linuxBleBluezCache.commandField"; fallback: "BlueZ recovery commands"; visible: false }
+
+        contentItem: Column {
+            spacing: Theme.spacingLarge
+
+            Text {
+                text: trBluezCacheTitle.text
+                font: Theme.subtitleFont
+                color: Theme.textColor
+                anchors.horizontalCenter: parent.horizontalCenter
+                wrapMode: Text.Wrap
+                width: parent.width
+                horizontalAlignment: Text.AlignHCenter
+            }
+
+            Text {
+                text: trBluezCacheMessage.text
+                wrapMode: Text.Wrap
+                width: parent.width
+                font: Theme.bodyFont
+                color: Theme.textColor
+            }
+
+            Rectangle {
+                width: parent.width
+                color: Theme.backgroundColor
+                radius: Theme.cardRadius / 2
+                border.width: 1
+                border.color: Theme.primaryContrastColor
+                height: bluezCmdText.implicitHeight + 2 * Theme.spacingMedium
+
+                TextEdit {
+                    id: bluezCmdText
+                    anchors.fill: parent
+                    anchors.margins: Theme.spacingMedium
+                    text: "MAC=$(bluetoothctl devices | awk '/DE1/ {print $2; exit}')\nbluetoothctl remove \"$MAC\"\nsudo systemctl restart bluetooth"
+                    readOnly: true
+                    selectByMouse: true
+                    wrapMode: TextEdit.Wrap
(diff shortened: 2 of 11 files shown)


skialpine:
### Code review

Found 4 issues:

1. **Main-thread UI freeze on first BLE error** (CLAUDE.md says "Never run database or disk I/O on the main thread. Use `QThread::create()` ... to run queries on a background thread").

    `logLinuxBtDiagnosticsOnce()` is invoked from `BleTransport::onControllerError` and `QtScaleBleTransport::onControllerError`, both main-thread slots. Inside, it runs three sequential `QProcess::waitForFinished(2000)` calls (`getcap`, `bluetoothctl --version`, `hciconfig -a`) — worst case ~6 s of UI hang the first time a BLE connect fails. `std::call_once` caps it at once per session, but that once is blocking. Move the whole `std::call_once` body onto a detached `QThread::create(...)`; `qInfo()` is already handler-chain safe.

    https://github.com/Kulitorum/Decenza/blob/61f36ecd89796201be76334da89b02b132a2e13e/src/ble/blecapability.cpp#L99-L125

2. **Data race on `g_checked` / `g_missing`** — unsynchronized `bool` globals in `ensureChecked()` can now be read/written from any thread because `BtLogFilter::messageHandler` calls `BleCapability::linuxMissing()` on whichever thread emitted the log. The eager `linuxMissing()` call in `BLEManager::BLEManager()` closes the window in practice (Qt Bluetooth doesn't log the flagged string before BLEManager has constructed controllers), but per the C++ memory model the write-then-read pattern is still a race and a TSAN would flag it. Cheap fix: make `g_checked` a `std::atomic_bool` or wrap the probe in `std::call_once`.

    https://github.com/Kulitorum/Decenza/blob/61f36ecd89796201be76334da89b02b132a2e13e/src/ble/blecapability.cpp#L17-L26

3. **Handler-chain diagram is wrong** — the comment depicts `... CrashHandler → BtLogFilter → AsyncLogger` with `ShotDebugLogger` at the top of the static chain, but the actual `main.cpp` install order is AsyncLogger → CrashHandler → WebDebugLogger → BtLogFilter, so BtLogFilter is the topmost static handler and ShotDebugLogger isn't in the static chain at all (it installs lazily at shot start). Either drop the diagram or match it to the code: `qDebug() → BtLogFilter → WebDebugLogger → CrashHandler → AsyncLogger` (static); ShotDebugLogger, when active, sits on top of BtLogFilter.

    https://github.com/Kulitorum/Decenza/blob/61f36ecd89796201be76334da89b02b132a2e13e/src/core/btlogfilter.h#L10-L15

4. **Issue number in code comments** (CLAUDE.md says "Don't reference the current task, fix, or callers ... since those belong in the PR description and rot as the codebase evolves"). Two places reference `issue #804`:

    https://github.com/Kulitorum/Decenza/blob/61f36ecd89796201be76334da89b02b132a2e13e/src/main.cpp#L331-L335

    https://github.com/Kulitorum/Decenza/blob/61f36ecd89796201be76334da89b02b132a2e13e/src/core/btlogfilter.h#L6-L10

<details>
<summary>3 below-threshold issues (50–74)</summary>

- [Worth a look] `BLEManager::requestBluezCacheHint()` consumes the one-shot token before checking `m_disabled`; in simulator mode the token is burned and the dialog can never fire later in the same process. Add an `if (m_disabled) return;` before `takeBluezCacheHintToken()`. Same class of issue fixed in #805 review. (score: 50)
- [Worth a look] QML `Connections` comment says the signal fires "on a DE1 connection attempt", but `QtScaleBleTransport::onControllerError` also emits it. One-line wording fix. (score: 72)
- [OK to skip] A couple of header doc-comments lean toward restating "what" the API does rather than the load-bearing "why" (most are fine, but `logLinuxBtDiagnosticsOnce`'s doc and the second paragraph of `btlogfilter.h` could be trimmed). Stylistic per CLAUDE.md. (score: 72)

</details>

skialpine:
### Code review

No issues at or above the 75 threshold in commit [`18593ab5`](https://github.com/Kulitorum/Decenza/pull/810/commits/18593ab5a9ae51194b0a1b18bcc644db7e77e7a6). All five findings from the prior review were addressed:

- Main-thread QProcess blocking → moved onto detached `QThread`
- `g_checked` data race → wrapped in `std::call_once`
- `m_disabled` gate in `requestBluezCacheHint` → added
- Wrong chain diagram in `btlogfilter.h` → rewritten
- `#804` references in code → removed

<details>
<summary>4 below-threshold issues (25–74)</summary>

- [Worth a look] The new worker thread in `logLinuxBtDiagnosticsOnce` is parentless and untracked — if the app exits within the ~6 s subprocess window on first BLE error, the `QThread` destructor can `qFatal` with "Destroyed while thread is still running". One-liner fix: store the `QThread*` in a file-local `QPointer` and `wait(500)` on `QCoreApplication::aboutToQuit`. (score: 72)

  https://github.com/Kulitorum/Decenza/blob/18593ab5a9ae51194b0a1b18bcc644db7e77e7a6/src/ble/blecapability.cpp#L121-L127

- [OK to skip] Pre-existing comment in `main.cpp` still names `ShotDebugLogger` as a static handler "above" `AsyncLogger` and doesn't mention the newly-installed `BtLogFilter`. CLAUDE.md asks for pre-existing fixes in touched files, but the comment is clearly an accuracy nit. (score: 42)

  https://github.com/Kulitorum/Decenza/blob/18593ab5a9ae51194b0a1b18bcc644db7e77e7a6/src/main.cpp#L301-L307

- [OK to skip] Rewritten `btlogfilter.h` comment dropped the chain diagram entirely and says the filter "suppresses before it reaches any other handler". Technically accurate for suppressed messages; mildly ambiguous if read as a general statement. (score: 38)

- [OK to skip] `std::call_once` lambda early-returns on `/proc/self/status` open/parse failure leaving `g_missing = false`; subsequent runs never re-probe. This is intentional conservative fallback (don't nag users whose /proc is unreadable) and pre-dates this commit. (score: 25)

</details>

Issue #123, opened 2026-02-01
[Feature] Add "Screen Off" time interval option(s)


LagaV:
Would be great to have, besides the wonderful videos playing on standby screen, options to turn off the screen during defined time ranges (e.g. at night).

PR #811, opened 2026-04-20
Mode-aware shot analysis: kill lever false positives, catch puck failures, add shot_eval corpus

skialpine:
## Summary

- **Mode-aware channeling + grind detection** — phase-mode windows (pressure-mode uses pressure goal stationarity, flow-mode uses flow goal stationarity) replace the old "whole pour / unrestricted" analysis. Kills the lever-ramp and D-Flow pour false positives that were catching 53% of shots in my history as Sustained.
- **New `detectPourTruncated` detector** — fires when peak pressure inside the pour window stays below 2.5 bar. Catches catastrophic puck failures (shot 868: 0.6 bar peak / 7 s / 37 g yield) that the dC/dt + grind-direction detectors are structurally blind to because conductance saturates at the clamp.
- **End-of-pour skip (1.5 s)** mirrors the existing 2 s start-skip. Kills fast-shot tail false positives where flow naturally accelerates from puck erosion.
- **Decoupled verdicts** — channeling no longer borrows the flow-grind delta to fabricate a "Puck collapsed — grind too fine" message. Each signal names itself only when it fires.
- **`tools/shot_eval/`** — desktop CLI that links the real production `ShotAnalysis` + `Conductance` code. Run against a directory of shot JSONs to diff detector output before/after an algorithm change.
- **`tests/data/shots/` regression corpus** — 12 curated shots + `manifest.json` with expected verdicts per shot. Wired as the `shot_corpus_regression` ctest entry so `ctest -R shot_corpus_regression` catches any future verdict flip on a golden case.

## Impact on my 1038-shot corpus

| | Before | After | Change |
|---|---:|---:|---:|
| Baseline Sustained (unrestricted) | 546 | — | — |
| Mode-aware Sustained | — | 43 | − |
| Mode-aware Transient | — | 53 | − |
| Grind issues | — | 97 | − |
| Pour-truncated (new) | — | 2 | + |
| **Total flagged by any signal** | ≈550 | **190 (20.5%)** | **−65%** |
| Shots relaxed from Sustained → None | — | 478 | — |
| Shots tightened (False → True Sustained) | — | 0 | — |

Spot-checked ~25 flagged shots manually; remaining flags look legitimate (Londinium gusher, Blooming choker, Malabar too-fine) or defensible (weight-plateau D-Flow shots that may be real micro-channels or scale artifacts).

## Code layout

- **`src/ai/shotanalysis.{h,cpp}`** — new `DetectionWindow` struct, `buildChannelingWindows()`, `analyzeFlowVsGoal()` shared helper, `detectPourTruncated()`, and the updated `detectChannelingFromDerivative()` + `detectGrindIssue()` signatures. Generate-summary verdict block rewritten so channeling, grind-direction, and pour-truncation each speak for themselves.
- **`src/ai/conductance.{h,cpp}`** — new. Conductance `C = F²/P` and the 9-point Gaussian dC/dt derivative extracted so `ShotDataModel` (live per-sample) and `shot_eval` (batch offline) share one implementation.
- **`src/models/shotdatamodel.cpp`** — delegates to `Conductance::derivative` / `Conductance::sample`. No behavior change for live shots.
- **`src/ai/shotsummarizer.cpp`**, **`src/history/shothistorystorage.cpp`** — all `detectChannelingFromDerivative` + `detectGrindIssue` callsites updated to pass phase markers and flags so badge and popup verdicts stay aligned.
- **`tests/tst_shotanalysis.cpp`** — 1 → 18 test cases: window builder happy + sad paths, end-skip, empty-windows contract, flow-mode vs pressure-mode grind check, beverage-type skip, pour-truncation variants.
- **`tools/shot_eval/main.cpp`** — loader handles both Decenza/upload and visualizer-download JSON shapes. Outputs a table or JSON; `--validate` mode compares against a manifest and exits non-zero on any mismatch.
- **`tests/data/shots/*.json` + `manifest.json`** — 12-shot corpus covering each detector path. Local-corpus shots have bean brand / barista / notes / UUIDs stripped; public visualizer shots copied as-is.
- **`tests/CMakeLists.txt`** — `src/ai/conductance.cpp` added to `PROFILEMANAGER_SOURCES`, `HISTORY_SOURCES`, and `tst_settling` directly so every ShotDataModel-linking test resolves the symbol. New `shot_corpus_regression` ctest entry.
- **`CMakeLists.txt`** — `shot_eval` target in the desktop-only dev-tools block alongside `profile_sync`.

## Test plan

- [x] `ctest -R tst_shotanalysis` — all 18 pass
- [x] `ctest -R shot_corpus_regression` — all 12 corpus shots pass
- [x] Full `cmake --build . --target all` — clean (Decenza + all tests)
- [x] Spot-checked `shot_eval` output on the user's 1038-shot export directory — no regressions from pre-fix behavior, 2 new pour-truncated catches
- [x] Validated against 9 additional public visualizer shots covering profile families not in the user's corpus (E61, Classic Italian, Gagné Adaptive, TurboBloom, Best Practice Light, Blooming, Cremina, Damian's LRv3)
- [ ] Eyeball a few of the still-flagged D-Flow / Q shots in the app to confirm the weight-plateau pattern is real micro-channeling vs scale artifact (optional follow-up — no blocker)

## Follow-ups (out of scope for this PR)

- `pourTruncated` isn't yet a stored DB flag / badge — it only shows in the generated summary popup and the AI prompt. A follow-up PR could add a proper SQL column + QML badge.
- One confirmed true-positive channeling shot on a flat-pressure profile (E61 / Classic Italian) with visible channel spritzers would be a valuable addition to the regression corpus.


skialpine:
### Code review

Found 6 issues:

1. Stale docstring + inline comment claim "empty windows ⇒ unrestricted fallback" but the implementation returns `None`. The actual contract is: `buildChannelingWindows` emits a whole-pour window when `phases.isEmpty()` (legacy data), and `detectChannelingFromDerivative` returns `None` on empty windows. The promise of a legacy unrestricted fallback when callers pass an empty vector is not delivered — anyone reading the header and doing so gets silent zero-detection.

https://github.com/Kulitorum/Decenza/blob/1816beadf71e568ef66177dcb1c35bcada99f1bc/src/ai/shotanalysis.h#L95-L103

https://github.com/Kulitorum/Decenza/blob/1816beadf71e568ef66177dcb1c35bcada99f1bc/src/ai/shotanalysis.cpp#L491-L500

2. `int i` iterating over `QJsonArray::size()` in `buildSeries` (CLAUDE.md says "Use `qsizetype` (not `int`) for container sizes — ... assigning to `int` causes `-Wshorten-64-to-32` warnings"). `QJsonArray::size()` returns `qsizetype` in Qt 6.

https://github.com/Kulitorum/Decenza/blob/1816beadf71e568ef66177dcb1c35bcada99f1bc/tools/shot_eval/main.cpp#L87-L94

3. `int sampleCount` in the new public `GrindCheck` struct holds the accumulated count of `QVector<QPointF>` samples. Same CLAUDE.md rule as above, and this is in a header (public surface) rather than a local variable — the rest of the new code (conductance.cpp, buildChannelingWindows, etc.) uses `qsizetype` consistently.

https://github.com/Kulitorum/Decenza/blob/1816beadf71e568ef66177dcb1c35bcada99f1bc/src/ai/shotanalysis.h#L143-L153

4. The refactor extracts conductance math into `Conductance::derivative` and has `ShotDataModel` delegate, but `ShotHistoryStorage::computeDerivedCurves` still contains a verbatim inline copy of the same 9-point Gaussian + clamp logic. Future tuning of the kernel or clamp range would have to be updated in two places or the live-shot path and legacy-recompute path diverge — the stated goal of a single source of truth for the formula is only half done.

https://github.com/Kulitorum/Decenza/blob/1816beadf71e568ef66177dcb1c35bcada99f1bc/src/history/shothistorystorage.cpp#L2020-L2060

5. `shot_eval` silently disables channeling detection on visualizer shots whose goal curves are absent/all-sentinel. `inferPhasesFromGoals` inserts a single `isFlowMode=false` fallback marker and its comment claims this "keeps the detector in its legacy unrestricted branch", but with phases non-empty and no qualifying goal data, `buildChannelingWindows` returns empty → `detectChannelingFromDerivative` returns `None`. The fallback needs to either return an empty `phases` list (so the whole-pour window fires) or have shot_eval pass a whole-pour `DetectionWindow` directly.

https://github.com/Kulitorum/Decenza/blob/1816beadf71e568ef66177dcb1c35bcada99f1bc/tools/shot_eval/main.cpp#L153-L168

6. Multiple comments reference specific shot-ID / fix-ordinal context that will rot (CLAUDE.md says "Don't reference the current task, fix, or callers ('used by X', 'added for the Y flow', 'handles the case from issue #123'), since those belong in the PR description and rot as the codebase evolves"). Affected locations:

https://github.com/Kulitorum/Decenza/blob/1816beadf71e568ef66177dcb1c35bcada99f1bc/src/ai/shotanalysis.h#L43-L50

https://github.com/Kulitorum/Decenza/blob/1816beadf71e568ef66177dcb1c35bcada99f1bc/tools/shot_eval/main.cpp#L459-L465

https://github.com/Kulitorum/Decenza/blob/1816beadf71e568ef66177dcb1c35bcada99f1bc/tests/tst_shotanalysis.cpp#L238-L246

<details>

- [Worth a look] End-skip of 1.5 s is applied in `detectChannelingFromDerivative` but not mirrored in `buildChannelingWindows` — windows can extend to `pourEnd` while sample analysis stops at `pourEnd - 1.5`. Harmless today (under-counts, never over-counts), but trivially fixable by mirroring the trim when building windows. (score: 50)
- [OK to skip] `tests/CMakeLists.txt` uses `COMMAND shot_eval` rather than `$<TARGET_FILE:shot_eval>`; CMake generates an absolute path in the test file so it works in practice, and the codebase's other dev-tool test registrations use the same pattern. (score: 50)

</details>

Issue #812, opened 2026-04-20
Startup log spam: ShotHistoryExporter re-parses every shot's profile on every launch


skialpine:
## Problem

On every app startup with `exportShotsToFile` enabled, `ShotHistoryExporter`'s constructor calls `startBulkExport()` which iterates all stored shots and re-parses each shot's embedded `profileJson`. On my 1039-shot library this produces:

- **~1,365 `Syncing espresso_temperature from X to first frame temp Y` debug lines** in ~6 seconds during boot
- **~30 `WARN jsonToDouble: failed to parse string "" - using default 2`** warnings in the same burst

Both fire from inside `Profile::fromJson()` parsing of stored shot snapshots. The `Syncing` log was added in [9c4b195a](https://github.com/Kulitorum/Decenza/commit/9c4b195a39d81c7ab3ee7f23e76322fd7eaf7798) ("Fix wrong profile and temperature on startup"); the correction is silent-by-design, so the log is pure noise.

Across my last 46 sessions the log file contains **6,825 total occurrences** of the sync message — roughly every launch spams this.

## Root cause

`src/history/shothistoryexporter.cpp:47-49`:

```cpp
if (m_settings->exportShotsToFile()) {
    startBulkExport();
}
```

`startBulkExport()` → iterates all shots → `writeShotJson()` → `loadShotRecordStatic()` → parses `record.profileJson` via `Profile::fromJson()` → fires sync log / jsonToDouble warnings for any legacy stored profile that has mismatched temperature fields or empty string fields in `limiter`/`exit` blocks.

About 10 distinct `(from, to)` temperature pairs in the corpus → roughly 10 profile variants in my history carry this legacy-format issue: `(93→82)`, `(93→84)`, `(93→85)`, `(93→86)`, `(93→88)`, `(93→89)`, `(93→94)`, `(94→84)`, `(94→90)`, `(94→92.5)`. The typical pattern is a D-Flow-family or 80's-Espresso-family profile where a legacy top-level `espresso_temperature` doesn't match the first-frame temp.

The bulk export also does ~5 seconds of redundant disk work on every launch even when the exported files already exist and are current — that's a separate perf issue this surfaces.

### 1. Silence the logs
One-line changes, cheapest fix.

- `src/profile/profile.cpp:569` — drop the `qDebug() << "Syncing espresso_temperature..."` since the correction is silent-by-design
[...]

PR #813, opened 2026-04-20
fix(history): skip bulk shot export when on-disk file is already current

skialpine:
## Summary
- Addresses fix option **#3** from #812: ShotHistoryExporter now skips rewriting an exported shot file when the on-disk copy is already current (file exists, non-empty, `mtime >= shots.updated_at`). On a warm start no shot records are loaded and no profile JSON is parsed, so the ~1,365-line "Syncing espresso_temperature..." debug spam and the ~5 s of redundant disk work during startup both go away for users with large libraries.
- Bulk enumeration query now selects `(id, timestamp, COALESCE(updated_at, 0))` so the worker can make the freshness decision without hitting `loadShotRecordStatic()` unless a write is actually needed.
- `bulkExportFinished` signal widened to `(int written, int skipped, int failed)` — no external subscribers today. Header comment updated to describe the self-heal behaviour.

## Test plan
- [ ] Build Release in Qt Creator (macOS/Windows).
- [ ] With `exportShotsToFile` enabled and a populated history folder, launch the app and confirm no "Syncing espresso_temperature..." lines and no `jsonToDouble: failed to parse string ""` warnings in the log.
- [ ] Delete one exported JSON, relaunch, confirm it reappears and the rest are still skipped.
- [ ] Edit a shot's metadata (e.g. enjoyment / notes) while running, confirm the single-shot exporter still rewrites that file.
- [ ] Toggle `exportShotsToFile` off then on, confirm files that already exist are kept and nothing is rewritten if unchanged.
- [ ] Manually set the mtime of one exported file to the epoch, relaunch, confirm only that one is rewritten.

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -3067,15 +3067,21 @@ ApplicationWindow {
 
     Connections {
         target: ShotHistoryExporter
-        function onBulkExportFinished(ok, failed) {
+        function onBulkExportFinished(written, skipped, failed) {
+            // Suppress the toast when nothing was actually written — a warm
+            // startup where every file is already current is the common case
+            // and a silent no-op is the honest signal there.
+            if (written === 0 && failed === 0) {
+                return
+            }
             if (failed > 0) {
                 shotExportToastText = TranslationManager.translate(
                     "main.toast.exportShotsPartial",
-                    "Exported %1 shots; %2 failed").arg(ok).arg(failed)
+                    "Exported %1 shots; %2 failed").arg(written).arg(failed)
             } else {
                 shotExportToastText = TranslationManager.translate(
                     "main.toast.exportShotsDone",
-                    "Exported %1 shots").arg(ok)
+                    "Exported %1 shots").arg(written)
             }
             shotExportToast.opacity = 1
             shotExportToastTimer.restart()
--- a/src/history/shothistoryexporter.cpp
+++ b/src/history/shothistoryexporter.cpp
@@ -9,6 +9,7 @@
 #include <QDateTime>
 #include <QDir>
 #include <QFile>
+#include <QFileInfo>
 #include <QSaveFile>
 #include <QSet>
 #include <QSqlQuery>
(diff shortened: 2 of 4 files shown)


skialpine:
### Code review

Found 2 issues:

1. `bulkExportFinished` signal widened from `(int, int)` to `(int written, int skipped, int failed)`, but the QML slot in `qml/main.qml` still declares `function onBulkExportFinished(ok, failed)`. QML binds positionally, so `failed` now receives the **skipped** count. On every warm startup the user will see a false "Exported X shots; Y failed" toast because `skipped > 0` is the normal case after this change.

https://github.com/Kulitorum/Decenza/blob/0e7ba9f2479d440823bb1a45de4fd0231b8b9335/qml/main.qml#L3067-L3086

2. The freshness check trusts `shots.updated_at` as a proxy for "exported JSON is current", but several `UPDATE shots` paths in `src/history/shothistorystorage.cpp` don't bump `updated_at`: the bulk grinder rename at [line 3822](https://github.com/Kulitorum/Decenza/blob/0e7ba9f2479d440823bb1a45de4fd0231b8b9335/src/history/shothistorystorage.cpp#L3818-L3833), the badge detector at [line 1764](https://github.com/Kulitorum/Decenza/blob/0e7ba9f2479d440823bb1a45de4fd0231b8b9335/src/history/shothistorystorage.cpp#L1762-L1775), and the `beverage_type` backfills. `grinder_model` is in the exported payload, so a user who renames a grinder via the bulk path will see the DB update but the on-disk JSONs will be permanently skipped by the next warm start. Consider either bumping `updated_at` on every mutating UPDATE or adding an `AFTER UPDATE` trigger as a safety net.

https://github.com/Kulitorum/Decenza/blob/0e7ba9f2479d440823bb1a45de4fd0231b8b9335/src/history/shothistoryexporter.cpp#L107-L114

<details>

- [Worth a look] Same-second race: `mtime >= updated_at` can skip a metadata edit that lands within the same wall-clock second as a bulk write; changing `>=` to `>` closes the narrow catch-up-pass window (score: 65).
- [OK to skip] `COALESCE(updated_at, 0)` means shots with NULL `updated_at` (legacy upgrade paths) treat any existing file as fresh, so a corrupted non-empty file (size > 0 but invalid JSON) can't self-heal. Narrow and low-frequency (score: 65).

</details>

PR #814, opened 2026-04-20
fix: stop leaking TDS/EY into next shot metadata

skialpine:
## Summary

- `PostShotReviewPage.saveEditedShot()` was syncing the edited TDS and EY back to `Settings.dyeDrinkTds` / `Settings.dyeDrinkEy`.
- `MainController::onShotEnded` (`src/controllers/maincontroller.cpp:1681-1682`) builds every new shot's metadata from those same Settings values, and correctly resets them to 0 after saving — but the post-shot review handler was immediately un-resetting them.
- Result: shot N's TDS/EY would silently appear on shot N+1 until the user typed new numbers.

Fix: drop the two sticky-sync assignments. TDS and EY are per-shot values, same as enjoyment and notes (which were already excluded). Updated the surrounding comment to document the rule and why it matters.

## Test plan

- [ ] Record a shot, open post-shot review, enter a TDS value (EY auto-calculates), save.
- [ ] Pull a second shot, open post-shot review: TDS and EY fields should be empty, not carrying shot 1's values.
- [ ] Confirm bean/grinder/dose/drink weight still persist across shots as before.

### DIFF
--- a/qml/pages/PostShotReviewPage.qml
+++ b/qml/pages/PostShotReviewPage.qml
@@ -260,8 +260,10 @@ Page {
         }
         MainController.shotHistory.requestUpdateShotMetadata(editShotId, metadata)
 
-        // Sync sticky metadata back to Settings (bean/grinder info) for the next shot
-        // but NOT enjoyment/notes which are shot-specific
+        // Sync sticky metadata back to Settings (bean/grinder info) for the next shot.
+        // Per-shot fields (enjoyment, notes, TDS, EY) are NOT synced — otherwise they
+        // would leak into the next shot's metadata, since MainController builds shot
+        // metadata from these Settings values at shot end.
         Settings.dyeBeanBrand = editBeanBrand
         Settings.dyeBeanType = editBeanType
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Verified the fix's core claim: [`MainController::onShotEnded`](https://github.com/Kulitorum/Decenza/blob/77f0cea42b5a5f245c11c2776848c96b35d9f55c/src/controllers/maincontroller.cpp#L1680-L1733) builds shot metadata from `Settings.dyeDrinkTds`/`dyeDrinkEy` and resets them to 0 post-save; writing them back from [`saveEditedShot`](https://github.com/Kulitorum/Decenza/blob/77f0cea42b5a5f245c11c2776848c96b35d9f55c/qml/pages/PostShotReviewPage.qml#L263-L280) would leak into shot N+1. No QML consumer reads those Settings fields, so the sync removal has no other side effects. The retained `dyeDrinkWeight` sync is safe since MainController unconditionally overwrites it with each shot's actual final weight.

PR #819, opened 2026-04-21
fix(steam): persist live steam warning banner until dismissed

skialpine:
## Summary
- The live steam pressure/temperature warning banner on `SteamPage` auto-dismissed after 5 seconds. Users focused on their pitcher routinely missed it — and since the underlying signal is latched per session in `SteamHealthTracker`, a missed warning was gone for the rest of the session.
- Banner now persists until the user taps it. Text appends "(tap to dismiss)" as a visible affordance, and an `AccessibleMouseArea` handles screen reader activation (first-tap-announce / second-tap-activate via TalkBack/VoiceOver).
- Any stale warning is cleared on the rising edge of `isSteaming` so it doesn't linger into the next steam session.

Only affects the live-session banner (pressure > 8.0 bar / temp > 180 °C hard limits). The post-session `Dialog` alerts (`onDescaleWarning`, `onTemperatureWarning`, `onScaleBuildupWarning`) were already modal with an OK button and are unchanged.

## Test plan
- [ ] Trigger a live pressure warning (steam pressure > 8 bar) — banner appears and stays until tapped.
- [ ] Tap the banner — it dismisses.
- [ ] Start a new steam session while a warning from a previous session is still visible — banner clears automatically.
- [ ] TalkBack/VoiceOver: first tap announces "Warning: steam pressure is too high", second tap dismisses.
- [ ] Temperature variant: same behavior with the "steam temperature is too high" text.

### DIFF
--- a/qml/pages/SteamPage.qml
+++ b/qml/pages/SteamPage.qml
@@ -182,13 +182,16 @@ Page {
         }
     }
 
-    // Warning banner (auto-dismiss after 5 seconds — allowed per CLAUDE.md for UI auto-dismiss)
+    // Live safety warning banner. Persists until the user taps it — this is a
+    // one-shot per-session alert (latched in SteamHealthTracker), so if it
+    // auto-dismissed the user could easily miss it while focused on the pitcher.
     property string warningText: ""
     property bool warningVisible: false
-    Timer {
-        id: warningDismissTimer
-        interval: 5000
-        onTriggered: warningVisible = false
+
+    // Clear any stale warning when a new steam session starts.
+    onIsSteamingChanged: {
+        if (isSteaming)
+            warningVisible = false
     }
 
     // Warning connections
@@ -199,13 +202,11 @@ Page {
             warningText = TranslationManager.translate("steam.warning.pressureHigh",
                 "Warning: steam pressure is too high")
             warningVisible = true
-            warningDismissTimer.restart()
         }
         function onTemperatureTooHigh() {
             warningText = TranslationManager.translate("steam.warning.temperatureHigh",
                 "Warning: steam temperature is too high")
             warningVisible = true
-            warningDismissTimer.restart()
         }
         function onDescaleWarning() {
             steamWarningDialog.warningMessage = TranslationManager.translate("steam.warning.descale",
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 2 issues:

1. Duplicate `onIsSteamingChanged` handler on the page root — the new handler at lines 191–195 collides with the existing handler at line 91, which manages `wasSteaming`, `steamSoftStopped`, `_lastAnnouncedSteamWeight`, and the session-end preset/heater shutdown (`MainController.sendSteamTemperature(0)` when `keepSteamHeaterOn=false`). QML does not merge duplicate signal handlers on the same object — one is silently dropped, or the page fails to load on Qt 6.10. Either way a real regression (CLAUDE.md QML Gotchas section warns against duplicate property/handler bindings: "Property has already been assigned a value"). Merge the `warningVisible = false` into the existing handler's `if (isSteaming)` branch at line 91 and delete the standalone block.

https://github.com/Kulitorum/Decenza/blob/e3f3f6ceee6b0d9ef6a3c748cf0da00f33e3b0f7/qml/pages/SteamPage.qml#L88-L130

https://github.com/Kulitorum/Decenza/blob/e3f3f6ceee6b0d9ef6a3c748cf0da00f33e3b0f7/qml/pages/SteamPage.qml#L189-L196

2. `AccessibleMouseArea.accessibleName` omits the "(tap to dismiss)" affordance — sighted users see it appended to the banner text (line 478), but screen-reader users only hear the warning with no hint that tapping dismisses it (`docs/CLAUDE_MD/ACCESSIBILITY.md` calls for `Accessible.description` describing secondary/activation actions on interactive elements). Either append the localized hint to `accessibleName`, or add `Accessible.description` on the parent `warningBanner` Rectangle (the established AccessibleMouseArea-with-parent pattern per ACCESSIBILITY.md).

https://github.com/Kulitorum/Decenza/blob/e3f3f6ceee6b0d9ef6a3c748cf0da00f33e3b0f7/qml/pages/SteamPage.qml#L484-L492

<details>
<summary>3 below-threshold issues (50–74)</summary>

- [OK to skip] i18n string concatenation `warningText + "  " + translate("...tapToDismiss")` — both parts are individually translated and the pattern is widely used elsewhere in the codebase, but it does prevent translator reordering (score: 50)
- [OK to skip] Comment "Clear any stale warning…" uses "stale" loosely — the banner isn't stale in the undismissed sense, it's just being cleared on new-session rising edge (score: 50)
- [OK to skip] Comment "one-shot per-session alert (latched in SteamHealthTracker)" is accurate for the two banner signals but the adjacent `Connections` block also handles three non-latched modal-dialog signals — minor scoping ambiguity (score: 50)

</details>

PR #827, opened 2026-04-22
fix(firmware): prompt for power-cycle immediately after verify

skialpine:
## Summary
- Captured two real flashes (upgrade 1333→1352 and downgrade 1352→1333 on the same DE1/firmware pair); on both, the \`auto-reboot window expired — prompting user to power-cycle\` log line fired at exactly +30 s because the DE1 kept running the old firmware. The 30 s wait added nothing and the prompt's wording implied something had gone wrong.
- Set \`m_needsManualReboot = true\` the moment the A009 verify notification confirms the new bank. Drop \`m_autoRebootWaitTimer\`, \`m_autoRebootWaitMs\`, and the \`onAutoRebootWaitTimeout\` handler.
- Reword the root-level power-cycle dialog and the Settings → Firmware awaiting-reboot strip to describe the next step plainly (flip the switch to load the new firmware) instead of describing a failure to auto-reboot.
- The disconnect + reconnect + version-match success path is untouched, so a DE1 that does auto-reboot (none observed) would still complete cleanly.

## Test plan
- [ ] Tap \"Update now\" on an available firmware → verify the power-cycle dialog appears within ~1 s of the verify-success notification, not 30 s later
- [ ] Power-cycle the DE1 → reconnect handshake completes, dialog closes, Settings → Firmware shows \"Update complete — DE1 is on v…\"
- [ ] Do nothing (don't power-cycle) → dialog stays up until the 180 s verify-disconnect grace expires and we fall back to the standard \"DE1 did not reconnect after verify\" failure (existing path, unchanged)
- [ ] Start a flash, let it verify, swipe the dialog-ack button → the Firmware tab strip still says \"please power-cycle the DE1\" with a warning-coloured border
- [ ] Unit tests: \`ctest -R tst_firmwareupdater\` (no test calls \`setAutoRebootWaitMs\`, so no test updates needed)

### DIFF
--- a/docs/CLAUDE_MD/FIRMWARE_UPDATE.md
+++ b/docs/CLAUDE_MD/FIRMWARE_UPDATE.md
@@ -24,7 +24,7 @@ Two buttons:
 - **Phase 1 — Erase**: `FWMapRequest(erase=1, map=1)` sent on characteristic `A009`. DE1 erases the inactive flash bank. Modern firmware sends a single erase-complete notification (`fwToErase=0`). We then wait a fixed 10 s (Android) / 1 s (elsewhere) per de1app's convention before starting the chunk pump.
 - **Phase 2 — Upload**: 28,992 sequential 20-byte packets streamed to characteristic `A006`. Each packet is `[len=16][addr24 BE][16-byte payload]`. Length byte is 16 (`0x10`) — same format as MMR writes which use length 4. Address is 24-bit **big-endian**. ACK-driven progress so the bar tracks bytes on the wire, not bytes in the BLE queue.
 - **Phase 3 — Verify**: `FWMapRequest(erase=0, map=1, FirstError=FF,FF,FF)` on `A009`. DE1 verifies the inactive bank and replies with a 3-byte `FirstError`. `{0xFF, 0xFF, 0xFD}` = success; anything else = the byte offset of the first corrupt block.
-- On success, the DE1 remaps the inactive bank as active and auto-reboots. Decenza's auto-reconnect logic re-establishes BLE.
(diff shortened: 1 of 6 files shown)


skialpine:
### Code review

Found 3 issues:

1. `FirmwareUpdater::stateText()` still returns "Waiting for DE1 to restart" for `AwaitingReboot` — the word "restart" implies the DE1 will act on its own, which directly contradicts this PR's thesis and the updated dialog wording. `stateText` is a `Q_PROPERTY` surfaced on the Firmware tab, so this string is user-visible. Suggest "Power-cycle the DE1 to load new firmware" to match the dialog.

https://github.com/Kulitorum/Decenza/blob/b3ab12043a65bcdd1d79c308ef813fa6ed3812ce/src/controllers/firmwareupdater.cpp#L308-L313

2. Stale comment on `m_verifyDisconnectGrace` — says a Verifying-phase disconnect "commonly means a successful reboot rather than a failure". Under this PR's model the DE1 never auto-reboots, so that framing is out of date; the normal path is user-initiated reconnect from `AwaitingReboot`.

https://github.com/Kulitorum/Decenza/blob/b3ab12043a65bcdd1d79c308ef813fa6ed3812ce/src/controllers/firmwareupdater.h#L210-L216

3. Stale inline comment in the disconnect path of `onDeviceConnectionChanged` — still reads `(successful reboot, either auto or user-initiated)`. The "auto" branch is the one the rest of the PR just retired; drop "either auto or".

https://github.com/Kulitorum/Decenza/blob/b3ab12043a65bcdd1d79c308ef813fa6ed3812ce/src/controllers/firmwareupdater.cpp#L225-L234

<details>

- [Worth a look] `m_verifyDisconnectGraceMs` comment frames the grace window as a defensive fallback against a "spurious failure", but after this PR it's the primary success path (user power-cycles, DE1 reconnects within this window). Easy one-liner in the header right next to the other comments being refreshed. (score: 50)

</details>

PR #829, opened 2026-04-22
fix(migration): make device discovery reliable on Wi-Fi

skialpine:
## Summary
- **Root cause**: Data-migration device discovery was intermittent (~50% of the time on the Decent tablet). Logs from the 192.168.10.163 tablet showed zero `Discovery request from …` lines for the whole session — the Android OS was silently dropping UDP broadcasts because the app didn't hold a `WifiManager.MulticastLock`. Compounding that, the client only sent one broadcast at the start of a 3-second window with no retries, so any packet loss anywhere was fatal.
- **Client (`DataMigrationClient`)**: send the discovery broadcast in 4 bursts at `t = 0 / 500 / 1200 / 2500 ms` across a `5 s` window (up from a single burst / 3 s). Each log line now tags the burst number, so the log tells you which send the responder actually heard.
- **Server (`ShotServer`, Android only)**: acquire a non-reference-counted `MulticastLock` while the discovery listener is bound; release it when the server stops. Adds `CHANGE_WIFI_MULTICAST_STATE` to the manifest (normal permission, granted at install).
- **Multi-device**: existing dedup-by-`serverUrl` already supports listing multiple servers; retransmission materially improves the odds that every responder on the LAN is seen in a single scan.

## Test plan
- [ ] Build both tablet and desktop from this branch; fresh-install the APK on the tablet (normal permission requires install, not upgrade).
- [ ] From desktop, run the data-migration "search for devices" flow with the tablet powered and Wi-Fi awake — confirm the tablet appears on every (or near-every) scan, and that the tablet log shows `ShotServer: Discovery request from …` for each scan.
- [ ] Verify tablet log shows `ShotServer: Acquired Wi-Fi MulticastLock for discovery` on start and `Released Wi-Fi MulticastLock` on stop.
- [ ] With two Decenza instances on the LAN, confirm both appear in the discovery list (dedup still works, no duplicate rows per device).
- [ ] On macOS/iOS/Windows/Linux builds, confirm the JNI block compiles out and discovery still works.
- [ ] Verify `stopDiscovery()` cancels the retransmit timer mid-burst (cancel a search quickly from the UI and confirm no stray broadcasts after).

### DIFF
--- a/android/AndroidManifest.xml.in
+++ b/android/AndroidManifest.xml.in
@@ -13,6 +13,9 @@
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.INTERNET" />
+    <!-- Allows us to hold a WifiManager.MulticastLock so incoming UDP broadcast
+         discovery datagrams aren't filtered by the OS on Wi-Fi. -->
+    <uses-permission android:name="android.permission.CHANGE_WIFI_MULTICAST_STATE" />
     <!-- Storage permissions -->
     <!-- Android 9 (SDK 28): needs WRITE_EXTERNAL_STORAGE -->
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="29" />
--- a/src/core/datamigrationclient.cpp
+++ b/src/core/datamigrationclient.cpp
@@ -34,6 +34,7 @@ DataMigrationClient::DataMigrationClient(QNetworkAccessManager* networkManager, 
 DataMigrationClient::~DataMigrationClient()
 {
     *m_destroyed = true;
+    stopDiscovery();
     cancel();
     delete m_tempDir;
 }
--- a/src/core/datamigrationclient.cpp
+++ b/src/core/datamigrationclient.cpp
@@ -1054,22 +1054,52 @@ void DataMigrationClient::startDiscovery()
 
     connect(m_discoverySocket, &QUdpSocket::readyRead, this, &DataMigrationClient::onDiscoveryDatagram);
 
-    // Send broadcast discovery message
-    QByteArray discoveryMessage = "DECENZA_DISCOVER";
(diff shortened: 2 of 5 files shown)


skialpine:
### Code review

Found 1 issue:

1. The retransmit timer fires discovery bursts on a fixed schedule (0/500/1200/2500ms) to compensate for UDP packet loss. Two reviewers flagged this against CLAUDE.md's timer rule: *"Never use timers as guards/workarounds. Timers are fragile heuristics that break on slow devices and hide the real problem... Only use timers for genuinely periodic tasks (polling, animation, heartbeats) and UI auto-dismiss."* Scheduled UDP retransmission is a widely-used network-programming pattern (mDNS, SSDP, DHCP), but the CLAUDE.md carveouts don't explicitly cover it, and the scheduled bursts are arguably a workaround for the real fix (MulticastLock + an event-driven retransmit-if-empty fallback on timeout). Worth a judgment call from the author.

https://github.com/Kulitorum/Decenza/blob/23f97c2396883e2acadb436e224630b36fe9a0e4/src/core/datamigrationclient.cpp#L1140-L1157

<details>

- [Worth a look] Destructor doesn't stop the discovery retransmit timer — `~DataMigrationClient()` calls `cancel()` which does not touch discovery state, so a queued `singleShot` timeout on `m_discoveryRetransmitTimer` could still fire between the destructor body returning and Qt's child cleanup. The `!m_searching` guard mitigates most cases, but calling `stopDiscovery()` from the destructor would close the gap. One-line fix. (score: 68)  https://github.com/Kulitorum/Decenza/blob/23f97c2396883e2acadb436e224630b36fe9a0e4/src/core/datamigrationclient.cpp#L33-L40
- [Worth a look] Cross-file comment reference in `shotserver.h` points at `src/core/datamigrationclient.cpp` retransmit logic. Per CLAUDE.md: *"Don't reference the current task, fix, or callers … since those belong in the PR description and rot as the codebase evolves."* The retransmit is independent of MulticastLock's rationale; the comment will drift if either is renamed. Simpler wording: drop the cross-reference. (score: 65)  https://github.com/Kulitorum/Decenza/blob/23f97c2396883e2acadb436e224630b36fe9a0e4/src/network/shotserver.h#L126-L132

</details>

PR #817, opened 2026-04-21
feat(firmware): add DE1 firmware update over BLE

Kulitorum:
## Summary

- **Adds DE1 firmware update as a first-class feature** — Decenza can now check Decent's GitHub for new firmware, download it, and flash the DE1 over BLE end-to-end. Mirrors de1app's protocol byte-for-byte after cross-checking against de1app's Tcl source and Kal Freese's Python reference implementation.
- **Validated on real hardware**: successful v1333 → v1352 flash on a Decent tablet + DE1PRO (PCB 1.3). ~16-minute upload at ~30 ACK/sec (the Android BLE throughput limit).
- **63 unit tests** covering protocol packet builders, file header parsing, HTTP asset cache helpers, DE1Device extensions, and the full state machine including error/retry/verify-disconnect paths.

## How it works

Settings → Firmware tab shows installed version, available version (from Decent's `main`), Check Now, and Update Now. Auto-check runs 30 s after app start and every 168 h thereafter.

Three-phase BLE flash matching de1app:
1. **Erase** — `FWMapRequest(erase=1, map=1)` to A009, wait 10 s (Android) / 1 s elsewhere
2. **Upload** — stream 16-byte chunks as `[0x10][addr24 BE][payload]` to A006 via the WriteToMMR path; ACK-driven progress
3. **Verify** — `FWMapRequest(erase=0, map=1, FirstError=FF,FF,FF)` to A009; bootloader responds with `{FF,FF,FD}` on success then reboots

Safety: the DE1's dual-bank flash layout means a failed update falls back to the running firmware. Our BoardMarker check on the downloaded file + the bootloader's own verify phase are the two gates that prevent activation of a bad image.

## Bugs found and fixed in this branch

In order of discovery during real-hardware testing:

1. QML context property name was `MainController` (capital M), not `mainController` — whole UI was silently null.
2. Firmware chunk address was little-endian; de1app and Kal's Python both use big-endian. LE-encoded chunks landed at wildly wrong addresses, triggered the DE1's pumps mid-upload, and got rejected by the bootloader.
3. Erase phase was gating progression on a notification that modern firmware doesn't always send. de1app just waits a fixed delay — we now do too.
4. Byte 0 of the chunk packet is a *length* field (16), not a magic "opcode 0x10" as the original research agent reported. Same field as MMR's `0x04` for 4-byte writes.
[...]

PR #834, opened 2026-04-22
refactor(settings): replace magic-number tab indexes with symbolic ids (#833)

### LINKED ISSUE #833: Replace magic-number settings tab indexes with symbolic names
## Problem

Settings tab navigation uses raw integer indexes in 5+ places that must all be kept in lockstep whenever a tab is added, removed, or reordered:

- `qml/pages/SettingsPage.qml` — the tabBar button list and StackLayout Loader list (implicit ordering)
- `qml/pages/SettingsPage.qml:105–119` — the `tabNames[]` array in the accessibility announcer
- Every `goToSettings(N)` caller (currently `qml/main.qml:1640` and `qml/main.qml:3650`)
- Every `tabIndex: N` entry in `qml/components/SettingsSearchIndex.js` (47 entries today)
- `// Tab N: ...` comments next to each Loader (documentation only, but also drifts)

This has already caused at least one user-visible bug: in **v1.7.0**, clicking "Update Now" on the update-available popup opened the Firmware tab instead of the About tab. `#817` (feat: firmware update over BLE) inserted a new Firmware tab at index 11 without updating the `goToSettings(11)` caller in `main.qml`, which had been written when About lived at that index. The issue was accidentally fixed by `#831` (Firmware merged into About), but the underlying fragility remains.

## Proposal

Replace integer indexes with symbolic names. Two viable shapes:

**Option A — named constants exported from `SettingsPage.qml`**
```qml
// SettingsPage.qml
readonly property int tabConnections: 0
readonly property int tabMachine: 1
// ...
readonly property int tabAbout: 11
```
Callers use `SettingsPage.tabAbout` etc. Same for the search index.

**Option B — string keys resolved inside `SettingsPage`**
```qml
goToSettings(\"about\")
```
`SettingsPage.qml` owns a map from name → index. Announcer and search index use the same names.

Option B is cleaner: adding a tab then only requires updating the resolver map, and callers don't need to know indexes at all. Option A is closer to the existing code and a smaller diff.

## Acceptance

- No integer literal indexes in `goToSettings` calls, `SettingsSearchIndex.js` entries, or the accessibility announcer's `tabNames[]`
- Adding or reordering a tab requires changes in only one place (the tab definitions in `SettingsPage.qml`)
- Existing behavior unchanged

## Out of scope

- No functional changes to Settings behavior
- Debug-tab visibility logic (`Settings.isDebugBuild`) stays as-is

skialpine:
Closes #833.

## Summary
- Adds a `SettingsTabs` QML singleton that owns the ordered tab definitions (id, translation key, fallback, Loader source, `loadSync`, `debugOnly`) plus `indexOf(id)`, `tabName(id)`, `tabLabels`, `visibleTabs()`, and `visibleTabNames()` helpers.
- `SettingsPage.qml` now drives both the `TabBar` buttons and the `StackLayout` Loaders from Repeaters keyed to `SettingsTabs.visibleTabs()`. The per-tab magic numbers in `active: N in loadedTabs` are gone, along with the hand-maintained `tabNames[]` announcer array.
- Navigation uses symbolic ids: `goToSettings(\"about\")`, `goToSettings(\"historyData\")`. `SettingsPage.requestedTabIndex` became `requestedTabId` (string).
- `SettingsSearchIndex.js` entries use `tabId` strings instead of `tabIndex` integers; `SettingsSearchDialog` emits `resultSelected(tabId, cardId)` and reads localized names via `SettingsTabs.tabName(tabId)`. The old `getTabName(index, tr)` helper is gone.
- Bonus fix: language-tab label branch now picks up language changes (reads `translationVersion`).

## Why it matters
The tab order used to live in ~5 places kept in lockstep by convention. #817 inserted a new Firmware tab and the \`goToSettings(11)\` caller in \`main.qml\` silently routed \"Update Now\" to the wrong tab — exactly the failure mode the issue describes. Adding or reordering a tab now only requires editing \`SettingsTabs.qml\`.

## Test plan
- [ ] Build & launch; open Settings, confirm all 12 (or 13 in debug builds) tabs render in the same order and localized labels match.
- [ ] Tap every tab; confirm lazy-loading still works (first visit loads content, subsequent visits are instant).
- [ ] From idle, trigger the update-available dialog → \"Update Now\" → lands on the **About** tab.
- [ ] Trigger the empty-database dialog → \"Restore\" → lands on **History & Data**.
- [ ] Open Settings search; pick a result in each tab and confirm navigation + scroll-to-card still works; verify the badge on the Language tab still appears for non-English locales with untranslated strings.
- [ ] TalkBack/VoiceOver: announce each tab on selection reports the correct localized name.
- [ ] Switch language; confirm all tab labels update.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -586,8 +586,9 @@ if(IOS)
     target_include_directories(Decenza PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/ios)
 endif()
 
-# QML module - mark Theme.qml as singleton
+# QML module - mark singletons
 set_source_files_properties(qml/Theme.qml PROPERTIES QT_QML_SINGLETON_TYPE TRUE)
+set_source_files_properties(qml/components/SettingsTabs.qml PROPERTIES QT_QML_SINGLETON_TYPE TRUE)
 
 # Define QML files list for dependency tracking
 set(QML_FILES
@@ -711,6 +712,7 @@ set(QML_FILES
     qml/components/DateUtils.js
     qml/components/qrcode.js
     qml/components/SettingsSearchIndex.js
+    qml/components/SettingsTabs.qml
 
     qml/components/layout/LayoutItemDelegate.qml
     qml/components/layout/LayoutBarZone.qml
--- a/qml/components/SettingsSearchIndex.js
+++ b/qml/components/SettingsSearchIndex.js
@@ -23,7 +23,11 @@ function getSearchEntries(tr) {
           description: tr("settings.search.scaleConnectionDesc", "Pair a Bluetooth scale or refractometer"),
           keywords: ["scale", "acaia", "decent", "felicita", "weight", "pair", "tds", "refractometer", "atago"] },
 
-        // Screensaver (auto-sleep and auto-wake moved from Machine tab)
+        // Screensaver — auto-sleep and auto-wake also live on the Screensaver tab
+        { tabId: "screensaver", cardId: "screensaver",
+          title: tr("settings.screensaver.settings", "Screensaver Settings"),
+          description: tr("settings.search.screensaverDesc", "Choose screensaver type and settings"),
+          keywords: ["screensaver", "attractor", "pipes", "clock", "video", "image", "dim"] },
         { tabId: "screensaver", cardId: "autoSleep",
           title: tr("settings.preferences.autoSleep", "Auto-Sleep"),
           description: tr("settings.preferences.autoSleepDesc", "Put the machine to sleep after inactivity"),
@@ -162,12 +166,6 @@ function getSearchEntries(tr) {
           title: tr("settings.layout.title", "Home Screen Layout"),
           description: tr("settings.search.layoutDesc", "Customize idle screen widgets and zones"),
           keywords: ["layout", "widget", "zone", "customize", "idle", "home", "editor"] },
-
-        // Screensaver
-        { tabId: "screensaver", cardId: "screensaver",
-          title: tr("settings.screensaver.settings", "Screensaver Settings"),
-          description: tr("settings.search.screensaverDesc", "Choose screensaver type and settings"),
-          keywords: ["screensaver", "attractor", "pipes", "clock", "video", "image", "dim"] },
 
         // Visualizer
         { tabId: "visualizer", cardId: "visualizer",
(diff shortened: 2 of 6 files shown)


skialpine:
### Code review

Found 3 issues:

1. `themesLoader` is no longer defined. The refactor replaced the named `Loader`s with a `Repeater { id: tabLoaders }`, but the Save Theme dialog still references the old id in two places. Typing Enter in the theme-name field or clicking **Save** will throw `ReferenceError: themesLoader is not defined` before `saveThemeDialog.close()` runs — the theme saves (first call), but the dialog never closes and the presets list never refreshes.

https://github.com/Kulitorum/Decenza/blob/83085b290f6f480bead794b9213aeb620eb0d479/qml/pages/SettingsPage.qml#L304-L312

https://github.com/Kulitorum/Decenza/blob/83085b290f6f480bead794b9213aeb620eb0d479/qml/pages/SettingsPage.qml#L331-L340

Fix: look the loader up via the Repeater, e.g. `var themesLoader = tabLoaders.itemAt(SettingsTabs.indexOf("themes"))` and move `saveThemeDialog.close()` out of the `if (themesLoader?.item)` guard so the dialog always closes.

2. Stale section comment. `// Tab content area - all tabs preload in background` contradicts the actual behaviour (only `connections` has `loadSync: true`; everything else lazy-loads on first visit — correctly documented 14 lines below). CLAUDE.md's comment guidance warns against misleading comments.

https://github.com/Kulitorum/Decenza/blob/83085b290f6f480bead794b9213aeb620eb0d479/qml/pages/SettingsPage.qml#L194-L213

3. `SettingsSearchIndex.js` has two separate `// Screensaver` section headers (auto-sleep/auto-wake block near the top, plain screensaver block further down). A future editor adding a screensaver entry is likely to add it to the wrong block or miss the split.

https://github.com/Kulitorum/Decenza/blob/83085b290f6f480bead794b9213aeb620eb0d479/qml/components/SettingsSearchIndex.js#L25-L60

https://github.com/Kulitorum/Decenza/blob/83085b290f6f480bead794b9213aeb620eb0d479/qml/components/SettingsSearchIndex.js#L163-L172

<details>
<summary>2 below-threshold issues discarded</summary>

- [OK to skip] `visibleTabs()` as a function allocates a fresh array on each call — binding deps (`tabs`, `Settings.isDebugBuild`) are constant, so Repeaters never re-diff; the allocations only happen in low-frequency signal handlers (score: 25)
- [OK to skip] `id` as a JS-model key might shadow QML's `id` keyword — `id` is not in CLAUDE.md's reserved-names list, and QML ids are compile-time only (score: 5)

</details>

PR #838, opened 2026-04-23
feat(auto-favorites): show target yield and split cards per recipe

skialpine:
## Summary
Two related changes to make the Auto-Favorites list reflect actual recipes:

1. The \"dose to yield\" chip now displays the **target** yield (from \`yield_override\`) instead of the most recent shot's scale reading, so what's on the card is what tapping the favorite will load.
2. Cards are now split by target yield (exact) and dose rounded to the nearest 0.5 g. Previously an 18 g to 36 g shot and an 18 g to 40 g shot on the same bean collapsed into one card. Now each distinct recipe gets its own card. Dose tweaks like 18.1 / 18.2 stay in the same \"18 g\" bucket; 18 g vs 18.5 g correctly split.

## How
- \`shothistorystorage.cpp\`: \`requestAutoFavorites\` adds \`ROUND(COALESCE(dose_weight, 0) * 2) / 2.0\` (dose bucket) and \`COALESCE(yield_override, 0)\` to every groupBy mode. The outer SELECT returns the bucketed dose as \`dose_weight\` and the group's \`yield_override\` so the displayed values are the ones actually loaded.
- \`MainController::loadShotWithMetadata\` gains an optional \`doseOverride\` parameter. When set, the dose is applied via a queued \`setDyeBeanWeight\` so it runs *after* \`ProfileManager::loadProfile\`'s own deferred \`setDyeBeanWeight(recommendedDose)\` — otherwise the profile's recommended dose could overwrite the favorite's dose on profiles that declare one.
- \`AutoFavoritesPage.qml\`:
  - New \`recipeYield(yieldOverride, finalWeight)\` helper prefers \`yield_override\`, falls back to \`finalWeight\` for legacy rows where no target was stored (matches the legacy branch in \`applyLoadedShotMetadata\`).
  - Recipe chip and accessible text go through that helper.
  - Load button passes the bucketed \`model.doseWeight\` so the DYE dose matches the card.

## Why \`yield_override\` is the target
\`setBrewYieldOverride\` is called with the profile's \`targetWeight()\` every profile load (\`profilemanager.cpp:1063-1064\`), so by shot-save time \`m_settings->brewYieldOverride()\` already holds the user's manual override OR the profile default. It's only 0 for legacy volume/timer-based profiles or rows saved before this behavior existed — those fall back to \`finalWeight\` in both display and load paths.

## Test plan
- [ ] Build in Qt Creator.
- [ ] Open Auto-Favorites. Verify each card shows \`dose g to target g\` where \`target\` is the profile's target (or the user's yield override) on the latest shot.
- [ ] Run two shots on the same bean/profile: 18 g to 36 g then 18 g to 40 g. Refresh. Expect two cards.
- [ ] Run three shots on the same bean/profile/yield with doses 18.0, 18.1, 18.2 g. Expect a single \"18.0 g\" card.
- [ ] Run two shots on the same bean/profile/yield with doses 18.0 g and 18.5 g. Expect two cards.
- [ ] Tap an 18 g to 36 g card: DYE dose should load as 18.0 g (even if the latest underlying shot was 18.2 g). Yield target should load as 36.0 g.
- [ ] Tap a favorite whose profile declares a \`recommendedDose\`: confirm the card's bucketed dose wins over the profile's recommended dose.
- [ ] Legacy shot with \`yield_override = 0\` (if any exist in your DB): card should show \`finalWeight\` and tapping loads that same value.
- [ ] VoiceOver/TalkBack reads the target yield (not the actual) in the recipe summary.

### DIFF
--- a/qml/pages/AutoFavoritesPage.qml
+++ b/qml/pages/AutoFavoritesPage.qml
@@ -63,8 +63,15 @@ Page {
         }
     }
 
+    // Target yield that will be loaded when this favorite is tapped — matches the
+    // priority applied in applyLoadedShotMetadata: yield_override if > 0, else the
+    // last shot's finalWeight (the old-volume-profile fallback).
+    function recipeYield(yieldOverride, finalWeight) {
+        return yieldOverride > 0 ? yieldOverride : (finalWeight || 0)
+    }
+
     // Build accessible text based on current groupBy setting
-    function buildGroupByText(beanBrand, beanType, profileName, grinderBrand, grinderModel, grinderSetting, doseWeight, finalWeight, shotCount, avgEnjoyment) {
+    function buildGroupByText(beanBrand, beanType, profileName, grinderBrand, grinderModel, grinderSetting, doseWeight, yieldOverride, finalWeight, shotCount, avgEnjoyment) {
         var includes = getGroupByIncludes()
         var parts = []
 
@@ -85,7 +92,7 @@ Page {
         }
 
         // Always include recipe summary
-        parts.push((doseWeight || 0).toFixed(1) + "g to " + (finalWeight || 0).toFixed(1) + "g")
+        parts.push((doseWeight || 0).toFixed(1) + "g to " + recipeYield(yieldOverride, finalWeight).toFixed(1) + "g")
         parts.push(shotCount + " " + TranslationManager.translate("autofavorites.shots", "shots"))
         if (avgEnjoyment > 0)
             parts.push(avgEnjoyment + "% enjoyment")
@@ -196,7 +203,8 @@ Page {
                 property string _groupByText: autoFavoritesPage.buildGroupByText(
                     model.beanBrand, model.beanType, model.profileName,
                     model.grinderBrand, model.grinderModel, model.grinderSetting,
-                    model.doseWeight, model.finalWeight, model.shotCount, model.avgEnjoyment)
+                    model.doseWeight, model.yieldOverride, model.finalWeight,
+                    model.shotCount, model.avgEnjoyment)
 
                 // Whole card announces full details based on groupBy setting
                 AccessibleMouseArea {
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

No issues found at the reporting threshold. Checked for bugs and CLAUDE.md compliance.

<details>

- [Worth a look] `recipeYield()` comment overstates parity with `applyLoadedShotMetadata`. The loader only falls back to `finalWeight` when the current profile's `targetWeight() <= 0`; the QML helper falls back unconditionally when `yieldOverride == 0`. For a modern profile with `yield_override == 0` the card can show `finalWeight` while the load path actually keeps the profile default — the comment should clarify this approximation. (score: 72)

  https://github.com/Kulitorum/Decenza/blob/6fe5a061ff16762bc5cb3c0db6f8311f5bbdba3e/qml/pages/AutoFavoritesPage.qml#L66-L71

- [Worth a look] Brew-dialog dose flicker when `autoFavoritesOpenBrewSettings` is on. `applyLoadedShotMetadata` queues the dose write via `Qt::QueuedConnection`, but `shotMetadataLoaded(shotId, true)` is emitted synchronously, so the page pop → brew dialog open runs before the queued setter. The dialog briefly shows the profile-recommended dose (or the prior value) until the queued event drains. Consider emitting the signal inside the queued callback, or explicitly noting the flicker. (score: 50)

  https://github.com/Kulitorum/Decenza/blob/6fe5a061ff16762bc5cb3c0db6f8311f5bbdba3e/src/controllers/maincontroller.cpp#L463-L475

- [OK to skip] Three numeric defaults use `||` instead of `??` on weight fields. CLAUDE.md's QML section says \"Use `??` not `||` for numeric defaults where 0 is valid\", but here every fallback is itself `0`, so behaviour is identical; purely stylistic. (score: 50)

  https://github.com/Kulitorum/Decenza/blob/6fe5a061ff16762bc5cb3c0db6f8311f5bbdba3e/qml/pages/AutoFavoritesPage.qml#L67-L72

- [OK to skip] FIFO guarantee comment could be more explicit. The dose-queue comment says \"ours wins\" without spelling out that Qt delivers queued events in posting order within a thread; a future maintainer refactoring `loadProfile` to a direct call could silently break the invariant. (score: 72)

  https://github.com/Kulitorum/Decenza/blob/6fe5a061ff16762bc5cb3c0db6f8311f5bbdba3e/src/controllers/maincontroller.cpp#L463-L475

</details>

PR #839, opened 2026-04-23
refactor(auto-favorites): opt into recipe splitting via Weight group-by

skialpine:
## Summary
- PR #838 unconditionally split Auto-Favorites cards by (rounded dose, target yield) and changed the "X g to Y g" chip to read `yield_override` for every Group by mode. Users expected the four existing modes ("Bean only", "Profile only", "Bean + Profile", "Bean + Profile + Grinder") to keep collapsing shots the way they did before #838.
- This PR moves both behaviours behind a new fifth Group by option **"Bean + Profile + Grinder + Weight"**. The four existing options revert to their pre-#838 behaviour: no dose/yield splitting, and the chip falls back to the latest shot's `finalWeight`.
- Storage layer already had WIP for this (under the internal name `recipe`); the commit renames it to `weight` and wires the fifth option into the QML Group by combo + `getGroupByIncludes()`.
- No controller changes. `loadShotWithMetadata`'s `doseOverride` param and the queued `shotMetadataLoaded` emit from #838 are no-ops when `doseOverride` equals the shot's own dose (the non-weight case), so they stay in place and only do work in weight mode.

## Test plan
- [ ] Build in Qt Creator.
- [ ] Open Auto-Favorites → settings gear → Group by. Confirm the dropdown now lists five options; the new one is "Bean + Profile + Grinder + Weight".
- [ ] With any of the four existing options selected, confirm an 18 g → 36 g shot and an 18 g → 40 g shot on the same bean/profile/grinder share one card and the chip shows the latest shot's `finalWeight` (pre-#838 behaviour).
- [ ] With "Bean + Profile + Grinder + Weight" selected, confirm those same shots split into separate cards; 18.1 g and 18.2 g stay in the 18.0 g bucket while 18 g and 18.5 g split.
- [ ] Tap a card in weight mode and confirm DYE loads with the bucketed dose shown on the card.
- [ ] Tap a card in a non-weight mode and confirm DYE loads with the shot's saved dose (no regression vs pre-#838).
- [ ] Smoke-test `ShotHistoryPage`'s Load flow — still calls `loadShotWithMetadata(shotId)` with no override and should behave identically.

### DIFF
--- a/qml/pages/AutoFavoriteInfoPage.qml
+++ b/qml/pages/AutoFavoriteInfoPage.qml
@@ -18,6 +18,8 @@ Page {
     property string grinderBrand: ""
     property string grinderModel: ""
     property string grinderSetting: ""
+    property real doseBucket: 0
+    property real yieldOverride: 0
     property int avgEnjoyment: 0
     property int shotCount: 0
 
@@ -38,7 +40,8 @@ Page {
             MainController.shotHistory.requestShot(shotId)
 
         MainController.shotHistory.requestAutoFavoriteGroupDetails(
-            groupBy, beanBrand, beanType, profileName, grinderBrand, grinderModel, grinderSetting)
+            groupBy, beanBrand, beanType, profileName, grinderBrand, grinderModel, grinderSetting,
+            doseBucket, yieldOverride)
     }
 
     // Handle async shot data and group details
--- a/qml/pages/AutoFavoritesPage.qml
+++ b/qml/pages/AutoFavoritesPage.qml
@@ -347,9 +347,10 @@ Page {
                                 ". " + favoriteDelegate._groupByText
                             accessibleItem: infoButton
                             onAccessibleClicked: {
-                                // In weight mode, model.doseWeight is the group's bucketed dose
-                                // and model.yieldOverride is the group's exact target yield — pass
-                                // them through so the Info page can scope stats to the same bucket.
+                                // In weight mode, model.doseBucket is the group's rounded dose
+                                // (used to scope stats) while model.doseWeight is the latest
+                                // shot's raw dose (shown on the card). Pass the bucket so the
+                                // Info page's averages cover the same shots the card aggregates.
                                 pageStack.push(Qt.resolvedUrl("AutoFavoriteInfoPage.qml"), {
                                     shotId: model.shotId,
                                     groupBy: Settings.autoFavoritesGroupBy,
@@ -359,7 +360,7 @@ Page {
                                     grinderBrand: model.grinderBrand || "",
                                     grinderModel: model.grinderModel || "",
                                     grinderSetting: model.grinderSetting || "",
-                                    doseBucket: model.doseWeight || 0,
+                                    doseBucket: model.doseBucket || 0,
                                     yieldOverride: model.yieldOverride || 0,
                                     avgEnjoyment: model.avgEnjoyment || 0,
                                     shotCount: model.shotCount || 0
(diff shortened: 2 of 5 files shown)


skialpine:
### Code review

Found 2 issues:

1. `ShotHistoryStorage::requestAutoFavoriteGroupDetails` has no branch for the new `bean_profile_grinder_weight` mode. `AutoFavoritesPage.qml` passes `Settings.autoFavoritesGroupBy` unchanged to this function, and any value other than `"bean"`, `"profile"`, or `"bean_profile_grinder"` falls through to the `bean_profile` default — stripping grinder filtering entirely. The Info page for a weight-split card will therefore aggregate avgTds / avgEy / avgDuration / avgDose / avgYield / avgTemperature / notes across every grinder and every dose/yield bucket that shares the card's bean + profile, showing a misleading superset of what the card represents. Fully fixing it also requires passing doseBucket + yieldOverride through the function signature so the stats actually match the card's scope.

https://github.com/Kulitorum/Decenza/blob/b7adbcd91eb3c710bf29543f0f5727ba08e99263/src/history/shothistorystorage.cpp#L2825-L2844

2. The card's `Show` button builds a `ShotHistoryPage` `initialFilter` from `getGroupByIncludes()`, which only returns bean / profile / grinder booleans. In `bean_profile_grinder_weight` mode the card also represents a specific (dose bucket, target yield), but neither `model.doseWeight` nor `model.yieldOverride` is added to `filter`, and `ShotHistoryPage.buildFilter()` only pulls `profileName`, `beanBrand`, `beanType`, `grinderBrand`, `grinderModel`, `grinderSetting` from `initialFilter`. Tapping Show on a weight-split card therefore opens shot history scoped only to bean + profile + grinder — the user sees shots outside the card's bucket.

https://github.com/Kulitorum/Decenza/blob/b7adbcd91eb3c710bf29543f0f5727ba08e99263/qml/pages/AutoFavoritesPage.qml#L386-L409

<details>

- [Worth a look] `recipeYield()` block comment says `yield_override` is populated "(modern shots always save it)" and the `finalWeight` fallback is for "legacy rows" — but the new SQL emits `0 AS yield_override` literally for every mode except `bean_profile_grinder_weight`, so in four of five modes the fallback is now the normal path regardless of whether the shot is legacy. Worth tightening the wording. (score: 50)

</details>

skialpine:
### Code review

Found 2 issues in the latest two commits:

1. Show button yield filter in weight mode filters against `final_weight` instead of `yield_override`, so it almost always matches zero shots. `minYield`/`maxYield` in the backend map to `final_weight >= ? AND final_weight <= ?` (src/history/shothistorystorage.cpp:1426-1427), but the card's `model.yieldOverride` is the user's target yield — actual pour weight rarely equals the target to float precision. The Show button in weight mode will return an empty list despite the card showing shotCount > 0. Fix either adds a new `yieldOverride` filter key that matches on `COALESCE(yield_override, 0) = ?`, or widens `minYield`/`maxYield` to a tolerance (e.g. ±1g) and accepts the semantic drift.

https://github.com/Kulitorum/Decenza/blob/17a38733b2563e9997e0f0a24c3ad26ebbf320f9/qml/pages/AutoFavoritesPage.qml#L419-L427

2. New filter-banner fragments "g dose" and "g yield" are not wrapped in `TranslationManager.translate(...)` while every other label in the same banner is. CLAUDE.md says [`All user-visible text must be internationalized. Use TranslationManager.translate("section.key", "Fallback text") for property bindings and inline expressions.`](https://github.com/Kulitorum/Decenza/blob/17a38733b2563e9997e0f0a24c3ad26ebbf320f9/CLAUDE.md).

https://github.com/Kulitorum/Decenza/blob/17a38733b2563e9997e0f0a24c3ad26ebbf320f9/qml/pages/ShotHistoryPage.qml#L500-L510

<details>

- [Worth a look] Comment at `AutoFavoritesPage.qml:463` still says "Pass the bucketed dose" but since commit 17a38733 the code passes the raw latest dose (score: 50)
- [OK to skip] `minYield === maxYield` in the filter banner uses exact float equality — currently safe since both come from the same source assignment, but fragile if any future caller introduces a tolerance range (score: 50)
- [Worth a look] Comment in `requestAutoFavoriteGroupDetails` weight branch still says "the (dose bucket, target yield) that the card represents" — the card represents the raw latest dose now, the bucket is only the group boundary (score: 50)

</details>

PR #845, opened 2026-04-24
fix(update): prevent downgrade install; add settings crash debug logging

skialpine:
## Summary

- **UpdateChecker downgrade fix**: `downloadAndInstall()` now returns early if `m_updateAvailable` is false, preventing a stale download URL (e.g. latest stable v1.6.3 while running v1.7.2 beta) from installing an older version. Also clears the cached APK path immediately when a version check flips `updateAvailable` to false.
- **Crash debug logging**: `SettingsPage` now logs which async tab starts loading, finishes, or errors. The next SIGSEGV in `QQmlConnections::connectSignalsToMethods()` (issue #844) will show the exact tab name as the last debug line before the crash.

## Context

Issue #844 crash report showed v1.7.2 (beta/prerelease) downloading and installing v1.6.3 (older stable) during a settings browsing session. The `downloadAndInstall()` function had no guard for `m_updateAvailable`, so a stale URL could be used even after the version check determined no update was needed. The crash itself (SIGSEGV during async QML incubation) requires another reproduction to identify the exact `Connections` block; the logging change makes that identification trivial.

## Test plan

- [ ] Run as a beta version with beta updates disabled — confirm update checker finds stable release, confirm `downloadAndInstall()` logs warning and does nothing
- [ ] Run as older version — confirm normal update flow still works (downloads and installs correctly)
- [ ] Browse all settings tabs — confirm `SettingsPage: async loading tab <id>` / `tab ready <id>` logs appear in debug output for each tab visited

### DIFF
--- a/qml/pages/SettingsPage.qml
+++ b/qml/pages/SettingsPage.qml
@@ -222,6 +222,15 @@ Page {
                 asynchronous: !modelData.loadSync
                 source: modelData.source
 
+                onStatusChanged: {
+                    if (status === Loader.Loading)
+                        console.log("SettingsPage: async loading tab", tabId)
+                    else if (status === Loader.Ready)
+                        console.log("SettingsPage: tab ready", tabId)
+                    else if (status === Loader.Error)
+                        console.log("SettingsPage: tab load error", tabId)
+                }
+
                 onLoaded: {
                     // Themes tab emits a signal requesting the Save Theme dialog
                     if (tabId === "themes" && item) {
--- a/qml/pages/SettingsPage.qml
+++ b/qml/pages/SettingsPage.qml
@@ -228,7 +228,7 @@ Page {
                     else if (status === Loader.Ready)
                         console.log("SettingsPage: tab ready", tabId)
                     else if (status === Loader.Error)
-                        console.log("SettingsPage: tab load error", tabId)
+                        console.warn("SettingsPage: tab load error", tabId)
                 }
 
                 onLoaded: {
--- a/src/core/updatechecker.cpp
+++ b/src/core/updatechecker.cpp
@@ -204,6 +204,7 @@ void UpdateChecker::parseReleaseInfo(const QByteArray& data)
     QJsonDocument doc = QJsonDocument::fromJson(data);
     if (!doc.isArray()) {
         m_errorMessage = "Invalid response from GitHub";
+        qWarning() << "UpdateChecker:" << m_errorMessage << "- response:" << data.left(200);
         emit errorMessageChanged();
         return;
     }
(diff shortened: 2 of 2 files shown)


skialpine:
### Code review

Found 3 issues:

1. Missing `!m_installInFlight` guard in the new APK cache cleanup block. `checkForUpdates()` and `onPeriodicCheck()` only guard on `!m_downloading`, not `!m_installInFlight`, so a periodic check can run while PackageInstaller is actively streaming the APK. If that check flips `m_updateAvailable` to false, the new block clears `m_downloadedApkPath` while the install session is in flight. When `onInstallStatus` fires for a failed install, its cleanup branches check `!m_downloadedApkPath.isEmpty()` — but the path is already gone, so the file is silently left on disk. The comment at line 328 says all three cleanup sites must be kept consistent; the other two (`dismissUpdate`, `onInstallStatus`) both respect install state.

https://github.com/Kulitorum/Decenza/blob/6f9ce11a916d670eb40a1622107ea4ac17b3f064/src/core/updatechecker.cpp#L327-L335

Fix: add `&& !m_installInFlight` to the condition at line 330.

2. `Loader.Error` logged at `console.log` level instead of `console.warn`. The `scrollToCard` function in the same file already uses `console.warn` for `Loader.Error`, establishing the severity precedent. Using `console.log` for the error case makes it invisible in filtered log views and in Qt Creator's warnings-only output — which matters because this logging exists specifically to diagnose a crash.

https://github.com/Kulitorum/Decenza/blob/6f9ce11a916d670eb40a1622107ea4ac17b3f064/qml/pages/SettingsPage.qml#L230-L232

https://github.com/Kulitorum/Decenza/blob/6f9ce11a916d670eb40a1622107ea4ac17b3f064/qml/pages/SettingsPage.qml#L386-L389

3. Comment at line 328 says `dismissUpdate()`, `onInstallStatus()`, and the new block "contain equivalent cleanup; all three must be kept consistent" — but they are not equivalent. `dismissUpdate()` calls `QFile::remove()` to delete the file; the new block explicitly leaves it on disk ("The file itself is left for OS cleanup"). The "keep consistent" directive is misleading: a maintainer aligning the three sites could either inadvertently add a file deletion here, or remove it from `dismissUpdate()`.

https://github.com/Kulitorum/Decenza/blob/6f9ce11a916d670eb40a1622107ea4ac17b3f064/src/core/updatechecker.cpp#L326-L335

---

<details>

- [Worth a look] APK cleanup block only fires on `m_updateAvailable` true→false transition; if a download happened, the app was sideloaded to a newer version, and the next check finds `wasAvailable == false` already, `m_downloadedApkPath` is never cleared (score: 63)
- [OK to skip] `console.log` for `Loading`/`Ready` fires on every settings page visit in production — intentional temporary diagnostic, but no removal plan or comment (score: 50)
- [Worth a look] New `qWarning` for missing platform asset fires on iOS during periodic checks because the asset loop has no iOS branch and `onPeriodicCheck` has no iOS guard (score: 65)

</details>

PR #846, opened 2026-04-25
[codex] Avoid async settings tab incubation

skialpine:
## Summary

- Keep Settings tabs lazy-loaded, but instantiate visited tabs synchronously instead of through async QML incubation.
- Update the SettingsTabs comment so `loadSync` describes eager startup loading rather than async behavior.

## Why

Issue #844 still has an unknown crash after the downgrade fix. The crash stack dies in `QQmlConnections::connectSignalsToMethods()` under `QQmlIncubationController::incubateFor()` while the user was browsing settings. Settings tabs were the main place where we lazily created many `Connections` objects through async loaders, and this repo already has a documented workaround for the same Qt crash class by avoiding signal activity during QML incubation.

This change preserves lazy loading, but removes async incubation from Settings tab creation.

## Validation

- `git diff --check`

### DIFF
--- a/qml/components/SettingsTabs.qml
+++ b/qml/components/SettingsTabs.qml
@@ -12,7 +12,7 @@ QtObject {
     // key       — TranslationManager key for the localized tab label
     // fallback  — English fallback used when no translation is available
     // source    — QML source loaded into the tab's StackLayout page
-    // loadSync  — if true, the Loader loads synchronously on startup (first tab only)
+    // loadSync  — if true, the Loader loads eagerly on startup (first tab only)
     // debugOnly — if true, the tab is hidden unless Settings.isDebugBuild is set
     readonly property var tabs: [
         { id: "connections",    key: "settings.tab.connections",    fallback: "Connections",       source: "settings/SettingsConnectionsTab.qml",     loadSync: true,  debugOnly: false },
--- a/qml/components/SettingsTabs.qml
+++ b/qml/components/SettingsTabs.qml
@@ -12,7 +12,7 @@ QtObject {
     // key       — TranslationManager key for the localized tab label
     // fallback  — English fallback used when no translation is available
     // source    — QML source loaded into the tab's StackLayout page
-    // loadSync  — if true, the Loader loads eagerly on startup (first tab only)
+    // loadSync  — if true, the tab is activated immediately when the page opens rather than on first visit (asynchronous: false is enforced globally in SettingsPage.qml)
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

Found 3 issues:

1. Stale "async Loader" comment in `scrollToCard` — the `else` branch says "Wait for async Loader to finish via statusChanged signal" but `asynchronous` is now hardcoded to `false`; the comment describes a mode that no longer exists and will mislead future maintainers about what that branch does

https://github.com/Kulitorum/Decenza/blob/f44203d55af04adb6ae20e38d345e1116849c213/qml/pages/SettingsPage.qml#L383-L396

2. Debug `console.log` for `Loader.Loading` and `Loader.Ready` fires on every settings tab visit; these were added in #845 for crash diagnosis but the PR retains them with no comment indicating they are temporary or have a removal condition — consider adding a `// TODO: remove after #844 is confirmed resolved` or gating behind a debug build flag

https://github.com/Kulitorum/Decenza/blob/f44203d55af04adb6ae20e38d345e1116849c213/qml/pages/SettingsPage.qml#L229-L236

3. New comment says "keeps Connections setup on the tab-switch event" — with `asynchronous: false`, Connections are set up during synchronous instantiation triggered by `active` flipping to `true`, not specifically tied to any named event; the phrasing implies an event-based guarantee that doesn't actually exist and doesn't clarify that `loadSync` no longer drives sync/async behavior (it now only controls `active`)

https://github.com/Kulitorum/Decenza/blob/f44203d55af04adb6ae20e38d345e1116849c213/qml/pages/SettingsPage.qml#L221-L227

---

<details>

- [Worth a look] `loadSync` comment in SettingsTabs.qml updated from "synchronously" to "eagerly" is correct, but a reader of SettingsTabs.qml in isolation has no indication that `asynchronous` is globally forced to `false` in SettingsPage.qml — the two files' semantics now diverge silently (score: 50)

https://github.com/Kulitorum/Decenza/blob/f44203d55af04adb6ae20e38d345e1116849c213/qml/components/SettingsTabs.qml#L14-L16

- [OK to skip] `onLoaded` for the Themes tab now fires synchronously during a tap; the Themes tab is 497 lines so the main-thread instantiation cost is real, but this is the acknowledged trade-off of the fix and the comment block explains the intent (score: 75, accepted trade-off)

https://github.com/Kulitorum/Decenza/blob/f44203d55af04adb6ae20e38d345e1116849c213/qml/pages/SettingsPage.qml#L237-L245

</details>

PR #848, opened 2026-04-25
feat(saw): per-(profile, scale) batched SAW learning (#847)

### LINKED ISSUE #847: [Bug] SAW Weight Stop Timing learning should be per-profile (portafilter-dependent overshoot/undershoot when switching profiles)
### What happened?

When switching between an open portafilter and a double-spout portafilter, the stop-at-weight timing (SAW learning / Weight Stop Timing) adapts globally and learning from one setup affects the other. Most of the time I use a double-spout portafilter with a frequently-used profile, leading to a slight overshoot compared to the target (more coffee than intended). When I use a less-frequent profile with an open portafilter, it tends to underdose (stop too early, less coffee). This is repeatable when interchanging portafilters and profiles — it's clear the learning is not profile-specific. Expected: SAW learning (Weight Stop Timing adjustment) should be per-profile, or at least have a per-profile calibration option, so switching portafilter/profile does not cause systematic error.

### Steps to reproduce

1. Use a double-spout portafilter with a frequently-used profile, let SAW learning adapt.
2. Switch to a less-used profile with an open portafilter.
3. Observe that the less-used profile with the open portafilter consistently underdoses (settled yield < target).
4. Switch back to the double-spout / frequent profile and see a slight overshoot (settled > target).

### Decenza version

1.7.1 (build 3312)

### Platform

Android

### Device / OS version

Samsung Tablet

### Scale (if applicable)

Decent Scale v1

_No response_

### Screenshots or video

_No response_

### Anything else?

Profile-specific learning or at least a way to reset/adjust SAW compensation per profile would solve this.

skialpine:
## Summary

Fixes #847: users alternating between profiles paired with different portafilters (open spout vs. double spout) experience systematic SAW overshoot/undershoot because the global model is pulled toward the dominant profile's drip dynamics.

**Root cause (from real device data, Apr 2026):**

| Shot | Profile | expectedDrip | Actual drip | Result |
|------|---------|-------------|-------------|--------|
| 1047 | D-Flow / Q | 1.24 g | ~0.8 g | **0.4 g over-predicted** |
| 1059 | 80's Espresso | 1.13 g | 1.1 g | accurate |

Shot 1047 is the smoking gun — D-Flow/Q's lag is consistently ~0.8 g but the global model predicts 1.24 g because 80's Espresso (the more frequent profile) dominates the shared history.

## What Changes

- **Per-(profile, scale) history** keyed by `"<profileFilename>::<scaleType>"` — each combo learns its own drip dynamics in isolation
- **Batched median commits (N=5)**: model only updates after 5 shots, same as flow cal (#739). Median is outlier-resistant by construction
- **IQR dispersion gate**: batch dropped if IQR > 1.0 s or any lag deviates > 1.5 s from batch median — catches mid-batch condition changes (new beans, grind)
- **Global bootstrap lag**: IQR-fenced median of graduated pairs on the same scale; new profile/scale combos cold-start from this instead of the static `sensorLag()` constant
- **Fallback chain**: `perProfile` (≥3 medians) → `globalBootstrap` → `globalPool` (legacy) → `scaleDefault`
- **Legacy pool preserved**: every committed batch median is mirrored into `saw/learningHistory` so `isSawConverged()` and pre-update data keep working — no migration needed
- **Per-shot debug log**: two new lines per shot — `[SAW] model: source=… lag=… profile=… scale=… historyN=k` at extraction start and `[SAW] accuracy: predictedDrip=… actualDrip=… delta=…` at settling — so any historic shot retrieved via `shots_get_debug_log` tells the SAW accuracy story end-to-end
- **Calibration tab**: source suffix `(per-profile)` / `(global bootstrap)` / `(global)` / `(default)` + per-profile reset action alongside the existing reset-all
- **MCP tool**: `reset_saw_learning_for_profile` clears one pair's history and pending batch, leaves other pairs intact

## Architecture

This mirrors the auto flow calibration design from #739 exactly: per-profile storage, batched median commits, global bootstrap from the pool of graduated pairs, and a graceful fallback chain. Reusing the same patterns reduces risk — they have a year of production evidence behind them. See `docs/CLAUDE_MD/SAW_LEARNING.md` for full analysis including the empirical data that motivated the design.

## Files

| File | Change |
|------|--------|
| `src/core/settings.h` / `.cpp` | New per-pair API, batch accumulator, IQR check, bootstrap recompute, fallback chain |
| `src/main.cpp` | Profile keyed into WeightProcessor snapshot + `addSawLearningPoint`; new per-shot debug lines |
| `src/mcp/mcptools_control.cpp` | `reset_saw_learning_for_profile` tool |
| `qml/pages/settings/SettingsCalibrationTab.qml` | Source suffix + per-profile reset |
| `tests/tst_saw_settings.cpp` | 9 new unit tests (isolation, batch, IQR, bootstrap, fallback chain, resets) |
| `tests/CMakeLists.txt` | `tst_saw_settings` target |
| `docs/CLAUDE_MD/SAW_LEARNING.md` | Design doc with empirical evidence and rationale |

## Test plan

- [ ] `ctest -R tst_saw_settings` — all 9 cases pass
- [ ] Pull 5 shots on profile A → system log shows `[SAW] accumulated drip=… (n/5)` then `committed median lag=…`
- [ ] Switch to profile B → log shows `globalBootstrap` or `scaleDefault` fallback (not A's lag)
- [ ] Pull 5 shots on B → B graduates with its own lag
- [ ] Switch back to A → A's lag unchanged from before B shots (**this is the issue #847 fix**)
- [ ] Calibration tab shows correct source suffix for active profile
- [ ] "Reset this profile" clears only the active pair; other pairs intact
- [ ] `reset_saw_learning_for_profile` MCP tool works via wscat

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -145,6 +145,7 @@ Detailed documentation lives in `docs/CLAUDE_MD/`. Read these when working in th
 | `EMOJI_SYSTEM.md` | Twemoji SVG rendering, adding/switching emoji sets |
 | `ACCESSIBILITY.md` | TalkBack/VoiceOver rules, focus order, anti-patterns, implementation plan |
 | `AUTO_FLOW_CALIBRATION.md` | Auto flow calibration algorithm, batched median updates, windowing, convergence |
+| `SAW_LEARNING.md` | Per-(profile, scale) stop-at-weight learning — analysis, batched median commits, IQR dispersion rejection, global bootstrap |
 | `FIRMWARE_UPDATE.md` | DE1 firmware update flow, source URL, validation rules, failure modes, simulator behaviour |
 
 Also in `docs/`:
--- a/docs/CLAUDE_MD/SAW_LEARNING.md
+++ b/docs/CLAUDE_MD/SAW_LEARNING.md
@@ -58,7 +58,7 @@ Three QSettings keys, each a JSON object keyed by `"<profileFilename>::<scaleTyp
 |-----|-------|------|---------|
 | `saw/perProfileHistory` | array of committed batch-median entries `{drip, flow, overshoot, scale, profile, ts, batchSize}` | 10 medians (~50 shots-worth) | Source of truth for `sawLearnedLagFor` / `getExpectedDripFor` once the pair has graduated (≥ 3 medians). |
 | `saw/perProfileBatch` | array of pending raw entries `{drip, flow, overshoot, scale, profile, ts}` (target size 5) | 5 (commit point) | Pending accumulator; flushed on commit or rejection. |
-| `saw/globalBootstrapLag/<scaleType>` | scalar `double` (seconds) | n/a | IQR-fenced median of last committed median lag from each graduated pair on this scale. Used as first-shot default for new pairs. |
+| `saw/globalBootstrapLag/<scaleType>` | scalar `double` (seconds) | n/a | IQR-fenced median of last committed median lag from each pair on this scale with at least one committed batch-median. Used as first-shot default for new pairs. (Graduation for the per-profile *read* path is a stricter ≥ 3 medians; the bootstrap is a cold-start prior, so it accepts pairs with any committed history — IQR fencing handles the rest.) |
 
 The legacy `saw/learningHistory` key is preserved as a **global pool**: every committed batch-median is mirrored into it (trim 50). This keeps `isSawConverged()` and the legacy convergence-divergence detection working without changes, and provides a final read-path fallback for users with pre-update data.
 
--- a/docs/CLAUDE_MD/SAW_LEARNING.md
+++ b/docs/CLAUDE_MD/SAW_LEARNING.md
@@ -1,0 +1,183 @@
+# SAW (Stop-At-Weight) Learning
+
+## Problem
+
+The DE1's stop-at-weight feature has to send the stop command *before* the target weight is reached, because there is unavoidable lag between the command and pump shutdown:
+
+```
+total drip ≈ flow_at_stop × (BLE_round_trip_lag + DE1_machine_lag) + post_shutoff_drip
+```
+
+The two hardware lags are scale-specific (`Settings::sensorLag()` documents them; e.g., Decent Scale 0.38 s, Acaia 0.69 s, Bookoo 0.50 s) and the post-shutoff drip is profile- and portafilter-specific. Together they yield a total "drip after stop trigger" the controller has to predict in advance, then trigger the stop when `current_weight + predicted_drip ≥ target`.
+
+For years the predictor learned a single drip-vs-flow curve per scale type, stored as a rolling history of `{drip, flow, scale, overshoot}` records in QSettings (`saw/learningHistory`). That model worked for users on a single profile but gradually accumulated systematic error for users who alternate between profiles with different drip dynamics.
+
+[Issue #847](https://github.com/Kulitorum/Decenza/issues/847) reported the failure mode in unambiguous terms: a user alternating between an open-spout portafilter with one profile and a double-spout portafilter with another saw the model overshoot for the dominant pairing and undershoot for the secondary pairing. The learning was being pulled by whichever profile-portafilter combo the user pulled most often, contaminating the prediction for the rarer one.
+
+## Empirical evidence (this device, Apr 2026)
+
+Pulling SAW debug logs from real shots on a Decent Scale showed the contamination directly:
+
+| Shot | Profile | Trigger flow | expectedDrip | Settled drip | Result |
+|------|---------|--------------|--------------|--------------|--------|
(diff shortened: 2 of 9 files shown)


skialpine:
### Code review

Found 5 issues:

1. **`recomputeGlobalSawBootstrap` includes under-graduated pairs** — bootstrap accumulator uses `!pairHistory.isEmpty()` (≥ 1 median) but graduation everywhere else is `pairHistory.size() >= 3`. Pairs with only 1 committed median (5 shots, well below the "graduated" threshold) contribute to the bootstrap, and the debug log emits "median of N graduated pairs" where N counts non-graduated pairs. Fix: change the `continue` guard to `if (pairHistory.size() < 3) continue;`

https://github.com/Kulitorum/Decenza/blob/e33880f05d9fd1ce2dc422a59a7c528d03d3e49b/src/core/settings.cpp#L4836-L4841

2. **`clazy-range-loop-detach` on `lags` in two loops** — PR #843 was specifically landed to fix this warning class. Two new range-for loops iterate over a non-const local `QVector<double> lags` without `std::as_const()`, which clazy will flag. Fix: `for (double l : std::as_const(lags))` and `for (double v : std::as_const(lags))`.

https://github.com/Kulitorum/Decenza/blob/e33880f05d9fd1ce2dc422a59a7c528d03d3e49b/src/core/settings.cpp#L4760-L4765

https://github.com/Kulitorum/Decenza/blob/e33880f05d9fd1ce2dc422a59a7c528d03d3e49b/src/core/settings.cpp#L4859-L4863

3. **`int` instead of `qsizetype` for container-size loops in two new functions** (CLAUDE.md says "Use `qsizetype` (not `int`) for container sizes — `QVector::size()` returns `qsizetype` (64-bit on iOS/macOS). Assigning to `int` causes `-Wshorten-64-to-32` warnings.") — `sawLearnedLagFor` uses `int count = 0` and `getExpectedDripFor` uses `for (int i = 0; i < entries.size(); ++i)` plus `qMax(1, entries.size() - 1)` mixing `int` with `qsizetype`. Fix: `qsizetype count = 0` and `for (qsizetype i = 0; ...)` and `qMax(qsizetype{1}, entries.size() - 1)`.

https://github.com/Kulitorum/Decenza/blob/e33880f05d9fd1ce2dc422a59a7c528d03d3e49b/src/core/settings.cpp#L4631-L4636

https://github.com/Kulitorum/Decenza/blob/e33880f05d9fd1ce2dc422a59a7c528d03d3e49b/src/core/settings.cpp#L4669-L4676

4. **RowLayout child reads `parent._modelSource` / `parent._sourceSuffix` by `parent` reference** — if the `Text` items are ever wrapped in an intermediate container (e.g., for spacing), `parent` silently resolves to the new container instead of the `RowLayout`, breaking the binding with no compile-time or runtime error. Fix: give the `RowLayout` an explicit `id` and use that id in child references.

https://github.com/Kulitorum/Decenza/blob/e33880f05d9fd1ce2dc422a59a7c528d03d3e49b/qml/pages/settings/SettingsCalibrationTab.qml#L193-L199

5. **Auto-reset comment says it "mirrors the single-shot auto-reset" but the batched threshold is 5× longer** — the comment at line 4775 claims parity with the legacy 2-shot path. But since each batch median represents 5 shots, two consecutive bad batch medians means 10 shots must overshoot ≥ 6g before a reset fires, vs. 2 shots in the legacy path. The comment should document the actual behavior rather than claim equivalence to a different time scale.

https://github.com/Kulitorum/Decenza/blob/e33880f05d9fd1ce2dc422a59a7c528d03d3e49b/src/core/settings.cpp#L4773-L4781

---

<details>
<summary>5 below-threshold issues (50–74)</summary>

- [OK to skip] Dispersion check uses ratio-of-medians (`medianDrip/medianFlow`) as reference for median-of-ratios (`drip[i]/flow[i]`) values — mathematically distinct but conservative 1.5 s threshold and consistent espresso flow rates make divergence rare in practice (score: 55)
- [Worth a look] Entries with `flow <= 0.5` are excluded from `lags` (IQR check) but included in `drips`/`flows`/`overs` (committed median) — low-flow entry can skew the persisted median while bypassing the outlier filter; rare in practice since espresso flow > 0.5 ml/s almost always (score: 55)
- [OK to skip] `confirmed` parameter appears in schema but handler doesn't check it — false positive; `McpServer` validates and strips `confirmed` before calling the handler, matching the existing `reset_saw_learning` pattern (score: 55)
- [OK to skip] `iqrOf` returns 0 when fewer than 4 entries have `flow > 0.5`, disabling the IQR gate — per-entry deviation check still runs; requires 2+ preinfusion-only stops in one 5-shot batch, extremely unlikely in normal use (score: 50)
- [Worth a look] Comment on `sawLearnedLagFor` in settings.h says "overload" but it is a companion function with a different name, not a C++ overload — one-word fix to "variant" (score: 50)

</details>

Issue #743, opened 2026-04-14
refactor: split settings.h to reduce recompile blast radius


skialpine:
## Problem

`settings.h` is 1,104 lines and is included by 41+ `.cpp` files (123 when counting transitive includes). Any change to the header — even adding a single method — recompiles most of the project. It's the largest non-template header by 2x and has the widest include footprint.

For comparison:
| Header | Lines | Direct includers |
|--------|-------|-----------------|
| **settings.h** | **1,104** | **41** |
| shothistorystorage.h | 455 | 20 |
| de1device.h | 281 | 24 |
| profile.h | 318 | 12 |
| maincontroller.h | 282 | 10 |

## Impact

Recent example: PR #742 added 3 methods for flow calibration batch storage. This caused a near-full recompile despite the new methods only being used by `maincontroller.cpp`.

## Options

1. **Split by domain** — `CalibrationSettings`, `EspressoSettings`, `UISettings`, etc. Each consumer only includes the settings it needs. Biggest win, biggest refactor.

2. **Pimpl pattern** — Move `QSettings` member and internal caches behind a private impl pointer. The header becomes stable (just method signatures), so adding internal state doesn't trigger recompiles.

3. **Extract targeted additions** — For self-contained features (like the flow cal batch store), put them in their own class that only the consumer includes. Incremental, low-risk.

## Not urgent

This is a build-time quality-of-life issue, not a correctness issue. Worth addressing when there's a natural opportunity (e.g., a larger refactor touching settings).

skialpine:
Done. The split landed in three tiers (archived under `openspec/changes/archive/`):

- Tier 1 + Tier 2: extracted MQTT, theme, network, AI, MCP, autowake, brew, app, hardware, DYE, visualizer into `settings_<domain>.h/.cpp`
- Tier 3 (today, e927c65b): extracted SAW + flow-cal into `settings_calibration`

Result:
- `src/core/settings.h`: **1,104 → 200 lines** (82% reduction). Now a thin façade.
- 12 domain headers: `settings_ai/app/autowake/brew/calibration/dye/hardware/mcp/mqtt/network/theme/visualizer`.

[...]

PR #852, opened 2026-04-25
Split Settings into 7 domain sub-objects (PR1 + Tier 1 of PR2)

skialpine:
## Summary

Decomposes the 1,170-line `Settings` class into composition-based domain classes to reduce recompile blast radius (issue #743). Also extracts shot-history struct types into their own header for the same reason.

Each sub-object is a standalone `QObject` with its own `QSettings` instance and `Q_PROPERTY` declarations. `Settings` owns the sub-objects and exposes them via `Q_PROPERTY` for QML access:

```qml
// Before
checked: Settings.mqttEnabled
// After
checked: Settings.mqtt.mqttEnabled
```

`Q_DECLARE_OPAQUE_POINTER` lets `settings.h` keep forward declarations of the sub-object types — so domain headers aren't transitively pulled in via `settings.h`.

## What's in this PR

**Domain classes added (7 of 11):**

| Domain | Properties | Key consumer dropped settings.h? |
|---|---|---|
| `SettingsMqtt` | 10 MQTT broker/publish | – (still needs steam state) |
| `SettingsAutoWake` | 4 wake schedule | ✅ AutoWakeManager |
| `SettingsHardware` | 6 heater calibration | ✅ DE1Device::setSettings |
| `SettingsAI` | 8 provider/keys + configurationChanged aggregate signal | – (uses settings.h for other things via raw value()) |
| `SettingsTheme` | ~25 theme/shader/font/brightness + invokables | – |
| `SettingsVisualizer` | 8 upload + defaultShotRating | – |
| `SettingsMcp` | 4 MCP server | – |

**Shot history types extracted:**
- `src/history/shothistory_types.h` containing `ShotRecord`, `HistoryShotSummary`, `ShotFilter`, `ShotSaveData`, `GrinderContext`, `HistoryPhaseMarker`
- `shothistorystorage.h` re-exports it
- `shotfileparser.h` switched to the lite types header

**OpenSpec proposal** committed to `openspec/changes/split-headers-by-domain/`.

## Metrics

- `settings.h`: **1,170 → 887 lines** (24% reduction)
- ~35 QML files migrated to `Settings.{domain}.{prop}` access
- ~15 C++ consumer files updated
- `tst_Settings`: **14/14 pass**
- macOS test build: **clean**

## What's deferred

4 remaining domains — `SettingsBrew`, `SettingsDye`, `SettingsNetwork`, `SettingsApp` — are documented as deferred in `tasks.md`. They're larger and have more cross-domain coupling (e.g., `setDefaultShotRating` ↔ `setDyeEspressoEnjoyment`); doing them well needs a dedicated session. The composition pattern is now established, so the follow-up should mostly be mechanical.

One behavior change worth flagging: `setDefaultShotRating()` no longer auto-syncs `dyeEspressoEnjoyment`. `MainController` already does this sync at the relevant call sites (lines 1834, 2101); QML/MCP callers that previously relied on the side effect now need to update both explicitly. Documented inline in `settings_visualizer.cpp`.

## Test plan
- [x] macOS Debug build (Decenza app + all test executables)
- [x] `tst_Settings` 14/14 pass
- [ ] Smoke-test the QML migrations (Mqtt tab, AutoWake schedule, Theme color editor, AI keys, Visualizer upload, MCP toggle, Steam/HotWater pages for hardware properties) on a real machine
- [ ] Confirm settings persistence — toggle a value in each migrated domain, restart app, verify it survived


skialpine:
### Code review

Found 3 issues:

1. **Broken `Connections { target: Settings }` block — theme editor swatches won't refresh**

`SettingsThemesTab.qml` listens for `onEditingPaletteChanged`, `onCustomThemeColorsChanged`, and `onIsDarkModeChanged` on `Settings`, but those signals moved to `SettingsTheme`. QML silently ignores handlers when the target doesn't emit the named signal, so the theme editor's swatch refresh + dark/light auto-switch are broken. Should target `Settings.theme`.

https://github.com/Kulitorum/Decenza/blob/1d98ca01637912a9f156ce6e2ea03d67f32f0ed3/qml/pages/settings/SettingsThemesTab.qml#L92-L108

2. **`setDefaultShotRating` side-effect dropped on QML/MCP paths**

The comment in `setDefaultShotRating()` claims callers handle the cross-domain sync to `dyeEspressoEnjoyment`, but no caller actually wires it: `SettingsVisualizerTab.qml` writes the value with no follow-up, `mcptools_write.cpp` does the same, and `MainController` only syncs in shot-save callbacks (after a shot finishes), not on the live setting change. Pre-PR, changing the default rating in the Visualizer tab pre-filled the current shot's enjoyment immediately. Suggest wiring `connect(m_visualizer, &SettingsVisualizer::defaultShotRatingChanged, this, [this]{ setDyeEspressoEnjoyment(m_visualizer->defaultShotRating()); });` in `Settings::Settings()`.

https://github.com/Kulitorum/Decenza/blob/1d98ca01637912a9f156ce6e2ea03d67f32f0ed3/src/core/settings_visualizer.cpp#L88-L98

3. **`m_settings ?` guards a `m_settingsMqtt->X()` dereference in mqttclient.cpp**

After the split, several null guards check the wrong pointer. E.g., `QString clientId = m_settings ? m_settingsMqtt->mqttClientId() : "";` — the guard on `m_settings` doesn't actually protect the `m_settingsMqtt` dereference. Doesn't crash today (both are always constructed together) but the invariant the guard suggests is wrong. Replace `m_settings ?` with `m_settingsMqtt ?` for the MQTT-only reads.

https://github.com/Kulitorum/Decenza/blob/1d98ca01637912a9f156ce6e2ea03d67f32f0ed3/src/network/mqttclient.cpp#L100-L120

skialpine:
### Code review

Found 1 issue in the 4 commits since the prior review:

1. **Two more `m_settings ?` guards still around `m_settingsMqtt->X()` derefs in mqttclient.cpp**

The earlier sed-style fix caught the `m_settings ? m_settingsMqtt->X() : default` ternary form (9 sites) but missed the negated-`if-return` form. Same bug pattern, same risk profile (latent — both pointers always non-null in practice from the single MainController call site, but the guard checks the wrong invariant). Quick fix: change both to `m_settingsMqtt`.

`connectToBroker` — guard is on `m_settings`, deref is on `m_settingsMqtt`:
https://github.com/Kulitorum/Decenza/blob/1224b355c1c425f561103afed9a9e14c81068fde/src/network/mqttclient.cpp#L283-L295

`publishHomeAssistantDiscovery` — same pattern:
https://github.com/Kulitorum/Decenza/blob/1224b355c1c425f561103afed9a9e14c81068fde/src/network/mqttclient.cpp#L866-L872

The 3 issues from the prior review are otherwise correctly fixed. The QObject* sub-object accessor approach (`1224b355`) successfully restores the build win — touching `settings_mqtt.h` now recompiles 9 .cpp files in the main Decenza target instead of 41.

<details>

- [Worth a look] Stale `Q_DECLARE_OPAQUE_POINTER in settings.h` reference in [src/main.cpp comment](https://github.com/Kulitorum/Decenza/blob/1224b355c1c425f561103afed9a9e14c81068fde/src/main.cpp#L1614-L1620) — the macro was removed in commit `1224b355`, comment wasn't updated (score: 72)
- [Worth a look] [settings.h Q_PROPERTY comment](https://github.com/Kulitorum/Decenza/blob/1224b355c1c425f561103afed9a9e14c81068fde/src/core/settings.h#L26-L34) explains the runtime metaObject mechanism but omits that `qmlRegisterUncreatableType<SettingsMqtt>(...)` in main.cpp is the prerequisite — without it the same bug recurs (score: 50)
- [OK to skip] [settings.cpp connect comment](https://github.com/Kulitorum/Decenza/blob/1224b355c1c425f561103afed9a9e14c81068fde/src/core/settings.cpp#L354-L362) says "applies immediately to the next shot" — slightly imprecise (also affects in-progress shot's enjoyment field), but intent is clear (score: 50)

</details>

PR #859, opened 2026-04-26
feat(saw): drop per-profile graduation from 3 medians to 2

skialpine:
## Summary

Follow-up to #848. A (profile, scale) pair previously needed 3 committed batch-medians (15 SAW shots) before its own data took over from the global fallbacks. For users pulling 1–2 SAW shots/day on a given combo, that's 1–3 weeks per pair — exactly the slow-ramp pain that #847 was trying to address.

This drops the read-path graduation threshold to 2 batch-medians (10 SAW shots).

## Why 2 is enough

Each committed median already represents:
- 5 SAW shots that **survived the IQR dispersion gate** (IQR ≤ 1.0 s on lags, no single lag deviating > 1.5 s from the batch median).
- A median of 5, which is robust against single-shot outliers by construction.

Two such medians = 10 shots through that filter. That is a substantially stronger signal than the single-shot legacy \`globalPool\` entries we'd otherwise fall back to during the gap.

The auto-reset path (2 consecutive medians with overshoot < -6 g, \`settings.cpp:1326\`) is unchanged and still catches regime changes after graduation.

## Risk

Worst case is a "honeymoon" where the user's first 10 shots on a new bean land at similar-but-biased lag, the pair graduates, then real-world shots drift. In practice the recency-weighted average in \`getExpectedDripFor\` (line 1207) means each new median pushes the model — post-graduation it self-corrects within a few committed batches.

## Changes

| File | Change |
|------|--------|
| \`src/core/settings.cpp\` | Lift the magic \`3\` into \`kSawMinMediansForGraduation = 2\` with a doc comment; update the 4 read-path sites and the \`recomputeGlobalSawBootstrap\` comment |
| \`src/core/settings.h\` | Update the \`sawLearnedLagFor\` comment to reference the named constant |
| \`tests/tst_saw_settings.cpp\` | Existing tests drop from 3 batches to 2; add \`singleMedianStillFallsBackToBootstrap\` to guard the lower-bound boundary |
| \`docs/CLAUDE_MD/SAW_LEARNING.md\` | Reword graduation references to use the named constant |

## Test plan

- [x] \`ctest -R tst_saw_settings\` — all 10 cases pass (9 existing + 1 new)
- [x] \`ctest -R tst_saw\` — legacy SAW tests still pass
- [ ] Pull 10 SAW shots on a new (profile, scale) combo → log shows \`[SAW] model: source="perProfile"\` after the 2nd batch commits
- [ ] Existing pairs that had 1 or 2 committed medians under the old code immediately become \`perProfile\` after upgrade (data shape unchanged — same per-pair history is just read past a lower gate)

### DIFF
--- a/docs/CLAUDE_MD/SAW_LEARNING.md
+++ b/docs/CLAUDE_MD/SAW_LEARNING.md
@@ -37,7 +37,7 @@ The same architecture that fixed the analogous problem in flow calibration ([AUT
 2. **Batched median commits** — accumulate 5 shots' worth of pending entries before changing the model. The median is robust to single-shot outliers (channeling, scale glitches, cup interaction). 5 was chosen to match flow cal.
 3. **Batch-level dispersion check** — if the 5 lags within a batch are too spread out (IQR > 1.0 s, or any single lag > 1.5 s from the batch median), the whole batch is dropped. Dispersion that high indicates the user changed conditions mid-batch (different beans, different grinder, manual stop).
 4. **Global bootstrap** — the median lag across all graduated `(profile, scale)` pairs on the same scale is published as `saw/globalBootstrapLag/<scaleType>`. New pairs use this as their first-shot default instead of the scale's hardware-only `sensorLag()`.
-5. **Read-path fallback chain** — `perProfile` (≥ 3 committed batches) → `globalBootstrap` → `globalPool` (legacy entries) → `scaleDefault`. New users / new pairs degrade gracefully.
+5. **Read-path fallback chain** — `perProfile` (≥ `kSawMinMediansForGraduation` committed batches, currently 2 = 10 SAW shots) → `globalBootstrap` → `globalPool` (legacy entries) → `scaleDefault`. New users / new pairs degrade gracefully.
 
 ### Why this is the right shape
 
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Found 1 issue:

1. New constant `kSawMinMediansForGraduation` is declared as `int` but is compared against `QJsonArray::size()` (returns `qsizetype`) at four sites in the same file. CLAUDE.md says: "Use `qsizetype` (not `int`) for container sizes — `QVector::size()`, `QList::size()`, `QString::size()` etc. return `qsizetype` (64-bit on iOS/macOS). Assigning to `int` causes `-Wshorten-64-to-32` warnings." PR #848's review flagged the same recurring pattern in adjacent SAW code; this introduces a new instance.

https://github.com/Kulitorum/Decenza/blob/6e1a400b4ee3d617a9f3517c5fa767f101a81181/src/core/settings.cpp#L685-L687

Fix: `static constexpr qsizetype kSawMinMediansForGraduation = 2;`

---

<details>

- [Worth a look] Comment at `src/core/settings.cpp` L1319-L1325 says "the threshold is effectively 10 consecutive bad shots" — with the new graduation threshold defined just above, "the threshold" becomes ambiguous; "the auto-reset trigger is effectively 10 consecutive bad shots" reads cleaner (score: 50).
- [OK to skip] Pre-existing `int count` / `for (int i = 0; i < entries.size(); ++i)` patterns in `Settings::sawLearnedLag` (L699) and `Settings::getExpectedDrip` (L849-L853) are loop counters/comparisons, not narrowing assignments — outside the literal CLAUDE.md `-Wshorten-64-to-32` rule even though same-spirited (score: 50).

</details>

PR #866, opened 2026-04-26
fix(analysis): repair corpus regressions from #864

skialpine:
## Summary
PR #864 was over-aggressive — it killed two real-positive cases in the regression corpus that I missed because the Qt Creator MCP `run_tests` only runs QTests, not the `shot_corpus_regression` ctest target. Linux CI caught it on the v1.7.2 retag. This PR repairs both.

## What broke

**londinium_gusher** — channeling went **Sustained → None**. The "signed positive only" rule from #864 silenced the post-channel collapse signature (flow drops 7.5 → 0.5 ml/s while pressure stays near 3 bar, conductance plummets, dC/dt strongly negative). The textbook channel is positive, but the corpus relies on the detector also catching the failure tail.

**puck_failure_short_gusher** — grindIssue went **true → false**. The new pump-ramp trim (0.5 s off the first flow-mode phase) consumed the entire ~0 s flow-mode range that this 2.2 s extreme puck-failure shot exposes after phase reconstruction. Zero samples → silent `no-data`.

## Three fixes

1. **Restore `std::abs()` in `detectChannelingFromDerivative`** so both signs of dC/dt count. Discriminating puck-failure (negative dC/dt under *stable* pressure) from lever-rise (negative dC/dt under *fast-rising* pressure) is now the window builder's job.

2. **`buildChannelingWindows`: fence flow-mode samples where pressure is rising fast** (> `WINDOW_STATIONARY_REL` across the ±0.75 s lookup). Direction matters — only **rises** are masked. Bloom transitions and pressure-mode → flow-mode handoffs (pressure falling) stay in-window because those are legitimate channeling-style transients.

3. **`analyzeFlowVsGoal`: both pump-ramp and limiter-tail trims now require ≥ 1 s of range left**. Trimming a fractional-second flow-mode phase down to nothing is strictly worse than leaving the raw samples — extreme puck-failure shots have just enough data in that short window to register a clear gusher delta.

## Tests
- Drop the obsolete `channelingFromDerivative_negativeSpikeNotFlagged` test (invalidated by the `abs` restore).
- Add `channelingWindows_flowModeRisingPressure_isExcluded` — rising pressure during flow-mode phase masks the samples.
- Add `channelingWindows_flowModeFallingPressure_isIncluded` — falling pressure (bloom/handoff) stays in-window.
- Add `flowVsGoal_shortShot_skipsTrimToPreserveSignal` — pump-ramp trim bypassed when range too short.

## Validation
- `shot_corpus_regression`: 10/12 → **12/12**
- `tst_ShotAnalysis`: 1701 → **1704** (3 new tests, no regressions)

## Process gap
Qt Creator MCP `run_tests` doesn't include `shot_corpus_regression` (separate ctest target driven by `shot_eval`). For analysis changes I should have run `ctest -R shot_corpus_regression` against the macOS-Debug build before pushing — added to my mental checklist for next time.

## Test plan
- [x] `tst_ShotAnalysis` full suite — 1704/1704
- [x] `shot_corpus_regression` — 12/12 in local Debug build
- [ ] Pull a fresh lever-style shot post-merge, confirm channeling badge silent on clean pours
- [ ] Pull a known-channeled flat-pressure shot, confirm Sustained still fires
- [ ] Pull a known fine-grind D-Flow shot, confirm grind issue still detected
- [ ] CI Linux build passes (currently failing on `dbe3ec49`)

### DIFF
--- a/src/ai/shotanalysis.cpp
+++ b/src/ai/shotanalysis.cpp
@@ -87,14 +87,11 @@ ShotAnalysis::ChannelingSeverity ShotAnalysis::detectChannelingFromDerivative(
     // channel signature is positive (flow surges, pressure dips,
     // conductance jumps up), but the post-channel collapse — flow drops
     // toward zero while pressure stays roughly stable — produces sustained
-    // negative dC/dt of equal diagnostic value. Examples in the regression
-    // corpus: londinium_gusher (puck gave up at 3.4 bar, flow collapsed
-    // 7.5 → 0.5 ml/s, dC/dt strongly negative throughout the pour).
-    // Distinguishing this real signal from the lever pressure-rise
-    // pattern (also negative dC/dt, but caused by intentional pressure
-    // ramping rather than puck failure) is the window builder's job —
-    // see buildChannelingWindows, which masks out samples where pressure
-    // is rapidly climbing.
+    // negative dC/dt of equal diagnostic value. Distinguishing that real
+    // signal from the lever pressure-rise pattern (also negative dC/dt, but
+    // caused by intentional pressure ramping rather than puck failure) is
+    // the window builder's job: see buildChannelingWindows, which masks out
+    // samples where pressure is rapidly climbing.
     for (const auto& pt : conductanceDerivative) {
         if (pt.x() < analysisStart) continue;
         if (pt.x() > analysisEnd) break;
@@ -219,11 +216,11 @@ QVector<ShotAnalysis::DetectionWindow> ShotAnalysis::buildChannelingWindows(
         // channeling.
         //
         // Direction matters: real puck failures collapse flow under
-        // approximately stable pressure (londinium_gusher) so pressure
-        // stays in-window. Bloom transitions and pressure-mode → flow-mode
-        // handoffs see pressure FALL rapidly — those are legitimate
-        // channeling signals (or expected transients) and must not be
-        // masked, so we only fence on rises here.
+        // approximately stable pressure, so pressure stays in-window.
+        // Bloom transitions and pressure-mode → flow-mode handoffs see
+        // pressure FALL rapidly — those are legitimate channeling signals
+        // (or expected transients) and must not be masked, so we only
+        // fence on rises here.
         if (isFlowMode) {
             const double pressureNow = lookupOrNaN(pressure, t);
             const double pressureFut = lookupOrNaN(pressure, t + WINDOW_HALF_SEC);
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 3 issues:

1. New test `flowVsGoal_shortShot_skipsTrimToPreserveSignal` does not exercise the new `kMinPostTrimRangeSec` bypass it claims to validate. With `phases = [(0.0, Fill), (2.0, End)]`, the guard math is `(2.0 − 0.0) − GRIND_PUMP_RAMP_SKIP_SEC (0.5) = 1.5 ≥ kMinPostTrimRangeSec (1.0)` → guard passes → trim IS applied. The bypass branch is never taken. The test passes for the wrong reason and the new gate has no unit-test coverage (only the corpus harness covers it). To exercise the bypass, the flow-mode phase needs to be < 1.5 s long.

https://github.com/Kulitorum/Decenza/blob/71560a09dd5ec91d614a9c85d2e316c40a8f68e6/tests/tst_shotanalysis.cpp#L464-L488

2. Production-code comments hardcode corpus-fixture filenames (`londinium_gusher`, `puck_failure_short_gusher`). Root [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/71560a09dd5ec91d614a9c85d2e316c40a8f68e6/CLAUDE.md) says: "Don't reference the current task, fix, or callers ('used by X', 'added for the Y flow', 'handles the case from issue #123'), since those belong in the PR description and rot as the codebase evolves." The fixture names will go stale if anything in `tests/data/shots/` is ever renamed; replace with behavioral descriptions (e.g. "puck collapse under stable pressure"). Three locations in `src/ai/shotanalysis.cpp`:

https://github.com/Kulitorum/Decenza/blob/71560a09dd5ec91d614a9c85d2e316c40a8f68e6/src/ai/shotanalysis.cpp#L88-L97

https://github.com/Kulitorum/Decenza/blob/71560a09dd5ec91d614a9c85d2e316c40a8f68e6/src/ai/shotanalysis.cpp#L218-L225

https://github.com/Kulitorum/Decenza/blob/71560a09dd5ec91d614a9c85d2e316c40a8f68e6/src/ai/shotanalysis.cpp#L382-L390

3. Header doc-comment for `buildChannelingWindows` enumerates two inclusion criteria (goal stationary, actual converged) but the PR adds a third — the rising-pressure exclusion gate for flow-mode phases. The header now describes a contract that doesn't match the implementation, and a maintainer reading only the header won't know flow-mode samples can be excluded for pressure rises.

https://github.com/Kulitorum/Decenza/blob/71560a09dd5ec91d614a9c85d2e316c40a8f68e6/src/ai/shotanalysis.h#L69-L89

<details>

- [Worth a look] Rising-pressure test assertion `windows.first().start >= 3.0` is looser than its comment claims (gate stops firing at t≈3.4 because `pressureFut` plateaus at 10 bar, ratio drops below 1.15 once pNow ≥ ~8.7); a regression weakening the gate could slip through. Tightening to `>= 4.0` would catch it. (score: 50)
- [Worth a look] The `pressureNow > 0.5` floor in the new gate is an undocumented magic number; bloom phases at 0.5–0.8 bar that rise sharply could be inconsistently masked. Not exercised by current corpus. (score: 50)

</details>

PR #868, opened 2026-04-26
fix(timing): use wall clock for shot/steam time, not 16-bit BLE timer

skialpine:
## The bug

DE1 BLE protocol encodes `ShotSample.timer` as a 16-bit value in 1/100 s units. It wraps every **65536 / 100 = 655.36 s** of DE1 uptime.

[`MainController::onShotSample`](https://github.com/Kulitorum/Decenza/blob/main/src/controllers/maincontroller.cpp#L2247) has been computing shot-elapsed time as raw subtraction since the initial commit (`a591aa1c`):

```cpp
double time = sample.timer - m_shotStartTime;
```

If the timer wraps mid-cycle, every sample after the wrap produces a negative `time`. That feeds `addPhaseMarker` and gets persisted to the shot record.

## The downstream damage (real shot, captured today)

Shot 887 on Mike's DE1 — 80's Espresso, 53 s, clean — shipped with these stored phases (note positions and times):

| pos | label | frame | time | reason |
|---|---|---|---|---|
| 0 | rise and hold | 2 | **−647.61** | pressure |
| 1 | decline | 3 | **−642.86** | time |
| 2 | Start | 0 | 0 | |
| 3 | preinfusion start | 0 | 0 | |
| 4 | preinfusion | 1 | 2.5 | time |
| 5 | End | −1 | 53.27 | |

The deltas (−647.61 vs +6, −642.86 vs +10) are ≈ 655.36, which is the 16-bit max. Single-wrap signature.

Cascading badge effects on this clean shot:
- `analyzeFlowVsGoal` walked the markers; with frame 2/3 sitting at negative times at the head of the array, "preinfusion"'s next phase resolved to "End" (53.27 s), stretching the flow-mode analysis range across the entire pour and firing **grindIssue=true** on flow vs the now-stale 7.5 ml/s preinfusion goal.
- `detectSkipFirstFrame` reads `phases.first()`, saw frame 2, and reported **skipFirstFrameDetected=true**.
- Phase summaries computed durations against the broken stamps and rendered "decline" duration as 696.1 s on a 53 s shot.

`channelingDetected` happened to stay false because mode-aware windows can't form on the corrupted phase data — but that's coincidence, not protection.

## What de1app does

[`espresso_millitimer`](https://github.com/decentespresso/de1app/blob/main/de1plus/vars.tcl) at `de1plus/vars.tcl:393` uses Tcl's `clock milliseconds` — wall clock, not the BLE field — for shot-elapsed timing. They also explicitly handle the 16-bit wrap on the *inter-sample* delta path at `de1plus/de1_de1.tcl:554`:

```tcl
# Unwrap 16-bit unsigned int SampleTime
set dhc [expr { $ShotSample(SampleTime) - $::de1::_previous_shotsample_update_time }]
if { $dhc < 0 } { set dhc [expr { $dhc + 65536 }] }
```

de1app's simulator at `de1plus/gui.tcl:1238` even comments the field's nature directly: `# SampleTime is a 16-bit counter of zero crossings`. They've known this for a long time. Decenza missed it.

## The fix

Two state variables converted from `sample.timer` units to wall-clock ms:
- `m_shotStartTime` → `m_shotStartTimeMs` (espresso cycle base)
- `m_steamStartTime` → `m_steamStartTimeMs` (steam session base)

Both stored as `qint64` ms-since-epoch. The rename makes the units self-documenting at every callsite. The first-sample anchor swaps `sample.timer` for `QDateTime::currentMSecsSinceEpoch()`; subsequent reads compute `(now - base) / 1000.0`.

`m_lastSampleTime` stays in `sample.timer` units because its only consumer is a self-correcting bounded check at [`maincontroller.cpp:2196`](https://github.com/Kulitorum/Decenza/blob/main/src/controllers/maincontroller.cpp#L2196):

```cpp
if (deltaTime > 0 && deltaTime < 1.0) {
```

A wrap there gives a huge negative delta, the check fails, the affected frame is silently skipped — no persistent corruption.

## Validation

- Full ctest suite (33 targets, 1700+ QTests) passes locally on macOS Debug
- 12/12 corpus regression passes
- Manual trace through the new wall-clock path: anchor at first espresso-phase sample, monotonic forever

## Test plan
- [x] ctest (1700+ unit tests + corpus regression)
- [ ] Pull a real shot post-merge, confirm phase markers all positive and ordered correctly
- [ ] Reproduce the wrap in simulator (advance DE1 timer near 655 s before starting a shot) and confirm phases stay clean

## Followups (not in this PR)
- `DirectController::onShotSample` at `src/controllers/directcontroller.cpp:186` has the same `sample.timer - m_shotStartTime` pattern; same wrap risk for direct-control mode.
- Shot 887 on Mike's DE1 is a real-world capture of this bug. Worth saving as a corpus regression fixture so a future regression can't re-introduce wrap-poisoned phase data without the corpus catching it.

### DIFF
--- a/src/controllers/maincontroller.cpp
+++ b/src/controllers/maincontroller.cpp
@@ -146,18 +146,18 @@ MainController::MainController(QNetworkAccessManager* networkManager,
                 m_settings->brew()->setSteamDisabled(false);
             }
 
-            // Steam session ended — run post-session analysis. m_steamStartTime
+            // Steam session ended — run post-session analysis. m_steamStartTimeMs
             // is only set when isFlowing() was true (Steaming/Pouring substates),
             // so this fires after all flowing samples have been collected even
             // though the Steaming phase persists through Puffing/Ending substates.
-            if (phase != MachineState::Phase::Steaming && m_steamStartTime > 0) {
+            if (phase != MachineState::Phase::Steaming && m_steamStartTimeMs > 0) {
                 if (m_steamHealthTracker && m_steamDataModel) {
                     m_steamHealthTracker->onSessionComplete(
                         m_steamDataModel,
                         m_settings->brew()->steamFlow(),
                         static_cast<int>(m_settings->brew()->steamTemperature()));
                 }
-                m_steamStartTime = 0;
+                m_steamStartTimeMs = 0;
                 if (m_steamHealthTracker)
                     m_steamHealthTracker->resetSession();
             }
@@ -1625,7 +1625,7 @@ void MainController::onEspressoCycleStarted() {
     }
 
     // Clear the graph for the new espresso cycle (previous shot is now saved)
-    m_shotStartTime = 0;
+    m_shotStartTimeMs = 0;
     m_lastShotTime = 0;
     m_extractionStarted = false;
     m_lastFrameNumber = -1;
@@ -2211,8 +2211,8 @@ void MainController::onShotSampleReceived(const ShotSample& sample) {
     bool steamFlowing = (phase == MachineState::Phase::Steaming
                          && m_machineState->isFlowing());
     if (steamFlowing && m_steamDataModel) {
-        if (m_steamStartTime == 0) {
-            m_steamStartTime = sample.timer;
+        if (m_steamStartTimeMs == 0) {
+            m_steamStartTimeMs = QDateTime::currentMSecsSinceEpoch();
             m_steamDataModel->clear();
             // Add flow goal line from current settings
             double flowGoal = m_settings->brew()->steamFlow() / 100.0;
@@ -2221,7 +2221,7 @@ void MainController::onShotSampleReceived(const ShotSample& sample) {
             if (m_steamHealthTracker)
                 m_steamHealthTracker->resetSession();
         }
-        double t = sample.timer - m_steamStartTime;
+        double t = (QDateTime::currentMSecsSinceEpoch() - m_steamStartTimeMs) / 1000.0;
         m_steamDataModel->addSample(t, sample.groupPressure, sample.groupFlow, sample.steamTemp);
 
         // Live threshold warnings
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

Found 2 issues:

1. Garbled comment in header — the new block-doc ends with "but no other persistent state should differ samples." Two reviewers independently flagged it; the phrase is grammatically broken and doesn't communicate the intended constraint (likely "no other persistent state should use sample.timer as its base"). Root [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/5df8ada128f0f22328a96ff346866346dbd6f0ec/CLAUDE.md) says comments should explain "the WHY (a hidden constraint, a subtle invariant)" — an incoherent comment fails that bar.

https://github.com/Kulitorum/Decenza/blob/5df8ada128f0f22328a96ff346866346dbd6f0ec/src/controllers/maincontroller.h#L251-L259

2. `MainController` and `ShotTimingController` anchor their t=0 origins at different moments. `MainController` sets `m_shotStartTimeMs` at the **first espresso-phase sample** (`maincontroller.cpp:2246-2247`); `ShotTimingController::onShotSample` resets `m_displayTimeBase` at the **first frame transition** (`shottimingcontroller.cpp:172`). In practice these are within tens of ms — the first Preinfusion-phase sample typically lands the same tick as the frame-0 transition — but they're not guaranteed equal. Pre-existing structurally; the PR doesn't introduce it but inherits it. Phase markers (`MainController`'s time) and graph data (`ShotTimingController`'s time) will diverge by the gap. A short follow-up that publishes one anchor as a signal the other consumes would close this.

https://github.com/Kulitorum/Decenza/blob/5df8ada128f0f22328a96ff346866346dbd6f0ec/src/controllers/maincontroller.cpp#L2244-L2252

<details>

- [Worth a look] Redundant `m_lastSampleTime = sample.timer` inside `if (m_shotStartTimeMs == 0)` — same line is also written unconditionally a few lines up (line 2204). Harmless but dead code in code being modified. (score: 35)
- [OK to skip] `m_lastSampleTime` carve-out drops one `onFlowSample` call per BLE timer wrap (`maincontroller.cpp:2194-2197`). Pre-existing, the PR's header comment accurately documents it, and it's ~40 ms of flow data lost per ~655 s wrap window — not a regression. (score: 30)

</details>

skialpine:
### Code review

Found 1 issue:

1. `m_lastSampleTime` is never reset between shots. `onEspressoCycleStarted` clears `m_lastShotTime`, `m_extractionStarted`, `m_lastFrameNumber`, etc., but leaves `m_lastSampleTime` carrying the prior shot's last `sample.timer` value. On the first dispensing sample of every subsequent shot, `deltaTime = sample.timer - stale` produces a value the bounded check `> 0 && < 1.0` silently drops. The new in-shot unwrap doesn't cover this case — unwrapping a 300+ s gap leaves a 350+ s delta, still outside the gate. Net effect: one `onFlowSample` call (≈one BLE frame) of FlowScale integration lost at the start of every shot after app launch. Pre-existing, but the PR is touching the surrounding code and the fix is a one-liner: add `m_lastSampleTime = 0;` to the reset block.

https://github.com/Kulitorum/Decenza/blob/e994ba27597a20da65d75d5c6a30d2061d817b7b/src/controllers/maincontroller.cpp#L1627-L1646

Manifests at the unwrap path here:

https://github.com/Kulitorum/Decenza/blob/e994ba27597a20da65d75d5c6a30d2061d817b7b/src/controllers/maincontroller.cpp#L2196-L2213

<details>
<summary>1 below-threshold issue (50–74)</summary>

- [Worth a look] At `maincontroller.cpp:2273-2278`, the new comment refers to "its `m_extractionStarted` flag" without qualifying which controller's flag — both `MainController` and `ShotTimingController` have one, and the latter is checked a few lines down (line 2292). Stylistic only; CLAUDE.md doesn't require it. (score: 50)

</details>

PR #870, opened 2026-04-26
saw: tune Gaussian flow-similarity sigma from 1.5 to 0.25

skialpine:
## Summary

Phase 1 of [openspec/tune-saw-old-prediction](openspec/changes/tune-saw-old-prediction/proposal.md). Drops the SAW Gaussian flow-similarity σ from 1.5 → 0.25 ml/s at three call sites (`Settings::getExpectedDrip`, `Settings::getExpectedDripFor`, `WeightProcessor::getExpectedDrip`) — the only numeric change is `flowDiff² / 4.5` → `flowDiff² / 0.125`.

Phase 0 (recorded in `analysis.md`) measured this on a 63-shot real-shot corpus through real production code via `tools/saw_parity/`:
- Overall MAE: 0.370 → 0.348 g (−6%)
- High-flow MAE: 0.686 → 0.592 g (−14%)
- Shot 887 over-prediction: +0.90 → +0.57 g (−37%)

Pool data, clamps (`drip ∈ [0.5, 20]`), and fallback paths (lag-fallback when pool is degenerate) are unchanged. Migration: zero.

## What's included

- σ change at the 3 sites (single numeric constant + comment update each)
- 3 new σ-exercising tests in `tst_saw_settings.cpp` — the existing 52 SAW tests train and query at the same flow, leaving σ entirely uncovered. The new tests defend against future regressions in either direction (σ widened or narrowed).
- Phase 2 (shadow logging) explicitly dropped; Phase 3 reframed as a qualitative decision. Rationale documented in `tasks.md`.

## Test plan

- [x] Full ctest suite passes (1707 / 1707, no warnings)
- [x] `tools/saw_parity/` confirms overall MAE = 0.34811 g and shot 887 signed error = +0.573 g (matching Phase 0 baseline through real production code)
- [x] `openspec validate tune-saw-old-prediction --strict --no-interactive` — clean
- [ ] Daily-driver soak on Jeff's machine for 2 weeks per Phase 3 decision criteria

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #874, opened 2026-04-26
saw: extract shared expected-drip prediction helper (#873)

### LINKED ISSUE #873: saw: WeightProcessor::getExpectedDrip carries an untested copy of the σ math
## Background

`WeightProcessor::getExpectedDrip` ([`src/machine/weightprocessor.cpp:414-454`](https://github.com/Kulitorum/Decenza/blob/main/src/machine/weightprocessor.cpp#L414-L454)) is a hand-copied snapshot of the Gaussian-weighted-average algorithm in `Settings::getExpectedDrip` / `Settings::getExpectedDripFor`. It runs on the dedicated weight worker thread for the live SAW threshold during a shot, so it is on the actual production path (called from [`weightprocessor.cpp:239`](https://github.com/Kulitorum/Decenza/blob/main/src/machine/weightprocessor.cpp#L239)).

PR #870 tuned σ from 1.5 → 0.25 at all three sites. The new `tst_saw_settings.cpp` tests cover the σ change in `Settings::getExpectedDripFor`, but `WeightProcessor::getExpectedDrip` carries an independent copy of the constant and has no σ-specific test. A future merge or copy-paste mistake that resets just this site to the old constant would not be caught by the existing suite.

## Risk

Diverging σ values between the live SAW threshold (in `WeightProcessor`) and the post-shot prediction used for learning-pool feedback (in `Settings`) would produce inconsistent behaviour without any test signal. The two sites must stay in lockstep for the SAW system to work coherently.

## Options

Listed in rough order of preference:

1. **Drop the duplicate.** Refactor `WeightProcessor` to delegate to `Settings::getExpectedDrip` / `getExpectedDripFor` rather than carry its own copy. Eliminates the lockstep problem at the source. Surface area is small but non-trivial because `WeightProcessor` runs on a worker thread and `Settings` is main-thread-owned — needs threading review.
2. **Make `WeightProcessor`'s σ math testable in isolation.** Extract the inner loop to a free function or static helper that takes `(entries, flow, sigmaSquaredTimes2)` and returns the predicted drip. Test it directly, leaving the live state pipeline alone.
3. **Add an integration test** that drives `WeightProcessor` through a contrived shot sequence and asserts the σ-gated behaviour from outside. Highest fidelity, highest cost.

Either #1 or #2 is enough to close this issue.

## Out of scope for #870

This is intentionally separated from PR #870 (the σ tuning itself) because the fix is a refactor or new test infrastructure, not a one-constant tweak. Filing here so it doesn't get lost.

skialpine:
## Summary
- Pulls the Gaussian flow-similarity + recency-weighted-average inner loop out of `WeightProcessor::getExpectedDrip` and both `Settings::getExpectedDrip[For]` sites into a shared header `src/machine/sawprediction.h`. All three call sites now route through `SawPrediction::weightedDripPrediction`, so the σ constant lives in exactly one place and the live SAW threshold can no longer drift away from the post-shot learning feedback.
- Adds `tests/tst_sawprediction.cpp` with the σ-specific direct coverage that was missing for the WeightProcessor copy (#873): σ value lock-in, zero-`flowDiff` collapse, far-flow attenuation to the NaN fallback, two-sided flow separation, recency weighting, single-entry edge, empty input, the `[0.5 g, 20 g]` output clamp.

## Why
Closes #873. PR #870 narrowed σ from 1.5 → 0.25 at all three sites, but only the `Settings` versions had σ-specific tests in `tst_saw_settings`. `WeightProcessor::getExpectedDrip` carried a hand-copied snapshot of the same math with no direct test, so a future merge or copy-paste regression at that one site would have silently desynced the live SAW threshold from the learning pool feedback. Centralising the kernel addresses both the lockstep risk (option #1 in the issue) and the testability gap (option #2) in one move.

## Test plan
- [x] `tst_sawprediction` — 13 new tests, all pass
- [x] `tst_weightprocessor` — 24 pass (no regressions)
- [x] `tst_saw` — 14 pass (no regressions)
- [x] `tst_saw_settings` — 17 pass (no regressions)
- [x] `tst_settings`, `tst_settling` — pass
- [x] Full Qt Creator test sweep: 1720 passed, 0 failed, 0 warnings

### DIFF
--- a/src/core/settings.cpp
+++ b/src/core/settings.cpp
@@ -11,6 +11,7 @@
 #include "settings_network.h"
 #include "settings_app.h"
 #include "grinderaliases.h"
+#include "../machine/sawprediction.h"
 #include <algorithm>
 #include <QStandardPaths>
 #include <QDir>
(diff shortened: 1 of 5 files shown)


skialpine:
### Code review

Found 2 issues:

1. `int maxEntries` should be `qsizetype` (CLAUDE.md says "Use `qsizetype` (not `int`) for container sizes — `QVector::size()`, `QList::size()`, `QString::size()` etc. return `qsizetype` (64-bit on iOS/macOS). Assigning to `int` causes `-Wshorten-64-to-32` warnings.")

The PR rewrites this block and adds `const`, but leaves the `int` type — the `static_cast<qsizetype>(maxEntries)` two lines below confirms the type mismatch this rule exists to catch. The same `int maxEntries` line in `settings.cpp:817` is pre-existing in the file this PR touches and would benefit from the same fix.

https://github.com/Kulitorum/Decenza/blob/cd563a773a0ed6ce4a765b98b794e575631179b9/src/machine/weightprocessor.cpp#L425-L432

2. `getExpectedDripFor` hardcodes `recencyMin=3.0`, while the other two callers of the new helper use `converged ? 3.0 : 1.0` — the steeper gradient is meant to speed up adaptation when the model has not yet converged, and the per-pair path silently opts out. Pre-existing in this file, but the surrounding comment ("the same recency + flow-similarity scheme as the global getExpectedDrip()", `settings.cpp:1190-1192`) was true when both functions had identical bodies and is now misleading. A comment refresh — or a convergence-aware `recencyMin` to match — would close the loop.

https://github.com/Kulitorum/Decenza/blob/cd563a773a0ed6ce4a765b98b794e575631179b9/src/core/settings.cpp#L1188-L1212

<details>

- [Worth a look] `Q_ASSERT(drips.size() == flows.size())` precondition in `sawprediction.h` has no direct test — would only fire in Debug builds; a guard returning NaN (matching the empty-input branch) plus a test would close the gap (score: 50)

</details>

PR #878, opened 2026-04-27
fix(android): suppress crash log during APK install dispatcher race

skialpine:
## Summary

- Adds an opt-in `CrashHandler::setSuppressCrashLog(bool)` flag (signal-safe `volatile sig_atomic_t`) that skips writing `crash.log` + `debug.log` append while still re-raising the signal so the OS terminates normally.
- `UpdateChecker::installApk()` sets the flag the moment dispatch succeeds; `onInstallStatus()` clears it on user-cancel and on every failure status; `dismissUpdate()` clears it for the OEM-ROM path that skips `STATUS_FAILURE_ABORTED`.

## Context

Closes #865 (identical signature: #844, auto-closed without fix). Both crash logs end with `UpdateChecker: APK install intent launched` followed by `QSocketNotifier: Invalid socket <N> with type Read, disabling...` and SIGSEGV in `QSocketNotifier::setEnabled` ← `QEventDispatcherUNIXPrivate::markPendingSocketNotifiers` ← `activateSocketNotifiers`.

Root cause: `installApk()` does **not** quit the app — Qt keeps spinning the event loop while Android prepares to replace the process. The OS reaps fds backing `ShotServer` (the only `QSocketNotifier` user) or `QNetworkAccessManager`, then Qt processes a pending event for one and dereferences freed state. The backtrace is entirely Qt-internal and the process is about to be replaced anyway, so this is teardown noise we can't fix in our code.

The flag scope is narrow: only set once Android install dispatch succeeds, only cleared on paths where the app keeps running. On `STATUS_SUCCESS` it stays set until the OS swaps the process — by design, the race window is exactly that interval.

iOS / macOS / Windows / Linux behavior unchanged.

## Test plan

- [ ] Build clean (Qt Creator) — confirm no compile/warn regressions
- [ ] Android: trigger an update install on a test device → confirm install flow is unchanged (system dialog, success path replaces app, new app starts cleanly)
- [ ] Android: trigger an install, cancel the system dialog → confirm `onInstallStatus` fires `STATUS_FAILURE_ABORTED`, app keeps running, deliberately-induced SIGSEGV afterward still produces a `crash.log`
- [ ] Watch crash dashboard — new auto-filed reports with the `APK install intent launched` + `QSocketNotifier::setEnabled` signature should disappear

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #880, opened 2026-04-27
fix(android): close sockets before APK install to dodge dispatcher race

skialpine:
## Summary

Real fix for the recurring `QSocketNotifier::setEnabled` SIGSEGV that follows every Android self-update dispatch (#865, #844). Replaces the previous noise-filter approach in #878 (closed) which only suppressed the crash report.

The diagnosis: Android reaps fds out from under us during the PackageInstaller handover; Qt's UNIX event dispatcher then services a pending notifier for the now-invalid fd and crashes. Eliminate the race by closing everything we own that holds a long-lived `QSocketNotifier` *before* the JNI dispatch.

## Changes

- New signal `aboutToDispatchInstall` on `UpdateChecker` and `ShotServer`, emitted synchronously on the main thread immediately before `ApkInstaller.install` JNI is called
- `main.cpp` wires both to a single slot that:
  - `ShotServer::stop()` — closes the TCP listener + all client sockets
  - `QNetworkAccessManager::clearConnectionCache()` — closes HTTP keepalive sockets
  - `RelayClient::shutdown()` — closes the WebSocket
- No restore on cancel. Install succeeds → process replaced. Install fails → rare; user relaunches.

## Why these specifically

`grep -rn "QSocketNotifier" src/` finds only one direct user (the SSL-peek mechanism in `HttpRedirectSslServer`, owned by ShotServer), but every `QTcpSocket` / `QWebSocket` / `QNetworkReply` carries Qt-internal notifiers. The three things above are the only owners with persistent sockets in our process. `MqttClient` wraps Paho's C client and manages its own non-Qt socket — not affected.

## Test plan

- [ ] Build clean (Qt Creator on macOS / Windows / Android)
- [ ] Android self-update from UI: confirm install dialog still appears, install completes, new app launches
- [ ] Android self-update via web upload (ShotServer): confirm install dialog appears on device. Browser will show connection-closed (expected — we tore down the server)
- [ ] Cancel the system install dialog: confirm app keeps running but ShotServer / Relay are down (require app restart to recover, documented behavior)
- [ ] Watch for new auto-filed crashes with the `APK install intent launched` + `QSocketNotifier::setEnabled` signature — these should disappear

### DIFF
--- a/src/core/crashhandler.cpp
+++ b/src/core/crashhandler.cpp
@@ -295,6 +295,28 @@ void CrashHandler::uninstall()
     }
 }
 
+void CrashHandler::logOpenFileDescriptors(const QString& tag)
+{
+#ifdef Q_OS_ANDROID
+    QDir fdDir("/proc/self/fd");
+    if (!fdDir.exists()) {
+        qDebug() << "[fd dump:" << tag << "] /proc/self/fd not accessible";
+        return;
+    }
+    // /proc/self/fd entries are symlinks; the default QDir filter excludes
+    // symlinks-to-non-existent. Pass an explicit filter that keeps everything
+    // except `.` / `..`.
+    const auto entries = fdDir.entryList(QDir::AllEntries | QDir::Hidden | QDir::System | QDir::NoDotAndDotDot);
+    qDebug().noquote() << "[fd dump:" << tag << "]" << entries.size() << "open fds:";
+    for (const QString& entry : entries) {
+        const QFileInfo fi("/proc/self/fd/" + entry);
+        qDebug().noquote().nospace() << "  fd=" << entry << " -> " << fi.symLinkTarget();
+    }
+#else
+    Q_UNUSED(tag);
+#endif
+}
+
 QString CrashHandler::crashLogPath()
 {
     return QString::fromUtf8(s_crashLogPath);
--- a/src/core/crashhandler.h
+++ b/src/core/crashhandler.h
@@ -35,6 +35,14 @@ public:
     /// Get the last N lines of debug.log for context
     static QString getDebugLogTail(int lines = 50);
 
+    /// Walk /proc/self/fd and qDebug each fd's symlink target. Diagnostic
+    /// only — used to attribute the fd that gets reaped during the Android
+    /// APK install handover (#865) so the next crash log tells us which
+    /// service or socket family was responsible. No-op on non-Android.
+    /// `tag` is logged with each line so multiple calls in the same run
+    /// (e.g. before vs. after teardown) can be distinguished.
+    static void logOpenFileDescriptors(const QString& tag);
+
 private:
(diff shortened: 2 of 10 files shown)


skialpine:
### Code review

Found 1 issue:

1. `ShotServer::stop()` doesn't actually close the upload socket — the PR comment is wrong, and the QSocketNotifier the fix is meant to silence stays alive. `stop()` iterates `m_sseLayoutClients`, `m_sseThemeClients`, the discovery socket, and the listener, but it does not iterate `m_pendingRequests`, which is where the in-flight upload connection lives. So the comment "this closes the very TCP connection the upload arrived on" is false (the 200 response below it will reach the client just fine), and more importantly the socket whose notifier we're trying to drop is still around when Android reaps fds. Either teach `stop()` to also close `m_pendingRequests`, or in `ShotServer::installApk` explicitly close the upload socket before emitting the signal.

https://github.com/Kulitorum/Decenza/blob/eeafa5e342f865a56d795bd8e5d2158d3afe7bcc/src/network/shotserver_upload.cpp#L415-L428

For reference, the `stop()` body that doesn't touch `m_pendingRequests`:

https://github.com/Kulitorum/Decenza/blob/eeafa5e342f865a56d795bd8e5d2158d3afe7bcc/src/network/shotserver.cpp#L492-L529

<details>

- [Worth a look] `CrashReporter` ([crashreporter.h:47](https://github.com/Kulitorum/Decenza/blob/eeafa5e342f865a56d795bd8e5d2158d3afe7bcc/src/network/crashreporter.h#L46-L48)) and `LibrarySharing` ([librarysharing.h:138](https://github.com/Kulitorum/Decenza/blob/eeafa5e342f865a56d795bd8e5d2158d3afe7bcc/src/network/librarysharing.h#L137-L139)) each own a private `QNetworkAccessManager` that the slot doesn't touch — `clearConnectionCache()` only hits the shared one. Practical risk is low (those flows are short-lived), but it's a real gap given the PR's "every fd" framing (score: 67)
- [Worth a look] The synchronous-emit invariant is load-bearing but undocumented at the connection site in `main.cpp` — if either emitter ever moves to a worker thread, `Qt::AutoConnection` flips to queued and the fix silently breaks. One-line comment fix (score: 50)
- [OK to skip] If the JNI `ApkInstaller.install()` returns false after the teardown emit, the app is left without ShotServer / Relay until restart. PR description acknowledges this as a deliberate trade-off and the failure window is narrow (score: 50)
- [OK to skip] `QDir::entryList` on `/proc/self/fd` itself opens an `opendir` fd, so each snapshot will include a transient `fd=N -> /proc/self/fd` entry. Cosmetic noise in the diagnostic only (score: 50)

</details>

<sub>If this code review was useful, please react with a thumbs up. Otherwise, react with a thumbs down.</sub>

Issue #887, opened 2026-04-27
Adopt Qt 6.10 high-contrast accessibility hint in Theme.qml (deferred)


skialpine:
## Status: Deferred / Tracking

Split out from PR #886 (platform screen reader announcements). The technical design is solid but the work is too speculative for current user demand. Pick this up if (a) a user reports a contrast/legibility problem, (b) we're already touching `Theme.qml` for an unrelated rework, or (c) we want to ship a measurable a11y improvement and have design bandwidth.

## Why deferred

- No current evidence of demand. Decent customers skew enthusiast/luxury; the high-contrast-at-OS-level cohort is small and we have no reports.
- Touching every getter in `Theme.qml` is high blast radius for visual regressions — we have no automated visual testing.
- "What does Decenza look like in high contrast" is a designer question (palette choices, brand consistency at high contrast), not just engineering.
- The `readonly property color foo: ...` pattern in `Theme.qml` doesn't re-evaluate on NOTIFY signals, so adopting this requires a sweep + visual verification.

### Hint source

Qt 6.10 exposes `QStyleHints::accessibility()` → `QAccessibilityHints*`, which carries:
- `Qt::ContrastPreference contrastPreference` — tri-state (`NoPreference`, `HighContrast`, future-reserved).
- `contrastPreferenceChanged(Qt::ContrastPreference)` signal — no polling needed.

Treat `NoPreference` as **\"normal\"**; it's a real value meaning \"user has no preference\", not a missing one. This matters because most users will fall here even on platforms that expose the API.

### C++ shim

In \`AccessibilityManager\`:

\`\`\`cpp
Q_PROPERTY(Qt::ContrastPreference platformContrastPreference READ ... NOTIFY ...)
Q_PROPERTY(bool effectiveHighContrast READ ... NOTIFY ...)  // combines platform + user override
\`\`\`

Wire \`platformContrastPreference\` to \`QStyleHints::accessibility()->contrastPreferenceChanged\`. Wrap (don't expose \`QStyleHints\` directly to QML) so a future Qt rename is a one-line fix and we can override during testing.

### User override

\`accessibility/contrastMode\` setting with three values:
- \`\"system\"\` (default) — \`HighContrast\` → true, \`NoPreference\` → false.
- \`\"normal\"\` — false regardless of OS hint.
- \`\"high\"\` — true regardless of OS hint.

Settings UI: a mode picker next to the announcement mode picker on \`SettingsAccessibilityTab.qml\`.

[...]

PR #889, opened 2026-04-27
a11y: route announcements through screen reader when one is active

skialpine:
## Summary

- `AccessibilityManager::announce()` now auto-routes: when `QAccessible::isActive()` is true (TalkBack/VoiceOver/Narrator running) we dispatch a `QAccessibleAnnouncementEvent` and **suppress** our `QTextToSpeech` voice. When no screen reader is detected, we fall back to the existing TTS path gated by `ttsEnabled`.
- Fixes the long-standing bug where Decenza's TTS overlapped TalkBack/VoiceOver on every announcement, making the app effectively unusable with a screen reader on.
- No new setting, no migration, no Settings UI changes. The existing "Voice Announcements" toggle keeps its label and binding; its meaning narrows to "speak when no screen reader is detected." None of the ~25 existing `AccessibilityManager.announce(...)` call sites needed to change.
- Implements `openspec/changes/add-platform-screen-reader-announcements`. Note: the original proposal also bundled high-contrast theme adaptation (the actual Qt 6.10 `QStyleHints::accessibility()->contrastPreference` work) — that piece was previously split out into tracking issue #887 and remains deferred. This PR uses the Qt 6.8 `QAccessibleAnnouncementEvent` API.

## Behavior matrix

| Screen reader | `ttsEnabled` | Before | After |
|---------------|--------------|--------|-------|
| On            | On           | Both speak (overlap bug) | TalkBack only |
| On            | Off          | TalkBack only | TalkBack only |
| Off           | On           | TTS speaks | TTS speaks |
| Off           | Off          | Silent | Silent |

## Implementation notes

- Routing rule lives in `announce()`. Two protected virtuals (`dispatchPlatformAnnouncement`, `dispatchTtsAnnouncement`) plus a virtual `isScreenReaderActive()` provide a unit-test seam — tests subclass and record calls without touching real Qt accessibility / TTS state.
- Platform dispatcher null-guards an empty `topLevelWindows()` list (very early startup / shutdown) and logs the dropped announcement at debug level.
- Every announcement logs path + `isActive()` reading at qInfo level so user reports of "missed announcements" are debuggable from transcripts.
- `QApplication` was an unused include in `accessibilitymanager.cpp` and is removed (it was preventing the test target from building without Qt::Widgets).

## Test plan

- [x] `tst_accessibility_announcements` (new): 10 passed, 0 failed, 0 warnings — covers all matrix combinations, interrupt → assertive mapping, the polite/assertive helpers, and the disabled-manager case.
- [ ] Manual TalkBack on Android: with TalkBack ON, no double-speak. With TalkBack OFF and `ttsEnabled` ON, Decenza's TTS speaks as before.
- [ ] Manual VoiceOver on iPad: same matrix.
- [ ] Existing test suite continues to pass (no production code paths changed beyond `announce()`).

### DIFF
--- a/docs/CLAUDE_MD/ACCESSIBILITY.md
+++ b/docs/CLAUDE_MD/ACCESSIBILITY.md
@@ -3,11 +3,22 @@
 ## Current Implementation
 
 The app has accessibility support via `AccessibilityManager` (C++) with:
-- Text-to-speech announcements via `AccessibilityManager.announce()`
+- Announcements via `AccessibilityManager.announce()` — auto-routed (see "Announcement routing" below)
 - Tick sounds for frame changes
 - `AccessibleTapHandler` and `AccessibleMouseArea` for touch handling
 - Extraction announcements (phase changes, weight milestones, periodic updates)
 - User-configurable settings in Settings → Accessibility
+
+### Announcement routing (`AccessibilityManager.announce()`)
+
+`AccessibilityManager.announce(text, interrupt=false)` is the only announcement entry point used from QML (~25 call sites). At dispatch time it picks one of two paths automatically:
+
+- **Platform path** — when `QAccessible::isActive()` returns true (TalkBack on Android, VoiceOver on iOS/macOS, Narrator on Windows). It builds a `QAccessibleAnnouncementEvent` against the root `QQuickWindow` and posts it via `QAccessible::updateAccessibility()`. `interrupt=true` maps to `Assertive` politeness; default maps to `Polite`. **`QTextToSpeech` does not also speak in this path** — that's the bug fix that this routing replaced (we used to overlap TalkBack).
+- **TTS fallback path** — when no screen reader is detected. Speaks via `QTextToSpeech` if the user has the existing `ttsEnabled` toggle on; stays silent otherwise. This preserves the "spoken extraction progress without TalkBack" use case for sighted users who want audio feedback during a shot.
+
+There is **no user-visible delivery-mode setting**. Routing is automatic; the existing `ttsEnabled` toggle keeps its place in Settings → Language → Accessibility but only matters on the no-screen-reader fallback path.
+
+`announcePolite(text)` / `announceAssertive(text)` are convenience wrappers that map to `announce(text, false)` / `announce(text, true)`. New call sites can use them when the politeness intent is worth being explicit; existing call sites don't need to change.
 
 ### Key QML Components
 
@@ -20,7 +31,7 @@ The app has accessibility support via `AccessibilityManager` (C++) with:
 
 ### Key C++ Component
 
-`src/core/accessibilitymanager.h/cpp` — TTS engine, tick sounds, settings persistence.
+`src/core/accessibilitymanager.h/cpp` — announcement routing (platform vs TTS), tick sounds, settings persistence.
 
 ---
 
--- a/openspec/changes/add-platform-screen-reader-announcements/design.md
+++ b/openspec/changes/add-platform-screen-reader-announcements/design.md
@@ -77,7 +77,7 @@ Notes:
 
 - `ttsEnabled` keeps its existing meaning ("speak via Decenza TTS"), but now only matters when no screen reader is detected. Sighted users who enable it for spoken extraction progress (no TalkBack) keep that capability unchanged.
 - When a screen reader IS detected, we suppress TTS unconditionally — that's the bug fix. There is no "both" mode in the new design.
-- `lastAnnouncedItem` de-duplication runs on every path (platform and TTS) — its semantics don't change.
+- `lastAnnouncedItem` is a QML-side de-duplication mechanism (set/checked from `AccessibleTapHandler`, `AccessibleButton`, etc.); `announce()` does not read or write it, so the new routing has no interaction with it.
 
 ### 6. No migration
 
--- a/openspec/changes/add-platform-screen-reader-announcements/design.md
+++ b/openspec/changes/add-platform-screen-reader-announcements/design.md
@@ -7,7 +7,7 @@
 2. Speaks **in addition** to the OS screen reader when one is active — both voices overlap.
 3. Doesn't participate in the screen reader's queue, so a Polite announcement can't wait for the user's swipe-read to finish.
 
-This change addresses (1)–(3) by routing announcements through `QAccessibleAnnouncementEvent`.
+This change addresses (1)–(3) by routing announcements through `QAccessibleAnnouncementEvent` whenever a platform screen reader is detected, and falling back to the existing `QTextToSpeech` path otherwise.
 
 The originally-proposed high-contrast theme adaptation has been split out — see "Deferred work" at the end.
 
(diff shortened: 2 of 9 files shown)


skialpine:
### Code review

Found 6 issues:

1. `setEnabled()` and `toggleEnabled()` still call `m_tts->say()` directly, bypassing the new `isScreenReaderActive()` routing — the "Accessibility enabled/disabled" confirmation will double-speak when TalkBack/VoiceOver is on, which is exactly the bug this PR claims to fix (`docs/CLAUDE_MD/ACCESSIBILITY.md` says "fix pre-existing violations in any file you touch").

https://github.com/Kulitorum/Decenza/blob/3f5fb47d07c1fec614c51998e268e8e2fda7f63b/src/core/accessibilitymanager.cpp#L139-L151

https://github.com/Kulitorum/Decenza/blob/3f5fb47d07c1fec614c51998e268e8e2fda7f63b/src/core/accessibilitymanager.cpp#L353-L364

2. `announceLabel()` is `Q_INVOKABLE` and called from 12 QML places (TemperatureItem, BatteryLevelItem, ScaleBatteryItem, etc.) — also speaks via `m_tts->say()` directly, no `isScreenReaderActive()` check. Same overlap bug as #1, much higher frequency.

https://github.com/Kulitorum/Decenza/blob/3f5fb47d07c1fec614c51998e268e8e2fda7f63b/src/core/accessibilitymanager.cpp#L319-L338

3. The dropped-window log uses `qDebug()` while the other three paths (`platform`, `tts`, `silent`) use `qInfo()`. `tasks.md` 1.6 explicitly says paths should be at qInfo level "so transcripts capture user-reportable 'missed announcement' cases" — production builds typically filter qDebug.

https://github.com/Kulitorum/Decenza/blob/3f5fb47d07c1fec614c51998e268e8e2fda7f63b/src/core/accessibilitymanager.cpp#L283-L298

4. `tst_accessibility_announcements` constructs `FakeAccessibilityManager`, which inherits the base constructor and unconditionally runs `initTts()`. On Linux CI without a TTS engine, the connected `stateChanged(Error)` lambda fires `qWarning() << "TTS error:"`. `docs/CLAUDE_MD/TESTING.md` is explicit: "Every `qWarning()` emitted during a test run must either be fixed at the source or explicitly marked as expected." No `QTest::ignoreMessage` / `ScopedWarningFilter` in the test.

https://github.com/Kulitorum/Decenza/blob/3f5fb47d07c1fec614c51998e268e8e2fda7f63b/tests/tst_accessibility_announcements.cpp#L52-L56

5. `dispatchTtsAnnouncement()` only checks `if (!m_tts) return;` — no `m_shuttingDown` guard, while every peer method (`announce`, `playTick`, `toggleEnabled`, `announceLabel`, `setEnabled`) starts with one. Pattern from commit `926f704`. The narrow current call path is safe, but a subclass / future caller could TTS during teardown.

https://github.com/Kulitorum/Decenza/blob/3f5fb47d07c1fec614c51998e268e8e2fda7f63b/src/core/accessibilitymanager.cpp#L310-L317

6. Every announcement logs the full text at qInfo to logcat / stderr / web debug log. Announcement text often includes user-entered content (bean brand, grinder, profile names). No CLAUDE.md rule covers this, but it's worth at least truncating since the PR introduced the log lines.

https://github.com/Kulitorum/Decenza/blob/3f5fb47d07c1fec614c51998e268e8e2fda7f63b/src/core/accessibilitymanager.cpp#L255-L269

<details>

- [Worth a look] Window lookup picks first \`QQuickWindow\` instead of \`focusWindow()\` first — only impacts debug builds where the GHC simulator window is open (score: 65)
- [Worth a look] design.md §5 claims \`lastAnnouncedItem\` de-dup runs in \`announce()\`, but it's a pure QML-side mechanism — doc-only inconsistency (score: 65)
- [OK to skip] \`initTts()\` is unconditional in the ctor while \`initTickSound()\` is gated — pre-existing, separate from this change (score: 50)

</details>

Generated with /review.

skialpine:
### Code review (follow-up on commit 06fac548)

Found 3 issues:

1. `toggleEnabled()` now double-speaks on the screen-reader path. `setEnabled(!m_enabled)` fires `routeAnnouncement(..., interrupt=false)` (Polite), then `toggleEnabled` immediately fires `routeAnnouncement(..., interrupt=true)` (Assertive) with the same text. The TTS path is fine because `stop()+say()` cancels the first utterance. The platform path has no equivalent cancellation — TalkBack/VoiceOver dispatches both `QAccessibleAnnouncementEvent`s and speaks the message twice. The test acknowledges this with `QVERIFY(platformCalls.size() >= 1)` instead of `== 1`. Fix: either suppress the announcement inside `setEnabled` when called from `toggleEnabled` (private overload with `announce=false`), or drop `toggleEnabled`'s second route call.

https://github.com/Kulitorum/Decenza/blob/06fac5484dd9015090be0deedd59c3f8cd0a147a/src/core/accessibilitymanager.cpp#L418-L435

2. The new `TestSkipAudioInit` ctor opens the real system QSettings store via `m_settings("Decenza", "DE1")` even though `loadSettings()` is skipped. Tests then call `setEnabled()` and `setTtsEnabled()` (via the new `setup()` helper), and both invoke `saveSettings()` — every test run writes to `~/Library/Preferences/Decenza.DE1.plist` / `~/.config/Decenza/DE1.conf`. `tst_settings.cpp` already has the established pattern (save originals in `init()`, restore in `cleanup()`); this test should follow it.

https://github.com/Kulitorum/Decenza/blob/06fac5484dd9015090be0deedd59c3f8cd0a147a/src/core/accessibilitymanager.cpp#L24-L35

3. Dead `wasEnabled` + `Q_UNUSED(wasEnabled)` in `toggleEnabled()` — leftover from the old logic that compared pre/post state. The new code doesn't read it. Root CLAUDE.md says \"Don't add backwards-compatibility hacks like renaming unused _vars\" — applies here. Just delete the two lines.

https://github.com/Kulitorum/Decenza/blob/06fac5484dd9015090be0deedd59c3f8cd0a147a/src/core/accessibilitymanager.cpp#L421-L425

<details>

- [Worth a look] Several inline comments are inaccurate (\`routeAnnouncement\` header comment omits the m_shuttingDown guard it does check; toggleEnabled comment claims \"no overlap\" but issue #1 above shows it does overlap on the platform path; \`disabledManagerDispatchesNothing\` test comment says \"don't bother calling resetCalls()\" but the test calls it anyway). Doc-only, easy to fix in the same touch (score: 82, reduced because doc-only — no functional impact).

</details>

Generated with /review.

skialpine:
### Code review (follow-up on commit 6a0c0444)

Found 1 issue:

1. The new `initTestCase()` redirection of QSettings is a no-op. `AccessibilityManager`'s `m_settings("Decenza", "DE1")` uses the two-string `QSettings` constructor, which is hardcoded to `NativeFormat` per [Qt docs](https://doc.qt.io/qt-6/qsettings.html#QSettings-2): _\"the format is set to QSettings::NativeFormat (i.e. calling setDefaultFormat() before calling this constructor has no effect)\"_. `setPath(IniFormat, ...)` only affects QSettings instances explicitly constructed with IniFormat, so test writes still hit `NSUserDefaults` / Windows registry / Linux native conf — i.e. the developer's real preferences. Established cross-platform pattern in this repo is `tst_settings.cpp`'s `init()`/`cleanup()` snapshot-and-restore.

https://github.com/Kulitorum/Decenza/blob/6a0c044459f946ff68dd16a73d12377247de408f/src/core/accessibilitymanager.cpp#L14-L22

https://github.com/Kulitorum/Decenza/blob/6a0c044459f946ff68dd16a73d12377247de408f/tests/tst_accessibility_announcements.cpp#L75-L86

<details>

- [Worth a look] `setEnabledRoutesThroughPlatformWhenScreenReaderActive` relies on the `TestSkipAudioInit` ctor leaving `m_enabled=false` rather than calling `mgr.setEnabled(false); mgr.resetCalls();` defensively. Sibling `toggleEnabled` test in same commit already does this — copying the pattern is one line. (score: 55)
- [OK to skip] Stale comment in `routeAnnouncement` (`.h`) names `setEnabled()` and `toggleEnabled()` as the bypass callers, but after the round-3 split the actual bypass is in `setEnabledImpl()`. Doc-only. (score: 60)
- [OK to skip] Stale "Don't bother calling resetCalls()" comment in `disabledManagerDispatchesNothing` test contradicted the line below it that calls resetCalls(). (score: 55)

</details>

Generated with /review.

PR #891, opened 2026-04-27
fix(analysis): detect choked-puck on pressure-mode pours via grind detector

skialpine:
## Summary
- Grind detector previously only ran flow-vs-goal averaging across flow-mode pour phases. Pressure-mode profiles (80's Espresso, Cremina, Londinium) have no flow-mode pour, so a catastrophic choke went unflagged.
- Concrete miss: shot 890 (80's Espresso, 20 g dose, 36 g target) held 6.6 bar for 50 s, mean pour flow 0.3 ml/s, **yield 1.1 g** — analyzer returned a "clean shot" verdict.
- Add a fallback inside `analyzeFlowVsGoal`: when no flow-mode window qualifies and pressure is available, flag if pour spent ≥15 s at ≥4 bar with mean flow <0.5 ml/s. Sets `chokedPuck=true` on the `GrindCheck`, synthesizes a large-negative delta so the existing `grindIssueDetected` badge fires through this path. `generateSummary` swaps in tailored verdict copy ("Puck choked — grind way too fine") that pre-empts the generic puck-integrity verdict.
- One detector, one badge, two paths to it (per the design discussion).

## Test plan
- [x] 3 new unit tests: choked-puck signature fires, normal lever pour does NOT fire, flow-mode pour takes precedence over the fallback
- [x] Full unit suite: 1738 passed, 0 failed, 0 warnings
- [x] `shot_eval --validate` against the regression corpus: 12/12 shots still pass — no regressions on `cremina_clean`, `damian_lrv3_clean`, `e61_long_clean`, `classic_italian_clean`, `gagne_adaptive_clean`, etc.
- [ ] Spot-check on Jeff's machine: shot 890 should pick up the choked-puck verdict on next analysis-dialog open

### DIFF
--- a/src/ai/shotanalysis.cpp
+++ b/src/ai/shotanalysis.cpp
@@ -410,8 +410,7 @@ ShotAnalysis::GrindCheck ShotAnalysis::analyzeFlowVsGoal(
         }
     }
     // Flow-vs-goal averaging path. Skipped when no flow-mode windows
-    // qualify or when the profile carries no flow goal — the choked-puck
-    // fallback below covers those pressure-mode pours.
+    // qualify or when the profile carries no flow goal.
     if (!flowModeRanges.isEmpty() && !flowGoal.isEmpty()) {
         auto inFlowMode = [&flowModeRanges](double t) {
             for (const auto& r : flowModeRanges) {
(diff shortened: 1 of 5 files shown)


skialpine:
### Code review

5 issues found and fixed in 574c14d20c3611c70c0bd1e3725d2deddef0b389. 1 below-threshold deferred.

**Fixed (75+):**

1. `sampleCount` not assigned on the choked-puck path — header doc promised it on both paths, but only the flow-vs-goal branch populated it. `shot_eval` was reporting 0 pressurized samples on every choked shot. ([shotanalysis.cpp#L495-L504](https://github.com/Kulitorum/Decenza/blob/535555fb7b798723444c53c370caaa039d053fa0/src/ai/shotanalysis.cpp#L495-L504))

2. `generateSummary` block comment said the choked-puck check "falls through" when flow-vs-goal had no data — that was the v1 design, not the post-review additive design. ([shotanalysis.cpp#L730-L740](https://github.com/Kulitorum/Decenza/blob/535555fb7b798723444c53c370caaa039d053fa0/src/ai/shotanalysis.cpp#L730-L740))

3. Threshold-constants comment in the header still described the path as a "fallback" gated on "no flow-mode window." ([shotanalysis.h#L153-L161](https://github.com/Kulitorum/Decenza/blob/535555fb7b798723444c53c370caaa039d053fa0/src/ai/shotanalysis.h#L153-L161))

**Fixed (50–74):**

4. Test `grindIssue_flowModePourTakesPrecedenceOverChokedFallback` doc claimed the choked-puck path "only runs when flow-vs-goal has no qualifying samples" — false post-restructure; the test passes because the per-sample `CHOKED_FLOW_MAX_MLPS` gate prevents firing. Renamed and rewrote the comment to describe the real mechanism. ([tst_shotanalysis.cpp#L697-L703](https://github.com/Kulitorum/Decenza/blob/535555fb7b798723444c53c370caaa039d053fa0/tests/tst_shotanalysis.cpp#L697-L703))

5. `skipFirstFrame > chokedPuck` verdict precedence had no documented rationale. Added a comment: a frame-skip bug can synthesise extraction dynamics that look like a choke, so fixing the machine is the prerequisite. ([shotanalysis.cpp#L818-L823](https://github.com/Kulitorum/Decenza/blob/535555fb7b798723444c53c370caaa039d053fa0/src/ai/shotanalysis.cpp#L818-L823))

<details>
<summary>3 below-threshold issues (28–72)</summary>

- [Worth a look] No corpus fixture for the motivating shot 890 — synthetic unit tests are solid, but a real-shot regression in the corpus would catch threshold drift; deferred because it requires uploading to visualizer.coffee or hand-converting from local DB. ([manifest.json](https://github.com/Kulitorum/Decenza/blob/535555fb7b798723444c53c370caaa039d053fa0/tests/data/shots/manifest.json)) (score: 72)
- [OK to skip] `pressureModeRanges` builder doesn't clamp to `pourStart` like `flowModeRanges` does — the outer flow-sample loop gates correctly so no wrong samples are counted, but ranges with no pour overlap still get appended; latent inconsistency only (score: 35)
- [OK to skip] Lever profile with all `isFlowMode=true` phases would let `inPressureMode` return true everywhere — the per-sample 4 bar gate + 0.5 ml/s threshold prevents false positives empirically; `cremina_clean` / `damian_lrv3_clean` both pass (score: 28)

</details>

Verified: 1740/1740 unit tests pass, 12/12 corpus shots pass.

PR #892, opened 2026-04-27
fix(analysis): catch moderate grind-too-fine via yield/target ratio

skialpine:
## Summary
- The choked-puck check had only the flow arm (mean pressurized flow < 0.5 ml/s). That caught the catastrophic case (shot 890, 1.1 g yield) but missed moderate chokes — same diagnosis, less severe.
- **Concrete miss:** shot 883 — 80's Espresso on grind 3.5, 25.3 g yield from a 36 g target, 60 s pressurized pour at 6.3 bar. Mean pressurized flow landed at 0.59 ml/s, narrowly above the flow threshold. Verdict fell through to "Puck integrity issue — improve distribution" via the channeling badge, pointing the user at the wrong fix when the real diagnosis was simply "grind too fine, coarsen."
- Add a **yield-ratio arm** on the same `chokedPuck` check: when `targetWeight > 0` and `finalWeight / targetWeight < 0.85` (with the existing 15 s pressurized-duration gate), set `chokedPuck=true`. Same flag, same verdict copy, two ways into it (severe-by-flow / moderate-by-yield).
- **Works headless.** `finalWeight` is populated either by a real BLE scale or by Decenza's `FlowScale` virtual scale (dose-aware flow integration), so the arm fires whether or not a Decent Scale is connected.

## Test plan
- [x] 2 new unit tests: yield-shortfall fires (shot 883 signature), clean ristretto doesn't (shot 888 signature, 99 % yield)
- [x] Full unit suite: 1742/1742 pass, 0 warnings
- [x] `shot_eval --validate` against the regression corpus: 12/12 shots still pass — corpus shots don't carry `target_weight` in visualizer.coffee format, so the new arm is a no-op there and the existing flow arm continues to drive the corpus expectations
- [ ] Spot-check on Jeff's machine: shot 883 should pick up the choked-puck verdict and badge on next analysis-dialog open


skialpine:
### Code review

5 issues addressed in 7947962e, plus 1 user-requested doc cleanup and a corpus fixture for the yield arm.

**Fixed (75+):**

1. `GrindCheck::chokedPuck` field doc said "puck held pressure but no flow" — true for the v1 flow arm, false for the new yield arm where flow can be moderate. Rewrote to describe both arms. ([shotanalysis.h#L188-L196](https://github.com/Kulitorum/Decenza/blob/9307e59bb5fa6e0b6e3c9e74e2e8d76eee5ba9bd/src/ai/shotanalysis.h#L188-L196))

**Fixed (50–74):**

2. Threshold-block comment said "both arms share the pressurizedDuration ≥ 15s gate" — actual gate also requires `flowSamples ≥ 5`. ([shotanalysis.h#L162-L168](https://github.com/Kulitorum/Decenza/blob/9307e59bb5fa6e0b6e3c9e74e2e8d76eee5ba9bd/src/ai/shotanalysis.h#L162-L168))

3. `analyzeFlowVsGoal` header doc "both default to 0 (arm disabled)" could be misread as AND. The arm disables when EITHER is 0. ([shotanalysis.h#L218-L222](https://github.com/Kulitorum/Decenza/blob/9307e59bb5fa6e0b6e3c9e74e2e8d76eee5ba9bd/src/ai/shotanalysis.h#L218-L222))

**User-requested cleanup:**

4. Documented `yieldOverride` semantics in `ShotRecord` — three reviewers misread it as brew-ratio-only. Updated `mcptools_dialing.cpp` comment to clarify the profile-JSON fallback there is defensive for imported shots (which `shotimporter.cpp` doesn't populate `yieldOverride` for). ([shothistory_types.h#L77-L86](https://github.com/Kulitorum/Decenza/blob/9307e59bb5fa6e0b6e3c9e74e2e8d76eee5ba9bd/src/history/shothistory_types.h#L77-L86))

5. **Added corpus fixture** `80s_choked_moderate.json` — shot 883 downloaded from visualizer.coffee with `target_weight: 36` hand-added (visualizer format doesn't carry it). `shot_eval` now reads `target_weight` and threads it through. Manifest entry expects `grindIssue: true, channeling: None, pourTruncated: false`. **13/13 corpus shots pass** including the new fixture. Real-shot regression coverage now exists for the yield arm. ([manifest.json](https://github.com/Kulitorum/Decenza/blob/7947962e/tests/data/shots/manifest.json))

<details>

- [Worth a look] 0.85 yield-ratio threshold may false-positive on user-stopped shots at 80–85 % — empirical data is two points (70 % choke, 99 % clean), no obvious better value without more shot data (score: 60)
- [OK to skip] Yield arm pressure floor — agent's own justification confirmed the per-sample 4 bar gate is stricter than mean-pressure averaging, design is sound (score: 80 nominal, false alarm in justification)
- [OK to skip] `yieldOverride` semantic mismatch — verified at `maincontroller.cpp:1702-1707` that the field is always populated correctly for shots saved by Decenza; the MCP fallback at `mcptools_dialing.cpp:122` is defensive for imported shots (now documented) (score: 85 nominal, false alarm in justification)

</details>

Verified: 1742/1742 unit tests pass, 13/13 corpus shots pass.

PR #893, opened 2026-04-27
fix(analysis): always recompute every quality badge on shot load

skialpine:
## Summary
- The load path was already re-deriving `grindIssue` and `skipFirstFrame` on every shot load — but `channeling` and `temperatureUnstable` were re-derived ONLY for pre-migration-10 shots (those missing stored conductance). Any modern shot kept its save-time badge values forever, even when the detector has been improved since.
- Concrete miss in the audit: shots 884 and 889 (both clean 35–36 g extractions on 80's Espresso) carry stored `channelingDetected=true`. Running the current mode-aware detector against shot 889's curves via `shot_eval` reports `channeling=None`. The badge is stale from a pre-PR-#864 save-time detector.
- Drop the `needsDerivedCurves` gate so all four badges re-evaluate on every load. `conductanceDerivative` is either loaded from the DB or filled by `computeDerivedCurves()` above, so it's always available. Fold the now-symmetric grind block into the same scope while we're here.
- Net result: every detector improvement we land flows automatically to existing shots on next view — no one-shot re-analyze needed.

## Test plan
- [x] 1742/1742 unit tests pass
- [x] 13/13 regression corpus shots pass
- [x] `shot_eval` against shot 889's visualizer JSON confirms current detector reports `channeling=None`
- [ ] Spot-check on Jeff's machine: shots 884 and 889 should drop the `channelingDetected` badge on next analysis-dialog open

### DIFF
--- a/src/history/shothistorystorage.cpp
+++ b/src/history/shothistorystorage.cpp
@@ -2266,7 +2266,11 @@ ShotRecord ShotHistoryStorage::loadShotRecordStatic(QSqlDatabase& db, qint64 sho
             }
         }
 
-        // Grind direction (flow-vs-goal + choked-puck arms)
+        // Grind direction (flow-vs-goal + choked-puck arms). Reset before the
+        // skip-check so filter/pourover/tea/steam/cleaning shots clear any
+        // stale stored value (the channeling/temp resets above are the same
+        // pattern).
+        record.grindIssueDetected = false;
         if (!ShotAnalysis::shouldSkipChannelingCheck(record.summary.beverageType, record.flow, pourStart, pourEnd)) {
             record.grindIssueDetected = ShotAnalysis::detectGrindIssue(
                 record.flow, record.flowGoal, record.phases,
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

1 issue found and fixed in 0d335b2f. 1 below-threshold worth a follow-up.

**Fixed (75+):**

1. `grindIssueDetected` wasn't reset before the `shouldSkipChannelingCheck` guard, so filter / pourover / tea / steam / cleaning shots kept their stale DB value while the channeling and temp badges in the same block correctly cleared. Asymmetric and contradicted the PR's "always recompute" intent. Same fix applied to `skipFirstFrameDetected` for consistency. ([shothistorystorage.cpp#L2268-L2284](https://github.com/Kulitorum/Decenza/blob/9668e38b657f0c0c4cf85f3e5fb3a40ec5beda99/src/history/shothistorystorage.cpp#L2268-L2284))

<details>

- [Worth a look] DB columns now permanently divergent from in-memory recomputed badges — filter queries (e.g. "show only channeling shots") still match stored values, so the history list can disagree with what the shot detail view shows. Pre-existing for legacy shots; PR widens the gap. Follow-up: either also write back from the load path, or run a one-shot `requestReanalyzeBadges` sweep. (score: 60)
- [OK to skip] `flowGoal.isEmpty()` guard removed — flagged as a perf concern but verified intentional, paralleling PR #891's same change in `requestReanalyzeBadges` so the choked-puck arm can fire on shots without flowGoal (score: ~30 after verification)
- [OK to skip] Comment-clarity nits — `needsDerivedCurves` lost its inline rationale; new block comment slightly overstates conductanceDerivative as the key invariant when it only matters for the channeling sub-block (score: ~50)

</details>

Verified: 1742/1742 unit tests pass, 13/13 corpus shots pass.

PR #896, opened 2026-04-27
fix(analysis): persist drifted quality badges on every shot load

### LINKED ISSUE #894: Stored badge columns drift from in-memory recomputed badges (history-list filters disagree with shot-detail badges)
## Background

PR #893 made `loadShotRecordStatic` always recompute the four quality badges (`channelingDetected`, `temperatureUnstable`, `grindIssueDetected`, `skipFirstFrameDetected`) from the loaded curve data, so detector improvements take effect on existing shots without requiring a one-shot re-analyze sweep.

That fix is correct for the **shot detail view** and any consumer that goes through `loadShotRecordStatic` — including the MCP `shots_get_detail` tool. But it only updates the in-memory `ShotRecord`. The DB columns (`channeling_detected`, `temperature_unstable`, `grind_issue_detected`, `skip_first_frame_detected`) keep their save-time values until something writes them back.

## Status: lazy-persist-on-view is already shipped

Option 2 below is already wired in:

- `ShotDetailPage.qml:71` calls `requestReanalyzeBadges(id)` after every `onShotReady`.
- `PostShotReviewPage.qml:148` calls it after a fresh shot lands on the post-shot review.
- `requestReanalyzeBadges` (`shothistorystorage.cpp:1724-1832`) recomputes badges on the worker thread and issues an `UPDATE` if any flag differs from the stored value, then emits `shotBadgesUpdated`.

So opening a shot in the detail view self-heals its DB row. The 884/889 channeling-on-clean-shots case is fixed permanently after each is viewed once.

## Remaining gap

Shots the user never opens stay stale in the DB, so:

- The history-list filter (`requestShotsFiltered` → `buildFilterQuery()` around `shothistorystorage.cpp:1463-1474`) keeps returning them under the save-time badge values.
- The MCP `shots_list` tool reads the same stored columns, so external AI agents see the same stale data.

For users with hundreds of shots, lazy persist never fully converges.

## Options for closing the gap

1. **One-shot reanalyze sweep** at app startup or as a post-update migration: walk every shot, run `requestReanalyzeBadges` per row, persist updated columns. Heavy on first run (~881 shots locally) but makes the list view immediately consistent. Could be gated by a settings flag (e.g. `badges_resweep_vN`) so it only runs once per detector revision.
2. **Manual \"Re-analyze all shots\" admin button** (settings? shot-list overflow menu?) that triggers the bulk sweep on demand. Users who notice the inconsistency can fix it; the rest pay nothing.
3. **Hybrid**: option 1 + option 2 (already shipped), with the manual button as a debug/admin escape hatch.

## Decision

**Holding off on options 1/2/3 for now.** Detector tuning is still active (see the recent run of #890 / #891 / #892 / #893) and more fixes are expected. Running a bulk resweep now would just need to be re-run after the next detector change. Revisit once the badge detectors are stable.

In the meantime, lazy persist on view is doing the work, and any user who hits a visible inconsistency can resolve it by opening the affected shots.

## Notes

- `requestReanalyzeBadges` already implements the per-shot write+signal semantics — reusable for any future bulk path.
- A one-shot migration could ride alongside the next schema migration so it only runs once.

Discovered during the PR #893 review (https://github.com/Kulitorum/Decenza/pull/893).

skialpine:
## Summary
- Closes the residual gap from #894: when the on-load badge recompute introduced in #893 finds drift between the stored DB columns and the current detector verdict, `loadShotRecordStatic` now issues an `UPDATE` on the same connection before returning. The DB converges with detector improvements as shots are viewed/queried instead of needing a separate bulk-resweep migration.
- Every consumer of shot detail goes through this one path — UI (`requestShot`), MCP (`shots_get_detail`, `shots_compare`, dialing tools), the HTTP shot server, the comparison/calibration models, and AI prompt building — so all of them now persist drift on read, not just the QML detail pages.
- `requestReanalyzeBadges` shrinks to a thin wrapper around `loadShotRecordStatic` via a new `outBadgesPersisted` out-param. The duplicate ~70-line detector block is gone; the worker just emits `shotBadgesUpdated` when the load wrote. QML wiring on `ShotDetailPage` / `PostShotReviewPage` is unchanged — the post-`onShotReady` call to `requestReanalyzeBadges` is now redundant in the common case (the load already persisted) but harmless.

## Why this matters

Issue #894 documented that the always-recompute-on-load behavior from #893 only updated the in-memory `ShotRecord`. The DB columns kept their save-time values, so `requestShotsFiltered` / `buildFilterQuery()` (the history-list filter chips) and the MCP `shots_list` tool kept returning stale verdicts for shots that had not been opened in `ShotDetailPage` since the last detector change.

`requestReanalyzeBadges` already did the right thing — recompute and persist — but only when QML explicitly called it on detail-page open. Other read paths (MCP, shot server, AI context loads, comparisons) never persisted. The intent stated when this came up was: "Reading a shot detail was supposed to always update the DB both from the UI and the MCP." This change makes that true.

## Effect

- Open a stale shot once via the detail view → DB row converges (already true before #893; still true after).
- Call `shots_get_detail` once via MCP → DB row converges (new).
- Call `shots_compare` for a stale shot → DB rows converge (new).
- Browse the history list with a filter chip → the filter still reads stored columns, but every detail open from that list now writes back, so successive filter passes converge.
- Loop through shots for AI prompt context → drifted badges get rewritten as a side effect (new).

## Test plan
- [x] Build clean — `mcp__qtcreator__build` reports 0 errors / 0 warnings, ~47 s.
- [x] `mcp__qtcreator__run_tests` — 1742 passed, 0 failed, 0 warnings.
- [ ] On-device: open a stale-badge shot in the detail view, confirm the badge chip matches the current detector and the row's `updated_at` advances.
- [ ] On-device: hit `shots_get_detail` for a stale shot via MCP and confirm the same.
- [ ] On-device: call `shots_compare` for a list including a stale shot and confirm the same.
- [ ] Verify `requestReanalyzeBadges` from QML still emits `shotBadgesUpdated` when something actually drifted (e.g., after a future detector tweak).

Closes #894.

### DIFF
--- a/qml/pages/PostShotReviewPage.qml
+++ b/qml/pages/PostShotReviewPage.qml
@@ -144,8 +144,10 @@ Page {
                 // arrived via R2 before the shot data was ready, or where the DB already
                 // has a non-zero TDS from a previous session).
                 calculateEy()
-                // Recompute quality badges in background (handles stale values after KB updates)
-                MainController.shotHistory.requestReanalyzeBadges(shotId)
+                // Quality badges already arrived recomputed in `shot` via
+                // loadShotRecordStatic, which also persists drift to the DB
+                // and emits shotBadgesUpdated when it does. onShotBadgesUpdated
+                // below catches the persist event.
             }
         }
         function onShotBadgesUpdated(shotId, channeling, tempUnstable, grindIssue, skipFirstFrame) {
--- a/qml/pages/ShotDetailPage.qml
+++ b/qml/pages/ShotDetailPage.qml
@@ -20,7 +20,6 @@ Page {
     property bool advancedMode: Settings.boolValue("shotReview/advancedMode", false)
     property int swipeDirection: 0  // 1 = going older, -1 = going newer; swipeDirection: 1 exits left, enters from right; -1 exits right, enters from left
     property bool navigating: false  // true only during a navigateToShot transition; guards enterAnimation from firing on non-navigation loads
-    property bool reloadingFromVisualizer: false  // true when loadShot() was called from onVisualizerInfoUpdated; suppresses duplicate badge reanalysis
 
     // Pick up toggle changes made on any other page sharing this setting
     // (Post-Shot Review, Shot Comparison, Espresso view selector).
@@ -68,7 +67,6 @@ Page {
             // loadShotRecordStatic, which also persists drift to the DB and
             // emits shotBadgesUpdated when it does. No extra reanalyze call
             // needed here — onShotBadgesUpdated below catches the persist event.
-            shotDetailPage.reloadingFromVisualizer = false
         }
         function onShotDeleted(deletedId) {
             if (deletedId === shotDetailPage.shotId)
@@ -77,7 +75,6 @@ Page {
         function onVisualizerInfoUpdated(id, success) {
             if (id !== shotDetailPage.shotId) return
             if (success) {
-                shotDetailPage.reloadingFromVisualizer = true
                 loadShot()
             } else {
                 console.warn("ShotDetailPage: Failed to save visualizer info for shot", id)
@@ -142,7 +139,6 @@ Page {
         }
         function onUpdateSuccess(visualizerId) {
             if (shotDetailPage.shotId > 0) {
-                shotDetailPage.reloadingFromVisualizer = true
                 loadShot()
             }
         }
@@ -170,7 +166,6 @@ Page {
                 shotId = shotIds[currentIndex]
                 contentSlide.x = shotDetailPage.swipeDirection * Theme.scaled(50)
                 shotDetailPage.navigating = true
-                shotDetailPage.reloadingFromVisualizer = false
                 loadShot()
             }
         }
--- a/qml/pages/ShotDetailPage.qml
+++ b/qml/pages/ShotDetailPage.qml
@@ -64,11 +64,10 @@ Page {
                 if (wasNavigating)
                     enterAnimation.start()
             })
-            // Recompute quality badges in background (handles stale values after KB updates).
-            // Skip when reloading after a visualizer update — badges didn't change and the
-            // visualizer path already triggers a second onShotReady via loadShot().
-            if (!shotDetailPage.reloadingFromVisualizer)
-                MainController.shotHistory.requestReanalyzeBadges(id)
+            // Quality badges already arrived recomputed in `shot` via
+            // loadShotRecordStatic, which also persists drift to the DB and
+            // emits shotBadgesUpdated when it does. No extra reanalyze call
+            // needed here — onShotBadgesUpdated below catches the persist event.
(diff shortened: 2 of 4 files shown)


skialpine:
### Code review

Found 1 high-confidence issue. Fixed in 040eed80.

1. **Redundant double load on every shot detail open** (score 75). With persist moved into `loadShotRecordStatic`, `requestShot` now writes drift on the same connection it loads from. The QML `onShotReady` handlers in `ShotDetailPage` and `PostShotReviewPage` were still calling `requestReanalyzeBadges`, which spawned a second background thread that opened a temp DB connection, decompressed the sample blob, re-ran all four detectors, and found nothing to write — pure waste, ~50–200 ms on the Decent tablet, on every shot detail open. The fix routes the `shotBadgesUpdated` emission through `requestShot` itself via the new `outBadgesPersisted` out-param, and drops the QML reanalyze call.

   https://github.com/Kulitorum/Decenza/blob/a5838aac1b4a96a7b00e6814eaa5cc0c0ca34c34/qml/pages/ShotDetailPage.qml#L67-L72

   https://github.com/Kulitorum/Decenza/blob/040eed803b02058d1c9c1beb79ce1b289934df22/src/history/shothistorystorage.cpp#L1693-L1733

<details>

- [Worth a look] `shotBadgesUpdated` is suppressed when the UPDATE itself fails (e.g., SQLITE_BUSY beyond the busy_timeout, disk full); the qWarning logs the failure but no signal fires. The UI display is still correct because `shotReady` carried the recomputed values. The fix above narrows the window further (signal now emits from the same worker that did the write), but a transient write failure still leaves no signal. Acceptable given UI is correct via `shotReady`. (score: 68)
- [OK to skip] Original concern that `shotBadgesUpdated` would never fire on first-open due to a race between `requestShot` and the QML-issued `requestReanalyzeBadges` — resolved by the same fix above (the redundant QML call is gone, the signal now fires from `requestShot` itself when the write happened). (score: 25)
- [OK to skip] `updated_at` bumped on viewed-and-drifted shots could cause exporter re-export work. Verified via grep that `updated_at` is not used as a sync cursor in `datamigrationclient.cpp`, `databasebackupmanager.cpp`, or `shotserver_backup.cpp`. The exporter freshness check (`shothistoryexporter.cpp:114`) treats this correctly: badge changed → file is stale → re-export, which is the intended behavior. (score: 22)

</details>

False positives ruled out during the review:

- A "critical" finding from one agent that the UPDATE was gated on `outBadgesPersisted` being non-null — verified against `shothistorystorage.cpp:2243-2247`, the gate is `if (flagsChanged)` and the `outBadgesPersisted` reference is only inside the success branch.
- A claim that the UPDATE failure path was silent — verified the `qWarning(...)` else branch at `shothistorystorage.cpp:2260-2262`.

Issue #900, opened 2026-04-27
Need a way to repair shot phase rows corrupted by the firmware time wrap


skialpine:
## Background

Live verification of v1.7.2 prerelease against the shot DB surfaced shot 887 — a clean 35.6 g / 36 g, 59 s extraction on the 80's Espresso profile that fires both \`grindIssueDetected\` and \`skipFirstFrameDetected\` badges. Investigation showed the badges are correct given the input data; the input data itself is the problem.

The shot's \`shot_phases\` rows on the device DB look like:

| frame | label           | time_offset |
|------:|-----------------|------------:|
|     0 | Start           |        0.00 |
|     0 | preinfusion start |      0.00 |
|     1 | preinfusion     |        2.50 |
|     2 | rise and hold   |    **−647.61** |
|     3 | decline         |    **−642.86** |
|    −1 | End             |       53.27 |

The negative times are real markers from this shot whose \`time_offset\` got broken by the firmware time wrap (the same 16-bit value rollover that the v1.7.2 release notes say was fixed for live timer display). Adding the wrap constant back recovers sane values:

- \`−647.61 + 655.36 = 7.75 s\` ← matches \"rise and hold\" entering after the preinfusion exit (~6 s preinfusion-start + ~2 s preinfusion-then-pressure-exit)
- \`−642.86 + 655.36 = 12.50 s\` ← matches +4 s rise-and-hold frame

The downstream effects:

- \`detectSkipFirstFrame\` walks phases linearly and skips a leading \`Start\`/frame-0 marker. With the negatives sorted to the front, the leading marker is frame 2 (not Start), so the skip fails and the loop hits frame 2 immediately with \`sawFrameZero == false\`. Falls into the FW-bug branch which returns \`phase.time < 2.0\` — \`−647.61\` trivially satisfies it. False positive.
- \`analyzeFlowVsGoal\` still false-fires even after a hypothetical in-memory filter that drops the negatives, because the *real* frame-2 / frame-3 transitions only exist as the negative phantoms — once filtered, the detector treats frame 1 (preinfusion) as running for the entire 50 s pour and reads the 2–3 mL/s flow during what was actually pressure-mode as a strong fine-grind signal.

The shot's badge columns therefore stay wrong on every load — the recompute uses the corrupted phases, produces the same wrong result, the persist (PR #896) sees no change, and the user keeps seeing the bad chips.

## Why this issue exists separately

[...]

PR #898, opened 2026-04-27
fix(analysis): gate temperatureUnstable on the shot reaching real extraction

skialpine:
## Summary
- Live verification of v1.7.2 prerelease against shot history surfaced a clean false positive: shot 885 — a 2.3 s shot aborted during preinfusion-start — was firing `temperatureUnstable=true`. The detector was averaging the machine's preheat ramp (68 → 75 °C against an 82 °C goal) across an extraction that never actually started, producing ~9 °C average deviation and tripping the 2 °C threshold.
- Add `ShotAnalysis::reachedExtractionPhase(phases, shotDuration)` and gate all three temperature-unstable call sites on it. The function returns true only when at least one `frameNumber >= 1` marker sits more than `TEMP_MIN_EXTRACTION_SEC` (1.0 s) before the shot's last sample.
- The duration buffer matters: phase-marker presence alone is not enough, because the firmware sometimes records the 0 → 1 transition in the final ms of an aborted shot. Shot 885 is exactly that — a frame-1 marker exists at `time=2.5` but the shot ended at `2.284`, so no samples landed in frame 1. The gate now correctly excludes that case.

## Detection results

| Shot | Duration | Frame ≥ 1 entered at | Past frame 1 | New gate |
|------|----------|---------------------:|-------------:|----------|
| 885 (aborted preheat) | 2.284 s | 2.5 s | -0.216 s | **skip** |
| 891 (clean 80's, 36 g) | 20.189 s | 1.873 s | 18.316 s | run |
| 890 (puck failure, 1.1 g yield) | 59.554 s | ~2 s | ~58 s | run |

## Three call sites updated

- `ShotHistoryStorage::saveShotStatic` — save-time write of `temperatureUnstable` column.
- `ShotHistoryStorage::loadShotRecordStatic` — load-time recompute (the path that PR #896 made canonical).
- `ShotAnalysis::generateSummary` — the `Temperature drifted X°C from goal on average` observation in the Shot Summary dialog.

## Why not just \"don't save aborted shots\"

That's a separate, larger discussion (user-visible behavior change, settings toggle, threshold definitions, telemetry). Even if we add a save-time filter for future shots, legacy aborts already in the DB still need this gate. Issue forthcoming.

## Test plan
- [x] `mcp__qtcreator__build` — clean, 0 errors / 0 warnings.
- [x] `mcp__qtcreator__run_tests` — 1742 passed, 0 failed, 0 warnings.
[...]

PR #901, opened 2026-04-27
fix(analysis): apply rising-pressure channeling gate in pressure-mode phases too

skialpine:
## Summary
- Live verification of v1.7.2 prerelease against shot history surfaced a clean false positive: shot 889 — a 35.6 g / 36 g, 59 s extraction on 80's Espresso — fired \`channelingDetected\` despite landing on target with no puck issue.
- Root cause: \`buildChannelingWindows\`'s rising-pressure gate (added in #866) was scoped flow-mode-only. The pressure-mode \"rise and hold\" tail produces the same intentional-ramp dynamic that the gate was built to suppress in flow-mode lever shots — pressure climbs the last leg toward a stationary goal under a permissive 15 % convergence threshold, dragging conductance hard negative, and ~11 dC/dt samples land in the inclusion window during that stretch (just over the >10 sustained threshold).
- Drop the \`isFlowMode\` guard so the gate applies in both phase modes. Direction-of-rise is preserved — the gate still fences only on rising pressure, never on falling, so bloom transitions and channel-induced pressure dips remain detectable.

## Why this is safe

Real channeling in pressure mode looks the **opposite** of an intentional ramp: flow surges and pressure dips as the controller compensates. That produces strongly *positive* dC/dt and falling pressure — both untouched by a rising-pressure gate. The \`shot_corpus_regression\` test confirms no fixture verdict changed; none of the 12 curated corpus shots have a clean pressure-mode rise-and-hold tail that previously triggered this path.

## What the data showed

For shot 889 around the convergence-but-still-rising stretch:

| t (s) | actual P | goal P | |P − G| / G | future P (t+0.75) | rising 15 % ahead? |
|---:|---:|---:|---:|---:|---|
| 7.704 | ~6.5 | 7.687 | within 15 % | 7.45 | yes |
| 7.901 | 7.122 | 7.75 | 8.0 % | 7.75 | yes |
| 8.111 | 7.457 | 7.812 | 4.5 % | 8.09 | yes |
| 8.325 | 7.749 | 7.812 | 0.8 % | 8.18 | barely |
| 8.5+  | ~8.1 | 7.812 | < 5 % | flat | no |

Samples from ~7.7 s to ~9.8 s passed the existing window checks (stationary goal + converged actual) but were the active pressure ramp itself — not puck behavior. With the gate now applied in pressure mode, those samples drop out of the inclusion window and channeling correctly stays silent.

## Test plan
- [x] \`mcp__qtcreator__build\` — clean, 0 errors / 0 warnings.
- [x] \`mcp__qtcreator__run_tests\` — 1742 passed, 0 failed, 0 warnings.
- [x] \`shot_corpus_regression\` — passes (no corpus verdict changed).
- [ ] On-device: open shot 889 in detail view, confirm channel chip no longer shows.
- [ ] On-device: confirm a real channeling shot (next time one appears) still flags — the gate doesn't suppress positive-dC/dt patterns, only intentional-ramp negative ones.

## Related

- PR #866 — the original flow-mode rising-pressure gate (this PR generalizes it).
- PR #898 — fixes the unrelated \`temperatureUnstable\` false positive on the same verification run (shot 885).
- Issue #900 — separate save-time data corruption tracked for shot 887.

### DIFF
--- a/docs/SHOT_REVIEW.md
+++ b/docs/SHOT_REVIEW.md
@@ -91,13 +91,25 @@ included only when, at that time:
    / goal(t) ≤ 0.15` on both sides.
 2. The actual value has **converged** onto the goal: `|actual − goal| / goal ≤
    0.15`.
-3. For **flow-mode phases only**, pressure is not rising fast: when
+3. Actual pressure is not rising fast in either mode: when
    `pressureNow > 0.5` bar AND `pressureFut > pressureNow * 1.15` over 0.75 s
-   ahead, the sample is disqualified. Falling pressure is allowed — bloom
-   transitions and pressure-mode → flow-mode handoffs are legitimate
-   channeling signals. The 0.5 bar precondition gates out the
-   near-atmospheric window where small absolute jitter would otherwise read
-   as a 15 % relative rise.
+   ahead, the sample is disqualified. Two ramp dynamics produce the same
+   conductance-drop signature and are both gated here:
+   - Flow-mode lever rise (Cremina, Damian LRv3, 80's Espresso preinfusion):
+     flow goal stationary while the puck builds pressure under a pressure-
+     ceiling exit.
+   - Pressure-mode rise-and-hold final leg: pressure goal locked at target
+     but actual pressure still ramping toward it. The 15 %-of-goal
+     convergence check admits these samples because actual is close enough
+     to goal, even though it is still climbing fast — `dC/dt` clamps at the
+     negative floor as conductance drops with pressure. (Shot 889 on 80's
+     Espresso was the motivating false positive.)
+   Falling pressure is always allowed — bloom transitions and
+   pressure-mode → flow-mode handoffs are legitimate channeling signals,
+   and the dC/dt detector counts both signs of conductance change as
+   diagnostic. The 0.5 bar precondition gates out the near-atmospheric
+   window where small absolute jitter would otherwise read as a 15 %
+   relative rise.
 
 Contiguous qualifying times collapse into windows; gaps ≤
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 1 issue:

1. Stale doc comment on `buildChannelingWindows` says the rising-pressure gate is flow-mode-only, but this PR removes that guard so the gate now runs for all phases. The header still describes the prior invariant — a future refactor reading only the header could re-introduce the regression this PR is fixing.

[src/ai/shotanalysis.h#L73-L82](https://github.com/Kulitorum/Decenza/blob/110c89f37675a5f371267ad74d32292e9abb92ca/src/ai/shotanalysis.h#L73-L82)

---

<details>

- [Worth a look] No corpus fixture added for the motivating shot 889 (clean 80's Espresso pressure-mode rise-and-hold). The PR's commit message acknowledges "none of the 12 curated corpus shots have a clean pressure-mode rise-and-hold tail that previously triggered this path," meaning a future re-narrowing of the gate would slip past `shot_corpus_regression`. Per [docs/CLAUDE_MD/TESTING.md](https://github.com/Kulitorum/Decenza/blob/110c89f37675a5f371267ad74d32292e9abb92ca/docs/CLAUDE_MD/TESTING.md), the corpus is the documented mechanism for guarding against detector regressions. (score: ~50)
- [OK to skip] Comment claims "real channeling in pressure mode manifests as flow surge with pressure dip... producing strongly POSITIVE dC/dt." The detector explicitly counts both signs of dC/dt; a puck restriction (low flow under sustained pressure) produces negative dC/dt and is also a real channeling manifestation. Comment overstates the canonical pressure-mode signature. (score: ~40)

</details>

PR #909, opened 2026-04-28
fix(mcp): harden SSE socket lifetime to prevent macOS CFSocket UAF

skialpine:
## Summary
- MCP SSE clients are now stored as `QList<QPointer<QTcpSocket>>` so a dropped socket reads as null instead of dangling on iteration — addresses the macOS `qt_mac_socket_callback` SIGSEGV observed after \"SSE client disconnected\" in the log.
- `McpServer::isSseClient()` lets `ShotServer::onReadyRead` skip MCP SSE sockets next to the existing layout/theme early-returns.
- `McpServer::probeSseKeepalives()` is invoked from `onCleanupTimerTick` so dead MCP connections are detected on the same 30 s cadence as the other SSE types. `ShotServer` remains the sole destroyer via `onDisconnected → deleteLater`.

## Why
A user-reported SIGSEGV on macOS crashed at:
```
__CFSocketPerformV0 → qt_mac_socket_callback → QCoreApplication::sendEvent → notifyInternal2
```
with `McpServer: SSE client disconnected, remaining: 0` as the last log line. Tracing showed three problems with how MCP SSE sockets were managed:
1. Raw `QTcpSocket*` storage in `m_sseClients` and a single signal-driven removal — no defense against teardown ordering races.
2. MCP SSE sockets were not in the SSE-aware early-return in `onReadyRead`, unlike layout/theme SSE clients.
3. No keepalive probe for MCP SSE clients — silently-broken connections sat until the OS surfaced the disconnect.

## Test plan
- [ ] Build cleanly on macOS (verified — `mcp__qtcreator__build` succeeded, 0 warnings)
- [ ] Connect an MCP client (e.g. `mcp-remote`), let it idle past the 30 s cleanup tick, confirm keepalive comments are visible on the wire
- [ ] Disconnect the MCP client and confirm `m_sseClients` empties without crash
- [ ] Run for an extended idle session on macOS (the original crash reproduced after several hours of idle SSE reconnects)

### DIFF
--- a/src/mcp/mcpserver.cpp
+++ b/src/mcp/mcpserver.cpp
@@ -322,8 +322,15 @@ void McpServer::handleHttpRequest(QTcpSocket* socket, const QString& method,
             return;
         }
 
-        // SSE stream for server-initiated notifications
-        if (static_cast<int>(m_sseClients.size()) >= MaxSseConnections) {
+        // SSE stream for server-initiated notifications. Count only live entries —
+        // a QPointer that has gone null (socket destroyed before our disconnect
+        // lambda ran) still occupies a slot until probeSseKeepalives() GCs it on
+        // the next 30 s tick, and we don't want stale nulls to falsely trip the
+        // limit and reject a legitimate client.
+        int liveSseCount = 0;
+        for (const QPointer<QTcpSocket>& p : std::as_const(m_sseClients))
+            if (!p.isNull()) ++liveSseCount;
+        if (liveSseCount >= MaxSseConnections) {
             sendHttpResponse(socket, 429, "Too many SSE connections", "text/plain");
             return;
         }
--- a/src/mcp/mcpserver.cpp
+++ b/src/mcp/mcpserver.cpp
@@ -146,10 +146,11 @@ void McpServer::broadcastSseNotification(const QString& resourceUri)
 
     // Send only to sessions that subscribed to this resource URI.
     // Sessions without any subscriptions receive all notifications (backward compat).
-    QList<QTcpSocket*> dead;
-    for (QTcpSocket* client : std::as_const(m_sseClients)) {
-        if (client->state() != QAbstractSocket::ConnectedState) {
-            dead.append(client);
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 1 issue:

1. The precondition comment at `shotserver.cpp:356-363` enumerates SSE registration sites that must call `m_keepAliveTimers.take()` and lists exactly two: `/api/theme/subscribe` and `/api/layout/events`. This PR adds a third SSE registration site (MCP via `m_mcpServer->handleHttpRequest`) without updating the comment. The invariant still holds in practice — MCP `/mcp` routes bypass `sendResponse`/`resetKeepAliveTimer` entirely, so MCP SSE sockets never enter `m_keepAliveTimers` — but the comment is now an incomplete enumeration of a load-bearing rule, and a future maintainer adding a fourth SSE endpoint reading this comment as exhaustive could miss the requirement. The CLAUDE.md note "If removing the comment wouldn't confuse a future reader, don't write it" implies this comment exists because the rule is non-obvious; an outdated enumeration is a maintenance hazard.

https://github.com/Kulitorum/Decenza/blob/d48124fa98607ce4fa0e10b0cf209152a1ace0bb/src/network/shotserver.cpp#L355-L364

<details>

- [OK to skip] Comment at `shotserver.cpp:938` ("ShotServer remains the sole destroyer via onDisconnected → deleteLater") is correct but doesn't explicitly document that this contract depends on `onNewConnection` wiring `disconnected → onDisconnected` for every accepted socket. Adjacent comments cover the synchronous-close path adequately. (score: 60)
- [Worth a look] `MaxSseConnections` gate at `mcpserver.cpp:326` uses `m_sseClients.size()`, which counts null QPointers from sockets destroyed without going through the disconnect lambda. In the exact scenario this PR defends against, 4 stale nulls could reject legitimate SSE connections for up to 30 s until `probeSseKeepalives` GCs them. Self-correcting; one-line fix to count non-null entries. (score: 65)

</details>

PR #910, opened 2026-04-28
feat(analysis): add yield-overshoot ("gusher") arm to grind detector

### LINKED ISSUE #904: No badge fires on gusher shots that significantly overshoot target weight
## Background

Surfaced during the third 50-shot verification batch (IDs 781-841) on the v1.7.2 prerelease. At least one clear gusher shot is silent on all four badges despite a 41% yield overshoot.

## Concrete case

**Shot 835** — D-Flow / Q profile, 18 g dose → 40 g target, **56.3 g yield in 35.4 s**, grind 11, peak pressure 6.04 bar. SAW didn't fire (yield blew well past target). All four badges report **F**. The summary verdict probably lands somewhere in the \"minor issues\" or \"clean shot\" range — wrong.

This is a textbook grind-too-coarse signature: the puck offered too little resistance, water blew through, the shot finished much heavier than intended. The user reads the bag and the bottom of the cup says \"failure,\" but the badge chip says \"clean.\"

## Why it's silent under current detectors

- `detectChannelingFromDerivative` → silent. dC/dt evolution looks normal in absolute terms; the puck wasn't channeling, it just had no resistance.
- `analyzeFlowVsGoal` flow-vs-goal arm → silent. Actual flow during D-Flow's Pouring phase (flow-mode, goal 1.8 mL/s) tracks the goal because the flow controller can pull back when actual exceeds goal. The over-yield happens during Filling/Infusing (pressure-mode), not Pouring.
- `analyzeFlowVsGoal` choked-puck arms → silent. Both arms gate on `pressurizedDuration ≥ CHOKED_DURATION_MIN_SEC` (15 s). For 835, pressure stayed above 4 bar for only ~4 s. The puck was so permeable that pressure couldn't build long enough for the gate to engage.
- Yield-ratio choked arm → silent. The arm fires when `yield/target < 0.85`. For 835, yield/target = 1.41 — *way over*, not under, so this arm doesn't apply by design.
- `detectPourTruncated` → silent. Peak pressure 6.04 bar > `PRESSURE_FLOOR_BAR` (2.5).
- `temperatureUnstable` → silent. Temp tracked goal fine.
- `skipFirstFrameDetected` → silent. Frame 0 ran normally.

So the entire detector suite is silent on a shot that visibly failed.

## Proposed: a yield-overshoot detector arm

Mirror the moderate-choke yield-ratio arm but in the *over* direction:

- Compute `yieldRatio = finalWeightG / targetWeightG`.
- Fire `gusherDetected` (or roll into `grindIssueDetected` with a different sub-flag) when:
  - `yieldRatio > YIELD_OVER_RATIO_MAX` (e.g., 1.20 — 20 % over target), AND
  - The shot wasn't a deliberate manual stop (autonomous DE1 stop, not user-cancelled), AND
  - Beverage type is espresso (not filter/pourover where bigger volumes are normal).

The existing yield-ratio choked arm requires `targetWeightG > 0 && finalWeightG > 0` — same precondition here. For shots with no recorded dose/target this stays silent (correct).

A lower threshold than the moderate-choke counterpart (`< 0.85`) makes sense because gushers are a less common but more obvious failure — by the time you're 20 % over, the user has visibly noticed.

## Threshold tuning

The existing CHOKED arms use:

- `CHOKED_DURATION_MIN_SEC = 15.0` — needs a long pressurized window
- `CHOKED_FLOW_MAX_MLPS = 0.5` — severe flow signal
- `CHOKED_YIELD_RATIO_MAX = 0.85` — moderate yield signal

For the gusher arm, the duration-min gate may be unnecessary — a gusher *by definition* doesn't pressurize for long. Just gate on the yield ratio plus the espresso/non-aborted check.

## Considered + rejected

- **Just rely on the user noticing the cup is heavy:** That's what we're doing now, and it means the badge UI and Shot Summary dialog both lie about the shot quality.
- **Add a low-peak-pressure detector instead:** Issue #903 already proposes one for under-yield cases (\"underpressurized-but-not-truncated\"). A peak-pressure-based gusher detector would conflict with profiles like D-Flow that intentionally run lower peaks. Yield ratio is profile-agnostic.

## Test plan

- Add fixture for shot 835 (or one matching shape) to `tests/data/shots/`.
- Add yield-overshoot fixtures spanning grind 9-11 to confirm threshold scales reasonably.
- Update `manifest.json` with the expected `grindIssueDetected=true` (or new gusher badge) verdict.
- Re-run `shot_corpus_regression`.

## Related

- Issue #903 — adjacent low-pressure-shot gap (under-yield direction).
- The verification batch where this surfaced: third 50 (IDs 781-841), shot 835 specifically.

skialpine:
## Summary

Closes #904. Adds a third arm to `analyzeFlowVsGoal` that fires when `finalWeight / targetWeight > 1.20`, catching shots like 835 (18 g → 40 g target, **56.3 g actual at grind 11**) where every existing detector stays silent:

- `analyzeFlowVsGoal` flow arm — flow tracks goal in flow-mode pour (controller pulls back).
- Choked-puck arms — both gated on `pressurizedDuration ≥ 15 s` at `pressure ≥ 4 bar`. Shot 835 only crosses 4 bar for ~4 s.
- Yield-ratio choked arm — fires when `yield/target < 0.85`. Shot 835 ratio is 1.41.
- `detectPourTruncated` — peak 6.04 bar > 2.5 bar floor.
- Channeling / temp / skip-first-frame — all silent (curves look normal in absolute terms).

Yield/target is the profile-agnostic signal that survives, and the existing yield-ratio arm already reads it for the `< 0.85` direction; this just adds the `> 1.20` mirror.

## Implementation notes

- Folds into `grind_issue_detected` (no DB migration). Existing badge chip / drift-on-load path light up for free.
- Arm has **no pressurized-window gate** — a gusher by definition can't sustain pressure long enough to satisfy the choked-puck `15 s × 4 bar` window.
- Verdict cascade gets a new tier above `chokedPuck`: gushers and chokes are mutually exclusive on yield ratio, but the gusher case deserves its own emphatic line ("Pour gushed past target — grind way too coarse. Grind much finer.") instead of falling through to the generic "Puck integrity issue" wording.
- `tools/shot_eval/main.cpp` — local mirror of `detectGrindIssue` updated to also check `grind.yieldOvershoot` so the `shot_corpus_regression` ctest target reflects the new arm.

## Out of scope (follow-up)

The saved `yield_override` column doesn't currently capture mid-shot +10g target bumps — `MainController::onShotEnded()` clears the bump before reading the SAW target. A bumped shot whose final weight matches the *bumped* target naturally has `yield/originalTarget > 1.20` → false positive on this arm. Tracked as a separate PR (see follow-up task in this session) that captures `m_machineState->targetWeight()` before clearing.

## Test plan

- [x] 38/38 pass in `tst_ShotAnalysis` (8 new methods covering threshold boundary, clean shot, zero target, non-espresso skip, empty curves, `grind_check_skip` flag suppression, and the summary verdict cascade).
- [x] 14/14 pass in `shot_corpus_regression` — added `synthetic_shot_835_gusher.json` fixture with simplified flat-plateau curves so the new arm is exercised in isolation. Replace with captured shot 835 curves once available.
- [ ] Open shot 835 in `ShotDetailPage` after build deploys and confirm the orange "Grind issue" chip appears + "Pour gushed past target" verdict in the Shot Summary dialog.


skialpine:
### Code review

<details>
<summary>7 below-threshold issues (50-74)</summary>

- [Worth a look] Manifest entry doesn't note that peak pressure > PRESSURE_FLOOR_BAR (2.5) is the load-bearing invariant keeping pourTruncated=false — a future replacement of the synthetic curves could quietly flip that without realizing (score: 60). https://github.com/Kulitorum/Decenza/blob/d8fdd9499dc61e38bdac6f50439e9f6a50e6405d/tests/data/shots/manifest.json#L31-L40
- [OK to skip] Observation block checks chokedPuck before yieldOvershoot but the verdict cascade checks yieldOvershoot first — internally inconsistent only in the physically-unreachable case where a gusher also satisfies the choked-puck flow arm's 15s × 4 bar × <0.5 mL/s gate (score: 50). https://github.com/Kulitorum/Decenza/blob/d8fdd9499dc61e38bdac6f50439e9f6a50e6405d/src/ai/shotanalysis.cpp#L811-L912
- [OK to skip] Catastrophic gusher (yield > 1.20× target AND peak < 2.5 bar) emits two parallel "warning" observation lines; verdict cascade still picks pourTruncated. Same dual-warning shape already exists for chokedPuck + pourTruncated (score: 50). https://github.com/Kulitorum/Decenza/blob/d8fdd9499dc61e38bdac6f50439e9f6a50e6405d/src/ai/shotanalysis.cpp#L811-L862
- [Worth a look] "Mutually exclusive with chokedPuck on the yield arm" reads as blanket exclusivity — chokedPuck can also fire from its flow sub-arm, so the comment should say "on the yield-ratio sub-arm" (score: 50). https://github.com/Kulitorum/Decenza/blob/d8fdd9499dc61e38bdac6f50439e9f6a50e6405d/src/ai/shotanalysis.h#L223-L228
- [Worth a look] YIELD_OVERSHOOT_RATIO_MIN doc lists "espresso beverage type" as a precondition, but the actual gate skips filter/pourover/tea/steam/cleaning — empty/unknown beverage types pass. Tightening to "non-skipped beverage type" matches the code (score: 50). https://github.com/Kulitorum/Decenza/blob/d8fdd9499dc61e38bdac6f50439e9f6a50e6405d/src/ai/shotanalysis.h#L195-L200
- [Worth a look] GrindCheck::delta doc says "Meaningful only when chokedPuck == false" but doesn't tell consumers to also check yieldOvershoot — a delta=0.0 from a yield-arm-only fire could be misread as "flow matched goal" (score: 50). https://github.com/Kulitorum/Decenza/blob/d8fdd9499dc61e38bdac6f50439e9f6a50e6405d/src/ai/shotanalysis.h#L213-L221
- [Worth a look] Gusher observation comment narrows the failure to "water shot through at the preinfusion goal" — the arm is yield-ratio-driven and phase-agnostic; comment should drop the preinfusion specificity (score: 50). https://github.com/Kulitorum/Decenza/blob/d8fdd9499dc61e38bdac6f50439e9f6a50e6405d/src/ai/shotanalysis.cpp#L820-L825

</details>

PR #912, opened 2026-04-28
feat(history): discard espresso shots that did not start (#899)

### LINKED ISSUE #899: Discussion: should we drop the save for shots that didn't start?
## Background

Live verification of v1.7.2 prerelease against the shot DB surfaced shot 885 — a 2.3 s shot that died during preinfusion-start. The DE1 firmware autonomously stopped the shot (\`Stop overlay: Profile complete - DE1 stopped the shot\` in the debug log) before any meaningful extraction happened: peak pressure 0.348 bar, 1.1 g final weight, no frame past 0 actually executed in samples.

Shots like this are saved into the DB today and pollute downstream analysis: they fire badge false positives (PR #898 just gated \`temperatureUnstable\` against this case), they show up in history filter chips, and they bias auto-favorites stats. The user question that prompted this issue: maybe we should not save them in the first place.

## What does \"didn't start\" mean

Some candidate definitions, in increasing strictness:

- **Duration-only**: `duration < 3 s` — catches the trivial case but probably also kills any deliberate 4 s turbo abort.
- **Frame-only**: no phase marker with `frameNumber >= 1` had samples land inside it — semantic, matches the gate PR #898 added for the temp detector.
- **Pressure-floor**: peak pressure during pour stayed below ~2 bar — overlaps with `detectPourTruncated` (which uses 2.5 bar) but covers a slightly different failure mode.
- **Conjunction**: ALL of (duration < 3 s) AND (yield < 0.5 g) AND (peak pressure < 2 bar) — most conservative, only drops the truly nothing-happened cases.

The conjunction probably catches the right population without false-dropping. Telemetry on how often each variant would fire against the existing DB would be informative before picking.

## Things we'd want to handle

1. **User feedback** — silent drops are surprising. Probably need a toast/dialog (\"Shot did not start, not recorded\") and a one-tap undo / save-anyway path.
2. **Settings toggle** — some users may want all shots saved regardless. \"Discard shots that did not start\" defaults to on, can be turned off.
3. **Profile/beverage-type carve-outs** — does this apply to non-espresso (filter / pourover / hot water / steam)? Almost certainly not for steam and hot water, where short \"shots\" are normal.
4. **Telemetry** — how often does this fire on Jeff's history? On other users'? Would inform threshold tuning before rollout.
5. **Migration** — legacy aborts already in the DB stay. PR #898 already handles that side via badge gating; this issue is about the going-forward save filter.

## Proposed next steps

1. Pick a candidate definition (probably the conjunction).
2. Run a one-shot dry-run query against Jeff's local DB: how many of the 881 existing shots would be dropped if applied retroactively? Eyeball the list to confirm none look legitimate.
3. Decide on the user-feedback path (toast vs. undo dialog vs. silent + log).
4. Land behind a settings toggle, default on.

## References

- PR #898 — gates `temperatureUnstable` on `reachedExtractionPhase`. Complementary; doesn't subsume this issue because legacy shots already in the DB still need to be loadable without firing badges incorrectly.
- Shot 885 in the v1.7.2 prerelease DB — the canonical \"did not start\" example; debug log includes the firmware-initiated stop sequence.

skialpine:
## Summary

- Drops espresso shots that the DE1 firmware autonomously stopped during preinfusion before any meaningful extraction (e.g. shot 885: 2.3 s / 1.1 g / 0.35 bar peak). These polluted history, biased auto-favorites, and triggered badge false positives.
- New conjunction classifier: `extractionDuration < 10 s AND finalWeight < 5 g` (both strict `<`). Validated against Jeff's 882-shot DB → 5/882 (0.57%) classify as aborted, **zero false-drops**. Long, low-yield chokes (e.g. 60 s / 1 g) are deliberately preserved as puck-prep diagnostics.
- Settings.brew.discardAbortedShots (default **on**) gates the behavior. Toggle exposed in **Settings → Calibration**, immediately below "Prefer Weight over Volume".
- When discarded, `MainController` emits `shotDiscarded()` and `main.qml` shows an 8-second toast: **"Shot did not start — not recorded"** with a **"Save anyway"** action that replays `saveShot()` from the cached payload. One-shot per discarded shot; cache invalidated when the next shot starts or the toast times out.
- New OpenSpec change `add-discard-aborted-shots` documents the design with a `shot-save-filter` capability spec.
- Closes #899.

## Threshold validation

| Rule | Discarded |
|------|-----------|
| `dur<3 AND yield<0.5g` (original strawman) | 1/882 — misses the canonical shot 885 |
| **`dur<10 AND yield<5g` (chosen)** | **5/882 (0.57%)** — all genuine "did not start" |
| `dur<20 AND yield<5g` | 6/882 — false-drops a 15-s diagnostic trace |

Matched: 885, 17, 850, 836, 1. All have peak pressure < 0.5 bar.
Preserved: 890 (60 s / 1.1 g choke), 708, 732, 117 (133 s / 3.8 g extreme choke), 868 (7 s / 37 g turbo).

## Test plan

- [x] **Unit tests** — `tst_aborted_shot_classifier.cpp` (14 cases): 5 corpus positives, 5 corpus negatives, boundary checks at exactly 10.0 s / 5.0 g (both → kept due to strict `<`), single-clause failures, idempotence loop. All pass.
- [x] Desktop build clean (Qt Creator, 0 warnings, 0 errors).
- [ ] Manual on-device: with toggle on, start an espresso and stop immediately. Verify toast appears with "Save anyway", and that the shot is **not** in history. Verify toggle off saves it normally.
- [ ] Manual: tap "Save anyway" and verify the shot appears in history with the same metadata. Confirm visualizer auto-upload runs if enabled.
- [ ] Manual: start a real shot after a discarded one; verify the toast dismisses cleanly and the new shot saves correctly.
- [ ] Other platforms (Windows / iOS / Android) — no platform-specific code in this change; should build unchanged.

### DIFF
--- a/openspec/changes/add-discard-aborted-shots/proposal.md
+++ b/openspec/changes/add-discard-aborted-shots/proposal.md
@@ -4,7 +4,7 @@
 
 The DE1 firmware can autonomously stop a shot during preinfusion-start (e.g. shot 885 in the v1.7.2 prerelease DB: 2.3 s, peak pressure 0.35 bar, 1.1 g final weight, no frame past 0 actually executed). These "did not start" shots are saved into the local history DB today and pollute downstream analysis: they fire badge false positives (PR #898 just gated `temperatureUnstable` against this case), they show up unfiltered in history chips, and they bias auto-favorites stats and dialing summaries.
 
-The user-visible question (issue #899): we shouldn't be persisting these as if they were real shots. The user is unlikely to want them, but we still need an undo path for the rare false-drop and a setting for users who genuinely want everything saved.
+The user-visible question (issue #899): we shouldn't be persisting these as if they were real shots. The user is unlikely to want them. A settings toggle is provided for users who do want everything saved; there is no per-shot undo because the dry-run shows zero false-drops and the toggle is the right escape hatch.
 
 ## Threshold validation
 
@@ -25,12 +25,12 @@ Suspicious data anomalies (Londonium ids 801 / 438: 2.7-s and 3.6-s with full 36
 ## What Changes
 
 - **ADD** an "aborted shot" classifier in the espresso save path with the validated conjunction: a shot is *aborted* iff `extractionDuration < 10.0 s` AND `finalWeight < 5.0 g`. Both must hold; either alone is not enough.
-- **ADD** save-time filter: when a shot classifies as aborted AND the new toggle is on, the espresso save path SHALL skip `ShotHistoryStorage::saveShot()` and the visualizer auto-upload, log the decision with both classifier values, and emit a UI signal to show a toast.
-- **ADD** a new toast surface in the post-shot flow: `"Shot did not start — not recorded"` with a `Save anyway` action. Tapping `Save anyway` runs the same save path that would have run, with the same metadata. Toast auto-dismiss is the only legitimate timer here per the project design rule.
+- **ADD** save-time filter: when a shot classifies as aborted AND the new toggle is on, the espresso save path SHALL skip `ShotHistoryStorage::saveShot()` and the visualizer auto-upload, log the decision with both classifier values, and emit a UI signal so a toast can notify the user.
+- **ADD** an informational toast `"Shot did not start — not recorded"` shown for a few seconds after a discard. **No "Save anyway" recovery path** — a discarded shot is intentionally gone. Users who want everything saved should turn the toggle off. Toast auto-dismiss is the only timer here per the project design rule.
 - **ADD** a settings toggle `discardAbortedShots` on `SettingsBrew`, default **on**. Exposed in the brew/recording settings page. When off, all shots save as before and the classifier is not consulted.
 - **CARVE OUT** non-espresso paths: steam, hot water, and flush operations do not flow through `MainController::endShot()` save logic, so no explicit guard is needed — but the proposal documents the carve-out so future refactors don't regress it.
 - **NO retroactive migration.** Legacy aborted shots already in the DB stay; they are handled via badge gating (PR #898). This change is going-forward only.
-- **ADD** observability: every classifier evaluation logs `duration`, `finalWeight`, the classifier verdict (`aborted` / `kept`), and the resulting action (`discarded` / `saved` / `saved-anyway`) via the async logger. Telemetry on real users is read out of the log; no new transport.
+- **ADD** observability: every classifier evaluation logs `duration`, `finalWeight`, the classifier verdict (`aborted` / `kept`), and the resulting action (`discarded` / `saved`) via the async logger. Telemetry on real users is read out of the log; no new transport.
 
 ### Out of scope (explicit)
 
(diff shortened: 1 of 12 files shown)


skialpine:
### Code review

Found 3 issues:

1. **Toast auto-dismiss does not release the cached payload, contradicting both code comments and the spec.** `discardedShotToastTimer.onTriggered` only sets `opacity = 0`; nothing in C++ clears `m_pendingDiscardedShot`. The header comment on the cache claims "Cleared on save-anyway, on next shot start, **or when the toast times out**" and `openspec/changes/add-discard-aborted-shots/specs/shot-save-filter/spec.md` (scenario "Toast auto-dismisses without action") says "the discarded shot's in-memory data SHALL be released". With the current wiring the cache stays live until the next shot starts.

   https://github.com/Kulitorum/Decenza/blob/188c98015924838e2ea4ce30a9ef46702fbbbc14/src/controllers/maincontroller.h#L307-L311

   https://github.com/Kulitorum/Decenza/blob/188c98015924838e2ea4ce30a9ef46702fbbbc14/qml/main.qml#L3325-L3334

2. **`saveAbortedShotAnyway()` omits the per-shot DYE metadata reset that the normal save callback performs.** The regular `shotSaved` lambda clears `dyeEspressoEnjoyment` (to default), `dyeShotNotes`, `dyeDrinkTds`, and `dyeDrinkEy` so the next shot starts with a clean slate; the save-anyway lambda only sets `dyeShotDateTime` and `dyeDrinkWeight`. After a "Save anyway", the next espresso will inherit the previous shot's notes/rating/TDS/EY — the exact leakage the explicit reset was added to prevent.

   https://github.com/Kulitorum/Decenza/blob/188c98015924838e2ea4ce30a9ef46702fbbbc14/src/controllers/maincontroller.cpp#L2003-L2014

   Compare with the normal path:

   https://github.com/Kulitorum/Decenza/blob/188c98015924838e2ea4ce30a9ef46702fbbbc14/src/controllers/maincontroller.cpp#L1879-L1894

3. **Discard branch leaves `m_pending*` state inconsistent.** `m_pendingShotEpoch` and `m_pendingDebugLog` are written unconditionally near the top of `onShotEnded()`, before the discard branch returns; but `m_hasPendingShot`, `m_pendingShotDuration`, `m_pendingShotFinalWeight`, and `m_pendingShotDoseWeight` are only updated on the saved-shot path. If a prior shot left `m_hasPendingShot = true` (user exited the post-shot review without uploading), a subsequent discarded shot will overwrite epoch + debug log while leaving the prior shot's timing/weight/dose. A later `uploadPendingShot()` would emit a mangled blend of the two shots. Either reset these fields in the discard branch, or move the `m_pendingShotEpoch` / `m_pendingDebugLog` writes below the classifier check so the discard path doesn't touch them.

   Unconditional writes:

   https://github.com/Kulitorum/Decenza/blob/188c98015924838e2ea4ce30a9ef46702fbbbc14/src/controllers/maincontroller.cpp#L1775-L1786

   Discard branch returns without resetting the rest:

   https://github.com/Kulitorum/Decenza/blob/188c98015924838e2ea4ce30a9ef46702fbbbc14/src/controllers/maincontroller.cpp#L1841-L1851

<details>

- [Worth a look] `tests/tst_aborted_shot_classifier.cpp` uses `QTEST_APPLESS_MAIN`, but `docs/CLAUDE_MD/TESTING.md` mandates `QTEST_GUILESS_MAIN(tst_YourTest)` for new tests, and every existing pure-function test in `tests/` (e.g. `tst_binarycodec`, `tst_recipeparams`, `tst_firmwarepackets`) follows that. The classifier is a pure function so `APPLESS` works today, but it diverges from the convention without a reason. One-line fix. (score: 50)

</details>

skialpine:
### Code review

Found 3 issues — all stale "Save anyway" references that survived the recovery-feature deletion. No functional bugs; the C++ ordering fix and the QML simplification are correct.

1. **`shotDiscarded` signal comment in `maincontroller.h` still promises "Save anyway available"**, but the entire recovery path was deleted in this commit. Misleads future readers who jump from the signal declaration into the Save-anyway flow that no longer exists.

   https://github.com/Kulitorum/Decenza/blob/575c9f644634383fa319bd1182e71fb3ff2fa88d/src/controllers/maincontroller.h#L210-L214

2. **User-visible settings helper text in `SettingsCalibrationTab.qml` still tells users "A toast with a Save anyway action will appear"** — but the toast no longer has that action. This is the only user-facing text in the flow, so the bad description ships to users who toggle the setting, not just to maintainers.

   https://github.com/Kulitorum/Decenza/blob/575c9f644634383fa319bd1182e71fb3ff2fa88d/qml/pages/settings/SettingsCalibrationTab.qml#L419-L426

3. **OpenSpec change docs still describe the removed feature.** Three locations:
   - `tasks.md` 6.1 (observability) lists `saveAbortedShotAnyway` as a log site and includes `saved-anyway` in the action enum — both deleted.
   - `tasks.md` 7.1 / 7.2 (manual verification) reference the non-existent "Save anyway" button and tap-to-recover flow.
   - `proposal.md`'s Impact bullet still says "toast/undo surface"; the spec only has a notification toast now.

   These will be archived as the canonical record of the change once it ships, so a fix here keeps the history honest.

   https://github.com/Kulitorum/Decenza/blob/575c9f644634383fa319bd1182e71fb3ff2fa88d/openspec/changes/add-discard-aborted-shots/tasks.md#L33-L41

   https://github.com/Kulitorum/Decenza/blob/575c9f644634383fa319bd1182e71fb3ff2fa88d/openspec/changes/add-discard-aborted-shots/proposal.md#L46-L50

For the record: the three earlier findings (cache release on toast timeout, DYE-reset divergence, `m_pending*` corruption) are all addressed cleanly. The reordering of `m_pendingShotEpoch` / `m_pendingDebugLog` past the discard gate is correct — nothing in the intervening window reads either member.

PR #917, opened 2026-04-29
fix(saw): capture per-shot log lines for SAW learning

skialpine:
## Summary
- `ShotTimingController::onSettlingComplete` emitted `shotProcessingReady` before `sawLearningComplete`. Because `shotProcessingReady` synchronously runs `MainController::onShotEnded`, which calls `ShotDebugLogger::stopCapture()`, the `[SAW] accuracy` / `accumulated` / `committed` `qDebug` lines that fire afterward were dropped from the per-shot log (system log only) — violating the contract in [`docs/CLAUDE_MD/SAW_LEARNING.md`](https://github.com/Kulitorum/Decenza/blob/main/docs/CLAUDE_MD/SAW_LEARNING.md).
- Wrap `shotProcessingReady` in a `QScopeGuard` so it always fires exactly once on function exit, after `sawLearningComplete` and the early-return `qWarning` paths. Auto flow cal already follows the same "log before stopCapture" ordering ([`maincontroller.cpp:1768`](https://github.com/Kulitorum/Decenza/blob/main/src/controllers/maincontroller.cpp#L1768)); SAW just took a different signal path and missed it.

Confirmed empirically on this machine: shots 891/892 logged `[SAW] accumulated drip=… (n/5)` to the persisted system log, but the per-shot debug logs only showed the `[SAW] model:` line (which fires at extraction *start*, before `stopCapture()`).

## Test plan
- [x] Existing `tests/tst_settling.cpp` still asserts exactly one `shotProcessingReady` emission per shot — `QScopeGuard` preserves that.
- [ ] Build via Qt Creator and pull a SAW shot; verify the saved shot's debug log now contains both `[SAW] accuracy: predictedDrip=…` and `[SAW] accumulated drip=…` (or `committed median lag=…` on every 5th shot).
- [ ] Verify a skipped-learning path (e.g., disconnect scale before settling completes) still saves the shot — `shotProcessingReady` must still fire on the early-return path. The scope guard guarantees this.

### DIFF
--- a/src/controllers/shottimingcontroller.cpp
+++ b/src/controllers/shottimingcontroller.cpp
@@ -4,6 +4,7 @@
 #include "../core/settings.h"
 #include "../machine/machinestate.h"
 #include <QDebug>
+#include <QScopeGuard>
 
 ShotTimingController::ShotTimingController(DE1Device* device, QObject* parent)
     : QObject(parent)
@@ -491,7 +492,13 @@ void ShotTimingController::onSettlingComplete()
     // Settling is done - stop display timer and notify UI
     m_displayTimer.stop();
     emit sawSettlingChanged();
-    emit shotProcessingReady();
+
+    // Defer shotProcessingReady until after sawLearningComplete fires. onShotEnded()
+    // calls ShotDebugLogger::stopCapture(), so any qDebug emitted after that point is
+    // dropped from the per-shot log. SAW_LEARNING.md requires the [SAW] accuracy /
+    // accumulated / committed lines to land in the per-shot log, so we run the SAW
+    // path first and emit shotProcessingReady on scope exit.
+    auto deferProcessing = qScopeGuard([this] { emit shotProcessingReady(); });
 
     // Check scale is still connected
     if (!m_scale || !m_scale->isConnected()) {
--- a/src/controllers/shottimingcontroller.cpp
+++ b/src/controllers/shottimingcontroller.cpp
@@ -493,11 +493,11 @@ void ShotTimingController::onSettlingComplete()
     m_displayTimer.stop();
     emit sawSettlingChanged();
 
-    // Defer shotProcessingReady until after sawLearningComplete fires. onShotEnded()
-    // calls ShotDebugLogger::stopCapture(), so any qDebug emitted after that point is
-    // dropped from the per-shot log. SAW_LEARNING.md requires the [SAW] accuracy /
-    // accumulated / committed lines to land in the per-shot log, so we run the SAW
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Found 3 issues:

1. The new comment names `ShotDebugLogger::stopCapture()` and `onShotEnded()` directly. CLAUDE.md says: *"Don't reference the current task, fix, or callers ('used by X', 'added for the Y flow'…), since those belong in the PR description and rot as the codebase evolves."* `stopCapture()` actually runs at [maincontroller.cpp:1777](https://github.com/Kulitorum/Decenza/blob/3c3cddb059f4ed79a79e343aecd3c73d44f868b5/src/controllers/maincontroller.cpp#L1775-L1779) — well into `onShotEnded()`, after `computeAutoFlowCalibration()` — and the comment's framing implies it fires at function entry. Trimming the cross-file reference to something like *"emit on scope exit so any qDebug from the SAW path lands in the per-shot log before it closes"* would say the same thing and not rot.

   https://github.com/Kulitorum/Decenza/blob/3c3cddb059f4ed79a79e343aecd3c73d44f868b5/src/controllers/shottimingcontroller.cpp#L495-L502

2. The ordering guarantee silently depends on both connections resolving to `Qt::DirectConnection` (same-thread, default `AutoConnection`). The connections in [main.cpp:527](https://github.com/Kulitorum/Decenza/blob/3c3cddb059f4ed79a79e343aecd3c73d44f868b5/src/main.cpp#L527-L552) and `:551` are direct today, but a future refactor that moves a receiver to a worker thread or adds an explicit `Qt::QueuedConnection` would silently break the contract — `sawLearningComplete`'s lambda would run after `shotProcessingReady`'s scope-guard emit had already triggered `stopCapture()`. Worth a one-line note in the comment.

   https://github.com/Kulitorum/Decenza/blob/3c3cddb059f4ed79a79e343aecd3c73d44f868b5/src/controllers/shottimingcontroller.cpp#L495-L502

3. No regression test enforces the new `sawLearningComplete`-before-`shotProcessingReady` ordering. [tst_settling.cpp:130-145](https://github.com/Kulitorum/Decenza/blob/3c3cddb059f4ed79a79e343aecd3c73d44f868b5/tests/tst_settling.cpp#L130-L145) only verifies `shotProcessingReady` count, not order. Two `QSignalSpy` instances + a timestamp comparison would lock the invariant in. Precedent: PR #689 (same file) added regression tests for new behavioural guarantees.

   https://github.com/Kulitorum/Decenza/blob/3c3cddb059f4ed79a79e343aecd3c73d44f868b5/tests/tst_settling.cpp#L130-L145

PR #920, opened 2026-04-29
fix(AI): keep conversation reply visible on mobile after inputDialog close

skialpine:
## Summary

The shot-review AI conversation would render the reply for ~1/8 sec and then appear blank until the user tapped the screen. Two race conditions combined to leave the render thread on a stale paint after returning from the half-screen input dialog.

- **Defer the send to `inputDialog.onClosed`.** The Send button used to call `conversationInput.sendFollowUp()` synchronously right after `inputDialog.close()`, so `conversation.followUp()` emitted `historyChanged` while the dialog was still animating away and the keyboard was still hiding — the overlay was being mutated on an obscured surface. The Send button now stages the message in `inputDialog._pendingMessage`, closes, and `onClosed` runs `sendFollowUp()` against a stable overlay.
- **Drive the scroll + repaint off `conversationText.onContentHeightChanged`.** The previous fix used `Qt.callLater`, which fires *before* the new Markdown layout pass settles, so `update()` was painting a pre-final-layout frame. Moving scroll/`update()` into `onContentHeightChanged` runs them after layout — and re-firing on async Markdown reflow keeps the render thread awake without a polling timer (per the CLAUDE.md "no timers as workarounds" rule).
- **Drop the redundant `conversationText.text = …` in `onResponseReceived`.** `aiconversation.cpp` emits `historyChanged` immediately before `responseReceived`, so `onHistoryChanged` already set the text. The second assignment just forced an extra layout pass on the new Markdown.

## Test plan

- [ ] iOS: open shot-review AI assistant, ask a follow-up via the input dialog, confirm the response appears and stays visible without tapping
- [ ] Android: same as above
- [ ] Desktop: confirm inline send (no inputDialog) still works and scrolls correctly
- [ ] Cancel / Escape from inputDialog: confirm `_pendingMessage` stays empty and no spurious send fires
- [ ] Multi-turn conversation: confirm `_preResponseHeight` scroll target lands at the top of the new response on each turn

### DIFF
--- a/qml/components/ConversationOverlay.qml
+++ b/qml/components/ConversationOverlay.qml
@@ -47,15 +47,23 @@ Rectangle {
 
     onVisibleChanged: {
         if (!visible && inputDialog.visible) {
+            // Drop any staged Send so inputDialog.onClosed doesn't fire a
+            // followUp() against this now-hidden overlay. The user explicitly
+            // dismissed the conversation; honour that and don't ship the
+            // message in the background.
+            inputDialog._pendingMessage = ""
             inputDialog.close()
         }
     }
 
     function open() {
         visible = true
-        Qt.callLater(function() {
-            conversationFlickable.contentY = Math.max(0, conversationFlickable.contentHeight - conversationFlickable.height)
-        })
+        // Scroll synchronously now that the overlay is visible. The TextArea
+        // bound `text` to getConversationText() at creation, so contentHeight
+        // is already current — no need to defer through Qt.callLater (which
+        // would race against any pending Markdown reflow).
+        conversationFlickable.contentY = Math.max(0,
+            conversationFlickable.contentHeight - conversationFlickable.height)
     }
 
     /**
@@ -239,6 +247,15 @@ Rectangle {
                             onClicked: {
                                 if (MainController.aiManager) {
                                     MainController.aiManager.clearCurrentConversation()
+                                    // Reset the scroll state machine. clearHistory() emits
+                                    // historyChanged but never responseReceived/errorOccurred,
+                                    // so without this the next user-send onHistoryChanged would
+                                    // see _waitingForResponse left at true and treat the send
+                                    // as a response — scrolling to a stale _preResponseHeight
+                                    // instead of the bottom.
+                                    overlay._waitingForResponse = false
+                                    overlay._pendingScrollKind = ""
+                                    overlay._preResponseHeight = 0
                                     // Re-fetch historical context on background thread
                                     if (overlay.shotId > 0) {
                                         overlay.historicalContext = ""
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 3 issues:

1. **Clear button leaves the scroll state machine stuck in "waiting for response"** (real bug — next user send + response will scroll to a stale `_preResponseHeight` instead of bottom).

`AIConversation::clearHistory()` (`src/ai/aiconversation.cpp`) emits `historyChanged` but never `responseReceived` / `errorOccurred`. The new `onHistoryChanged` unconditionally sets `_waitingForResponse = true` whenever it sees an idle-side history change, so after Clear, `_waitingForResponse` stays `true` and `_preResponseHeight` is captured at the moment of the clear. The next user send sees `isResponse = true` (stuck flag), skips capturing a fresh `_preResponseHeight`, sets `_pendingScrollKind = "preResponse"` instead of `"bottom"`, and scrolls to the stale height. Same on the response that follows.

Fix: reset `_waitingForResponse`, `_preResponseHeight`, and `_pendingScrollKind` in the Clear `onClicked`, or special-case empty-history in `onHistoryChanged`.

https://github.com/Kulitorum/Decenza/blob/a1d7f380c8cabb67ccc4732c850773142524dfa9/qml/components/ConversationOverlay.qml#L786-L803

2. **`open()` still uses `Qt.callLater` for its scroll-to-bottom — same race the PR explicitly fixes elsewhere in this file.**

The PR's stated rationale is "`Qt.callLater` fires before the layout pass settles." That same `Qt.callLater` scroll is still in `open()`, so a re-opened conversation with prior history can land at the wrong position on mobile. CLAUDE.md root says "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'." Easiest fix: set `_pendingScrollKind = \"bottom\"` before `visible = true` and let `onContentHeightChanged` apply it.

https://github.com/Kulitorum/Decenza/blob/a1d7f380c8cabb67ccc4732c850773142524dfa9/qml/components/ConversationOverlay.qml#L54-L60

3. **`_pendingMessage` can survive an overlay close-while-dialog-still-open and fire `sendFollowUp()` on an invisible overlay.**

Sequence: tap Send (sets `_pendingMessage`, starts dialog close) → tap Back (overlay `onVisibleChanged` calls `inputDialog.close()` again, which is harmless, but `inputDialog.onClosed` still fires once the close animation finishes and runs `sendFollowUp()` against a now-hidden overlay). The `AIConversation` is global so the network request goes out anyway and the auto-save persists it, which is arguably acceptable, but the ConversationOverlay code path is mutating state on a dismissed view. Defensive fix: clear `_pendingMessage` in the overlay's `onVisibleChanged` block before calling `inputDialog.close()`.

https://github.com/Kulitorum/Decenza/blob/a1d7f380c8cabb67ccc4732c850773142524dfa9/qml/components/ConversationOverlay.qml#L48-L52

<details>

- [Worth a look] Comment in `onResponseReceived` ("...so there's nothing left to do here") is only true if `contentHeight` actually changed during the text= assignment; if it didn't, `_pendingScrollKind` lingers as `"preResponse"` and an unrelated future layout change will consume it. Either reset `_pendingScrollKind` in `onResponseReceived` defensively, or document the caveat. (score: 67)

</details>

Issue #921, opened 2026-04-29
AI advisor reports wrong observations on puck-failure shots (ShotSummarizer doesn't suppress on pourTruncated)


skialpine:
## Background

PR for #903 added a `pourTruncated` quality flag with a suppression cascade in the user-facing chip/verdict path: when a shot peaks below `PRESSURE_FLOOR_BAR`, the channeling / temp / grind detectors are forced to false because the curves they read off don't mean what they normally mean (conductance saturates → derivative flat, flow tracks preinfusion goal → grind delta ≈ 0, temp drift measured against a pour that didn't really happen).

The same suppression logic does **not** apply to the AI advisor's analysis path. `src/ai/shotsummarizer.cpp` computes its own `channelingDetected` and `temperatureUnstable` flags on the `ShotSummary` struct ([`src/ai/shotsummarizer.h:88-89`](https://github.com/Kulitorum/Decenza/blob/main/src/ai/shotsummarizer.h#L88-L89)) via independent code paths (`detectChannelingInPhases`, `calculateTemperatureStability`). Those flags drive the \"Observations\" section of the AI prompt at [`src/ai/shotsummarizer.cpp:694-702`](https://github.com/Kulitorum/Decenza/blob/main/src/ai/shotsummarizer.cpp#L694-L702):

```
## Observations

- **Puck integrity**: Sustained channeling in the conductance derivative (dC/dt) during extraction…
- **Temperature deviation**: Average temperature deviates from target by more than 2°C…
```

On a puck-failure shot (e.g. shot 868), neither of those observations is diagnostically meaningful — yet the AI advisor will receive them as authoritative input and dial-in advice will be wrong (chase temp settings, treat the shot as a channeling problem) when the actual signal is \"this shot has no information.\"

## What needs to change

Apply the same suppression cascade in `ShotSummarizer::summarize()` / `summarizeFromHistory()` that PR for #903 applied to `ShotAnalysis::generateSummary` and the storage path:

1. Detect `pourTruncated` early in the summarize pipeline (using `ShotAnalysis::detectPourTruncated` — same instrument as the rest of the suppression cascade).
2. When it fires, force `summary.channelingDetected = false` and `summary.temperatureUnstable = false` regardless of what the independent detectors compute.
3. Add a new `summary.pourTruncatedDetected` flag and emit a corresponding observation in the AI prompt:

[...]

Issue #903, opened 2026-04-27
Low-pressure shots aren't surfaced as badges (pourTruncated is summary-only; underpressurized-but-not-truncated has no detector)


skialpine:
## Background

Surfaced during the second 15-shot verification batch against the v1.7.2 prerelease (after PR #896). Two adjacent gaps in the badge system both involve shots that finish too easily — the puck didn't build proper extraction pressure but the detectors stay silent (or fire the wrong chip).

### Shot 868 — \`temperatureUnstable\` chip on what's actually a puck failure

- 18.1 g → 36 g target, **37.4 g yield in 7.27 s**, grind 7 on 80's Espresso.
- The shot never reached frame 2 (rise-and-hold). Peak pressure: **0.63 bar**. The water blew through preinfusion at the 7.5 mL/s flow goal and SAW stopped it at ~37 g.
- Badges fired: \`temperatureUnstable=true\` only. Reason: temp climbed from 77 °C through 81 °C across the 7 s shot, against an 82 °C goal — average deviation ~3 °C, over the 2 °C threshold.
- \`detectPourTruncated\` does fire (peak 0.63 < 2.5 bar floor), so the **summary** verdict correctly says \"Pour never pressurized — puck failed.\" But the user sees a \"Temp unstable\" *chip* — wrong diagnosis.
- PR #898's \`reachedExtractionPhase\` gate doesn't help here: frame 1 was reached + 4.7 s of pour, comfortably above the 1 s threshold. The shot ran long enough to score temp, just on a puck that wasn't actually doing anything.

### Shot 870 — silent grind-too-coarse

- 18.1 g → 36 g target, **37.5 g yield in 12.3 s**, grind 10 on 80's Espresso.
- Peak pressure **2.75 bar** — *just* above \`PRESSURE_FLOOR_BAR\` (2.5), so \`detectPourTruncated\` returns false.
- Choked-puck arms also silent: pressure ≥ 4 bar never lasted 15 s (never even reached 4 bar). Flow-vs-goal averaging during preinfusion shows delta ≈ 0 (flow tracked goal). Yield ratio is *over* target, not under, so the moderate-choke arm doesn't apply.
- All four badges fire **F**. The summary verdict probably lands on \"Decent shot with minor issues to watch\" or similar — wrong. This is textbook grind-too-coarse, and the user gets no signal.

## Two adjacent gaps

These cases together suggest two related improvements:

[...]

PR #922, opened 2026-04-29
feat(badges): add puckFailed chip + suppression cascade (#903)

skialpine:
## Summary

Closes the chip half of [#903](https://github.com/Kulitorum/Decenza/issues/903): low-pressure shots get a dedicated red **Puck failed** chip and the channeling/temp/grind flags are suppressed under it, so users no longer see contradictory \"Temp unstable\" or \"Clean extraction\" badges on what's actually a puck failure.

- New \`pour_truncated_detected\` column (migration 13), drift-on-load fixes 15 historical shots automatically on next read (8 of which currently fire wrong-diagnosis \"Temp unstable\")
- Suppression cascade applied in three places: save block, \`loadShotRecordStatic\` recompute, and \`ShotAnalysis::generateSummary\`
- New chip in \`QualityBadges.qml\`, gated against \"Clean extraction\" so a suppressed shot can't fall through to the green chip
- Summary verdict reworded to lead with the meta-action (\"Don't tune off this shot — peak pressure never built, so the other quality signals are unreliable\") and the warning line includes actual peak pressure

## Why this shape

Discussion on the issue landed on combining what was originally PR A (suppression) and PR B (badge column) — suppression alone would let the green \"Clean extraction\" chip fire on puck-failure shots, which is worse than the wrong-orange-chip status quo. The two halves are a single design.

Gap 2 from the issue (the 2.5–5 bar underpressurized band, ~4 shots / 0.4 % incidence) is intentionally **not** in this PR. See the issue thread for the rationale.

## Test plan

- [x] \`ctest\` — all 1769 pass (1766 pre-existing + 3 new \`pourTruncated_summary_*\` cascade tests)
- [x] \`tst_dbmigration\` — schema version bump 12→13 + new column-presence assertion
- [x] Existing \`shot_corpus_regression\` — both puck-failure fixtures (\`80s_puck_failure.json\`, \`puck_failure_short_gusher.json\`) still pass; \`londinium_gusher.json\` peak-9-bar fixture still shows channeling (suppression doesn't fire when \`pourTruncated\` is false)
- [ ] Manual smoke: open shot 1052 in history (the canonical puck-failure case from the issue: 80's Espresso, 18.1g → 37.4g in 7.27s, peak 0.63 bar) — should now show red \"Puck failed\" instead of orange \"Temp unstable.\" Confirm \`pour_truncated_detected\` flipped via SQLite.
- [ ] Manual smoke: open a clean shot (e.g. anything peaking at 9 bar) — should still show green \"Clean extraction\" with no regressions

## Out of scope

- \`ShotSummarizer\` (AI advisor prompt context path) computes its own \`channelingDetected\` / \`temperatureUnstable\` flags via an independent pipeline. Same suppression logic should apply there — tracked as [#921](https://github.com/Kulitorum/Decenza/issues/921) for a follow-up PR.

### DIFF
--- a/docs/SHOT_REVIEW.md
+++ b/docs/SHOT_REVIEW.md
@@ -42,18 +42,30 @@ shared `shotReview/advancedMode` setting toggles information density on both.
 
 ### Quality badges
 
-Four independent flags, surfaced via `qml/components/QualityBadges.qml`. When
-any fire, those chips show; when none fire, a single green "Clean extraction"
-chip shows. None of the chips suppress the others. A tappable "Shot
-Summary" chip always sits at the end of the row — it's the entry point to
-the analysis dialog described in §3.
+Five flags, surfaced via `qml/components/QualityBadges.qml`. When any fire,
+those chips show; when none fire, a single green "Clean extraction" chip
+shows. A tappable "Shot Summary" chip always sits at the end of the row —
+it's the entry point to the analysis dialog described in §3.
 
 | Flag                       | Color  | Label                  | Source                       |
 | -------------------------- | ------ | ---------------------- | ---------------------------- |
+| `pourTruncatedDetected`    | red    | "Puck failed"          | `detectPourTruncated`        |
 | `channelingDetected`       | red    | "Channeling detected"  | `detectChannelingFromDerivative` |
 | `temperatureUnstable`      | orange | "Temp unstable"        | `avgTempDeviation` + threshold |
 | `grindIssueDetected`       | orange | "Grind issue"          | `detectGrindIssue` (`analyzeFlowVsGoal`) |
 | `skipFirstFrameDetected`   | red    | "First step skipped"   | `detectSkipFirstFrame`       |
+
+**Suppression cascade.** `pourTruncatedDetected` is dominant: when it fires,
+`channelingDetected` / `temperatureUnstable` / `grindIssueDetected` are
+forced to false at save time, in the load-time recompute, and inside
+`generateSummary`. The puck failed to build pressure, so the curves the
+other three detectors read off don't mean what they normally mean
+(conductance saturates → derivative flat, flow tracks preinfusion goal →
+grind delta ≈ 0, temp drift measured against a pour that didn't really
+happen). `skipFirstFrameDetected` is **not** suppressed — it's a
+machine/profile issue orthogonal to puck integrity. The clean-extraction
+green chip's visibility gate also includes `!pourTruncatedDetected` so a
+suppressed puck-failure shot can't fall through to the wrong all-clear.
 
 The badges are recomputed on every shot load (see §4) so detector improvements
 take effect on existing shots without a manual re-analyze.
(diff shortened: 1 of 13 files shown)


skialpine:
### Code review

Found 2 issues:

1. `docs/SHOT_REVIEW.md` is not updated alongside the detector promotion. The root [`CLAUDE.md`](https://github.com/Kulitorum/Decenza/blob/8dd1c5fea2dc359d9bf68420f20254f570c9cb50/CLAUDE.md?plain=1#L156-L159) says: *"It is the source of truth for detector internals, gate semantics, and the recompute-on-load contract; keep it in sync when changing any of the above."* This PR moves `pourTruncatedDetected` from a summary-only observation to a fully persisted badge with a new DB column, suppression cascade, an extra `shotBadgesUpdated` parameter, and a chip — multiple sections of `SHOT_REVIEW.md` (the four-detector table, persistence semantics, the four-arg signal description, and §2.5 which explicitly states "not wired to a separate badge column") become stale.

https://github.com/Kulitorum/Decenza/blob/8dd1c5fea2dc359d9bf68420f20254f570c9cb50/CLAUDE.md?plain=1#L155-L160

2. `shouldSkipChannelingCheck` masks the gusher arm on turbo puck-failures (pre-existing, but the PR is actively touching the gate). When avg flow > 3 mL/s, the outer gate at [shothistorystorage.cpp:1013-1015](https://github.com/Kulitorum/Decenza/blob/8dd1c5fea2dc359d9bf68420f20254f570c9cb50/src/history/shothistorystorage.cpp#L1011-L1017) (and the parallel recompute path at [:2304-2306](https://github.com/Kulitorum/Decenza/blob/8dd1c5fea2dc359d9bf68420f20254f570c9cb50/src/history/shothistorystorage.cpp#L2302-L2308)) skips `detectGrindIssue` entirely. But `analyzeFlowVsGoal`'s yield-overshoot arm (added in #910) is documented to fire *"independently of any pressure/flow window because a gusher by definition can't sustain pressure"* — exactly the turbo case. So a turbo gusher whose peak just clears `PRESSURE_FLOOR_BAR` (e.g. peak 2.6–3 bar) silently gets `grindIssueDetected=false` and no chip. Pre-existing from #910, but the [`CLAUDE.md`](https://github.com/Kulitorum/Decenza/blob/8dd1c5fea2dc359d9bf68420f20254f570c9cb50/CLAUDE.md?plain=1#L573-L575) rule applies (*"Fix pre-existing violations in any file you touch"*). One-line fix: gate the turbo skip out of the yield-overshoot path inside `analyzeFlowVsGoal`, or hoist the gusher arm above the outer call.

https://github.com/Kulitorum/Decenza/blob/8dd1c5fea2dc359d9bf68420f20254f570c9cb50/src/ai/shotanalysis.cpp#L390-L405

---

<details>

- [Worth a look] `qml/components/QualityBadges.qml:18` — comment says *"forces the other four flags to false"* but only three are suppressed (channeling, temp, grind); `skipFirstFrameDetected` is intentionally not gated because it's a machine/profile issue independent of puck integrity. One-word fix: `four` → `three`. (score: 65)
- [Worth a look] No new fixture in `tests/data/shots/manifest.json` for the new "Don't tune off this shot" verdict text. Detector flag is already covered by `80s_puck_failure.json` and `puck_failure_short_gusher.json`; the PR's 3 new unit tests cover the verdict text inline, but `shot_eval` only validates booleans — so future verdict-text drift has no corpus regression guard. (score: 58)

</details>

Issue #929, opened 2026-04-29
Refactor: dedupe synthesized 'Extraction' fallback phase in ShotSummarizer


skialpine:
## Background

`src/ai/shotsummarizer.cpp` has two near-identical blocks that synthesize a single `PhaseSummary` named "Extraction" spanning the full shot duration when no phase markers are present. One lives in `summarize()` (live path, reads from `ShotDataModel`), the other in `summarizeFromHistory()` (history path, reads from the persisted `QVariantMap`). Each is ~30 lines computing the same per-phase metrics (avgPressure, maxPressure, minPressure, pressureAtStart/Middle/End; same for flow; avgTemperature; weightGained).

## Why this is duplicated

The two paths read from different curve sources:

- `summarize()` uses local references like `pressureData`, `flowData`, `tempData`, `cumulativeWeightData` (typed `QVector<QPointF>` from `ShotDataModel`).
- `summarizeFromHistory()` uses `summary.pressureCurve`, `summary.flowCurve`, etc. (already converted from the `QVariantMap`).

Factoring out the common logic needs either a template parameterized on the curve-source binding, or a static helper that takes the four curves + `totalDuration` by const-reference and returns a populated `PhaseSummary`. The latter is simpler and matches the rest of the file's helper style.

## Suggested refactor

Add a static helper in `shotsummarizer.cpp`:

```cpp
static PhaseSummary makeWholeShotPhase(
    const ShotSummarizer& self,
    const QVector<QPointF>& pressure,
    const QVector<QPointF>& flow,
    const QVector<QPointF>& temperature,
    const QVector<QPointF>& weight,
    double totalDuration);
```

…and call it from both summarize paths in place of the inline blocks.

(The `self` parameter is needed because `findValueAtTime` / `calculateAverage` / `calculateMin` / `calculateMax` are private non-static member functions on `ShotSummarizer`; alternatively make those helpers static and drop the `self` arg.)

## Why deferred

Identified during the cleanup pass for #921 (PR for #921 — ShotSummarizer suppression cascade + AI-prompt parity refactor). The synthesized block is rarely hit in practice (legacy shots predating phase-marker capture, or shots with no markers at all), and the refactor would have expanded the diff scope. Bookmarked here so it doesn't get lost.

## Files

- `src/ai/shotsummarizer.cpp` — `summarize()` (around the `if (markers.isEmpty())` branch) and `summarizeFromHistory()` (around the `if (summary.phases.isEmpty())` branch).

PR #930, opened 2026-04-29
fix(ai): share Shot Summary detector cascade with AI advisor (#921)

### LINKED ISSUE #921: AI advisor reports wrong observations on puck-failure shots (ShotSummarizer doesn't suppress on pourTruncated)

skialpine:
## Summary

Closes #921. ShotSummarizer (the AI advisor's prompt path) ran its own channeling/temperature detectors with no puck-failure suppression, so on a low-peak-pressure shot the prompt would receive "Puck integrity" and "Temperature deviation" observations the user never saw on the badge UI. The AI then dialed in against curves that didn't actually pour.

## Approach: share, don't replicate

Rather than re-implementing the suppression cascade inside ShotSummarizer, this PR has it **delegate** to `ShotAnalysis::generateSummary(...)` — the same static call `ShotHistoryStorage::generateShotSummary` already makes for the in-app Shot Summary dialog. The full cascade (pour truncated → channeling/temp/grind forced false) lives in exactly one place; future detector tweaks land in both consumers automatically.

The prompt now emits `generateSummary`'s line list verbatim under a `## Detector Observations` section, with a preamble that frames the lines as **detector evidence** rather than advice to parrot:

```
The lines below come from the same deterministic detectors that drive the
in-app Shot Summary dialog the user sees. Treat them as diagnostic signals
(evidence), not your conclusions. Severity tags reflect detector confidence,
not your final assessment:

- [warning] high-confidence failure mode (sustained channeling, choked puck, pour truncated, frame skip)
- [caution] directional hint (grind drift, flow trend, temp drift)
- [good] positive signal (puck stable)
- [observation] context (preinfusion drip mass)
```

The verdict ships in its own `## Dialog Verdict` section with explicit instruction not to repeat it. This brings the AI six previously-missing observation types: flow trend, preinfusion drip, grind direction, pour truncated, skip first frame, and the verdict itself.

## Cleanup pass

While in the file:

- Deleted dead `ShotSummary` fields (`timeToFirstDrip`, `preinfusionDuration`, `mainExtractionDuration`) and `PhaseSummary::tempStability` — written but never read by any consumer.
- Deleted the helpers that only fed those fields (`findTimeToFirstDrip`, `calculateStdDev`).
- Deleted the two wrapper functions (`detectChannelingInPhases`, `calculateTemperatureStability`) — their job is now `generateSummary`'s.
- Consolidated the duplicate `profileJson` parse in `summarizeFromHistory` (parsed three times → once).
- Built `HistoryPhaseMarker` list in the same pass as `PhaseSummary` instead of re-iterating.
- Switched `summarize()` to typed `phaseMarkersList()` to drop a `QVariantMap` round-trip.
- Trimmed the `shothistorystorage.h` include to `shothistory_types.h` (only `HistoryPhaseMarker` was needed).

Net diff: +444 / -167 across the changed files; the cleanup itself saved ~50 LoC under the new preamble + delegation.

## Out of scope (filed)

#929 — the synthesized "Extraction" fallback phase (when no markers exist) is still duplicated between `summarize()` and `summarizeFromHistory()`. Different curve sources; would need a static helper or template. Bookmarked.

## Doc updates

- `CLAUDE.md` line 158: "four quality-badge detectors" → "five" + added "pour truncated" to the parenthetical (the `pourTruncatedDetected` badge from PR #922 was missing from the index sentence).

## Test plan

- [x] `tst_ShotSummarizer` (new) — 2 cases: (1) low-peak shot fires `pourTruncatedDetected`, suppresses channeling/temp lines, prompt contains "Pour never pressurized" but not "Puck integrity" / "Temperature deviation"; (2) healthy 9-bar shot leaves the flag false and renders both Observations + Verdict sections.
- [x] `tst_ShotAnalysis` — 41 / 41 pass (refactor didn't perturb the underlying heuristics).
- [x] `shot_corpus_regression` — 14 / 14 fixtures pass (lever-clean, puck-failure, gusher, choked, frame-skip).
- [x] Full debug build clean (0 errors, 0 warnings).
- [ ] Manual: open a known puck-failure shot in the AI advisor and confirm the prompt now leads with "Puck failed" + "Don't tune off this shot" rather than channeling/temp.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,7 +155,7 @@ Detailed documentation lives in `docs/CLAUDE_MD/`. Read these when working in th
 | `AI_ADVISOR.md` | AI dialing assistant design |
 | `SETTINGS.md` | Settings architecture: 7 domain sub-objects, how to add properties/domains, QML access pattern, build-blast rules |
 
-Read [`docs/SHOT_REVIEW.md`](https://github.com/Kulitorum/Decenza/blob/main/docs/SHOT_REVIEW.md) when working on the post-shot review / shot detail pages, the four quality-badge detectors (channeling, grind issue, temperature unstable, skip-first-frame), the Shot Summary dialog, badge persistence, or `src/ai/shotanalysis.{h,cpp}`. It is the source of truth for detector internals, gate semantics, and the recompute-on-load contract; keep it in sync when changing any of the above.
+Read [`docs/SHOT_REVIEW.md`](https://github.com/Kulitorum/Decenza/blob/main/docs/SHOT_REVIEW.md) when working on the post-shot review / shot detail pages, the five quality-badge detectors (pour truncated, channeling, grind issue, temperature unstable, skip-first-frame), the Shot Summary dialog, badge persistence, or `src/ai/shotanalysis.{h,cpp}`. It is the source of truth for detector internals, gate semantics, and the recompute-on-load contract; keep it in sync when changing any of the above.
 
 ## Development Environment
 
--- a/docs/SHOT_REVIEW.md
+++ b/docs/SHOT_REVIEW.md
@@ -419,6 +419,23 @@ take effect on summary text the same way they do on badges — no save-time
 freeze. There is no DB persistence for summary lines; they're regenerated
 on demand.
 
+### AI advisor consumes the same line list (PR #930)
+
+The in-app AI advisor's prompt is built by `ShotSummarizer::buildUserPrompt`,
+which ships the same `generateSummary` line list under a `## Detector
+Observations` section with a preamble framing the lines as detector
+evidence (severity tags `[warning]` / `[caution]` / `[good]` /
+`[observation]`). The `verdict` line is filtered out so the AI reasons
+from the same observations the verdict was built from rather than
+anchoring on the dialog's pre-cooked conclusion. The suppression cascade
+is enforced in exactly one place — `generateSummary` — so the badge UI,
+the dialog, and the AI advisor cannot drift.
+
+External MCP-connected agents currently see only the cascaded boolean
+flags via `convertShotRecord`, not the formatted line list. See
+Issue #931 for the design discussion (full parity vs intentional
+data-plane layering).
+
 ---
 
 ## 4. Persistence semantics
@@ -600,8 +617,27 @@ To add a fixture:
 - PR #922 / Issue #903 — fifth badge `pourTruncatedDetected` ("Puck failed"),
   suppression cascade across save / load / `generateSummary`,
   meta-action verdict ("Don't tune off this shot"), migration 13.
-- Issue #921 — `ShotSummarizer` (AI advisor prompt path) does not yet
-  share the suppression cascade; tracked as a follow-up to PR #922.
+- PR #930 / Issue #921 — `ShotSummarizer` (AI advisor prompt path) now
+  shares the suppression cascade. Detector orchestration delegates to
+  `ShotAnalysis::generateSummary`, the same call `ShotHistoryStorage::generateShotSummary`
+  makes for the dialog. The prompt's `## Detector Observations` section
+  emits `generateSummary`'s line list verbatim with severity tags
+  (`[warning]` / `[caution]` / `[good]` / `[observation]`) under a preamble
+  framing the lines as detector evidence. The `verdict` line is filtered
+  out before emission so the AI reasons from the same observations the
+  user sees in the dialog without anchoring on the dialog's prescriptive
+  conclusion. Per-phase `PhaseSummary::temperatureUnstable` markers are
+  gated on the same `reachedExtractionPhase` + `pourTruncatedDetected`
+  cascade the aggregate temp detector uses. Cleanup: dropped dead
+  `ShotSummary` fields (`channelingDetected`, `temperatureUnstable`,
+  `timeToFirstDrip`, `preinfusionDuration`, `mainExtractionDuration`)
+  and the wrapper helpers (`detectChannelingInPhases`,
+  `calculateTemperatureStability`) they fed.
+- Issue #931 — MCP `shots_get_detail` / `shots_compare` still expose
+  only the boolean badge flags (cascade applied), not the formatted
+  `summaryLines`. Followup to #921 / #930; design call pending between
+  full parity (Option A: emit lines on the MCP payload) and "data
+  plane" layering (Option B: external agents do their own analysis).
 
 External resources that informed the diagnostic patterns:
 
--- a/src/ai/shotsummarizer.cpp
+++ b/src/ai/shotsummarizer.cpp
@@ -1,9 +1,9 @@
 #include "shotsummarizer.h"
 #include "shotanalysis.h"
-#include "../history/shothistorystorage.h"  // HistoryPhaseMarker — passed to ShotAnalysis mode-aware detectors
+#include "../history/shothistory_types.h"  // HistoryPhaseMarker — passed to ShotAnalysis::generateSummary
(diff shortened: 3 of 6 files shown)


skialpine:
### Code review

Found 3 issues at score 75+, all fixed in [a18b11ef](https://github.com/Kulitorum/Decenza/commit/a18b11ef329db79a5489d217d2b2cebc125c95d8):

1. **`tst_shotsummarizer.cpp` used `QTEST_APPLESS_MAIN`** — `docs/CLAUDE_MD/TESTING.md` prescribes `QTEST_GUILESS_MAIN`, and every other test in the project follows that convention. The test links `Qt::Quick` + `Qt::Charts` transitively, which expect a `QCoreApplication`. Fixed.

   https://github.com/Kulitorum/Decenza/blob/38e7df71b1104a7ed2eccfee46f0b36375d6d997/tests/tst_shotsummarizer.cpp#L218-L222

2. **`markPerPhaseTempInstability` was missing the `reachedExtractionPhase` gate** — `ShotAnalysis::generateSummary` got that gate in PR #898 (so the aggregate "Temperature drifted X°C" caution correctly stays silent on aborted-during-preinfusion shots), but the per-phase prompt block populated by this helper bypassed it. Result: the AI prompt's per-phase markers would emit "Temperature instability" on the preheat ramp while the aggregate observation correctly suppressed. Fixed by gating both call sites and adding a regression test (`abortedPreinfusionDoesNotFlagPerPhaseTemp`).

   https://github.com/Kulitorum/Decenza/blob/38e7df71b1104a7ed2eccfee46f0b36375d6d997/src/ai/shotsummarizer.cpp#L86-L101

3. **`docs/SHOT_REVIEW.md` was stale** — root `CLAUDE.md` says "keep it in sync when changing... the Shot Summary dialog, badge persistence, or `src/ai/shotanalysis.{h,cpp}`." Lines 603-604 still tracked #921 as an open follow-up to PR #922. Fixed: §3 grows an "AI advisor consumes the same line list" subsection; §7 closes #921 and adds #931 (MCP parity follow-up).

   https://github.com/Kulitorum/Decenza/blob/38e7df71b1104a7ed2eccfee46f0b36375d6d997/docs/SHOT_REVIEW.md#L600-L605

Also folded a separate concern raised in the discussion thread: the `## Dialog Verdict` section in the AI prompt was anchoring the LLM on the dialog's prescriptive conclusion. Dropped from the prompt; the AI now reasons from the same observations the verdict was built from. Tests updated to assert the verdict text does not leak.

<details>
<summary>5 below-threshold issues (50)</summary>

- [Worth a look] `computePourWindow` comment claimed "Matches" generateSummary's fallback chain — the iteration semantics differ in edge cases. Reworded to "Approximates the phase-boundary logic" since the function's only caller is the per-phase temp gate where exact match isn't required (score: 50, fixed)
- [Worth a look] Unified marker/phase loop comment said "Keeps the two derived views in lockstep" — the lists can diverge when degenerate phases exist. Reworded to clarify they're consumed by different code paths and never joined by index (score: 50, fixed)
- [Worth a look] `yieldOverride` comment mentioned only the choked-puck arm — the gusher arm (PR #910) also uses `targetWeightG`. Comment expanded (score: 50, fixed)
- [Worth a look] Preamble `[warning]` legend omitted "yield overshoot/gusher" — `generateSummary` emits it as `type=warning`. Legend extended (score: 50, fixed)
- [OK to skip] `[observation]` legend describes only "preinfusion drip mass" but `generateSummary` also has a "Not enough data" early-return path — that path returns before the prompt section is rendered, so the legend gap doesn't surface to the AI in practice (score: 50)

</details>

Verified: `tst_ShotAnalysis` 41/41, `tst_ShotSummarizer` 4/4 (3 cases × QTest harness), `shot_corpus_regression` 14/14 — no detector drift.

PR #932, opened 2026-04-29
refactor(ai): dedupe Extraction fallback in ShotSummarizer (#929)

### LINKED ISSUE #929: Refactor: dedupe synthesized 'Extraction' fallback phase in ShotSummarizer

skialpine:
Closes #929.

## Summary
- Extract a `makeWholeShotPhase()` static helper that builds a single full-shot `PhaseSummary` from the four curves + duration.
- Replace the two near-identical ~30-line inline blocks in `summarize()` (live path) and `summarizeFromHistory()` (history path) with a single call to the helper.
- Flip `findValueAtTime` / `calculateAverage` / `calculateMin` / `calculateMax` to `static` — they don't read member state, and that lets the new helper call them without taking a `self`/instance arg.

No behavior change: the helper produces identical `PhaseSummary` values to the inline code it replaces. The synthesized-Extraction path is rarely hit in practice (legacy shots predating phase-marker capture, or shots aborted before frame 0 emitted).

## Test plan
- [x] Build clean (Qt Creator MCP)
- [x] Existing test suite green (1774 pre-existing tests, 0 failures)
- [x] `tst_ShotSummarizer` covers the live-path summary code that calls the new helper

### DIFF
--- a/src/ai/shotsummarizer.cpp
+++ b/src/ai/shotsummarizer.cpp
@@ -58,6 +58,42 @@ QString ShotSummarizer::profileTypeDescription(const QString& editorType)
     return QString();
 }
 
+PhaseSummary ShotSummarizer::makeWholeShotPhase(const QVector<QPointF>& pressure,
+                                                const QVector<QPointF>& flow,
+                                                const QVector<QPointF>& temperature,
+                                                const QVector<QPointF>& weight,
+                                                double totalDuration)
+{
+    PhaseSummary phase;
+    phase.name = QStringLiteral("Extraction");
+    phase.startTime = 0;
+    phase.endTime = totalDuration;
+    phase.duration = totalDuration;
+
+    phase.avgPressure = calculateAverage(pressure, 0, totalDuration);
+    phase.maxPressure = calculateMax(pressure, 0, totalDuration);
+    phase.minPressure = calculateMin(pressure, 0, totalDuration);
+    phase.pressureAtStart = findValueAtTime(pressure, 0);
+    phase.pressureAtMiddle = findValueAtTime(pressure, totalDuration / 2);
+    phase.pressureAtEnd = findValueAtTime(pressure, totalDuration);
+
+    phase.avgFlow = calculateAverage(flow, 0, totalDuration);
+    phase.maxFlow = calculateMax(flow, 0, totalDuration);
+    phase.minFlow = calculateMin(flow, 0, totalDuration);
+    phase.flowAtStart = findValueAtTime(flow, 0);
+    phase.flowAtMiddle = findValueAtTime(flow, totalDuration / 2);
+    phase.flowAtEnd = findValueAtTime(flow, totalDuration);
+
+    phase.avgTemperature = calculateAverage(temperature, 0, totalDuration);
+
+    if (!weight.isEmpty()) {
+        const double startWeight = findValueAtTime(weight, 0);
+        const double endWeight = findValueAtTime(weight, totalDuration);
+        phase.weightGained = endWeight - startWeight;
+    }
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

<details>
<summary>3 below-threshold issues (50)</summary>

- [OK to skip] Header comment at [shotsummarizer.h#L156-L160](https://github.com/Kulitorum/Decenza/blob/3e518f20094b727f670e748206929eda149e9e76/src/ai/shotsummarizer.h#L156-L160) frames the `static` promotion as "without a ShotSummarizer instance" — the deeper reason is that `makeWholeShotPhase` is itself static and a static member can't call non-static members regardless of instance availability. Wording-only nit. (score: 50)
- [OK to skip] `makeWholeShotPhase` doc at [shotsummarizer.h#L165-L174](https://github.com/Kulitorum/Decenza/blob/3e518f20094b727f670e748206929eda149e9e76/src/ai/shotsummarizer.h#L165-L174) describes the trigger as "no phase markers" but `summarizeFromHistory()` also reaches this path when markers exist with degenerate spans (`endTime <= startTime`). Doc accuracy nit. (score: 50)
- [OK to skip] The new helper adds an `if (!weight.isEmpty())` guard around `weightGained` that the previous `summarize()` inline block didn't have. Behavior is unchanged because `findValueAtTime` returns 0 on empty input — the guard just unifies the two paths to the more defensive form. Doc nit. (score: 50)

</details>

PR #939, opened 2026-04-30
perf(history): cache analyzeShot result on ShotRecord (D)

skialpine:
Implements OpenSpec change [\`cache-analyzeshot-on-shotrecord\`](https://github.com/Kulitorum/Decenza/blob/3baa8cf8/openspec/changes/cache-analyzeshot-on-shotrecord/proposal.md). First of three D/E/F follow-ups from the parity audit.

## Summary
- Add \`std::optional<ShotAnalysis::AnalysisResult> cachedAnalysis\` to \`ShotRecord\`.
- \`loadShotRecordStatic\` populates it (via \`std::move\`) after the badge projection.
- \`convertShotRecord\` reads from the cache when present; falls back to running \`analyzeShot\` inline for direct-construction callers (\`ShotHistoryExporter\`, tests).
- Halves detector CPU per shot detail load — eliminates the dual analyzeShot pass on the canonical \`requestShot\` → \`shotReady\` path.

## Why
\`loadShotRecordStatic\` and \`convertShotRecord\` ran \`analyzeShot\` independently on the same \`ShotRecord\` in immediate succession. Two passes for the same answer. Cache the first result; second call reuses it.

## Test plan
- [x] Build clean (Qt Creator MCP, 0 errors / 0 warnings)
- [x] All 1802 tests pass (1797 prior + 3 new methods covering cached / fallback / equivalence — QCOMPARE-in-loop expansion accounts for the +5 reported)
- [x] \`cachedAnalysis_isReusedByConvert\` — sentinel-based, fails if recomputation fires
- [x] \`noCachedAnalysis_fallsBackToInlineAnalyzeShot\` — direct-construction path still produces complete output
- [x] \`cachedAndFallback_paths_produceEquivalentOutput\` — byte-equal output across paths

## Risk surface
The cache is invalidated if any input curve on the \`ShotRecord\` is mutated post-load. Documented inline in the new field's docstring. Today no caller does that — \`requestShot\`'s callback doesn't mutate \`ShotRecord\` between \`loadShotRecordStatic\` and \`convertShotRecord\`.

### DIFF
--- a/docs/SHOT_REVIEW.md
+++ b/docs/SHOT_REVIEW.md
@@ -321,8 +321,12 @@ There are two consumer paths that share a single detector pass:
   of `{ text, type }` lines rendered by a `Repeater` with a colored dot
   per line.
 - **MCP path** (returns prose + structured detectors):
-  `convertShotRecord` → `ShotAnalysis::analyzeShot(...)` →
-  `AnalysisResult { lines, detectors }` → emitted as `summaryLines` plus a
+  `convertShotRecord` → reads `record.cachedAnalysis` populated by
+  `loadShotRecordStatic`'s earlier `analyzeShot` pass (single computation
+  per detail load). Falls back to `ShotAnalysis::analyzeShot(...)` inline
+  only for direct-construction callers (`ShotHistoryExporter`, tests) that
+  bypass `loadShotRecordStatic`. Either way the result is an
+  `AnalysisResult { lines, detectors }` emitted as `summaryLines` plus a
   nested `detectorResults` JSON object on every shot record served by
   `shots_get_detail` / `shots_compare`.
 
--- a/openspec/changes/cache-analyzeshot-on-shotrecord/tasks.md
+++ b/openspec/changes/cache-analyzeshot-on-shotrecord/tasks.md
@@ -2,34 +2,35 @@
 
 ## 1. Cache field on ShotRecord
 
-- [ ] 1.1 Add `std::optional<ShotAnalysis::AnalysisResult> cachedAnalysis;` to `ShotRecord` in `src/history/shothistory_types.h`. Document that it is populated by `loadShotRecordStatic` after the badge projection and consumed by `convertShotRecord`; reset to `std::nullopt` if any of the input curves are mutated after load.
-- [ ] 1.2 Confirm `<optional>` is included (or add to the header). `ShotAnalysis::AnalysisResult` is already public.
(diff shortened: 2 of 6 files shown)


skialpine:
### Code review

No high-confidence bugs found. 3 doc/style nits applied in 5e3f1db2:

1. **Include ordering** — root [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/5e3f1db2a795ac528d2e6ea1fad271769485690e/CLAUDE.md) doesn't explicitly mandate it, but the file's pre-existing convention was Qt/std headers first, then project-local. \`#include "ai/shotanalysis.h"\` is now placed after \`<optional>\`.

   https://github.com/Kulitorum/Decenza/blob/5e3f1db2a795ac528d2e6ea1fad271769485690e/src/history/shothistory_types.h#L1-L13

2. **Stale "ShotHistoryExporter" example** in the slow-path comment — the exporter actually goes through \`loadShotRecordStatic\` (see \`writeShotJson\`), so it hits the cache, not the fallback. Removed the misleading example from both the field docstring and the convertShotRecord comment.

   https://github.com/Kulitorum/Decenza/blob/5e3f1db2a795ac528d2e6ea1fad271769485690e/src/history/shothistory_types.h#L120-L126

3. **Test "by construction" wording** in \`cachedAndFallback_paths_produceEquivalentOutput\` — the two records are independent, not derived from a shared intermediate. Reworded to reflect the actual basis (deterministic analyzeShot + identical seed).

   https://github.com/Kulitorum/Decenza/blob/5e3f1db2a795ac528d2e6ea1fad271769485690e/tests/tst_shotrecord_cache.cpp#L146-L152

**Verified clean:**
- Cache lifecycle: \`applyBadgesToTarget\` reads detectors by const-ref before \`std::move(analysis)\`. Safe.
- Pointer-into-optional aliasing: \`analysisPtr\` is dereferenced inside the same scope as the optional's lifetime. Safe.
- Fallback scope: \`analysisOwned\` outlives every dereference of \`analysisPtr\`. Safe.
- No circular include from \`shothistory_types.h\` → \`ai/shotanalysis.h\`. Header coupling cost noted but acceptable.

<details>
<summary>1 below-threshold issue (50)</summary>

- [Worth a look] \`convertShotRecord\` fallback inlines \`getAnalysisFlags\` + \`profileFrameInfoFromJson\` instead of calling \`prepareAnalysisInputs\`. The helper lives in PR #941; once both merge, a small follow-up commit can collapse the fallback to one helper call. Not a blocker — fallback only triggers in tests today. (score: 50)

</details>

PR #944, opened 2026-04-30
refactor(ai): expose analyzeShot pour window on DetectorResults (H)

skialpine:
## Summary

- Adds `pourStartSec` / `pourEndSec` to `ShotAnalysis::DetectorResults`, populated by `analyzeShot` from the same phase-boundary range it already uses to gate every other detector.
- Switches both `ShotSummarizer::summarize()` and `summarizeFromHistory()` (slow branch) from `generateSummary` + a separate `detectPourTruncated` to one `analyzeShot` call, reading lines and `pourTruncated` from the returned `AnalysisResult`.
- Deletes the file-static `computePourWindow` helper — it was a duplicate of analyzeShot's internal logic kept just to gate `markPerPhaseTempInstability`.
- Surfaces the window through MCP `shots_get_detail` (additive `detectorResults.pourStartSec` / `pourEndSec`) and documents the new fields in `docs/CLAUDE_MD/MCP_SERVER.md`.
- Implements OpenSpec proposal `expose-pour-window-on-analysis-result` (proposal H of the G–K series).

## Why

`computePourWindow` re-derived analyzeShot's pour-window logic with slightly different fallbacks. If analyzeShot's heuristic ever tightened (new phase labels, different precedence) the gate that decides whether `markPerPhaseTempInstability` runs would silently disagree with the rest of the cascade. Easier to expose the value `analyzeShot` already computed than to keep a parallel implementation honest.

## Test plan

- [x] Added `analyzeShot_pourWindow_matchesPhaseBoundaries` — preinfusion + pour + end markers, asserts `pourStartSec == 8.0`, `pourEndSec == 28.0`.
- [x] Added `analyzeShot_pourWindow_noMarkers_spansWholeShot` — no markers, asserts whole-shot fallback.
- [x] Added `analyzeShot_pourWindow_preinfusionOnly_usesPreinfusionBoundary` — preinfusion-only, asserts pourStart picks up the preinfusion boundary.
- [x] Added `analyzeShot_pourWindow_insufficientData_defaultsToZero` — early-return path leaves both at `0.0`.
- [x] Existing `tst_ShotSummarizer::abortedPreinfusionDoesNotFlagPerPhaseTemp` still passes (rewired gate behaves identically).
- [x] Build clean, 0 warnings.
- [x] Full suite: 1806 passed, 0 failed (Qt Creator MCP).

### DIFF
--- a/docs/CLAUDE_MD/MCP_SERVER.md
+++ b/docs/CLAUDE_MD/MCP_SERVER.md
@@ -298,8 +298,10 @@ Both are populated by a single call to `ShotAnalysis::analyzeShot`, so they cann
   },
   "pourTruncated": false,
   "peakPressureBar": 9.1,        // present only when pourTruncated == true
-  "pourStartSec": 6.2,           // phase-boundary window analyzeShot used to gate every other detector
-  "pourEndSec": 32.8,            // both 0.0 when phase markers were absent and the no-data fallback ran
+  "pourStartSec": 6.2,           // phase-boundary window analyzeShot used to gate every other detector;
+                                 //   stays 0.0 when no "preinfusion"/"pour" markers are present (whole-shot fallback)
+  "pourEndSec": 32.8,            // defaults to shot duration when no "end" marker is present;
+                                 //   both fields are 0.0 only on the insufficient-data early return (pressure.size() < 10)
   "skipFirstFrame": false,
   "verdictCategory": "minorIssuesGrindFine"
 }
--- a/docs/CLAUDE_MD/MCP_SERVER.md
+++ b/docs/CLAUDE_MD/MCP_SERVER.md
@@ -298,6 +298,8 @@ Both are populated by a single call to `ShotAnalysis::analyzeShot`, so they cann
   },
   "pourTruncated": false,
   "peakPressureBar": 9.1,        // present only when pourTruncated == true
+  "pourStartSec": 6.2,           // phase-boundary window analyzeShot used to gate every other detector
+  "pourEndSec": 32.8,            // both 0.0 when phase markers were absent and the no-data fallback ran
   "skipFirstFrame": false,
   "verdictCategory": "minorIssuesGrindFine"
 }
--- a/openspec/changes/expose-pour-window-on-analysis-result/tasks.md
+++ b/openspec/changes/expose-pour-window-on-analysis-result/tasks.md
@@ -2,26 +2,30 @@
 
 ## 1. DetectorResults extension
 
-- [ ] 1.1 Add `double pourStartSec = 0.0;` and `double pourEndSec = 0.0;` fields to `ShotAnalysis::DetectorResults`. Default `0.0` matches the function's local-variable initial values when no phase markers are present.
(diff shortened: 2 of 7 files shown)


skialpine:
### Code review

Found 3 issues:

1. `MCP_SERVER.md`'s `pourEndSec` comment is factually wrong — it says "both 0.0 when phase markers were absent and the no-data fallback ran", but the new test `analyzeShot_pourWindow_noMarkers_spansWholeShot` asserts `pourEndSec == 30.0` (= duration) for that exact case. Both fields are `0.0` only on the `pressure.size() < 10` early-return path. An MCP consumer that follows the doc will mis-classify legacy shots that have no phase markers as "no analysis was possible" when in fact the whole-shot fallback ran successfully.

https://github.com/Kulitorum/Decenza/blob/e0a045c0f56f7b98c825153c9f5f7051e24319e7/docs/CLAUDE_MD/MCP_SERVER.md#L300-L303

2. The new `DetectorResults` block comment lists `markPerPhaseTempInstability` as a consumer of `pourStartSec`/`pourEndSec` ("downstream consumers (the per-phase temperature instability check in ShotSummarizer, MCP consumers, regression tests) read the same window the cascade used"). Post-PR `markPerPhaseTempInstability` iterates `summary.phases` directly and never touches the new fields — only the cascade gate (`!pourTruncatedDetected`) couples ShotSummarizer to the analyzer. The header overstates what the field exposure achieved; an inline comment in `summarize()` even contradicts it ("`markPerPhaseTempInstability` iterates summary.phases directly — it doesn't need a pour window").

https://github.com/Kulitorum/Decenza/blob/e0a045c0f56f7b98c825153c9f5f7051e24319e7/src/ai/shotanalysis.h#L356-L367

3. Stale "via the generateSummary wrapper" comments at two sites in `shotsummarizer.cpp`. Both call sites now invoke `analyzeShot` directly (and read `analysis.detectors.pourTruncated` from the result, so the "this caller only needs the prose lines" rationale is also wrong).

https://github.com/Kulitorum/Decenza/blob/e0a045c0f56f7b98c825153c9f5f7051e24319e7/src/ai/shotsummarizer.cpp#L255-L259

https://github.com/Kulitorum/Decenza/blob/e0a045c0f56f7b98c825153c9f5f7051e24319e7/src/ai/shotsummarizer.cpp#L475-L478

<details>

- [Worth a look] "Default 0.0 means 'not set'" wording in the header (lines 363-365) and the test docstring at line 1363 conflate "no phase data" with the `pressure.size() < 10` early return. The two cases are distinguishable only by `pourEndSec` — making "0.0 means not set" ambiguous as written. (score: 65)
- [Worth a look] Fast path of `summarizeFromHistory` doesn't repopulate `pourStartSec`/`pourEndSec` for shots saved before the PR — they'll come back as `0.0/0.0` from `detectorResults` blobs that don't carry the new keys. Additive, but worth deciding whether to back-fill or document. (score: 55)
- [Worth a look] `analyzeShot_pourWindow_noMarkers_spansWholeShot` asserts `pourEndSec == 30.0` (a literal) instead of `pourEndSec == duration` (the parameter) — if `pourEnd`'s default ever changes from `duration` to `pressure.last().x()` and the two coincidentally match, the test still passes. (score: 45)
- [Worth a look] `markPerPhaseTempInstability` is documented as a consumer of the new fields but the gate it actually uses (`!pourTruncatedDetected`) is unchanged. If the intent is for it to read the window directly, that wiring needs a follow-up. (score: 55)

</details>

PR #934, opened 2026-04-30
perf(dialog): dedup Shot Summary by reading pre-computed summaryLines (A)

skialpine:
Implements OpenSpec change [`dedup-shot-summary-dialog`](https://github.com/Kulitorum/Decenza/blob/c13f355d/openspec/changes/dedup-shot-summary-dialog/proposal.md). First of three follow-ups (A/B/C) from the parity audit on PR #933.

## Summary
- `ShotAnalysisDialog.qml` now reads `shotData.summaryLines` directly when present, falling back to `MainController.shotHistory.generateShotSummary(shotData)` only for legacy `shotData` maps that didn't flow through `convertShotRecord`.
- `convertShotRecord` (added in PR #933) already runs `analyzeShot()` per shot conversion and stores the prose lines in `summaryLines`. This change saves one full detector pass per dialog open.
- Both `ShotDetailPage.qml` and `PostShotReviewPage.qml` instantiate the dialog component, so the fix applies to both via the dialog's own binding — no per-page edits needed.
- Both fast and fallback paths ultimately call `analyzeShot`, so the rendered output is byte-identical by construction.

## Test plan
- [x] Build clean (Qt Creator MCP, 0 errors / 0 warnings)
- [x] All 1778 existing tests pass; the QML binding has no C++ test surface.
- [x] `analyzeShot` and `convertShotRecord` correctness already locked in by `tst_shotanalysis` (PR #933).
- [ ] Manual smoke (when reviewer is at the machine): open Shot Detail on a clean shot, a chokedPuck shot, and a pourTruncated shot — lines render identically to pre-change behavior.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #945, opened 2026-04-30
refactor(ai): extract runShotAnalysisAndPopulate helper (I)

skialpine:
## Summary

- Extracts `ShotSummarizer::runShotAnalysisAndPopulate(...)` to consolidate the detector orchestration tail (analyzeShot call + summaryLines copy + pourTruncatedDetected derivation + per-phase temp gate) shared by `summarize()` and the slow path of `summarizeFromHistory()`.
- Both call sites shrink to a single helper invocation; the orchestration shape is now defined in exactly one place.
- Implements OpenSpec proposal `dedup-shotsummarizer-input-adapter` (proposal I in the G–K series).

## Stacked on

This PR targets [`expose/pour-window-on-analysis-result`](https://github.com/Kulitorum/Decenza/pull/944) (H). When H merges to main, this PR's base auto-rebases.

## Why

After PR #944 (H) eliminated `computePourWindow`, the tail of `summarize()` and `summarizeFromHistory()` slow path became structurally identical: call `analyzeShot`, copy lines, derive `pourTruncatedDetected`, conditionally run `markPerPhaseTempInstability`. Two copies of the same orchestration is two places future cascade tweaks have to land — one place is enough.

The fast path of `summarizeFromHistory` (pre-computed `summaryLines`, PR #934) stays inline. It consumes a `QVariantMap` rather than typed `AnalysisResult` data; routing it through the helper via a `cachedAnalysis` parameter would re-introduce a parallel shape inside the helper itself, defeating the dedup goal.

## Test plan

- [x] Existing 1806 tests pass — the helper is exercised by both call sites, so the slow-path coverage doubles as a regression lock for the orchestration shape.
- [x] `pourTruncatedSuppressesChannelingAndTempLines`, `abortedPreinfusionDoesNotFlagPerPhaseTemp`, `summarizeFromHistory_fastAndSlowPathsAgree`, and `summarizeFromHistory_fastPathPreservesPourTruncatedCascade` all green.
- [x] Build clean, 0 warnings.

A direct live-path equivalence test (`summarize()` vs `summarizeFromHistory()` on the same shot) is scoped to proposal K, which adds the MockShotDataModel infrastructure direct `summarize()` tests require.

### DIFF
--- a/docs/CLAUDE_MD/MCP_SERVER.md
+++ b/docs/CLAUDE_MD/MCP_SERVER.md
@@ -298,8 +298,10 @@ Both are populated by a single call to `ShotAnalysis::analyzeShot`, so they cann
   },
   "pourTruncated": false,
   "peakPressureBar": 9.1,        // present only when pourTruncated == true
-  "pourStartSec": 6.2,           // phase-boundary window analyzeShot used to gate every other detector
-  "pourEndSec": 32.8,            // both 0.0 when phase markers were absent and the no-data fallback ran
+  "pourStartSec": 6.2,           // phase-boundary window analyzeShot used to gate every other detector;
+                                 //   stays 0.0 when no "preinfusion"/"pour" markers are present (whole-shot fallback)
+  "pourEndSec": 32.8,            // defaults to shot duration when no "end" marker is present;
+                                 //   both fields are 0.0 only on the insufficient-data early return (pressure.size() < 10)
   "skipFirstFrame": false,
   "verdictCategory": "minorIssuesGrindFine"
 }
--- a/docs/CLAUDE_MD/MCP_SERVER.md
+++ b/docs/CLAUDE_MD/MCP_SERVER.md
@@ -298,6 +298,8 @@ Both are populated by a single call to `ShotAnalysis::analyzeShot`, so they cann
   },
   "pourTruncated": false,
   "peakPressureBar": 9.1,        // present only when pourTruncated == true
+  "pourStartSec": 6.2,           // phase-boundary window analyzeShot used to gate every other detector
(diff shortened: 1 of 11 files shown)


skialpine:
### Code review

Found 1 issue:

1. Slow-path comment "Fall through to the inline detector orchestration" is now actively false — the slow path calls `runShotAnalysisAndPopulate(...)`, no inline orchestration remains. The cross-reference "see summarize() for rationale" on the next line points at the same now-obsolete comment in `summarize()`.

https://github.com/Kulitorum/Decenza/blob/c734e7bf615ae6ddd49ae339f18e42ab523058a1/src/ai/shotsummarizer.cpp#L487-L494

<details>

- [Worth a look] Helper docstring claims "the live and saved-shot paths can no longer drift on detector wiring." The fast path of `summarizeFromHistory` still has its own copy of the cascade gate inline (lines 482-484); changes to the gate must be applied in both places. Reword to "the live path and the saved-shot slow path can no longer drift" with an explicit note about the fast path. (score: 70)
- [Worth a look] Pre-existing stale "via the generateSummary wrapper" comments at lines 286 and 492 also became stale here (and were flagged on PR #944). Either PR could fix them; if they're left, they'll bait further drift. (score: 70)
- [Worth a look] `runShotAnalysisAndPopulate` reads `summary.beverageType`, `summary.totalDuration`, and `summary.finalWeight` off the `summary&` reference rather than as parameters. Both call sites set them before invoking the helper, but the precondition isn't documented in the helper's docstring — a future caller that forgets to set `summary.finalWeight` would silently get `0.0` and disable the grind yield-arm. (score: 60)

</details>

PR #946, opened 2026-04-30
test(ai): add direct unit tests for ShotSummarizer::summarize() (K)

skialpine:
## Summary

- Adds three live-path tests to `tst_shotsummarizer` that mirror the existing saved-shot tests: puck-failure cascade, aborted-preinfusion gate, healthy-shot sanity.
- Constructs real `ShotDataModel` instances through their public ingestion API; no `MockShotDataModel` or production-code friend declaration needed.
- Implements OpenSpec proposal `add-shotsummarizer-live-path-test` (proposal K, the final piece of the G–K series).

## Stacked on

This PR targets [`dedup/shotsummarizer-input-adapter`](https://github.com/Kulitorum/Decenza/pull/945) (I). When I merges to main, this PR's base auto-rebases.

## Why

The history-path tests covered `summarizeFromHistory` thoroughly, but the live path (`summarize(ShotDataModel*, ...)`) had no regression lock — bugs in the curve-extraction adapter from `ShotDataModel*` could regress without surfacing in CI.

After PR #945 (I) both paths share `runShotAnalysisAndPopulate`, so a divergence here would mean the live-path adapter built different inputs (curves, markers, frame info, target weight) for the same shot — not a detector-orchestration drift.

## Test plan

- [x] Three new live-path tests added: `summarize_pourTruncated_suppressesChannelingAndTempLines_live`, `summarize_abortedPreinfusion_doesNotFlagPerPhaseTemp_live`, `summarize_healthyShot_keepsObservations_live`.
- [x] Existing 1806 tests still pass; suite up to 1809 (Qt Creator MCP, 0 warnings).
- [x] Build clean.

The optional live ↔ history parity test was skipped per the proposal's note ("nice-to-have, not a contract this change needs to lock in") — the direct live-path tests already cover the adapter surface, and post-PR #945 the orchestration drift is structurally prevented.

### DIFF
--- a/docs/CLAUDE_MD/MCP_SERVER.md
+++ b/docs/CLAUDE_MD/MCP_SERVER.md
@@ -298,8 +298,10 @@ Both are populated by a single call to `ShotAnalysis::analyzeShot`, so they cann
   },
   "pourTruncated": false,
   "peakPressureBar": 9.1,        // present only when pourTruncated == true
-  "pourStartSec": 6.2,           // phase-boundary window analyzeShot used to gate every other detector
-  "pourEndSec": 32.8,            // both 0.0 when phase markers were absent and the no-data fallback ran
+  "pourStartSec": 6.2,           // phase-boundary window analyzeShot used to gate every other detector;
+                                 //   stays 0.0 when no "preinfusion"/"pour" markers are present (whole-shot fallback)
+  "pourEndSec": 32.8,            // defaults to shot duration when no "end" marker is present;
+                                 //   both fields are 0.0 only on the insufficient-data early return (pressure.size() < 10)
   "skipFirstFrame": false,
   "verdictCategory": "minorIssuesGrindFine"
 }
--- a/docs/CLAUDE_MD/MCP_SERVER.md
+++ b/docs/CLAUDE_MD/MCP_SERVER.md
@@ -298,6 +298,8 @@ Both are populated by a single call to `ShotAnalysis::analyzeShot`, so they cann
   },
   "pourTruncated": false,
   "peakPressureBar": 9.1,        // present only when pourTruncated == true
+  "pourStartSec": 6.2,           // phase-boundary window analyzeShot used to gate every other detector
+  "pourEndSec": 32.8,            // both 0.0 when phase markers were absent and the no-data fallback ran
   "skipFirstFrame": false,
   "verdictCategory": "minorIssuesGrindFine"
 }
--- a/openspec/changes/add-shotsummarizer-live-path-test/tasks.md
+++ b/openspec/changes/add-shotsummarizer-live-path-test/tasks.md
@@ -2,25 +2,22 @@
 
 ## 1. Mock ShotDataModel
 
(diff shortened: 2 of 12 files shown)


skialpine:
### Code review

Found 2 issues:

1. The pre-existing docstring "Cascade integrity through the fast path: when shotData carries a `detectorResults.pourTruncated == true`..." was meant to document `summarizeFromHistory_fastPathPreservesPourTruncatedCascade` (now at line 594) but the new live-path section header was inserted immediately after it on line 484, leaving the docstring orphaned above an unrelated section. A reader scanning the live-path block reads the fast-path docstring first; the actual fast-path test ends up with no comment.

https://github.com/Kulitorum/Decenza/blob/295182e867b57f7f99df3cbfe37de558b185eeaa/tests/tst_shotsummarizer.cpp#L478-L496

2. `summarize_pourTruncated_suppressesChannelingAndTempLines_live` asserts `!linesContain(..., "Sustained channeling")`, but the cascade isn't actually exercised for the channeling arm: every `LiveSample` is built with `pressureGoal=0.0, flowGoal=0.0`, so `addSample`'s `if (flowGoal > 0)` / `if (pressureGoal > 0)` guards never fire and the goal curves stay empty. `buildChannelingWindows` then returns empty windows and `detectChannelingFromDerivative` exits with `None` regardless of `pourTruncated`. The assertion passes for the wrong reason — if the cascade regressed and stopped suppressing channeling, this test would still be green.

   Fix: feed pressure-mode pour samples with non-zero `pressureGoal` so `buildChannelingWindows` produces a valid window and the cascade actually has something to suppress. (Same shape applies to the `_live` test's preinfusion: `LiveSample.isFlowMode=false` while the phase marker is `isFlowMode=true`, which is the inconsistency that lets goal curves stay silent.)

https://github.com/Kulitorum/Decenza/blob/295182e867b57f7f99df3cbfe37de558b185eeaa/tests/tst_shotsummarizer.cpp#L497-L526

<details>

- [Worth a look] `populateLiveShot()` synthesises weight as `s.t * 1.2`. For the aborted-preinfusion test that produces `3.6 g` at `t=3.0` while the test passes `finalWeight=0.5` to `summarize()` — the curve doesn't match the parameter. Also, `t=0.0` is silently dropped by `addWeightSample`'s `weight < 0.1` guard. The current tests don't assert on weight-derived detectors so it's not breaking anything yet, but the helper is documented as "enough to give findValueAtTime something to interpolate" without flagging the inconsistency. (score: 70)
- [Worth a look] `populateLiveShot()` docstring says "every test method owns its model and disposes via QObject parent ownership" — actually the model is stack-allocated with no parent; disposal is RAII. Misleading. (score: 65)
- [Worth a look] Live `summarize_pourTruncated_*_live` uses `isFlowMode=false` for samples but `isFlowMode=true` for the preinfusion phase marker — see issue 2 for why this also breaks the channeling-suppression assertion. (score: 60)
- [Worth a look] `summarize_abortedPreinfusion_doesNotFlagPerPhaseTemp_live` uses `isFlowMode=true` throughout, while its history-path mirror (`abortedPreinfusionDoesNotFlagPerPhaseTemp`) uses `isFlowMode=false`. The tests claim to mirror each other but don't, structurally. (score: 55)
- [OK to skip] New test names use snake_case fragments + `_live` suffix (`summarize_pourTruncated_suppressesChannelingAndTempLines_live`) while existing tests are pure camelCase. Stylistic deviation, not load-bearing. (score: 50)

</details>

PR #948, opened 2026-04-30
fix(history): move exported shot history dir out of profiles/ subdirectory

skialpine:
## Summary

- `userHistoryPath()` and `userHistoryPathIfExists()` were computing their base from `fallbackPath()` (`AppDataLocation/profiles`), so exported JSON shot history files landed at `…/profiles/history` instead of `…/history` (a sibling of `profiles/`, not nested inside it)
- Fix uses `AppDataLocation` directly on Mac, Windows, Linux, and iOS; Android (which uses `externalProfilesPath()` → `Documents/Decenza/`) was already correct
- Adds a one-time startup migration that renames the old `profiles/history` directory into the correct location using `QDir::rename()` (atomic on the same filesystem); no-op if the old directory doesn't exist

## Test plan

- [ ] Build and run on Mac
- [ ] Enable "Export shots to file" in settings
- [ ] Confirm exported files appear at `~/Library/Application Support/DecentEspresso/Decenza/history/`
- [ ] Confirm `profiles/history/` does not exist
- [ ] Migration test: manually create `profiles/history/` with a dummy `.json`, restart app, verify file moved to `history/`

### DIFF
--- a/src/core/profilestorage.cpp
+++ b/src/core/profilestorage.cpp
@@ -27,7 +27,8 @@ ProfileStorage::ProfileStorage(QObject* parent)
         migrateProfilesToExternal();
     }
 
-    // One-time migration: history dir was incorrectly nested inside profiles/
+    // One-time migration: history dir was incorrectly nested inside profiles/ on desktop.
+    // No-op on Android with external storage (history was never written to AppDataLocation there).
     migrateHistoryToNewPath();
 }
 
@@ -126,7 +127,10 @@ QString ProfileStorage::downloadedProfilesPath() const {
 }
 
 QString ProfileStorage::userHistoryPath() const {
-    QString basePath = externalProfilesPath(); // non-empty only on Android
+    // Use externalProfilesPath only when storage permission is confirmed — unlike other
+    // path methods, the fallback here is AppDataLocation (not fallbackPath/profiles/) so
+    // that history lands alongside profiles/ rather than inside it.
+    QString basePath = isConfigured() ? externalProfilesPath() : QString();
     if (basePath.isEmpty()) {
         basePath = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
     }
@@ -139,7 +143,7 @@ QString ProfileStorage::userHistoryPath() const {
 }
 
 QString ProfileStorage::userHistoryPathIfExists() const {
-    QString basePath = externalProfilesPath(); // non-empty only on Android
+    QString basePath = isConfigured() ? externalProfilesPath() : QString();
     if (basePath.isEmpty()) {
         basePath = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
     }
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

Found 7 issues:

1. **Migration silently orphans existing files when destination already exists** — `QDir::rename()` fails on Windows whenever `newPath` exists (empty or not), and on POSIX when `newPath` is a non-empty directory. On failure the code logs a warning and returns, but `userHistoryPath()` now exclusively resolves to the new path — all files remaining in `profiles/history` become permanently inaccessible. Realistic triggers: downgrade-then-upgrade, a cloud-sync tool pre-creating the target directory, or any scenario where `AppDataLocation/history` already exists when the migration runs.

   https://github.com/Kulitorum/Decenza/blob/fd763cf1630f1ec604868ea2009973f34d5b1c49/src/core/profilestorage.cpp#L155-L165

2. **Android: `userHistoryPath()` no longer falls back to internal storage when permission is denied** — `StorageHelper.getProfilesPath()` always returns a non-empty `Documents/Decenza` path regardless of permission state. The old implementation gated on `isConfigured()` so unpermissioned Android users fell back to always-writable internal storage; the new code calls `externalProfilesPath()` unconditionally. On Android 11+ without `MANAGE_EXTERNAL_STORAGE`, `mkpath()` and all subsequent writes silently fail. Every other path method (`userProfilesPath()`, `downloadedProfilesPath()`) retains the `isConfigured()` guard for exactly this reason.

   https://github.com/Kulitorum/Decenza/blob/fd763cf1630f1ec604868ea2009973f34d5b1c49/src/core/profilestorage.cpp#L128-L148

3. **`migrateHistoryToNewPath()` is declared `public` but should be `private`** — called only from the constructor, no `Q_INVOKABLE`. Contrast with `migrateProfilesToExternal()` which is `Q_INVOKABLE` for QML access; this method has no equivalent need.

   https://github.com/Kulitorum/Decenza/blob/fd763cf1630f1ec604868ea2009973f34d5b1c49/src/core/profilestorage.h#L71-L75

4. **`// non-empty only on Android` comment omits the reason `isConfigured()` is intentionally skipped** — all sibling path methods (`userProfilesPath()`, `downloadedProfilesPath()`) guard on `isConfigured()` before calling `externalProfilesPath()`. A future maintainer seeing the inconsistency may add the guard back, breaking the desktop path fix.

   https://github.com/Kulitorum/Decenza/blob/fd763cf1630f1ec604868ea2009973f34d5b1c49/src/core/profilestorage.cpp#L128-L133

5. **Constructor comment implies migration applies to all platforms; it is a no-op on Android with external storage configured** — history on Android was always written to `Documents/Decenza/history/` (via `externalProfilesPath()`), not under `AppDataLocation/profiles/`, so `oldDir.exists()` is always false there.

   https://github.com/Kulitorum/Decenza/blob/fd763cf1630f1ec604868ea2009973f34d5b1c49/src/core/profilestorage.cpp#L29-L32

6. **`migrateHistoryToNewPath()` header says "one-time" without explaining how idempotency is enforced** — there is no guard flag; the guarantee comes from the old directory ceasing to exist after a successful run. Readers expecting a `m_historyMigrated` flag will be confused.

   https://github.com/Kulitorum/Decenza/blob/fd763cf1630f1ec604868ea2009973f34d5b1c49/src/core/profilestorage.h#L73-L75

7. **`userHistoryPath()` header comment "sibling of the user profiles folder" is imprecise on desktop** — on desktop, `history/` is at `AppDataLocation/history/` while `userProfilesPath()` returns `AppDataLocation/profiles/user/`; history is a sibling of `profiles/`, not of `user/`.

   https://github.com/Kulitorum/Decenza/blob/fd763cf1630f1ec604868ea2009973f34d5b1c49/src/core/profilestorage.h#L60-L64

---

<details>
<summary>0 below-threshold issues (50-74)</summary>
</details>

PR #953, opened 2026-04-30
Adopt MCP spec 2025-11-25

skialpine:
## Summary

Bumps the Decenza MCP server's preferred protocol version from `2025-03-26` to `2025-11-25` and adopts the spec features that materially improve LLM client experience and tighten transport security. Implements the [`update-mcp-protocol-2025-11-25` OpenSpec proposal](openspec/changes/update-mcp-protocol-2025-11-25/proposal.md).

### What's in
- **Protocol version negotiation** — supported set bumped to `{"2025-11-25", "2025-06-18", "2025-03-26", "2024-11-05"}`; preferred = first.
- **`MCP-Protocol-Version` request header validation** (2025-06-18) — required on every HTTP request after `initialize`; mismatches return HTTP 400.
- **Origin allowlist** (2025-11-25) — DNS-rebinding protection. Loopback + the host's own LAN IPs are accepted; foreign origins return HTTP 403 before any JSON-RPC parsing. CORS responses now echo the validated origin instead of wildcarding.
- **Structured tool output** (2025-06-18) — every `tools/call` and `resources/read` response now carries a `structuredContent` field alongside the legacy `text` content block (kept for `2025-03-26` clients).
- **Resource link content blocks** (2025-06-18) — shot/profile/machine read tools emit `resource_link` blocks pointing at `decenza://shots/{id}`, `decenza://profiles/{filename}`, `decenza://machine/state`, etc. Inline payloads are kept so older clients aren't penalized.
- **`title` field on tools and resources** (2025-06-18) — auto-derived from `snake_case` programmatic names so the diff stays small (no per-tool churn).
- **`icons` field on tools and resources** (2025-11-25) — category-driven SVGs from `qrc:/icons/` embedded as `data:image/svg+xml;base64,…` URIs.
- **JSON Schema 2020-12 dialect** (2025-11-25) — every input schema declares `"$schema": "https://json-schema.org/draft/2020-12/schema"` via a registry-level helper.

### Bonus fix
- **Setup-link guard** in `qml/pages/settings/SettingsAITab.qml`: the "Setup page: /mcp/setup" link in Settings → AI was visible before the shot server's listener was bound. In that window the URL collapsed to `/mcp/setup`, which macOS resolved against the qrc base — popping a Finder "no application" dialog instead of opening a browser. The link now hides until `shotServer.url` is non-empty.

### Out of scope (deliberate)
- **Elicitation** — the in-app `m_pendingConfirmation` modal is tightly coupled to the existing `confirmationRequested` signal; migration cost outweighs conformance benefit.
- **Tasks (experimental, 2025-11-25)** — the current deferred-socket pattern in `sendAsyncToolResponse` works.
- **OAuth / RFC 8707 / OIDC discovery** — owned by the existing `add-remote-mcp-connector` change.
- **Removing JSON-RPC batching** — already a non-issue (`mcpserver.cpp` calls `doc.object()` and rejects arrays at parse time).

## Test plan

- [x] **Build (Qt Creator, Debug)**: 0 errors, 0 warnings, ~26 s.
- [x] **`tst_McpServerProtocol`** (new, 22 tests): protocol-version negotiation across the supported set, `MCP-Protocol-Version` mismatch → 400, Origin allowlist (loopback / 127.0.0.1 / foreign-with-malformed-body / empty), `tools/list` shape (`title`, `icons[].src`/`mimeType`/`sizes`, `inputSchema.$schema`), `tools/call` shape (`structuredContent` + legacy `text` block, `_resourceLinks` side-channel converted to `resource_link` blocks and stripped from `structuredContent`), pure helpers (`deriveTitle`, `withJsonSchemaDialect` idempotence). All 22 pass.
- [x] **No regression** in `tst_McpServerSession` (15), `tst_McpToolsProfiles` (8), `tst_McpToolsWrite` (4) — 54 MCP tests pass total.
- [x] **Runtime probe vs live `localhost:8888` and `192.168.10.195:8888`**: both endpoints negotiate `2025-11-25`, emit `MCP-Protocol-Version` response header, return `structuredContent` + `resource_link` blocks, reject `Origin: http://evil.example` with 403, reject mismatched `MCP-Protocol-Version` with 400, still negotiate `2025-03-26` for legacy clients.
- [ ] **End-to-end via `mcp-remote`** from Claude Desktop — pending; the bridge is the canonical client, worth a smoke-test before merging.
- [ ] **`scripts/test_mcp.sh`** integration suite (~200 cases) against the live server — recommended per [docs/CLAUDE_MD/TESTING.md](docs/CLAUDE_MD/TESTING.md) for any change to `src/mcp/`.


skialpine:
### Code review

Found 7 issues. All addressed in [`65c5b57d`](https://github.com/Kulitorum/Decenza/commit/65c5b57d).

1. **Auto-recovery + protocol version mismatch — clients permanently 400'd after session expiry.** When a stale `Mcp-Session-Id` forced an auto-create (multiple sessions present, can't pin to sole), the new session kept the default `2025-03-26` while the client kept sending its previously-negotiated header (e.g. `2025-11-25`). The immediately-following mismatch check then 400'd every request — defeating exactly the auto-recovery the surrounding comment claims to provide. Regression test added in the fix commit.

   https://github.com/Kulitorum/Decenza/blob/b02641f428677931d212a4dbf0accdf7c6bd9bbe/src/mcp/mcpserver.cpp#L380-L410

2. **`int` instead of `qsizetype` for `QString::indexOf()` return.** [Root CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/b02641f428677931d212a4dbf0accdf7c6bd9bbe/CLAUDE.md#L424) says: *"Use `qsizetype` (not `int`) for container sizes — `QVector::size()`, `QList::size()`, `QString::size()` etc. return `qsizetype` (64-bit on iOS/macOS). Assigning to `int` causes `-Wshorten-64-to-32` warnings."* Same return type as `size()`.

   https://github.com/Kulitorum/Decenza/blob/b02641f428677931d212a4dbf0accdf7c6bd9bbe/src/mcp/mcpserver.cpp#L98-L104

3. **Same qrc setup-link bug exists at a second site.** The PR's commit `a0b360b4` fixes the bug at the inline setup link, but the "Open Web Guide" button inside `mcpHelpDialog` (same file) still uses the old guard `Settings.mcp.mcpEnabled && MainController.shotServer` and concatenates an empty `shotServer.url` with `"/mcp/setup"` — producing the same `qrc:/mcp/setup` Finder dialog on macOS when the server is enabled but not yet listening. [Root CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/b02641f428677931d212a4dbf0accdf7c6bd9bbe/CLAUDE.md#L641) says: *"Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'."*

   https://github.com/Kulitorum/Decenza/blob/b02641f428677931d212a4dbf0accdf7c6bd9bbe/qml/pages/settings/SettingsAITab.qml#L953-L962

4. **`design.md` Risk #2 references an additional-origins setting that was removed from scope.** The Risks section reads *"the settings UI exposes an additional-origins field for advanced users"* but [`tasks.md`](https://github.com/Kulitorum/Decenza/blob/b02641f428677931d212a4dbf0accdf7c6bd9bbe/openspec/changes/update-mcp-protocol-2025-11-25/tasks.md#L21-L23) explicitly removes tasks 3.5/3.6 from scope. Stale doc artifact.

   https://github.com/Kulitorum/Decenza/blob/b02641f428677931d212a4dbf0accdf7c6bd9bbe/openspec/changes/update-mcp-protocol-2025-11-25/design.md#L72-L76

5. **`SessionTimeoutMinutes` inline comment is now wrong.** Reads *"safety net — orphan cleanup only runs during session creation"*, but the constructor wires `cleanupExpiredSessions` to a 60s `m_cleanupTimer` that fires unconditionally. The comment was accurate before this PR but isn't anymore.

   https://github.com/Kulitorum/Decenza/blob/b02641f428677931d212a4dbf0accdf7c6bd9bbe/src/mcp/mcpserver.h#L186-L191

6. **`HTTP/1.1 403 Unknown` instead of `403 Forbidden`.** `httpStatusText` switch maps 200/202/204/400/405/429 but not 403, which is the new Origin-rejection path's status code. Cosmetic per RFC 7230, but trivially fixable.

   https://github.com/Kulitorum/Decenza/blob/b02641f428677931d212a4dbf0accdf7c6bd9bbe/src/mcp/mcpserver.cpp#L1103-L1114

7. **Dead `want` parameter on header-extraction lambda in test.** The closure already captures `line`, so every call site passes a literal string that's silently ignored — `-Wunused-parameter` (Clang) / C4100 (MSVC).

   https://github.com/Kulitorum/Decenza/blob/b02641f428677931d212a4dbf0accdf7c6bd9bbe/tests/tst_mcpserver_protocol.cpp#L101-L112

PR #875, opened 2026-04-26
fix(ios): defer focus shift to dodge QIOSTapRecognizer dispatch_async race

skialpine:
## Summary

Fix iOS SIGSEGV crash inside Qt's `QIOSTapRecognizer` tap-dispatch block by deferring the focus-shift work in `SuggestionField.qml` with `Qt.callLater()`. Closes #871. Tracked upstream as [QTBUG-146020](https://bugreports.qt.io/browse/QTBUG-146020).

## What's going on

After symbolicating crashes #871 and #872 against the v1.7.1 dSYM, frame 3 was always `__-[QIOSTapRecognizer touchesEnded:withEvent:]_block_invoke`. Reading `qtbase/src/plugins/platforms/ios/qiostextinputoverlay.mm` (Qt 6.10.3) shows why:

`QIOSTapRecognizer::touchesEnded` (line 965) queues a block via `dispatch_async(dispatch_get_main_queue(), ^{ ... })` to show the iOS edit menu on the next runloop pass:
```objc
dispatch_async(dispatch_get_main_queue(), ^{
    if (_menuShouldBeVisible)
        showEditMenu(_focusView, touchPos.toPoint());   // ← reads self->_focusView lazily
    ...
});
```

But `setEnabled:NO` (line 915), called when focus leaves a text-editable item, **releases `_focusView` and sets it to nil**. And `showEditMenu`'s first line dereferences it without a nil check:
```objc
QWindow *qtWindow = quiview_cast(focusView).platformWindow->window();   // null deref → SIGSEGV
```

So the race is: user taps inside our QQuickWindow → `touchesEnded` queues the block → our QML tap handler synchronously calls `forceActiveFocus()` + `Qt.inputMethod.hide()`, moving focus off the text field → `_focusView = nil` → next runloop pass, the block fires and crashes.

This matches both user notes on the issue: "changing grinders / filling out shot information" and "pressing the up arrow next to coffee/grinder" both involve `SuggestionField` chevron taps that synchronously shift focus.

## What this PR does

Wraps the focus-shift work in `selectSuggestion()` and `openSuggestionsDialog()` in `Qt.callLater(...)` so the dispatch_async block runs first against a still-valid `_focusView`. Costs one event-loop tick per chevron tap or suggestion pick — imperceptible to users.

[...]

PR #956, opened 2026-04-30
fix(analysis): detect puck failure on shots aborted before first pour marker

skialpine:
## Summary
- `detectPourTruncated` early-returned when `pourEnd <= pourStart`, hiding puck-failure shots whose first pour-class phase marker landed after the shot was already aborted. Verified against shot 875 (80's Espresso, 2.28 s, 1.1 g, peak 0.35 bar): `pourStart=2.5 > pourEnd=2.28` — production gave the "Clean shot. Puck held well." verdict instead of the puck-failure cascade.
- Fall back to scanning the whole pressure series when the marker-derived window is degenerate. The 2.5 bar floor is conservative enough that any shot that built real extraction pressure crosses it somewhere; if nothing in the entire series reaches the floor, the puck demonstrably never built.
- New regression test `pourTruncated_invertedWindow_scansFullSeries` mirrors shot 875's series (12 samples, 0.35 bar peak, inverted window).

## Context
Surfaced while reviewing badge classifications across the last 100 shots in local history via the live Decenza MCP server. Of 100 shots, only one fell into this trap — but the failure mode (very-short aborts) is real and silent.

The non-espresso skip-list (filter / pourover / tea / steam / cleaning) still runs first, so legitimate low-pressure modes are unaffected. The `pressure.size() < 10` guard is preserved for truly empty shots.

## Test plan
- [x] `tst_shotanalysis` — 1835 passed, 0 failed
- [x] `shot_corpus_regression` ctest passes (no existing fixture regresses)
- [x] New unit test covers the inverted-window case
- [ ] Manual: re-load shot 875 in Shot Detail; confirm "Pour never pressurized" warning + "Don't tune off this shot" verdict (relies on lazy-recompute on detail open)

### DIFF
--- a/src/ai/shotanalysis.cpp
+++ b/src/ai/shotanalysis.cpp
@@ -597,16 +597,25 @@ bool ShotAnalysis::detectPourTruncated(const QVector<QPointF>& pressure,
         || bev == QStringLiteral("cleaning"))
         return false;
 
-    if (pressure.size() < 10 || pourEnd <= pourStart) return false;
+    if (pressure.size() < 10) return false;
 
-    // Scan the pour window for peak pressure. We look only inside the pour
-    // (not the entire sample range) because some profiles briefly spike
-    // during fill before the puck is engaged — that's not diagnostic of
-    // whether extraction actually built pressure.
+    // Scan the pour window for peak pressure. We normally look only inside
+    // the pour (not the entire sample range) because some profiles briefly
+    // spike during fill before the puck is engaged — that's not diagnostic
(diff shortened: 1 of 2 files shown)


skialpine:
Closing — on review with Jeff, the fix would convert a passive mis-verdict ("Clean shot") into an active misdiagnosis ("Puck failed — check prep") for aborted shots like #875 (2.3 s, 0 g, no preinfusion ever started). The right path is an aborted-shot gate ahead of the cascade that suppresses all badges and emits a neutral verdict. Will scaffold via OpenSpec.

PR #957, opened 2026-04-30
refactor(history): always discard aborted espresso shots (no toggle)

skialpine:
## Summary
- Remove the `Settings.brew.discardAbortedShots` toggle. The classifier (extractionDuration < 10 s AND finalWeight < 5 g) now runs unconditionally.
- Drops 75 lines, adds 4. UI row in the Calibration settings tab is gone; the C++ Q_PROPERTY / getter / setter / signal are all removed; `MainController::handleEspressoShotEnd` calls `decenza::isAbortedShot(...)` directly.
- Orphan QSettings key `espresso/discardAbortedShots` left in place on existing installs — QSettings silently ignores unknown keys, so no migration needed.

## Why
The toggle defaulted to on and was validated against an 882-shot corpus with 0 false-drops. Functionally always-on; removing the user knob removes a class of confusion (e.g. "why is my 2 s abort still showing as a Clean shot?" — answer: someone toggled it off).

Pruned 4 pre-fix aborted shots from local history (IDs 17, 826, 840, 875) before this change so they don't keep biasing badge audits / auto-favorites.

## Files
- `src/core/settings_brew.{h,cpp}` — remove Q_PROPERTY, getter, setter, signal.
- `src/controllers/maincontroller.cpp:1812-1828` — drop the `discardEnabled` gate; classifier always fires.
- `qml/pages/settings/SettingsCalibrationTab.qml` — remove the toggle Rectangle (lines 392-438 in the prior version).

## Test plan
- [x] `tst_aborted_shot_classifier` — 1834 passed, 0 failed
- [x] desktop build clean
- [ ] manual: confirm a deliberate ≤2 s abort produces the toast and is not in history

### DIFF
--- a/openspec/specs/shot-save-filter/spec.md
+++ b/openspec/specs/shot-save-filter/spec.md
@@ -7,16 +7,15 @@ TBD - created by archiving change add-discard-aborted-shots. Update Purpose afte
 
 When an espresso extraction ends and reaches the save path, the application SHALL classify the shot as *aborted* iff BOTH of the following hold: `extractionDuration < 10.0 s` AND `finalWeight < 5.0 g`. The two clauses form a conjunction; either alone is insufficient to classify the shot as aborted. Long-running low-yield shots (e.g. a choked puck producing 1 g over 60 s) MUST NOT classify as aborted, because their graphs are diagnostically valuable.
 
-When the `Settings.brew.discardAbortedShots` toggle is `true` (default) AND the classifier returns *aborted*, the application SHALL skip persisting the shot to `ShotHistoryStorage` AND SHALL skip any visualizer auto-upload for that shot. When the toggle is `false`, the classifier SHALL NOT be consulted and all shots SHALL save as they do today.
+When the classifier returns *aborted*, the application SHALL skip persisting the shot to `ShotHistoryStorage` AND SHALL skip any visualizer auto-upload for that shot. The classifier runs unconditionally — there is no user-facing opt-out.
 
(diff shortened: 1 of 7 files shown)


skialpine:
### Code review

Found 3 issues:

1. Live OpenSpec spec is now factually wrong — still describes the removed toggle as the gate, plus contains a scenario that's impossible after this PR. CLAUDE.md OpenSpec Stage 3 says "Update `specs/` if capabilities changed."

https://github.com/Kulitorum/Decenza/blob/80e9fd3fa801e3b98f056288c401bc7b310158ef/openspec/specs/shot-save-filter/spec.md#L8-L48

2. Stale path reference in the new comment — `// See openspec/changes/add-discard-aborted-shots.` points to a path that's been archived to `openspec/changes/archive/2026-04-30-add-discard-aborted-shots/`. CLAUDE.md says don't write references that "rot as the codebase evolves" — replace with `openspec/specs/shot-save-filter/spec.md`.

https://github.com/Kulitorum/Decenza/blob/80e9fd3fa801e3b98f056288c401bc7b310158ef/src/controllers/maincontroller.cpp#L1812-L1816

3. Test docstring references a gate that no longer exists — `// The toggle that gates whether a shot's verdict actually causes a drop lives in MainController` describes behavior removed by this PR. Same CLAUDE.md anti-pattern as #2 (comment rot referencing removed implementation detail).

https://github.com/Kulitorum/Decenza/blob/80e9fd3fa801e3b98f056288c401bc7b310158ef/tests/tst_aborted_shot_classifier.cpp#L84-L89

---

<details>

- [Worth a look] `shotDiscarded` signal comment in maincontroller.h doesn't note the always-on contract — one-line addition like "Always fires when the classifier triggers — no user opt-out" prevents future maintainers from looking for an opt-out path that no longer exists (score: 60)

</details>

PR #961, opened 2026-04-30
fix(profile): stop overriding espresso_temperature on every parse (#958)

### LINKED ISSUE #958: Profile::fromJson side-effect: "Syncing espresso_temperature" qDebug fires on every read (incl. shots_get_detail)
### Symptom

Every time `Profile::fromJson` parses a non-simple profile whose top-level `espresso_temperature` differs from `steps[0].temperature`, it emits

```
Syncing espresso_temperature from <old> to first frame temp <new>
```

at qDebug level. During an MCP audit that fetched 500 historical shots via `shots_get_detail`, this produced ~500 of these lines plus 500 `ShotServer: "POST" "/mcp"` log entries — visible in the live app log under normal operation.

### Why it's happening

[`src/profile/profile.cpp:559-574`](https://github.com/Kulitorum/Decenza/blob/main/src/profile/profile.cpp#L559-L574) (added in commit `9c4b195a`, "Fix wrong profile and temperature on startup"). The original intent was to heal profiles saved before an editor bug that let users edit per-frame temps without updating the profile-level field — without the fix, activating such a profile would upload the wrong group temperature to the DE1.

### Why it's noise (not corruption)

The mutation lives inside the freshly-parsed `Profile` object. It does NOT write back to QSettings or the active profile. For the active-profile adoption path the fix is load-bearing; for the read paths (`shots_get_detail` → `convertShotRecord`, profile listing, MCP responses, etc.) the parsed Profile is consumed for metadata extraction and discarded — the sync runs but its output is never used, and the qDebug just pollutes logs.

I verified during the audit that `Settings.espresso.espressoTemperatureC` did not change despite ~500 of these log lines firing.

### Suggested fix

Move the sync out of `Profile::fromJson` (parsers should be pure) and into the active-profile adoption path:

- **A. Smallest fix:** drop the `qDebug` line. Mutation stays where it is; logs go quiet. Tactical, doesn't address the architectural smell.
- **B. Right fix:** relocate the reconciliation into `ProfileManager` at the moment the active profile is committed (and into `Profile::saveJson` so legacy stored profiles get healed once on next save instead of re-healed on every read). Drop the `fromJson` block.
- **C. One-time migration:** sweep stored profiles on first run after the next release, normalize the stale top-level temps, then remove the workaround entirely.

I'd take **B** as a small PR — keeps the safety net for legacy profiles but stops the read-path mutation/log noise. **C** is cleaner long-term but more code.

### Repro

1. Run Decenza with at least one historical profile whose top-level `espresso_temperature` differs from `steps[0].temperature` (most legacy profiles, in practice).
2. Trigger any read path that parses a stored profile JSON — easiest: load Shot Detail in the app, or hit `shots_get_detail` via MCP.
3. Watch the log.

Surfaced during a 500-shot badge audit on PR #957's branch.

skialpine:
## Summary

- `Profile::fromJson` was healing the top-level `espresso_temperature` to `steps[0].temperature` on every parse — and logging `Syncing espresso_temperature…` each time.
- Mismatched top-level vs. frame[0] is **intentional** on many profiles (D-Flow / A-Flow built-ins ship with a cooler group preheat paired with a hotter preinfusion ramp). The heal silently overrode authored values.
- `fromJson` is on every read path (Shot Detail, MCP `shots_get_detail`, profile listing), so the log noise was constant under normal use and produced ~500 lines during a 500-shot MCP audit.
- de1app does no auto-sync on profile load, so removing the heal matches its behavior. The "All temps" slider in Decenza's editor already matches de1app's `change_espresso_temperature` (shifts all frames by delta, sets top-level to the new absolute frame[0]) — no change needed there.
- The defaulting branch (`if !contains("espresso_temperature")`) is kept so malformed JSON that omits the field doesn't end up with NaN/0.

Fixes #958.

## Test plan

- [x] Builds cleanly (Qt Creator MSVC, 0 errors / 0 warnings)
- [ ] Open Shot Detail on a historical shot — confirm no `Syncing espresso_temperature…` log lines
- [ ] Activate a D-Flow / A-Flow built-in (e.g. `a_flow_default_dark`) — confirm the profile-level temp shown matches the JSON (92°C, not 93°C) and the heater preheats to that value

### DIFF
--- a/src/profile/profile.cpp
+++ b/src/profile/profile.cpp
@@ -556,21 +556,12 @@ Profile Profile::fromJson(const QJsonDocument& doc) {
         }
     }
 
-    // Sync espresso_temperature from first frame for advanced profiles only.
-    // Simple profiles (settings_2a/2b) have explicit espresso_temperature that is
-    // authoritative — their frames may be stale from a prior conversion.
-    bool isSimpleProfile = (profile.m_profileType == "settings_2a" || profile.m_profileType == "settings_2b");
-    if (!isSimpleProfile && !profile.m_steps.isEmpty()) {
-        if (!obj.contains("espresso_temperature")) {
-            profile.m_espressoTemperature = profile.m_steps.first().temperature;
-        } else {
-            double firstFrameTemp = profile.m_steps.first().temperature;
-            if (qAbs(profile.m_espressoTemperature - firstFrameTemp) > 0.1) {
-                qDebug() << "Syncing espresso_temperature from" << profile.m_espressoTemperature
-                         << "to first frame temp" << firstFrameTemp;
-                profile.m_espressoTemperature = firstFrameTemp;
-            }
-        }
+    // Default espresso_temperature from first frame only when the JSON omits it,
+    // so we never store NaN/0. Otherwise the top-level value is authoritative —
+    // it can legitimately differ from steps[0].temperature (e.g. a cooler group
+    // preheat target paired with a hotter preinfusion ramp).
+    if (!obj.contains("espresso_temperature") && !profile.m_steps.isEmpty()) {
+        profile.m_espressoTemperature = profile.m_steps.first().temperature;
     }
 
     // De1app defaults NumberOfPreinfuseFrames to 0 when the field is missing
--- a/src/profile/profile.cpp
+++ b/src/profile/profile.cpp
@@ -557,9 +557,10 @@ Profile Profile::fromJson(const QJsonDocument& doc) {
     }
 
     // Default espresso_temperature from first frame only when the JSON omits it,
-    // so we never store NaN/0. Otherwise the top-level value is authoritative —
-    // it can legitimately differ from steps[0].temperature (e.g. a cooler group
-    // preheat target paired with a hotter preinfusion ramp).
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 1 issue:

1. Stale comment cross-reference. The comment at `src/profile/profile.cpp:1192-1193` reads `// Simple profiles have authoritative scalar temperature // (see fromJson guard at line ~548).` — but that guard is the very block this PR deleted. The new code at the same physical location is a JSON-missing fallback, not a simple-profile guard, and the framing "Simple profiles have authoritative scalar temperature" is no longer true after this PR (top-level is now authoritative for all profile types). Recommend trimming the comment to keep only the still-valid reasoning about `preinfusionTime=0` + `tempStepsEnabled=true` and dropping the stale anchor.

https://github.com/Kulitorum/Decenza/blob/648a3d0736899fe0d81df90dcda71a777cd3fe40/src/profile/profile.cpp#L1188-L1195

<details>

- [Worth a look] New comment at `profile.cpp:559-562` says top-level `espresso_temperature` is authoritative without noting that `regenerateFromRecipe` resyncs it from `steps[0]`. Optional one-line clarifier: "authoritative *at load time*". (score: 62)

</details>

Issue #967, opened 2026-05-01
MCP: apply units-in-name convention to remaining shots_get_detail fields


skialpine:
Follow-up to #964. After fix(shots): clarify target weight field landed (`yieldOverride` → `targetWeightG` + grind detector gates), six fields in the `shots_get_detail` / `shots_compare` payload still violate the CLAUDE.md MCP convention "Include units in field names":

| Current | Should be |
|---------|-----------|
| `finalWeight` | `finalWeightG` |
| `doseWeight` | `doseWeightG` |
| `duration` | `durationSec` |
| `temperatureOverride` | `temperatureOverrideC` |
| `drinkTds` | (keep — `Tds` already implies %) |
| `drinkEy` | (keep — `Ey` already implies %) |

(`drinkTds`/`drinkEy` are arguable; the others are clear violations.)

## Why

Issue #964 originally called this out:

> Audit the rest of the shot payload for fields that violate the units-in-name convention (`finalWeight` → `finalWeightG`, `doseWeight` → `doseWeightG`, `duration` → `durationSec`, etc.) — there are several. Decide whether to fix all or just `targetWeightG`.

We deferred during #964 to keep that PR focused on the single-field rename + grind-detector gates. This issue captures the deferred work.

The other MCP tools (`shots_list`, `dialing_get_context`) already use the suffix convention (`doseG`, `yieldG`, `durationSec`, `temperatureOverrideC`) — only the `convertShotRecord` shared projection consumed by `shots_get_detail` and `shots_compare` lags.

## Scope

Same playbook as the #964 PR — pure rename, no alias (per the project decision in #964 that "nobody will care"):

- **Emit site**: `src/history/shothistorystorage_serialize.cpp` — 4 lines.
- **Readers** (~50–100 references):
  - QML: `qml/pages/ShotDetailPage.qml`, `qml/pages/ShotHistoryPage.qml`, `qml/pages/AutoFavoritesPage.qml`, `qml/pages/AutoFavoriteInfoPage.qml`, `qml/components/ComparisonShotTable.qml`
  - C++: `src/network/shotserver*.cpp`, `src/models/shotcomparisonmodel.cpp`, `src/history/shothistorystorage_queries.cpp` (the filtered-list and recent-by-kbId projections also emit these keys)
  - Tests: `tests/tst_shotrecord_cache.cpp`, `tests/tst_shotsummarizer.cpp`
[...]

PR #969, opened 2026-05-01
refactor(history): rename yieldOverride to targetWeight; expose grind detector gates (#964)

### LINKED ISSUE #964: MCP: improve grind/shot field clarity (rename yieldOverride → targetWeightG; expose detector inputs)
The MCP `shots_get_detail` payload has two ergonomic problems that the 2026-04-30 500-shot audit (#963) surfaced — directly broke my own simulation tooling and would catch any external agent consumer the same way.

## Problem 1: target weight is buried in `yieldOverride`

The MCP payload exposes the target weight (the value that goes into `analyzeShot` as the `targetWeightG` parameter and gates the choked-yield arm) under the field name `yieldOverride`. That name reads as "user override of the recorded yield," not "target weight." A consumer scanning the JSON for "target" / "targetWeight" / "ratio target" will miss it entirely.

History: the underlying DB column is `yield_override` (named so because it was originally a per-shot user override of the profile's target). At save time it gets populated either from a real user override OR — more commonly — from `profile.targetWeight()` as a fallback ([`shothistorystorage_serialize.cpp:128`](https://github.com/Kulitorum/Decenza/blob/main/src/history/shothistorystorage_serialize.cpp#L128) passes `record.yieldOverride` directly as `targetWeightG`). So by the time it lands on the shot record, it's the effective target. The field name didn't catch up.

Concrete impact: my Gap A simulation's first pass read `detectorResults.grind.targetWeightG` (which doesn't exist) and silently produced 0/500 candidates. Re-pointed at `yieldOverride` it found 23. CLAUDE.md MCP_SERVER.md already says "Include units in field names" — `yieldOverride` violates that AND is mis-named.

## Problem 2: detector struct doesn't carry the inputs it judged against

`detectorResults.grind` exposes outputs (`direction`, `chokedPuck`, `yieldOvershoot`, `verifiedClean`, `coverage`, `deltaMlPerSec`, `sampleCount`) but not the inputs the gates evaluated. So consumers asking "why didn't this badge fire?" can't answer from the JSON — they need to dig into the source.

Specific missing fields:
- `yieldRatio` (`finalWeightG / targetWeightG`) — the value the moderate-choke arm gate compares to `CHOKED_YIELD_RATIO_MAX = 0.85`.
- `pressurizedDurationSec` — the value the choked-puck arm gate compares to `CHOKED_DURATION_MIN_SEC = 15.0`.
- `meanPressurizedFlowMlPerSec` — what the flow-arm compares to `CHOKED_FLOW_MAX_MLPS = 0.5`.

Concrete impact: shot 745 (Adaptive v2, yield ratio 0.64, well below the 0.85 moderate-choke threshold) silently gets `direction: "onTarget"` because `pressurizedDuration < 15.0 s`. Today no MCP consumer can tell why; you have to read the C++. Three audit iterations were spent answering this question.

### Fix A — rename target field at MCP surface (small, additive)

- Add `targetWeightG` as a top-level field on `shots_get_detail` next to `finalWeight` and `doseWeight`. Read from `record.yieldOverride`.
- Mark `yieldOverride` deprecated in `docs/CLAUDE_MD/MCP_SERVER.md` (keep emitting for backwards compat).
- Audit the rest of the shot payload for fields that violate the units-in-name convention (`finalWeight` → `finalWeightG`, `doseWeight` → `doseWeightG`, `duration` → `durationSec`, etc.) — there are several. Decide whether to fix all or just `targetWeightG`.
- Test: regression confirming both fields carry the same value.

Rough code scope:
- `src/history/shothistorystorage_serialize.cpp` — one line near [line 62](https://github.com/Kulitorum/Decenza/blob/main/src/history/shothistorystorage_serialize.cpp#L62).
- `docs/CLAUDE_MD/MCP_SERVER.md` — document the new field + deprecation.
- `tests/tst_shothistorystorage*.cpp` (or wherever convertShotRecord is tested) — one assertion.

Effort: small, ~30 minutes including PR.

### Fix B — expose detector inputs (medium)

Add a `gates` object to `detectorResults.grind`:

```json
{
  "checked": true,
  "hasData": true,
  "coverage": "verified",
  "verifiedClean": false,
  "direction": "onTarget",
  "chokedPuck": false,
  "yieldOvershoot": false,
  "deltaMlPerSec": -0.18,
  "sampleCount": 142,
  "yieldRatio": 0.64,
  "gates": {
    "pressurizedDurationSec": 8.76,
    "meanPressurizedFlowMlPerSec": 2.43,
    "minSamples": 5,
    "minPressurizedSec": 15.0,
    "minPressureBar": 4.0,
    "passed": false
  }
}
```

Then "why didn't this badge fire?" is answerable: `gates.passed=false` AND `pressurizedDurationSec < minPressurizedSec`. Same question that took three audit iterations on shot 745.

Code scope:
- `src/ai/shotanalysis.{h,cpp}` — populate `pressurizedDuration` and `meanPressurizedFlow` on `GrindCheck` (currently local variables in `analyzeFlowVsGoal`); copy to new `DetectorResults.grindGates` sub-struct.
- `src/history/shothistorystorage_serialize.cpp` — emit the new `gates` sub-object.
- `docs/CLAUDE_MD/MCP_SERVER.md` — document the shape.
- 2-3 unit tests covering the gate-fail / gate-pass / no-data cases.

Effort: medium, ~2-3 hours including spec/tests/PR. Probably worth bundling with Gap A's fix so the same code change exposes the gate values it's loosening.

## Recommendation

Do **Fix A** as a standalone small PR now (low-risk, eliminates the field-name footgun). **Fix B** when picking up Gap A from #963 — same code paths, similar change scope.

Both are MCP surface additions only — no breaking removals — so they don't need OpenSpec; just normal PRs.

## References

- #963 (the audit-findings tracker that surfaced both problems)
- #903 (predecessor low-pressure-shots tracker)
- `docs/CLAUDE_MD/MCP_SERVER.md` MCP data conventions section

skialpine:
Closes #964.

## Summary

- Rename in-app field `yieldOverride` → `targetWeight` across `ShotRecord`, `ShotSaveData`, `ShotFilter`, `ComparisonShot`, all C++ callers, and QML readers. The old name was a historical leftover — the field stores the effective target weight (user brew-by-ratio override OR profile `target_weight`, falling back to actual yield), not just a per-shot override.
- MCP `shots_get_detail` now emits `targetWeightG` (units-suffixed per CLAUDE.md MCP convention). No alias — clean break.
- DB column stays `yield_override` (no migration; column name predates the rename).
- Add `detectorResults.grind.gates` substruct exposing the choked-puck arm's inputs (`pressurizedDurationSec`, `meanPressurizedFlowMlPerSec`, `yieldRatio`, `flowSamples`) plus the thresholds it compared against. MCP consumers can now answer "why didn't this badge fire?" from JSON alone — directly addresses the audit finding on shot 745 (yield ratio 0.64 but `direction: onTarget` because pressurized duration < 15 s).

## Out of scope

Other unit-less MCP fields (`finalWeight`, `doseWeight`, `duration`, `temperatureOverride`) tracked in #967 as a follow-up — same rename playbook, kept separate to keep this PR's blast radius proportional to issue #964.

## Test plan

- [x] `tst_ShotRecordCache` — added `convertShotRecord_emitsTargetWeightG` and `detectorResults_grindGates_exposeArm2Inputs`
- [x] Existing `tst_ShotSummarizer`, `tst_DbMigration`, `tst_ShotAnalysis` — all pass (106 / 106 in the targeted suite)
- [x] Full test suite — 1837 pass; 3 pre-existing failures in `tst_Profile` (unrelated, related to recent commit #961's espresso_temperature change; verified failing on main with this branch's changes stashed)
- [ ] Manual: dial-in conversation against a real shot — confirm the AI sees `targetWeightG` and the `gates` substruct in the response
- [ ] Manual: open the in-app Shot Detail page on an espresso shot — confirm output shows "actual (target g)" correctly


skialpine:
### Code review

Found 2 issues:

1. `gatePassed` docstring is factually wrong about which fields fire on a gate-fail path. It says "When false, the choked-puck loop returns early without setting chokedPuck / yieldOvershoot / verifiedClean", but `yieldOvershoot` is set by an independent arm in [shotanalysis.cpp lines 399-403](https://github.com/Kulitorum/Decenza/blob/ac877b97b90d62713707868b923c76643e3458a4/src/ai/shotanalysis.cpp#L399-L403) **before** the gate is evaluated. A gusher with `gateRan == true && gatePassed == false && yieldOvershoot == true` is the exact case the gates substruct was meant to make legible — the comment will mislead consumers in the opposite direction. Drop `yieldOvershoot` from the list, or add a clarifier that the overshoot arm runs pre-gate.

https://github.com/Kulitorum/Decenza/blob/ac877b97b90d62713707868b923c76643e3458a4/src/ai/shotanalysis.h#L244-L249

2. `temperatureOverride` MCP key is not units-suffixed even though the PR explicitly renames the adjacent line to satisfy the convention ([MCP_SERVER.md says](https://github.com/Kulitorum/Decenza/blob/ac877b97b90d62713707868b923c76643e3458a4/docs/CLAUDE_MD/MCP_SERVER.md#L268-L271) "Numeric fields include their unit as a suffix — `temperatureC` (Celsius)…"). Trivial one-line fix at the same site, plus a few QML/web reader updates. Issue #967 already tracks the rest of the unit-suffix renames so this could go there, but `temperatureOverride` sits literally on the line above the rename and the same readers are already being touched in this PR.

https://github.com/Kulitorum/Decenza/blob/ac877b97b90d62713707868b923c76643e3458a4/src/history/shothistorystorage_serialize.cpp#L60-L63

<details>

- [OK to skip] OpenSpec `brew-overrides/spec.md` only references the unchanged DB column name `yield_override`, so spec stays accurate; the in-memory C++ field rename creates a small naming gap but no behavior change → CLAUDE.md OpenSpec Stage 3's "update specs/ if capabilities changed" rule doesn't trigger (score: 50)

</details>

PR #973, opened 2026-05-01
fix(profile_sync): eliminate false-positive diffs from plugin/simple profiles (#972)

### LINKED ISSUE #972: profile_sync: false-positive diffs from missing plugin awareness, simple-profile expansion, and silent buildDiff
While investigating #968 I ran \`profile_sync\` against the de1app source and got 8 \"diffs\" against shipping JSONs. Tracing them down, all 8 are false positives caused by tool limitations — **none reflect real on-disk drift** in the shipping profiles. The tool should be tightened so its output is trustworthy.

Run was \`profile_sync /Users/jeffreyh/Development/GitHub/de1app/de1plus/profiles /Users/jeffreyh/Development/GitHub/Decenza/resources/profiles\`.

### 1. profile_sync doesn't know about plugin profiles

A-Flow profiles ship under \`de1plus/plugins/A_Flow/profiles/\` (canonical, 9 frames, author Janek), but a stale 6-frame copy still lives at \`de1plus/profiles/A-Flow____*.tcl\` (author Decent). \`profile_sync\` only scans the second path, so it compared Decenza's correct 9-frame JSONs against the older 6-frame source and flagged 4 false diffs:

- \`a_flow_default_dark.json\`
- \`a_flow_default_like_dflow.json\`
- \`a_flow_default_medium.json\`
- \`a_flow_default_very_dark.json\`

**Suggested fix:** scan plugin \`profiles/\` subdirectories under \`de1plus/plugins/*/profiles/\` and have plugin copies override base copies (canonical wins). The README/CLAUDE.md path callout should also be updated once the tool handles this.

### 2. Simple profiles compared without runtime frame regeneration

The 4 remaining diffs (\`default\`, \`best_overall_pressure_profile\`, \`gentle_and_sweet\`, \`e61_with_fast_preinfusion_to_9_bar\`) are all \`settings_2a\`. Their JSONs ship with \`\"steps\": []\` (frames are regenerated at activation time by \`ProfileManager::regenerateSimpleFrames()\` — \`src/controllers/profilemanager.cpp:1754\`, \`:2677\`). \`Profile::loadFromTclString\` regenerates frames at parse time, so we end up comparing N frames vs 0 frames → \`functionallyEqual\` returns false even though every scalar field matches the source exactly.

**Suggested fix:** for simple profiles (\`settings_2a\`/\`settings_2b\`), call the equivalent of \`regenerateSimpleFrames()\` on the JSON-loaded \`Profile\` before comparing. Or alternatively compare scalar fields directly when both sides are simple profiles.

### 3. \`buildDiff\` is silent on header-level mismatches

When \`functionallyEqual\` returned false on those 4 simple profiles, \`buildDiff\` produced an empty body — the per-frame loop doesn't execute when one side has 0 frames, and step-count / \`preinfuseFrameCount\` mismatches aren't reported. The output looked like:

\`\`\`
DIFF: Best overall pressure profile (best_overall_pressure_profile.json)
DIFF: Gentle and sweet (gentle_and_sweet.json)
DIFF: Default (default.json)
DIFF: E61 with fast preinfusion to 9 bar (e61_with_fast_preinfusion_to_9_bar.json)
\`\`\`

…with no actionable detail. I had to grep the underlying TCL/JSON manually to figure out what was actually different.

**Suggested fix:** in \`tools/profile_sync.cpp:46\` \`buildDiff\`, also report:
- step count mismatch (it already prints this when both sides have frames; extend so it's the headline when one side is 0)
- \`preinfuseFrameCount\` mismatch
- (optional) profile-level scalar differences when applicable

## Why this matters

\`profile_sync\` is the canonical way to verify shipping JSONs match upstream. If it flags false positives, every downstream check has to be re-validated by hand, and real drifts will get lost in the noise. With the gaps above, today's run produces 8 noisy diffs and 0 real ones.

## Out of scope here

Not related to #968 / PR #961 — that bug never wrote to disk. This issue is purely about the developer tool.

## References

- Tool: \`tools/profile_sync.cpp\`
- App-side regeneration: \`src/controllers/profilemanager.cpp:1754\`, \`:2677\`
- Equality check: \`Profile::functionallyEqual\` at \`src/profile/profile.cpp:887\`
- CLAUDE.md path callout: \"de1app source: ... \`/Users/jeffreyh/Development/GitHub/de1app\`\"

skialpine:
Closes #972.

Tightens `profile_sync` so its output is trustworthy. Fixes 8 false-positive diffs against shipping JSONs caused by three independent tool gaps; none reflected real on-disk drift.

## Changes

- **Plugin profile scanning.** Walk `<de1plus>/plugins/*/profiles/*.tcl` (derived from the parent of the first arg). When a plugin profile produces the same output filename as a base profile, the plugin copy wins — the canonical 9-frame `A_Flow` plugin profiles now replace the stale 6-frame copies in `de1plus/profiles/`.
- **Simple-profile normalisation.** For `settings_2a`/`settings_2b`, regenerate frames before comparing (mirrors the app's runtime `regenerateSimpleFrames`) and recompute `preinfuseFrameCount` via `countPreinfuseFrames(steps)` on both sides. The TCL parser stores the literal `final_desired_shot_volume_advanced_count_start` field (0) while JSON load derives the count from frames (2) — normalising aligns the comparison with what the DE1 actually receives.
- **`buildDiff` header lines.** Always print step-count and `preinfuseFrameCount` mismatches, even when one side has 0 frames, so simple-profile diffs no longer render with empty bodies.
- **Docs.** Updated `CLAUDE.md` profile-sync callout and the tool's usage banner.

Bulk of the patch is the parse-and-compare loop becoming a parse-then-compare loop to support plugin overrides; the actual new logic is small.

## Verification

```
$ profile_sync /path/to/de1app/de1plus/profiles /path/to/Decenza/resources/profiles
Plugin profiles overriding base copies (canonical wins):
  a_flow_default_dark.json       ← .../plugins/A_Flow/profiles/A-Flow____default-dark.tcl
  a_flow_default_like_dflow.json ← .../plugins/A_Flow/profiles/A-Flow____default-like-dflow.tcl
  a_flow_default_medium.json     ← .../plugins/A_Flow/profiles/A-Flow____default-medium.tcl
  a_flow_default_very_dark.json  ← .../plugins/A_Flow/profiles/A-Flow____default-very-dark.tcl

Compare complete: 0 different, 0 missing, 89 in sync, 0 skipped
```

## Test plan

- [x] Build profile_sync (no warnings)
- [x] Run against current de1app source — 0 false-positive diffs, 89 in sync
- [x] Plugin override report lists the 4 A-Flow profiles
- [ ] Spot-check `--sync` mode does not modify any files (since 0 diffs)


skialpine:
### Code review

Found 1 issue:

1. Redundant WHAT comment on `findPluginTclFiles` — the second comment line just restates the function signature (CLAUDE.md says: "Don't explain WHAT the code does, since well-named identifiers already do that").

https://github.com/Kulitorum/Decenza/blob/e5f3d60a681990a3ae667dbd86dc1f4d19b169a6/tools/profile_sync.cpp#L135-L139

The first line ("Walk `<de1plus>/plugins/*/profiles/*.tcl` if `baseDir` looks like `<de1plus>/profiles`.") is the useful WHY/path-pattern part; the "Returns absolute file paths to every plugin TCL profile; empty list otherwise" line adds nothing the signature doesn't already say. Drop the second line.

<details>

- [Worth a look] File header (lines 12-15) claims "scanning happens when the parent of the first argument is named 'de1plus'", but `findPluginTclFiles` does no such name check — it just looks for a `plugins/` sibling directory, regardless of the parent's name. One-line doc fix in the same code block being modified. https://github.com/Kulitorum/Decenza/blob/e5f3d60a681990a3ae667dbd86dc1f4d19b169a6/tools/profile_sync.cpp#L11-L16 (score: 50)

</details>

PR #974, opened 2026-05-01
refactor(mcp): apply units-in-name convention to remaining shot detail fields (#967)

skialpine:
Closes #967.

## Summary

- Rename the last three `convertShotRecord` JSON projection keys that violated the CLAUDE.md MCP units-in-name convention: `finalWeight` → `finalWeightG`, `doseWeight` → `doseWeightG`, `duration` → `durationSec`.
- Same pattern as #964's `targetWeight` / `targetWeightG` and #969's `temperatureOverride` / `temperatureOverrideC` splits — C++ struct fields stay unsuffixed; only the externally-visible JSON key gets the unit.
- DB column names (`final_weight`, `dose_weight`, `duration_seconds`) are unchanged.
- No alias kept (per project decision in #964: "nobody will care").
- 16 files touched: 5 emit sites, plus C++/QML/JS readers, plus `tst_shotsummarizer`.

## Out of scope

`drinkTds` and `drinkEy` left as-is — `Tds` and `Ey` already imply percent per the issue note.

## Test plan

- [x] `tst_ShotSummarizer`, `tst_ShotRecordCache`, `tst_DbMigration`, `tst_ShotAnalysis` and full suite — 1755 / 1755 pass
- [ ] Manual: open Shot Detail page on an espresso shot — confirm dose / yield / duration display correctly
- [ ] Manual: ShotHistoryPage list — confirm rows show dose/yield/duration
- [ ] Manual: comparison page — confirm side-by-side metrics render
- [ ] Manual: visualizer upload of a historical shot — confirm `meta.in` / `meta.out` / `meta.time` populate correctly (visualizeruploader.cpp readers)
- [ ] Manual: dial-in MCP query — confirm `doseG` / `yieldG` / `durationSec` still appear in tool output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #976, opened 2026-05-01
refactor(history): replace shot projection map with ShotProjection Q_GADGET (#975)

### LINKED ISSUE #975: Refactor convertShotRecord projection toward typed keys (compile-time safety)
## Why

The \`shots_get_detail\` / \`shots_compare\` projection produced by [\`ShotHistoryStorage::convertShotRecord\`](https://github.com/Kulitorum/Decenza/blob/2ce54bfa356ac3843d7341744e2957acb679cbe0/src/history/shothistorystorage_serialize.cpp) is a stringly-typed \`QVariantMap\`. Every key is a free-form string both at the emit site and at every reader. The compiler can't help when keys are renamed; the type system can't help; and many of the consumer paths (MCP tools, AI candidate selection, QML→Visualizer upload) have no unit-test coverage. The result: across the three rename PRs in this sequence ([#969](https://github.com/Kulitorum/Decenza/pull/969), [#971](https://github.com/Kulitorum/Decenza/pull/971), [#974](https://github.com/Kulitorum/Decenza/pull/974)) we silently broke six consumer sites that the next reader review cycle had to catch and fix:

- \`AIManager::isMistakeShot\` (caught post-merge in #974 fix-up)
- \`AIManager::loadQualifiedShots\` (caught by #974 review)
- \`ShotSummarizer::summarizeFromHistory\` (caught by failing test)
- \`ShotSummarizer\` recent-shot history block (caught by failing test)
- \`mcptools_shots.cpp::shots_compare\` \`diffNumUnit\` (caught by #974 review)
- \`PostShotReviewPage.qml\` Visualizer update + first-upload payloads (caught by #974 review)

Two of those would have shipped silently without the review pass — \`shots_compare\` would have stopped emitting dose/yield/duration diffs, and Visualizer re-uploads would have dropped edited dose/yield values. Both are user-visible regressions that no automated test caught.

## What

Two viable scopes:

### Option A — \`ShotKeys\` namespace constants (light, recommended starting point)

Define a single header that exports key-name constants:

\`\`\`cpp
// src/history/shotprojection_keys.h
namespace decenza::ShotKeys {
    inline constexpr const char* TARGET_WEIGHT_G        = \"targetWeightG\";
    inline constexpr const char* FINAL_WEIGHT_G         = \"finalWeightG\";
    inline constexpr const char* DOSE_WEIGHT_G          = \"doseWeightG\";
    inline constexpr const char* DURATION_SEC           = \"durationSec\";
    inline constexpr const char* TEMPERATURE_OVERRIDE_C = \"temperatureOverrideC\";
    // ... ~30 keys total
}
\`\`\`

Migrate every \`result[\"X\"]\` and \`shot[\"X\"]\` site (C++ side only — QML stays unchanged because it's already accessing properties by name, not via map lookups). On the next rename, deleting \`FINAL_WEIGHT_G\` from the header turns missed readers into compile errors.

Estimated effort: ~2 hours, mechanical grep-and-replace. No behavioral change. Catches all six bugs from the recent rename history retroactively if a reader missed an update.

### Option B — \`ShotProjection\` Q_GADGET struct (full refactor)

Replace \`QVariantMap convertShotRecord(...)\` with a typed \`ShotProjection\` struct annotated with \`Q_GADGET\` so QML can still do \`shotData.finalWeightG\` via Q_PROPERTY. C++ readers use struct field access (\`shot.finalWeightG\`). JSON serialization centralizes in a single \`toJson()\` method that walks the struct fields once.

Bigger blast radius: every QML site that reads \`shotData.X\` keeps working (Q_PROPERTY makes the syntax identical), but the QML model bindings need verification. C++ \`shot[\"X\"].toDouble()\` reads become \`shot.X\` field access.

Estimated effort: ~4-6 hours. Provides compile-time safety AND struct-style ergonomics.

## Recommendation

Do **Option A** as a focused PR now — it's mechanical, low risk, catches the bug class fully, and doesn't preclude Option B. Open Option B as a deeper follow-up if struct ergonomics become valuable.

## Out of scope

- Time-series fields (\`pressure\`, \`flow\`, etc.) — those are nested objects of \`{x, y}\` points, not affected by the rename pattern.
- The write-metadata path used by \`requestUpdateShotMetadata\` (a separate stringly-typed contract from QML to \`updateShotMetadataStatic\`'s \`fieldMap\`) — distinct from the read projection. Could be done in a separate PR.
- Visualizer external schema (snake_case \`bean_weight\` etc.) — external API, not internal projection.

skialpine:
Closes #975.

## Summary

- Adds [`ShotProjection`](src/history/shotprojection.h) — a Q_GADGET value type that replaces the stringly-typed `QVariantMap` returned by `ShotHistoryStorage::convertShotRecord`. Renames now fail to compile at every C++ consumer site instead of silently dropping data — the bug class that #969, #971, and #974 each had to chase through six readers.
- QML keeps reading fields by name (`shotData.finalWeightG`) via Q_PROPERTY reflection; field names match the legacy map keys exactly. A registered `QVariantMap → ShotProjection` meta-converter lets `Q_INVOKABLE` methods accept JS objects (e.g. `PostShotReviewPage`'s `Object.assign`'d `editShotData`) without QML changes.
- C++ consumers migrated to field access: `AIManager` (`isMistakeShot`, `loadQualifiedShots`, `generateHistoryShotSummary`), `ShotSummarizer` (`summarizeFromHistory`, recent-shot block), MCP (`shots_get_detail`, `shots_compare` with member-pointer `diffStr`/`diffNum` helpers, `dialing_get_context`), `VisualizerUploader` (`uploadShotFromHistory`, `updateShotOnVisualizer`, `buildHistoryShotJson`), `ShotServer` (list + detail HTML), and `ShotHistoryExporter`.
- The `shotReady` signal now emits `const ShotProjection&`. Registered as a metatype in [main.cpp](src/main.cpp) so QueuedConnection across thread boundaries works.

## Out of scope (per the issue)

Time-series fields stay as `QVariantList<{x,y}>`, the metadata write path through `requestUpdateShotMetadata` is a separate stringly-typed contract, and visualizer.coffee's external snake_case schema is unchanged.

## Test plan

- [x] Full build clean (Qt Creator, macOS Debug)
- [x] Full test suite green: 1843 passed / 0 failed
- [ ] Smoke test on device: open a shot from history (PostShotReviewPage / ShotDetailPage), edit metadata, upload to Visualizer
- [ ] Smoke test MCP: `shots_get_detail`, `shots_compare`, `dialing_get_context` against a live shot
- [ ] Spot-check QML graph rendering on `LastShotItem` (uses `shot.pressure`, `shot.flow`, etc. via Q_PROPERTY)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,6 +293,7 @@ set(SOURCES
     src/history/shothistorystorage_internal.cpp
     src/history/shothistorystorage_serialize.cpp
     src/history/shothistorystorage_queries.cpp
+    src/history/shotprojection.cpp
     src/history/shotdebuglogger.cpp
     src/history/shotfileparser.cpp
     src/history/shotimporter.cpp
@@ -458,6 +459,7 @@ set(HEADERS
     src/history/shothistory_types.h
     src/history/shothistorystorage.h
     src/history/shothistorystorage_internal.h
+    src/history/shotprojection.h
     src/history/shotdebuglogger.h
     src/history/shotfileparser.h
     src/history/shotimporter.h
--- a/src/ai/aimanager.cpp
+++ b/src/ai/aimanager.cpp
@@ -353,7 +353,7 @@ QString AIManager::generateShotSummary(ShotDataModel* shotData,
     return m_summarizer->buildUserPrompt(summary);
 }
 
-QString AIManager::generateHistoryShotSummary(const QVariantMap& shotData)
+QString AIManager::generateHistoryShotSummary(const ShotProjection& shotData)
 {
     ShotSummary summary = m_summarizer->summarizeFromHistory(shotData);
     return m_summarizer->buildUserPrompt(summary);
@@ -368,12 +368,12 @@ void AIManager::setShotHistoryStorage(ShotHistoryStorage* storage)
 // Returns (timestamp, fullShot) pairs. Extracted from requestRecentShotContext
 // to reduce lambda nesting. NOT safe to call from the main thread (would conflict
 // with the primary DB connection).
-static QList<QPair<qint64, QVariantMap>> loadQualifiedShots(
+static QList<QPair<qint64, ShotProjection>> loadQualifiedShots(
     const QString& dbPath,
     const QString& beanBrand, const QString& beanType,
     const QString& profileName, int excludeShotId)
 {
-    QList<QPair<qint64, QVariantMap>> qualifiedShots;
+    QList<QPair<qint64, ShotProjection>> qualifiedShots;
 
     withTempDb(dbPath, "ai_context", [&](QSqlDatabase& db) {
         // 1. Look up the current shot's timestamp
@@ -444,7 +444,7 @@ static QList<QPair<qint64, QVariantMap>> loadQualifiedShots(
                 continue;
             }
 
-            QVariantMap fullShot;
+            ShotProjection fullShot;
             try {
                 ShotRecord record = ShotHistoryStorage::loadShotRecordStatic(db, c.id);
                 fullShot = ShotHistoryStorage::convertShotRecord(record);
@@ -452,14 +452,13 @@ static QList<QPair<qint64, QVariantMap>> loadQualifiedShots(
                 qWarning() << "  Shot id=" << c.id << "-> SKIPPED (exception:" << e.what() << ")";
                 continue;
             }
-            if (fullShot.isEmpty()) {
+            if (!fullShot.isValid()) {
                 qWarning() << "  Shot id=" << c.id << "-> SKIPPED (convertShotRecord returned empty)";
                 continue;
             }
 
             // Check targetWeight-based mistake filter (needs full record)
-            double targetWeight = fullShot.value("targetWeightG", 0.0).toDouble();
-            if (targetWeight > 0.0 && c.finalWeight < targetWeight / 3.0) {
+            if (fullShot.targetWeightG > 0.0 && c.finalWeight < fullShot.targetWeightG / 3.0) {
                 qDebug() << "  Shot id=" << c.id << "-> SKIPPED (mistake, weight < 1/3 target)";
                 continue;
             }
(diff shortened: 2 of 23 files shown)


skialpine:
Smoke-tested against a Decenza-Desktop debug build on macOS, MCP at `localhost:8888/mcp`:

- `shots_get_detail` (shot 884): all units-suffixed top-level keys present and populated — `finalWeightG=35.9`, `doseWeightG=20`, `durationSec=26.242`, `targetWeightG=36`, `temperatureOverrideC=82`, `timestampIso`, `dateTime`. Time-series arrays (pressure / flow / weight / temperature) all populated with `{x, y}` points. `detectorResults.grind.gates` exposes the full Arm-2 input set.
- `shots_compare` (882, 883, 884): diff loop emits `durationSec`, `yieldG`, `grinderSetting` rows; `doseG` correctly suppressed (all three shots are 20g). Member-pointer `diffStr`/`diffNum` accessors work end-to-end.
- `dialing_get_context` (shot 884, `history_limit: 3`): `shot` summary and `dialInHistory[*]` rows populate `doseG`/`yieldG`/`durationSec`/`enjoyment0to100`/`temperatureOverrideC`/`targetWeightG` correctly.

Both `dialing_get_context.dialInHistory` (which feeds from `loadRecentShotsByKbIdStatic` → `ShotProjection::fromVariantMap`) and the direct `convertShotRecord` path produce matching outputs, confirming the round-trip through `fromVariantMap` is lossless for the fields that matter to consumers.

skialpine:
### Code review

Found 4 issues:

1. `ShotProjection::toVariantMap()` emits the raw `timestamp` qint64 (Unix epoch seconds) alongside `timestampIso`, which propagates to MCP `shots_get_detail` / `shots_compare` JSON via `toJsonObject()`. CLAUDE.md MCP convention says "Never return Unix timestamps. Use ISO 8601 with timezone." (`docs/CLAUDE_MD/MCP_SERVER.md`). This is pre-existing (the prior `convertShotRecord` QVariantMap emitted the same key) but is in code being modified.

https://github.com/Kulitorum/Decenza/blob/74ec7f396f76a2c8a2aa1f70429c983acba19342/src/history/shotprojection.cpp#L10-L14

2. `ShotProjection::dateTime` has a dual identity that the refactor preserves but doesn't address. `convertShotRecord()` writes a locale-formatted string (`"yyyy-MM-dd HH:mm:ss"`), while `loadRecentShotsByKbIdStatic()` writes ISO 8601. The dialing tool reads `shot.dateTime` and emits it under MCP key `"timestamp"`. Today this path is ISO (compliant); but if a future caller swaps to the `convertShotRecord` projection, the locale string silently violates the "ISO 8601 with timezone" rule. Safer to read `shot.timestampIso` here (and populate `timestampIso` from the queries path too).

https://github.com/Kulitorum/Decenza/blob/74ec7f396f76a2c8a2aa1f70429c983acba19342/src/mcp/mcptools_dialing.cpp#L103-L107

3. The doc comment for `toVariantMap()` lists callers that don't actually use it. The shot-list HTML page (`shotserver_shots.cpp::generateShotListPage`) consumes via `ShotProjection::fromVariantMap` then reads typed fields directly — it never calls `toVariantMap()`. The visualizer/exporter path (`visualizeruploader.cpp::buildHistoryShotJson`) takes `const ShotProjection&` and reads fields directly. Only the MCP `toJsonObject()` callers actually delegate through it.

https://github.com/Kulitorum/Decenza/blob/74ec7f396f76a2c8a2aa1f70429c983acba19342/src/history/shotprojection.h#L160-L165

4. Stale slow-path comment in `shotsummarizer.cpp` references "any QVariantMap that didn't flow through convertShotRecord" — but `summarizeFromHistory` now takes `const ShotProjection&`, so callers cannot pass a raw `QVariantMap` across this boundary anymore.

https://github.com/Kulitorum/Decenza/blob/74ec7f396f76a2c8a2aa1f70429c983acba19342/src/ai/shotsummarizer.cpp#L483-L486

<details>

- [Worth a look] `shothistorystorage_serialize.cpp` lines 99-108 had its slow-path explanation removed, leaving the `analysisOwned` ownership pattern in the `else` branch undocumented — easy one-line fix in code already being modified (score: 55)

</details>

PR #981, opened 2026-05-01
fix(mcp): gate spec-versioned tool fields on negotiated protocol version

skialpine:
## Summary
Closes the entire #953 spec-version blast radius and a related HTTP-session leak that was wedging the session pool.

### Spec-version gating
`McpToolRegistry::listTools()` and four other surfaces unconditionally emitted spec-versioned fields, so strict 2024-11-05 validators (Claude Code) rejected the responses and surfaced zero tools.

| Surface | Field | Spec | Status |
|---|---|---|---|
| `tools/list` | `title` | 2025-06-18 | gated on `>= 2025-06-18` |
| `tools/list` | `icons` | 2025-11-25 | gated on `>= 2025-11-25` |
| `tools/list` | `inputSchema.$schema` | 2025-11-25 | moved from registration to `listTools()`, gated on `>= 2025-11-25` |
| `resources/list` | `title` | 2025-06-18 | gated on `>= 2025-06-18` |
| `resources/list` | `icons` | 2025-11-25 | gated on `>= 2025-11-25` |
| `tools/call` | `structuredContent` | 2025-06-18 | gated on `>= 2025-06-18` |
| `tools/call` | `resource_link` content blocks | 2025-06-18 | gated on `>= 2025-06-18` |
| `resources/read` | `structuredContent` | 2025-06-18 | gated on `>= 2025-06-18` |

Deferred response paths (chat confirmation, in-app confirmation, async tool dispatch, async resource read) capture the negotiated protocol version at request time so held responses honour the originating session's spec even after the session pointer may have changed.

### Session leak (HTTP transport)
`findOrCreateSession()` previously only purged sessions that had established *and lost* an SSE socket. Pure-HTTP transports (Claude Code's `"type": "http"`) never establish SSE, so each reconnect leaked a slot until the 30-min idle timeout, wedging the pool at `MaxSessions=8` with `Too many sessions` errors. Now also reaps non-SSE sessions whose `lastActivity` is older than 60s — matching the equivalent SSE-loss heuristic without double-tapping live clients.

### Regression introduced by
#953 ("Adopt MCP spec 2025-11-25") — added every field above without conditional emission.

## Test plan
- [x] Qt Creator build succeeds with 0 errors / 0 warnings
- [x] All 27 `tst_McpServerProtocol` tests pass (including 4 new pinning tests for 2024-11-05 contract)
- [x] `curl localhost:8888/mcp` with `protocolVersion: "2024-11-05"` returns tools with only `name`/`description`/`inputSchema` (no `title`, no `icons`, no `$schema`)
- [x] Same probe with `protocolVersion: "2025-11-25"` returns all fields
- [ ] After merge + rebuild + Claude Code restart, `mcp__decenza__*` tools surface in Claude Code's deferred-tool list

### DIFF
--- a/docs/MCP_TEST_PLAN.md
+++ b/docs/MCP_TEST_PLAN.md
@@ -33,6 +33,7 @@ If ORIGINAL_MODIFIED is true, warn the user — tests will modify the profile an
 | Date | Tools | Passed | Skipped | Failed | Notes |
 |------|-------|--------|---------|--------|-------|
 | 2026-03-20 | 37 | 41 | 1 | 0 | Initial run, simulator mode |
+| 2026-04-30 | 33 | 56 | 1 | 0 | Simulator, MCP 2025-11-25 negotiated. Plan refreshed for PR #976 field renames, `profiles_save` readonly refusal, `profiles_create` empty-filename, removal of `preferences` category, and §4.10 active-delete behavior. |
 
 ---
 
@@ -100,9 +101,11 @@ Verify: editorType values include at least "pressure", "flow", "dflow", "aflow",
 ### 3.2 profiles_get_active
 ```
 Call: profiles_get_active
-Expect: filename non-empty, editorType non-empty, modified is boolean, targetWeightG > 0
+Expect: filename non-empty, editorType non-empty, modified is boolean,
+        targetWeightG > 0, targetTemperatureC > 0, readOnly is boolean
 Save: note the filename and editorType as ORIGINAL_PROFILE and ORIGINAL_EDITOR
 ```
+Note: target temperature field is `targetTemperatureC` (suffix), not `targetTemperature`.
 
 ### 3.3 profiles_get_detail
 ```
@@ -168,7 +171,7 @@ Save: note tempStart as ORIGINAL_TEMP
 ```
 Call: profiles_edit_params (tempStart: ORIGINAL_TEMP+2, tempPreinfuse: ORIGINAL_TEMP+2, tempHold: ORIGINAL_TEMP+2, tempDecline: ORIGINAL_TEMP+2, confirmed: true)
 Expect: success=true, editorType="pressure", modified=true
-Verify: profiles_get_active → targetTemperature = ORIGINAL_TEMP+2, modified=true
+Verify: profiles_get_active → targetTemperatureC = ORIGINAL_TEMP+2, modified=true
 ```
 
 ### 4.2 Restore temperature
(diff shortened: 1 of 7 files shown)


skialpine:
Adding a third commit (`95f7bbd3`) for a related but distinct bug uncovered while verifying this PR.

**Symptom:** Even after the gating fixes in 6b968e02 + 46a04570, Claude Code surfaced **zero** decenza tools despite reporting the server as "connected" in `/mcp`.

**Root cause:** The shape of `icons[].sizes` was wrong, independent of gating. The MCP icon schema (HTML-style) requires `string[]`, but `iconsArrayFromQrc()` in `src/mcp/mcptoolregistry.h` emitted a bare string `"any"`. Claude Code's zod validator rejected every tool entry in `tools/list` and every entry in `resources/list` with `expected: "array", received: "string"`, dropping all 50 tools + 8 resources. The `initialize` handshake still succeeded, hence "connected" + empty tool surface.

This bug only manifests when a client negotiates ≥ 2025-11-25 (where icons are emitted). The earlier gating commits prevented the field from reaching strict 2024-11-05 clients, but newer clients that *do* receive icons hit this shape error.

**Fix:** one-line change to wrap `"any"` in a `QJsonArray`, plus a regression assertion in `toolsListIncludesTitleAndIcons` that pins `icon.sizes` to an array of strings so this can't silently regress.

**Verification:** Qt Creator build clean (0/0). `tst_McpServerProtocol`: 27/27 pass (was 27/27 before, same suite, with the new `sizes` assertion). After this commit + a `/mcp reconnect`, all 50 `mcp__decenza__*` tools surface in Claude Code's deferred-tool list.

skialpine:
Adding a fourth commit (\`2c97a89f\`) for the next bug uncovered while exercising MCP tools end-to-end.

**Symptom:** With the prior three commits + a \`/mcp reconnect\`, all 50 \`mcp__decenza__*\` tools surfaced (good — \`tools/list\` works). But every \`tools/call\` against any tool that ships resource links — \`shots_list\`, \`machine_get_state\`, \`profiles_list\`, \`profiles_get_active\`, \`profiles_get_detail\`, \`shots_get_detail\`, \`machine_get_telemetry\` — failed with \`Invalid input: expected string, received undefined\` at \`path: ["content", N, "name"]\`.

**Root cause:** MCP 2025-06-18 made \`resource_link\` content blocks share the \`Resource\` shape, where \`name\` is REQUIRED. The seven \`_resourceLinks\` emitter sites in \`src/mcp/mcptools_*.cpp\` only set \`uri\`/\`title\`/\`mimeType\` — \`name\` was never set, and \`buildToolCallResponse\` didn't synthesize one. Every tools/call response that ships resource links had its content[] entries rejected by strict clients (e.g. Claude Code's zod validator).

**Fix:** centralize in \`buildToolCallResponse\` — prefer a \`name\` from the side-channel entry when emitters supply one, otherwise derive it from the uri's last path segment (e.g. \`decenza://shots/884\` → \`"884"\`). This covers all seven existing emitters without per-site edits and gives future emitters a sensible default. Pin the contract by extending \`toolsCallEmitsResourceLinkBlocksFromSideChannel\` to assert the \`name\` field is present and matches the uri-tail derivation.

**Verification:** Qt Creator build clean (0/0). \`tst_McpServerProtocol\`: 27/27 pass (same suite, augmented assertions). End-to-end via Claude Code MCP tools is gated on relaunching the Decenza app so the running binary has the fix; will verify once that's done.

Bundling per the user's "fix all of it in the current pr" guidance — same #953 spec-version blast radius.

PR #1028, opened 2026-05-01
feat(mcp,ai): optimize dialing_get_context payload — PR 1 cost wins

skialpine:
## Summary

PR 1 of the [`optimize-dialing-context-payload` openspec change](../tree/feat/optimize-dialing-context-pr1-cost-wins/openspec/changes/optimize-dialing-context-payload). Pure cost wins — same facts, fewer tokens. PR 2 (canonical-source separation, `result.profile`, prose Profile/Coffee/Grinder strip, history-block render mode) follows separately so it can be reverted without losing PR 1's wins.

Tasks 1–7 of the spec implemented:

**JSON shape (4 wins):**
- `dialInSessions[].context` hoists shared shot identity (grinderBrand/grinderModel/grinderBurrs/beanBrand/beanType). Per-shot entries carry only fields that differ. New helper `hoistSessionContext` in `mcptools_dialing_helpers.h`. (Task 1)
- Drop `result["shot"]` JSON block — its fields are all already in `shotAnalysis` prose. (Task 2)
- `currentBean.beanFreshness` block replaces `daysSinceRoast` + `daysSinceRoastNote`. Carries `roastDate` + `freshnessKnown:false` + an imperative instruction to ASK about storage before quoting age. **No precomputed day count under any field name** — calendar age without storage context (frozen vs counter) was misleading data the AI was anchoring on. (Task 6)
- `tastingFeedback.recommendation` framing string moved to system prompt. (Task 4)
- `grinderContext.settingsObserved` is now bean-scoped via a new optional `beanBrand` parameter on `queryGrinderContext` (backwards-compatible default). When the bean-scoped query returns < 2 distinct settings the response also emits `allBeansSettings` (cross-bean) so the AI sees overall range — explicitly tagged. (Task 7)

**Prose shape (3 wins):**
- `## Detector Observations` legend preamble moved out of every `buildUserPrompt` body and into the system prompt's new `## Reading Detector Observations` section. Per-line `[warning]` / `[good]` tags still emit. (Task 3)
- "(N days since roast, ...)" parenthetical replaced with "(ask user about storage before reasoning about age)" — no day count. (Task 6.3)
- New system prompt section `## How to Read Structured Fields` teaches the AI to gate on `tastingFeedback.has*`, `currentBean.beanFreshness.freshnessKnown`, and `currentBean.inferredFields` — replacing per-call framing strings with a single per-conversation lesson. (Task 4.4)

## Empirical impact

Measured against `tests/data/dialing/northbound_80s_espresso_pre_change_baseline.json` (frozen pre-change capture, 14,652 chars / ~3,663 tokens of context per turn):

- "Detector Observations" legend repetitions: 0 (was 2) — saves ~860 chars
- Roast-date day-count parenthetical: 0 (was 4) — saves ~240 chars
- `tastingFeedback.recommendation`: 0 (was 1) — saves ~210 chars
- `currentBean.daysSinceRoast` precomputed day count: 0 (was 1) — removes the AI-anchoring failure mode

Total PR 1 reduction: **~1,200 chars (~300 tokens, ~8%)** on the iteration-2 turn. PR 2 will land the larger wins (Profile intent + Recipe hoisting, prose Coffee/Grinder strip) for the remaining ~25–30%.

## Test plan

- [x] `TstMcpToolsDialingHelpers`: 19 pass (8 existing groupSessions + 3 new `buildBeanFreshness` + 6 new `hoistSessionContext`)
- [x] `tst_ShotSummarizer`: 19 pass (16 existing + 3 new for prose roast-date format, detector legend migration, system-prompt structural-field guidance)
- [x] Full test suite: 1,885 passed, 0 failed, **0 warnings** (per `docs/CLAUDE_MD/TESTING.md` §Test warnings — added `qt_add_resources(... ai.qrc)` to the `tst_shotsummarizer` target so `ShotSummarizer::shotAnalysisSystemPrompt` finds its resources silently in the test build).
- [ ] Live verification: re-render the Northbound 80's Espresso 4-shot iteration through `dialing_get_context` against the simulator and confirm session.context hoisting + beanFreshness shape + per-bean grinderContext.

## Forward references

- The OpenSpec change is in this PR under `openspec/changes/optimize-dialing-context-payload/`. Review there for the full proposal, design, tasks, and spec delta.
- Empirical baselines are saved to `tests/data/dialing/` (frozen pre-change captures + a README documenting the schema, measured numbers, regression checks, and maintenance policy).
- PR 2 will follow on a separate branch once this lands.


skialpine:
### Code review

Found 2 issues:

1. **Spec internal contradiction on prose roast-date — implementation is consistent with one half of the spec, contradicts the other.** The "single canonical surface for the user's roast date" requirement states the prose SHALL NOT contain the literal `"roasted YYYY-MM-DD"` string, but an earlier requirement under task 6 / beanFreshness says the prose SHALL contain `"roasted YYYY-MM-DD"`. The implementation keeps the date in prose, matching PR 1 scope per `design.md` (strip-from-prose is deferred to PR 2). Fix the spec wording: either scope the "SHALL NOT" requirement to PR 2, or qualify it with a phasing note so a future reader doesn't treat the implementation as non-compliant.

   https://github.com/Kulitorum/Decenza/blob/8c4c86be6c0584dca7fafd1b8ad0498b7d2b64c2/openspec/changes/optimize-dialing-context-payload/specs/dialing-context-payload/spec.md#L228-L240

   https://github.com/Kulitorum/Decenza/blob/8c4c86be6c0584dca7fafd1b8ad0498b7d2b64c2/src/ai/shotsummarizer.cpp#L555-L568

2. **Silent inheritance of session.context for shot[0] when its identity field is empty.** In `hoistSessionContext`, when shot[0] has an empty grinder/bean field but a later shot has a value, `firstNonEmpty` picks the later shot's value as the session context, and shot[0]'s override is left empty (so the JSON serializer omits it). The AI then sees shot[0] inheriting a value the shot itself never recorded. The test `hoistSessionContext_firstShotEmptyForField_fallsBackToFirstNonEmpty` pins this as intentional, and in practice grinder identity rarely changes within a session, so the inferred value is usually correct. But the AI cannot distinguish "shot's field was unrecorded" from "shot agrees with session" — for legacy shots imported without burr metadata, this attaches a modern burr label as fact. Worth either documenting the inheritance contract explicitly in the system prompt's "How to Read Structured Fields" section, or having the override carry an explicit `null` for the unknown case.

   https://github.com/Kulitorum/Decenza/blob/8c4c86be6c0584dca7fafd1b8ad0498b7d2b64c2/src/mcp/mcptools_dialing_helpers.h#L60-L100

   https://github.com/Kulitorum/Decenza/blob/8c4c86be6c0584dca7fafd1b8ad0498b7d2b64c2/tests/tst_mcptools_dialing_helpers.cpp#L266-L304

<details>

- [Worth a look] When the bean-scoped `queryGrinderContext` returns 0 distinct settings (narrow case: imported shot with a bean brand never used elsewhere, or shot whose own `grinder_setting` is empty), the outer guard `if (!ctx.settingsObserved.isEmpty())` skips the entire `grinderContext` block — the `allBeansSettings` fallback never fires, so the response loses cross-bean context the pre-PR shape would have provided. One-line fix: hoist the cross-bean fallback above the outer guard, or duplicate the fallback at the empty-bean-scoped branch (score: 50)
- [OK to skip] `queryGrinderContext` declaration comment in `shothistorystorage.h:89` documents the `beanBrand` parameter but doesn't mention that the two-tier `allBeansSettings` fallback (when bean-scoped is sparse) lives in the **caller** (`mcptools_dialing.cpp`), not in this function. Could mislead a future maintainer who tries to implement the fallback inside the query function. Minor docs polish (score: 50)

</details>

PR #1027, opened 2026-05-01
feat(shotserver): add quality badges + shot summary, switch ratings to N%

skialpine:
## Summary
- Adds the quality-badge chips and Shot Summary modal to the web shot detail page, mirroring the in-app `QualityBadges` + `ShotAnalysisDialog`. Channeling, temp-unstable, grind-issue, puck-failed, first-step-skipped, or a green "Clean extraction" chip render alongside the Rating; the Shot Summary button opens a modal listing the analyzeShot() observations and verdict.
- Switches shot ratings from 5-star UI to numeric `N%` everywhere — detail page (`85%` / `-`), list page chip (hidden when unrated), and edit-mode 0-100 number input — matching `ShotDetailPage.qml`, `ShotHistoryPage.qml`, and `RatingInput.qml`. `data-rating` and the `rating:N+` search syntax keep using the numeric value.

Data flows through the existing `ShotProjection` (no backend changes) — `summaryLines` and detector flags are already populated by `convertShotRecord`.

## Test plan
- [ ] Open the shot list — rating chips show `N%` and disappear for unrated shots; clicking a chip filters with `rating:N+`
- [ ] Open a shot detail page — Rating reads `85%` (or `-`) instead of stars
- [ ] Verify badge chips appear next to Rating and reflect the same flags as the in-app shot detail
- [ ] Click Shot Summary — modal opens with the same observation lines + verdict as the in-app dialog
- [ ] Edit a shot — Rating field is a 0-100 number input, save persists the value, list/detail re-render with the new percentage
- [ ] Confirm the modal closes via OK, backdrop click, and survives edit-mode cancel/restore

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1029, opened 2026-05-01
fix(shotserver): replace stray `%%` literals in raw CSS/HTML with `%`

skialpine:
## Summary
Follow-up to #1027. Several CSS rules in the web shotserver shipped to browsers with `100%%`, `60%%`, `0%%`, and one cell with text `(%%)`, because the source had `%%` written for a literal `%` that was never going through `QString::arg()` (or, in the MCP setup page, was going through arg() — but Qt's arg() does NOT reduce `%%` back to `%`, so the doubling is wrong in both paths). Browsers reject `100%%` as an invalid CSS value and silently fall back to `auto`, so these rules were quietly broken:

- `.menu-dropdown { top: 100% }` (shot list page + shared `webtemplates/menu_css.h`)
- `.sort-dropdown { top: 100% }`
- `.help-table { width: 100% }`
- `.shot-beans { max-width: 60% }`
- `.progress-fill { height: 100%; width: 0% }` (restore page progress bar)
- MCP setup page table `width:100%`
- Search-help cell `Extraction yield (%)` (was rendering "(%%)")

Empirical Qt 6.10 behavior (minimal repro):

```
QString("width: 100%%").arg("X")     → "width: 100%%"
QString("width: 100%%; %1").arg("X") → "width: 100%%; X"
```

`%%` is preserved verbatim — the "double the percent" convention was cargo-cult and produced wrong output. The two `% → %%` doubling lambdas (`escapeForJs`, `jsEscape`) were already fixed in #1027; this PR cleans up the remaining literal occurrences in the source.

## Test plan
- [ ] Open the shot list — `.shot-beans` truncates beans column at 60% (was unbounded), and the burger-menu / sort dropdowns position directly below their buttons (was floating at `top: auto`)
- [ ] Open the restore-backup page — uploading a backup shows the progress bar fill with non-zero height (was rendering as a 0-height div)
- [ ] Open the MCP setup page — table fills the full width
- [ ] Open the search help on the shot list — "Extraction yield (%)" reads with one percent sign

### DIFF
--- a/src/network/shotserver.cpp
+++ b/src/network/shotserver.cpp
@@ -1426,7 +1426,7 @@ a{color:#6c8cff}
 <p>The MCP server runs inside Decenza on <strong>any platform</strong> (Android tablet, iOS, Windows, macOS, Linux). Claude Desktop on your computer connects to it over your local WiFi network.</p>
 
 <h2>Platform Compatibility</h2>
-<table style="width:100%%;border-collapse:collapse;margin:12px 0">
+<table style="width:100%;border-collapse:collapse;margin:12px 0">
 <tr style="border-bottom:1px solid #333"><td style="padding:8px"><strong>Decenza runs on</strong></td><td style="padding:8px">Any platform &mdash; tablet, phone, or desktop</td></tr>
 <tr style="border-bottom:1px solid #333"><td style="padding:8px"><strong>Claude Desktop connects from</strong></td><td style="padding:8px">macOS or Windows (same WiFi network)</td></tr>
 )HTML"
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Verified each of the 9 changed sites: 8 are in raw `html += R"HTML(...)"` blocks that never pass through `QString::arg()`, and the one site that does ([shotserver.cpp:1429](https://github.com/Kulitorum/Decenza/blob/73adc06d55361a6175a8c21856ab41e15dc4053a/src/network/shotserver.cpp#L1428-L1430)) has the new `%` followed by `;` — not a digit, so it cannot collide with the `%1`/`%2`/`%3` placeholders consumed at [shotserver.cpp:1575](https://github.com/Kulitorum/Decenza/blob/73adc06d55361a6175a8c21856ab41e15dc4053a/src/network/shotserver.cpp#L1574-L1578). False-negative scan across all `src/network/shotserver*.cpp` and `src/network/webtemplates/*.h` files: zero remaining `%%` literals outside comments.

PR #1030, opened 2026-05-01
feat(mcp,ai): canonical-source separation for dialing_get_context — PR 2

skialpine:
## Summary

PR 2 of the [`optimize-dialing-context-payload`](../tree/main/openspec/changes/optimize-dialing-context-payload) OpenSpec change. Builds on PR 1's cost wins (#1028) with the canonical-source separation step:

- **Hoist profile metadata to `result.profile`** (task 8). Replaces the legacy `result.currentProfile` block. New shape carries `filename`, `title`, `intent`, `recipe`, `targetWeightG`, `targetTemperatureC`, `recommendedDoseG`. The intent + recipe describe the resolved shot's profile (read from `dbResult.shotData.profileNotes` / `profileJson` — already in memory, no extra DB hit); targets describe the live profile loaded on the machine. Asymmetry is intentional — see `design.md`.
- **Strip shot-invariant identity from `shotAnalysis` prose** (task 9). Profile / Coffee / Grinder brand+model+burrs / roasted-date lines all leave the prose body. Per-shot variable `Grind setting:` stays. Bean identity now lives only in `currentBean.*`; profile identity only in `result.profile.*`; grinder model + burrs only in `currentBean.grinder*` / `dialInSessions[].context`.
- **Add `RenderMode { Standalone, HistoryBlock }` to `ShotSummarizer::buildUserPrompt`** (task 10). `HistoryBlock` skips ONLY the two top-level header lines (`## Shot Summary`, `## Detector Observations`) — body content (dose, yield, peaks, phase data, per-line detector tags, tasting feedback) is identical between modes. Default is `Standalone` (preserves single-shot dialog + `dialing_get_context.shotAnalysis` rendering).
- **Rewire `AIManager::requestRecentShotContext`** to emit a single `### Profile: <title>` + intent + recipe header at the top of the history section, plus a single `### Setup: <grinder> on <bean>` subsection when all shots in the history share grinder + bean. Per-shot blocks now use `HistoryBlock` mode.
- **`buildHistoryContext` symmetrically hoists** Profile + Recipe to a single header at the top.
- **System prompt teaches canonical sources** — `result.profile` for profile metadata, `currentBean` for bean / grinder identity, `dialInSessions[].context` for session-shared identity. Replaces per-call framing strings.

[...]

PR #1031, opened 2026-05-01
feat(mcp,ai): add ai_advisor_invoke MCP tool for prompt A/B testing

skialpine:
## Summary

New write-tier MCP tool that exercises the configured AI advisor end-to-end. Useful for:
- **Prompt A/B testing** — capture before/after responses on the same shot to validate prompt-side changes (like the canonical-source separation in #1030) don't regress advice quality.
- **Cost-free prompt preview** — pass \`dryRun: true\` to assemble the system + user prompts and return them without sending anywhere.
- **End-to-end advisor validation** — verify the configured provider + model + system prompt actually round-trip cleanly.

Always echoes \`systemPromptUsed\` + \`userPromptUsed\` in the response so the caller sees exactly what reached the provider — not just the response.

## Tool shape

\`ai_advisor_invoke\`, write tier:

\`\`\`json
{
  "shot_id": 42,                    // optional; defaults to most recent shot
  "dryRun": false,                  // optional; default false
  "userPromptOverride": "...",      // optional; replaces auto-built user prompt
  "systemPromptOverride": "..."     // optional; replaces auto-built system prompt
}
\`\`\`

Response:

\`\`\`json
{
  "shotId": 42,
  "provider": "anthropic",
  "model": "claude-sonnet-4-6",
  "systemPromptUsed": "...",
  "userPromptUsed": "...",
  "response": "...",                // when not dryRun
  "latencyMs": 4231,                // when not dryRun
  "dryRun": true                    // when dryRun
}
\`\`\`

Errors as \`{"error": "..."}\` for: unconfigured provider, busy advisor, missing shot, timeout (75s wrapper around the provider's 60s).

## Implementation notes

- Routes through \`AIManager::analyze\`, so the response also reaches the in-app conversation overlay (\`lastRecommendation\` / \`recommendationReceived\`). This is documented in the tool description; users can switch to a non-default provider via settings if they want isolation, or fire \`dryRun: true\` calls that skip the provider entirely.
- Pattern matches \`dialing_get_context\`: SQL on a worker thread, main-thread hop for AIManager + ShotSummarizer access (neither is thread-safe).
- 75s wrapper timeout sits just above the provider's own 60s \`ANALYSIS_TIMEOUT_MS\` so callers always get a clean reply rather than a dangling promise.
- Re-checks the busy gate after the background DB load completes — the user might have triggered an in-app advisor call between the gate check and the main-thread hop.
- \`done\` flag on the call state guards against double-fire (provider's own timeout could in theory fire alongside our wrapper's; rare but harmless).

## Test plan

- [x] Build via Qt Creator MCP — succeeded with 0 errors, 0 warnings
- [x] \`tst_McpServerSession\` (16 tests) — pass
- [x] \`tst_McpServerProtocol\` (29 tests) — pass
- [ ] Manual smoke test: configure a provider, invoke \`ai_advisor_invoke\` with \`dryRun: true\`, confirm the assembled prompts match what the in-app advisor sees
- [ ] Manual smoke test: same with \`dryRun: false\`, confirm round-trip works and \`response\` is populated
- [ ] Verify the in-app overlay receives the response (intended side effect — documented)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -322,6 +322,7 @@ set(SOURCES
     src/mcp/mcptools_devices.cpp
     src/mcp/mcptools_write.cpp
     src/mcp/mcptools_agent.cpp
+    src/mcp/mcptools_ai.cpp
     src/network/mqttclient.cpp
     src/network/webdebuglogger.cpp
     src/network/locationprovider.cpp
--- a/src/mcp/mcpserver.cpp
+++ b/src/mcp/mcpserver.cpp
@@ -54,6 +54,7 @@ void registerDeviceTools(McpToolRegistry* registry, BLEManager* bleManager, DE1D
 class MemoryMonitor;
 void registerDebugTools(McpToolRegistry* registry, MemoryMonitor* memoryMonitor);
 void registerAgentTools(McpToolRegistry* registry);
+void registerAITools(McpToolRegistry* registry, MainController* mainController);
 void registerMcpResources(McpResourceRegistry* registry, DE1Device* device,
                           MachineState* machineState, ProfileManager* profileManager,
                           ShotHistoryStorage* shotHistory, MemoryMonitor* memoryMonitor,
@@ -160,6 +161,7 @@ void McpServer::registerAllTools()
     registerDeviceTools(m_toolRegistry, m_bleManager, m_device);
     registerDebugTools(m_toolRegistry, m_memoryMonitor);
     registerAgentTools(m_toolRegistry);
+    registerAITools(m_toolRegistry, m_mainController);
     qDebug() << "McpServer: Registered"
              << m_toolRegistry->listTools(2, QStringLiteral("2025-11-25")).size() << "tools";
 }
--- a/src/mcp/mcptools_ai.cpp
+++ b/src/mcp/mcptools_ai.cpp
@@ -21,27 +21,24 @@
 #include <QThread>
 #include <QTimer>
 
-// Hard cap on a single advisor call. Provider-side timeout is 60s
-// (AIProvider::ANALYSIS_TIMEOUT_MS). This adds a small buffer so the MCP
-// caller gets a clean "timeout" reply rather than a dangling promise if
-// the provider's own timeout fires.
-static constexpr int kAdvisorMcpTimeoutMs = 75 * 1000;
+// Hard cap on a single advisor call, sized to outlast the slowest
+// provider's own timeout so the MCP caller always gets a clean reply
+// from us rather than a dangling promise. Cloud providers cap at 60s
+// (AIProvider::ANALYSIS_TIMEOUT_MS); OllamaProvider caps at 120s
+// (LOCAL_ANALYSIS_TIMEOUT_MS). 135s adds a small buffer above the
+// Ollama path.
+static constexpr int kAdvisorMcpTimeoutMs = 135 * 1000;
 
 void registerAITools(McpToolRegistry* registry, MainController* mainController)
 {
-    // ai_advisor_invoke
-    //
-    // Tier: write — makes a paid outbound call to the configured AI
-    // provider. Side effects: emits `recommendationReceived` (which the
-    // in-app advisor's QML overlay listens to) and updates AIManager's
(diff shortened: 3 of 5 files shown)


skialpine:
### Code review

Found 6 issues:

1. **Critical: `"write"` is not a recognized tool tier — every call to `ai_advisor_invoke` will be rejected.** `categoryMinLevel()` only knows `"read"` (0), `"control"` (1), `"settings"` (2); anything else returns 3 (deny-all). The tool was registered with `"write"`, so `callAsyncTool` will return "Access level insufficient" on every invocation. Fix: change to `"control"` (matches the pattern used by mutating tools in `mcptools_write.cpp`).

https://github.com/Kulitorum/Decenza/blob/4e8456f83acfcb63c9d30b42d947d2313b519bca/src/mcp/mcptools_ai.cpp#L266-L269

https://github.com/Kulitorum/Decenza/blob/4e8456f83acfcb63c9d30b42d947d2313b519bca/src/mcp/mcptoolregistry.h#L283-L292

2. **Ollama users get a premature 75 s timeout while the provider is still working.** `kAdvisorMcpTimeoutMs = 75 * 1000` was sized to sit just above `AIProvider::ANALYSIS_TIMEOUT_MS = 60s`, but `OllamaProvider::LOCAL_ANALYSIS_TIMEOUT_MS = 120s`. With Ollama selected the MCP wrapper fires first; the in-app overlay still receives the late provider response, creating a split-result situation. Fix: raise the cap to ~135 s, or read it from the active provider.

https://github.com/Kulitorum/Decenza/blob/4e8456f83acfcb63c9d30b42d947d2313b519bca/src/mcp/mcptools_ai.cpp#L24-L29

3. **QTimer leaks per call into `AIManager`'s child list.** `state->timeout = new QTimer(ai)` is parented to AIManager but only `stop()`-ed (not `deleteLater()`-ed) inside `finalize`. The struct holding the pointer is `delete`-ed but the QTimer object survives as a permanent AIManager child. The tool is built for repeated A/B testing — every call leaks one QTimer. Fix: `state->timeout->deleteLater()` (or stash the QTimer as a `unique_ptr<QTimer, QObjectDeleteLater>`).

https://github.com/Kulitorum/Decenza/blob/4e8456f83acfcb63c9d30b42d947d2313b519bca/src/mcp/mcptools_ai.cpp#L213-L245

4. **`CallState` leaks and the MCP caller hangs if AIManager is destroyed mid-flight.** All three connections that drive `finalize` use `ai` as the receiver/context, and the QTimer is parented to `ai`. On AIManager destruction Qt auto-disconnects the lambdas and destroys the timer — none of the finalize paths fire, `respond()` is never invoked, and `state` leaks. Practical impact is narrow (AIManager dies with the app), but trivially mitigated. Fix: `connect(ai, &QObject::destroyed, ...)` to call `finalize` with an error.

https://github.com/Kulitorum/Decenza/blob/4e8456f83acfcb63c9d30b42d947d2313b519bca/src/mcp/mcptools_ai.cpp#L246-L262

5. **Hardcoded `"75s"` in the timeout error message will desync from the constant.** The literal string drifts the moment `kAdvisorMcpTimeoutMs` changes. Fix: derive from the constant via `QString("Advisor call timed out after %1s").arg(kAdvisorMcpTimeoutMs / 1000)`.

https://github.com/Kulitorum/Decenza/blob/4e8456f83acfcb63c9d30b42d947d2313b519bca/src/mcp/mcptools_ai.cpp#L255-L259

6. **`dryRun: true` is described as having "no side effects" — but it still spawns a QThread, opens a SQLite connection via `withTempDb`, and runs the summarizer.** Accurate framing is "no network call, no token cost." Misleading callers about resource impact when they expect to call the dry-run path at high frequency for prompt iteration.

https://github.com/Kulitorum/Decenza/blob/4e8456f83acfcb63c9d30b42d947d2313b519bca/src/mcp/mcptools_ai.cpp#L51-L57

<details>

- [OK to skip] Block comment above `registerAsyncTool` partially restates content already present in the tool description string passed to MCP clients. The genuine WHY content (thread safety, race-condition guards, tier rationale) is valuable; the redundant WHAT could be trimmed but isn't blocking. (score: 50)

</details>

PR #1035, opened 2026-05-01
fix(ai): tighten Setup header builder against partial-DYE field shapes

skialpine:
## Summary

Follow-up to #1030. The Setup header builder in `AIManager::emitRecentShotContext` unconditionally emitted separators (`" "`, `" with "`, `" on "`, `" - "`) and relied on every field being populated to make them read cleanly. Partial-DYE shapes broke that assumption — flagged post-merge by a late-arriving review agent.

**Failure modes (pre-fix):**
- Burrs populated, grinder brand + model both empty: `### Setup:  with 63mm Mazzer Kony` (double space after colon).
- Bean type populated, bean brand empty: `### Setup: Niche Zero with 63mm on Spring Tour` (stray `on ` before just a cultivar).
- Grinder brand only / model only: similar awkward separators.

**Fix:** rebuild each segment (grinderName, beanName) as a complete fragment, then join with a single space. No segment owns a leading separator, so absent fields can't produce double-space artifacts. The `on` preposition is gated on grinder fields being non-empty, so a bean-only setup renders just the bean name.

**Coverage:** new data-driven test `emitRecentShotContext_setupHeader_partialFieldShapes` pins 8 partial-DYE shapes (full / burrs-only / beanTypeOnly / beanBrandOnly / grinderBrandOnly / grinderModelOnly / grinderOnly / beanOnly) plus a defensive `no-double-space` check on the Setup line.

## Test plan

- [x] Build via Qt Creator MCP — succeeded with 0 errors, 0 warnings
- [x] `tst_AIManager` (14 tests, including 8 new data rows) — pass
- [x] Existing setup tests still pass: `_legacyEmptyShotDoesNotSuppressSetup`, `_genuineConflictSuppressesSetup`, `_hoistsProfileAndSetupOnce`

### DIFF
--- a/src/ai/aimanager.cpp
+++ b/src/ai/aimanager.cpp
@@ -608,21 +608,35 @@ void AIManager::emitRecentShotContext(
 
         if (setupShared && (!setupGrinderBrand.isEmpty() || !setupGrinderModel.isEmpty()
                             || !setupBeanBrand.isEmpty() || !setupBeanType.isEmpty())) {
-            result += "### Setup: ";
-            if (!setupGrinderBrand.isEmpty()) result += setupGrinderBrand;
+            // Build each segment as a complete fragment, then join with " "
+            // — that way no segment owns a leading space, and absent fields
+            // don't produce double-space artifacts (e.g. burrs without a
+            // grinder brand+model used to render "### Setup:  with 63mm").
+            QStringList parts;
+            QString grinderName;
+            if (!setupGrinderBrand.isEmpty()) grinderName = setupGrinderBrand;
             if (!setupGrinderModel.isEmpty()) {
-                if (!setupGrinderBrand.isEmpty()) result += " ";
-                result += setupGrinderModel;
+                if (!grinderName.isEmpty()) grinderName += " ";
+                grinderName += setupGrinderModel;
             }
-            if (!setupGrinderBurrs.isEmpty())
-                result += " with " + setupGrinderBurrs;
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

Found 1 issue:

1. **`int` instead of `qsizetype` for `QString::indexOf` / `QString::mid` results** — three new locals in the test store qsizetype-returning Qt 6 calls into `int`. CLAUDE.md says: *"Use `qsizetype` (not `int`) for container sizes — `QVector::size()`, `QList::size()`, `QString::size()` etc. return `qsizetype` (64-bit on iOS/macOS). Assigning to `int` causes `-Wshorten-64-to-32` warnings."* Local clang doesn't fire here (small strings, value fits) but iOS/macOS CI may. One-line fix: change the two `const int` declarations to `const qsizetype`.

https://github.com/Kulitorum/Decenza/blob/4d34b7ee6671d3da1f81eea3ab06401b1e831227/tests/tst_aimanager.cpp#L317-L324

[CLAUDE.md rule](https://github.com/Kulitorum/Decenza/blob/4d34b7ee6671d3da1f81eea3ab06401b1e831227/CLAUDE.md)

<details>

- [Worth a look] Data-row names `burrsOnly`, `beanTypeOnly`, `beanBrandOnly` mislead — they actually mean "only this within its category" (e.g. `beanTypeOnly` has full grinder + bean type without bean brand). A maintainer copying `beanTypeOnly` could miss that grinder is fully populated. Suggest `burrsNoGrinderName`, `beanTypeNoBrand`, `beanBrandNoType`. (score: 58)

</details>

PR #1051, opened 2026-05-01
refactor(ai): read AIConversation per-shot fields from JSON envelope (#1039)

### LINKED ISSUE #1039: Replace AIConversation regex-based change-detection with direct JSON-field reads
## Context

`AIConversation::processShotForConversation` and `AIConversation::summarizeShotMessage` use regex constants to extract dose/yield/duration/grinder/score/notes from the per-shot user message:

```cpp
const QRegularExpression AIConversation::s_doseRe("\\*\\*Dose\\*\\*:\\s*([\\d.]+)g");
const QRegularExpression AIConversation::s_yieldRe("\\*\\*Yield\\*\\*:\\s*([\\d.]+)g");
const QRegularExpression AIConversation::s_durationRe("\\*\\*Duration\\*\\*:\\s*([\\d.]+)s");
const QRegularExpression AIConversation::s_grinderRe("\\*\\*Grinder\\*\\*:\\s*(.+?)\\n");
// + s_profileRe, s_shotLabelRe, s_scoreRe, s_notesRe
```

PR #1034 added `extractShotProse` so the regexes still work on the new JSON envelope — they parse the JSON, extract `shotAnalysis`, then run the regex against the unescaped prose. Functional but fragile.

## Why this matters

After PR #1034, dose / yield / duration / grinder / score / notes all live in structured fields:

- `currentBean.doseWeightG` — direct value, no parsing
- `currentBean.grinderBrand` / `.grinderModel` / `.grinderBurrs` / `.grinderSetting` — structured
- The dose / yield / duration also live in the (planned) structured `shot` block from the "Restructure shotAnalysis" follow-up issue
- Score and notes are stored separately in `ShotProjection`; AIConversation could read directly from there instead of re-parsing the user message

The regex path:

- Is fragile if the prose formatting drifts (e.g., a future `## Shot Summary` line wording change silently breaks change detection).
- Requires an `extractShotProse` JSON-parse step before each regex match — wasted work given the structured fields are right there.
- Forces `buildUserPrompt` to emit prose in a specific shape forever, locking out future structural changes.

## Suggested approach

Migrate change detection to read structured fields directly:

1. The user message (the JSON envelope) becomes the source of structured fields. `processShotForConversation` parses the JSON once, reads `currentBean.doseWeightG` / `.grinderSetting` directly, computes the diff against the previous shot's structured fields.
2. Drop the regex constants from `AIConversation`.
3. For legacy stored conversations whose user messages are still in prose-only format: keep `extractShotProse` + regexes as a fallback path with a deprecation comment, OR migrate stored conversation history at startup.

## Why not in PR #1034

PR #1034 was scoped to the user prompt migration; touching AIConversation's change-detection logic was deferred to keep the migration small. Should land after the "Restructure shotAnalysis" issue ships structured per-phase fields, since change-detection is a natural client of those fields.

## Acceptance criteria

- `AIConversation::processShotForConversation` and `::summarizeShotMessage` read dose / yield / duration / grinder / score / notes from JSON envelope fields directly, not via regex.
- Regex constants `s_doseRe` / `s_yieldRe` / `s_durationRe` / `s_grinderRe` / `s_scoreRe` / `s_notesRe` deleted (or relegated to a clearly-marked legacy fallback path).
- `extractShotProse` becomes a fallback-only utility OR is deleted entirely.
- Tests pin the new path; the legacy prose fallback (if kept) is exercised by an explicit "stored conversation history" test.

## References

- PR #1034 — `extractShotProse` and the JSON envelope this builds on.
- `src/ai/aiconversation.cpp::processShotForConversation` — the change-detection function.
- `src/ai/aiconversation.cpp::summarizeShotMessage` — sister function that also regex-extracts.

skialpine:
## Summary

- Adds a small `shot` block to the user-prompt envelope (`doseG`, `yieldG`, `durationSec`, `ratio`, `grinderSetting`, `enjoymentScore`, `notes`) so `AIConversation` can read what the user iterates on without prose substring searches
- Replaces brittle regex-based change-detection with `extractShotFields`, which parses the envelope and returns a typed `ShotFields` struct
- Demotes the regex constants to a clearly-marked legacy fallback path used only when JSON parsing fails (preserves stored conversations from before the JSON envelope migration)
- Diff strings remain byte-identical across the migration (e.g., `Dose 18.0g→20.0g`, `Grinder Niche Zero (63mm) at 4.0 → ...`)

## Test plan

- [x] 6 new tests under `tst_AIManager` cover the structured path, the regex fallback, grinder-string format equivalence, numeric precision, detector-flag echoes, and the empty-shot omission contract
- [x] Full suite passes — 1945 passed, 0 failed
- [x] Build via Qt Creator MCP — 0 errors, 0 warnings

Closes #1039

### DIFF
--- a/src/ai/aiconversation.cpp
+++ b/src/ai/aiconversation.cpp
@@ -4,17 +4,32 @@
 
 #include <QDebug>
 #include <QSettings>
+#include <QJsonArray>
 #include <QJsonDocument>
+#include <QJsonObject>
 #include <QJsonParseError>
+#include <QJsonValue>
 #include <QRegularExpression>
 
-// Static regex constants shared between processShotForConversation() and summarizeShotMessage()
+// Outer wrapper regex for the "## Shot (date)" header that
+// `addShotContext` prepends OUTSIDE the JSON envelope. The header is
+// not part of the envelope itself, so it stays a regex match. All
+// per-shot data fields (dose / yield / duration / grinder / score /
+// notes / profile) come from structured JSON fields now — see
+// extractShotFields() below. Issue #1039.
+const QRegularExpression AIConversation::s_shotLabelRe("## Shot \\(([^)]+)\\)");
+
+// Legacy fallback regexes for stored conversations whose user messages
+// were saved before issue #1034 introduced the JSON envelope and #1039
+// added the structured `shot` block. These run only when JSON parsing
+// fails AND the message looks like prose. New code should NOT add new
+// callers — read `shot.*` / `currentBean.*` / `profile.*` / `tastingFeedback.*`
+// from the parsed envelope instead.
 const QRegularExpression AIConversation::s_doseRe("\\*\\*Dose\\*\\*:\\s*([\\d.]+)g");
 const QRegularExpression AIConversation::s_yieldRe("\\*\\*Yield\\*\\*:\\s*([\\d.]+)g");
 const QRegularExpression AIConversation::s_durationRe("\\*\\*Duration\\*\\*:\\s*([\\d.]+)s");
 const QRegularExpression AIConversation::s_grinderRe("\\*\\*Grinder\\*\\*:\\s*(.+?)\\n");
 const QRegularExpression AIConversation::s_profileRe("\\*\\*Profile\\*\\*:\\s*(.+?)(?:\\s*\\(by|\\n|$)");
-const QRegularExpression AIConversation::s_shotLabelRe("## Shot \\(([^)]+)\\)");
 const QRegularExpression AIConversation::s_scoreRe("\\*\\*Score\\*\\*:\\s*(\\d+)");
 const QRegularExpression AIConversation::s_notesRe("\\*\\*Notes\\*\\*:\\s*\"([^\"]+)\"");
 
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Found 2 issues, both fixed in 6c73fa78:

1. The new `shot.enjoymentScore` field violates the CLAUDE.md MCP-naming rule that bounded values must carry their scale in the field name. The existing dialing-context payload uses `enjoyment0to100` everywhere (`mcptools_dialing_blocks.cpp:57`, `:219`); the new envelope field must match.

   Old: https://github.com/Kulitorum/Decenza/blob/cf4aaebc89d2a3dee98cd8e1a5e0f25b8e6580fe/src/ai/shotsummarizer.cpp#L596-L601

2. The structured-path grinder string used `(<burrs>) at <setting>`, which differs from the legacy prose format `<brand> <model> with <burrs> @ <setting>` that historical stored conversations contain. A multi-shot conversation that spans both eras (older shot's grinder regex-extracted from prose, newer shot's grinder built from JSON) would emit a spurious "grinder changed" diff every time. The parity test pinned the new format against itself, so it did not catch the drift.

   Old: https://github.com/Kulitorum/Decenza/blob/cf4aaebc89d2a3dee98cd8e1a5e0f25b8e6580fe/src/ai/aiconversation.cpp#L502-L519

Issue #1053, opened 2026-05-01
AI advisor: closed-loop coaching — recentAdvice block in user prompt for outcome attribution


skialpine:
## Context

Each call to the AI advisor is currently amnesic about its own past advice and whether the user followed it. The dialing-context envelope ships recent shot history and the previous shot's `changeFromPrev` diff, but nothing about what the *advisor* recommended on the prior turn or whether the recommendation worked.

In the May-2026 advisor testing run (post-#1041 / #1037), the advisor said "try grind 4.75" on shot 884 and "0.25 finer" on shot 855. On the next turn, the advisor has no structured way to know:

- Whether the user actually changed grinder setting to 4.75
- What the resulting shot was rated / tasted like
- Whether to revise direction (if "finer" hurt) or commit harder (if it helped)

Each turn re-derives the dial-in trajectory from raw history. The reasoning it produces is roughly "I see they went from 5.0 to 4.5, so they're going finer." That works in single-session dial-ins, but the *advisor's own contribution to the trajectory* is invisible. There is no signal "advisor recommended X, user did Y."

## Why this matters

Today's advisor is a one-shot analyst. Closing the loop turns it into a coach:

- **Adherence-aware:** "Last time I said go finer; you stayed at 4.5 — let's try the recommendation before I revise." vs. "You did go finer and it got worse — direction was wrong, reverse."
- **Self-correcting:** the advisor can notice when its own advice didn't move the outcome and adapt strategy mid-session ("two coarse-direction recommendations didn't help; this isn't a grind problem, let's check temperature").
- **Cumulative learning within a session:** by turn 3 the advisor has narrowed the search instead of restarting.

## Suggested approach

Add a `recentAdvice` block to the user-prompt envelope. Carries the last 1–3 advisor turns plus what happened next.

```json
{
  "recentAdvice": [
    {
      "turnsAgo": 1,
      "recommendation": "Try grind 4.75 to slow flow toward 1.0–1.5 ml/s",
      "structuredNext": {
        "grinderSetting": "4.75",
        "expectedDurationSec": [32, 38],
        "expectedFlowMlPerSec": [1.0, 1.5]
      },
      "userResponse": {
        "actualNextShotId": 887,
        "grinderSetting": "4.75",
        "adherence": "followed",
        "outcomeRating": 75,
        "outcomeNotes": "Better, less harsh",
        "outcomeInPredictedRange": { "duration": true, "flow": true }
      }
    }
  ]
}
```

[...]

PR #1057, opened 2026-05-02
feat(ai): structured nextShot recommendation alongside prose (#1054)

### LINKED ISSUE #1054: AI advisor: emit structured nextShot recommendation alongside prose response
## Context

The advisor today returns prose. The recommendation lives inside a paragraph — \"try 4.75\" or \"0.25 finer\" or \"keep grind at 4.0, let yield run to 40g.\" Useful for the human reader, opaque to everything else.

In the May-2026 testing run (post-#1041 / #1037), the responses ended with concrete actionable advice in every case, but the *app* couldn't extract \"the grinder should be 4.75 next\" or \"expect 32–38s, 1.0–1.5 ml/s\" without re-parsing prose. That gap blocks every downstream feature that wants to validate whether the advisor's prediction came true.

## Why this matters

A structured prediction unlocks several capabilities at once:

- **Closed-loop coaching (issue #TBD-1)** — without `expectedDurationSec` / `expectedFlowMlPerSec`, the next turn's `outcomeInPredictedRange` flag cannot be computed, and the advisor cannot self-correct.
- **Active proactive UI** — the app could show \"Recommended next: grind 4.75, expect 32–38s\" as a coachmark on the dial-in screen *before* the user pulls.
- **Quality scoring of advice** — over many turns, the app can compute how often advisor predictions land in their own predicted ranges. A signal for \"is the model getting better with the new envelope?\" that isn't subjective.
- **Multi-provider A/B** — comparing GPT/Claude/Gemini on the same prompt today requires reading the prose. With structured predictions the comparison is mechanical.

## Suggested approach

Two non-mutually-exclusive paths:

1. **Tool-call shape.** Provider supports it — emit `nextShot(...)` as a structured tool call alongside the prose response. Cleanest for Anthropic / OpenAI / Gemini.
2. **JSON suffix.** Append a fenced JSON block to the prose (system prompt asks for it):

```json
{
  \"nextShot\": {
    \"grinderSetting\": \"4.75\",
    \"doseG\": 20,
    \"profileFilename\": \"80s-espresso\",
    \"expectedDurationSec\": [32, 38],
    \"expectedFlowMlPerSec\": [1.0, 1.5],
    \"expectedPeakPressureBar\": [7, 8],
    \"successCondition\": \"score >= 70 OR (durationSec in [32,38] AND flowMlPerSec in [1.0,1.5])\",
    \"reasoning\": \"Slow flow toward profile target without going past the choke point\"
  }
}
```

Path 2 works across every provider including Ollama. Path 1 is cleaner where supported. Implementation should pick based on provider capability.

`expectedX` ranges are intentional — a single value would be either too precise (forcing the user to hit it) or too vague to score against.

`successCondition` is a string the app can evaluate (or just store) — lets the model express \"good if score is good even if curves drift\" or \"good if curves match even if user hasn't rated.\"

## Implementation notes

- System prompt change: add a section under \"Response Guidelines\" asking for the structured nextShot. Cost: ~150 chars in the prompt.
- Parser in `AIManager`: extract the JSON block from the response (after the prose). Tolerate absence — older models or one-off Q&A turns may skip it.
- Storage: persist alongside the assistant message in `AIConversation`'s storage. Issue #TBD-1 reads it back to compute adherence.
- Cache implication: putting the request in the system prompt is fine (cacheable); putting it in the user prompt costs no additional cache stability since the user prompt already changes per shot.

## Acceptance criteria

- Advisor responses on every standalone shot analysis include a parseable nextShot block (JSON or tool call).
- The block is exposed by `ai_advisor_invoke` in its response envelope as `structuredNext` (separate from `response` prose) so MCP consumers don't have to re-parse.
- Tests pin the contract: synthetic shot → assert response contains a `structuredNext` with at least `grinderSetting`, `expectedDurationSec` range, `successCondition`.
- Robustness: when the model's prose ends without a JSON block (older models, transient issue), the response still parses and `structuredNext` is null with a clear log line.

## Depends on / enables

- **Enables #TBD-1** (closed-loop coaching) — that issue cannot score adherence without this structured shape.

## References

- May-2026 advisor testing run — every response had an actionable recommendation in prose, none in machine-readable form

skialpine:
## Summary

- Teach the shot-analysis system prompt to append a fenced ` ```json ` block named `nextShot` at end of every response that recommends a concrete parameter change (grind / dose / profile).
- Add a tolerant `AIManager::parseStructuredNext` static helper. Only the **trailing** fenced block tagged `json` (case-insensitive) qualifies — mid-message fenced blocks are intentionally ignored. Malformed JSON returns `nullopt` with a `qWarning`. Absent block returns `nullopt` silently.
- Persist the parsed object on each assistant turn in `AIConversation` as a sibling of `role` / `content` (soft-schema extension — older conversations load unchanged). Reader `structuredNextForLastAssistantTurn()` exposes it for downstream consumers.
- Surface the parsed block on `ai_advisor_invoke`'s MCP envelope as a top-level `structuredNext` field; omitted (no null) when absent.

This is the load-bearing precondition for closed-loop coaching (#1053): `recentAdvice[].structuredNext` reads straight back from this stored field, with `expectedDurationSec` / `expectedFlowMlPerSec` driving the `outcomeInPredictedRange` computation.

OpenSpec proposal at `openspec/changes/add-structured-next-shot/` (validates strict).

Closes #1054.

## Test plan

- [x] Unit tests: 7 `parseStructuredNext_*` cases covering the trailing-block extraction, mid-message rejection, malformed-JSON handling, case-insensitive `JSON` tag, trailing-whitespace tolerance, and empty input.
- [x] Persistence tests: `addAssistantMessage` round-trip through QSettings, no-key-when-absent, legacy-conversation backward compat.
- [x] Build clean, all 51 `tst_AIManager` tests pass (the warning on `parseStructuredNext_returnsNulloptOnMalformedJson` is the expected `qWarning` from the malformed-JSON path).
- [ ] Manual smoke against a live provider response is deferred; the contract is enforced by tests.

## Notes for reviewers

- The block is intentionally **left in place** in the prose (not stripped). Stripping would require a second LLM call for non-conversational re-rendering, and seeing the JSON in the conversation overlay is a useful debug signal that the model is honoring the new contract.
- No UI surface for the structured block is added in this PR — it's data plumbing for #1053 (closed-loop) and for future dial-in coachmarks.

### DIFF
--- a/docs/CLAUDE_MD/AI_ADVISOR.md
+++ b/docs/CLAUDE_MD/AI_ADVISOR.md
@@ -101,6 +101,24 @@ The four DB-scoped blocks (`dialInSessions`, `bestRecentShot`, `sawPrediction`, 
 
 Note: `dialing_get_context` (the MCP read tool) shares the four `McpDialingBlocks::*` block builders but does **not** go through `enrichUserPromptObject` — it assembles its own response envelope (top-level `dialInSessions` / `bestRecentShot` / `sawPrediction` / `grinderContext` plus `currentBean`, `profile`, `tastingFeedback`, `shotAnalysis`). The block-level shape is shared by construction; the wrapper envelope is not.
 
+### Structured `nextShot` output (issue #1054)
+
+The shot-analysis system prompt asks the model to append a fenced ` ```json ` block named `nextShot` at the very end of any response that recommends a concrete parameter change (grind / dose / profile). The block carries:
+
+- `grinderSetting`, `doseG`, `profileFilename` — present only on the field(s) the recommendation moves
+- `expectedDurationSec`, `expectedFlowMlPerSec` — required `[low, high]` ranges
+- `expectedPeakPressureBar` — optional `[low, high]` range when the advice targets pressure
+- `successCondition` — short natural-language predicate (stored verbatim)
+- `reasoning` — one sentence explaining why
+
+`AIManager::parseStructuredNext(QString)` extracts the trailing block. The parser is conservative: only the **last** fenced block whose closing fence is followed by nothing but whitespace, and whose opener is tagged `json` (case-insensitive), qualifies. Mid-message fenced blocks (e.g., the model echoing prior advice for context) are intentionally ignored. Malformed JSON returns `nullopt` with a `qWarning`. Absent block returns `nullopt` silently — the model is supposed to omit the block on clarifying-question responses.
+
+The parsed object is persisted on the assistant message as `structuredNext` alongside `role` / `content` in `AIConversation::m_messages`. `AIConversation::structuredNextForLastAssistantTurn()` reads it back. Older saved conversations (no `structuredNext` key) load cleanly and the reader returns `nullopt`.
+
+`ai_advisor_invoke` (MCP) parses the trailing block from the assistant response and surfaces it as a top-level `structuredNext` field in the tool result envelope, alongside `response`. The field is **omitted** (no `null`) when the response carries no recommendation.
+
+This block is the load-bearing precondition for #1053's closed-loop coaching — `recentAdvice[].structuredNext` is read straight back from this stored field, with `expectedDurationSec` / `expectedFlowMlPerSec` driving the `outcomeInPredictedRange` computation.
+
 ---
 
 ## Lessons Learned: Profile Knowledge Doesn't Scale (March 2026)
--- a/docs/CLAUDE_MD/AI_ADVISOR.md
+++ b/docs/CLAUDE_MD/AI_ADVISOR.md
@@ -105,7 +105,7 @@ Note: `dialing_get_context` (the MCP read tool) shares the four `McpDialingBlock
 
 The shot-analysis system prompt asks the model to append a fenced ` ```json ` block named `nextShot` at the very end of any response that recommends a concrete parameter change (grind / dose / profile). The block carries:
 
-- `grinderSetting`, `doseG`, `profileFilename` — present only on the field(s) the recommendation moves
+- `grinderSetting`, `doseG`, `profileTitle` — present only on the field(s) the recommendation moves
 - `expectedDurationSec`, `expectedFlowMlPerSec` — required `[low, high]` ranges
 - `expectedPeakPressureBar` — optional `[low, high]` range when the advice targets pressure
 - `successCondition` — short natural-language predicate (stored verbatim)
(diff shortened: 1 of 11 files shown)


skialpine:
### Code review

Found 3 issues:

1. `parseStructuredNext` silently returns `nullopt` whenever the total count of triple-backtick markers in the message is odd, even when the trailing pair forms a valid `nextShot` block. A model that emits a stray ``` somewhere in the prose (e.g. an inline-code mishap) loses its recommendation. Removing the parity guard and pairing only the last two fences as opener/closer fixes this without weakening the mid-message-block protection — the closer-followed-by-only-whitespace check already enforces the trailing-block contract.

https://github.com/Kulitorum/Decenza/blob/10f92559d5ce84e0fc35bd2e48760a7095eb2526/src/ai/aimanager.cpp#L189-L196

2. The MCP end-to-end test `aiAdvisorInvokeSurfacesStructuredNext` listed in `tasks.md` is missing — the MCP `body.insert("structuredNext", *structured)` lambda branch in `mcptools_ai.cpp:307-313` has no coverage. (CLAUDE.md says "Implement tasks sequentially; confirm every item is done before updating statuses.")

https://github.com/Kulitorum/Decenza/blob/10f92559d5ce84e0fc35bd2e48760a7095eb2526/openspec/changes/add-structured-next-shot/tasks.md#L29-L35

3. All 9 boxes in `tasks.md` remain `- [ ]` even though the implementation, docs, and tests are landed. CLAUDE.md says "Set every task to `- [x]` after all work is complete."

https://github.com/Kulitorum/Decenza/blob/10f92559d5ce84e0fc35bd2e48760a7095eb2526/openspec/changes/add-structured-next-shot/tasks.md#L5-L37

<details>

- [Worth a look] `m_messages` doc-comment in `aiconversation.h:223` still reads "Array of {role, content} objects" — should mention the new optional `structuredNext` sibling (score: 55).
- [Worth a look] Tests use `QSettings(); settings.clear();` which depends on `QStandardPaths::setTestModeEnabled` to redirect; safer to scope the clear to the per-test prefix instead of the entire app store (score: 60).
- [OK to skip] System prompt could explicitly tell the model that `expectedDurationSec` / `expectedFlowMlPerSec` must be 2-element arrays even for a "single intended value" — current parser tolerates a scalar by silently producing an empty array on `.toArray()` (score: 65). Worth tightening if a future provider misbehaves; not load-bearing today.

</details>

(Verified: the bug-scan agent's "stray `)` causing compile break" claim in `mcptools_ai.cpp` is a false positive — the file reads `QJsonObject body{{"response", response}};` and the build was clean.)

PR #1058, opened 2026-05-02
feat(ai): closed-loop coaching with recentAdvice block (#1053)

### LINKED ISSUE #1053: AI advisor: closed-loop coaching — recentAdvice block in user prompt for outcome attribution

skialpine:
## Summary

Builds on #1057 (#1054 structured nextShot). Stack base is `fix/1054-structured-next-shot`; merging #1057 first will rebase this onto `main`.

- **Part A — Per-turn shot linkage in `AIConversation`**: soft-schema extension persists a `shotId` on each turn entry. `setShotIdForCurrentTurn` applies retroactively to the latest user turn AND latches onto the next-appended assistant turn so a pair shares one id. New readers `shotIdForTurn`, `recentAssistantTurns(max)`, and a static `loadRecentAssistantTurnsForKey(key, max)` for the MCP path.
- **Part B — `recentAdvice` block builder**: `DialingBlocks::buildRecentAdviceBlock` pairs each prior advisor turn (`shotId` + parsed `structuredNext`) with the user's actual follow-up shot on the same profile. Computes `adherence` (followed / partial / ignored) with a "did the user actually move from prior" guard, and `outcomeInPredictedRange` (`duration` + `flow`). `outcomeRating` is omitted when the follow-up is unrated.
- **Wiring**: `enrichUserPromptObject` takes a `recentAdvice` argument. Both surfaces (in-app advisor's bg closure and `ai_advisor_invoke`'s bg closure) compute the block and merge into the envelope. The MCP path uses the static loader to read the conversation that the in-app advisor wrote, so `recentAdvice` is byte-equivalent across surfaces for the same `(shot, conversation key)`.
- **System prompt teaching**: `shotAnalysisSystemPrompt` gains a `recentAdvice` subsection in "How to read structured fields" — adherence semantics, omitted-outcomeRating fallback, and the framing that `recentAdvice` is the LLM's own track record.

Side improvement: `ShotSummarizer::loadProfileKnowledge` now latches `s_knowledgeLoaded` on the resource-missing failure path so test binaries (which don't link the qrc) don't re-warn on every call. `TstDialingBlocks::initTestCase` pins the single expected warning per TESTING.md.

OpenSpec proposal at `openspec/changes/add-closed-loop-advisor/` (validates strict).

Closes #1053. Depends on #1054 (#1057) for the structuredNext field.

## Test plan

- [x] 6 new dialing-blocks tests covering `recentAdvice` qualifying turns / unrated follow-up / cross-profile filter / ignored adherence / empty turns / no-follow-up skip
- [x] 4 new `AIConversation` tests covering pair linkage / legacy turn skipping / max cap / static loader
- [x] All 1985 tests pass clean — **zero warnings** per TESTING.md
- [ ] Manual end-to-end against a live provider deferred; the contract is enforced by tests

## Notes for reviewers

- Pressure attribution (`outcomeInPredictedRange.pressure`) is intentionally omitted from the current implementation. ShotProjection doesn't carry peak pressure, and parsing the samples blob for every recentAdvice entry would be too expensive. The spec was written so `pressure` is OPTIONAL — a future change can fill it in when peak pressure becomes a first-class projection field.
- Adherence's "did the user move from prior" guard handles the realistic case where the model recommends `4.75` but the user's prior was already `5.0` — keeping `5.0` registers as `ignored`, not `followed`.


skialpine:
### Code review

Found 8 issues:

1. `AIConversation::ask()` clears `m_messages` on every advisor call, so the in-app advisor's conversation in QSettings is always exactly one pair — `recentAdvice` will never accumulate more than one entry, breaking the spec's "up to 3 entries" promise. The fix (route subsequent calls through `addShotContext` to accumulate) is a substantial design change that should land separately; for this PR, the spec/proposal need to be honest about the one-entry cap and document the follow-up.

https://github.com/Kulitorum/Decenza/blob/00fcf7868f453d36920b37b703b977a9b549a53b/src/ai/aiconversation.cpp#L62-L86

2. `addUserMessage` doesn't consult `m_pendingShotId`, so the production flow (`setShotIdForCurrentTurn(N)` → bg thread → `ask()` clears + addUser → addAssistant) ends up with `[{user, no shotId}, {assistant, shotId=N}]` — the spec scenario "User and assistant of the same turn pair share shotId" fails in production.

https://github.com/Kulitorum/Decenza/blob/00fcf7868f453d36920b37b703b977a9b549a53b/src/ai/aiconversation.cpp#L168-L174

3. MCP-naming convention violation (CLAUDE.md says "Include scale in field names for bounded values, `enjoyment0to100` instead of `enjoyment`") — `userResponse.outcomeRating` is a 0-100 value and should be `outcomeRating0to100`.

https://github.com/Kulitorum/Decenza/blob/00fcf7868f453d36920b37b703b977a9b549a53b/src/ai/dialing_blocks.cpp#L558-L562

4. `computeAdherence` compares `structuredNext.profileFilename` (a filename like `"de_30s_traditional.json"`) against `ShotProjection::profileName` (a profile title like `"80's Espresso"`). The two will never match, so any profile-switch recommendation always reads as `ignored`. Cleanest fix: rename the structuredNext field to `profileTitle` so both ends use the title.

https://github.com/Kulitorum/Decenza/blob/00fcf7868f453d36920b37b703b977a9b549a53b/src/ai/dialing_blocks.cpp#L411-L420

5. `setShotIdForCurrentTurn` retroactively rewrites the most-recent user turn's `shotId` even when that user turn already carried one — corrupts the prior pair when the conversation accumulates. Should only stamp when the user turn lacks a `shotId`.

https://github.com/Kulitorum/Decenza/blob/00fcf7868f453d36920b37b703b977a9b549a53b/src/ai/aiconversation.cpp#L218-L233

6. `kGrinderStepTolerance = 0.1` contradicts the spec's "within 0.25 of a numeric step." With the prior-movement guard added in this PR, the original 0.25 is correct (the guard catches the no-movement case). The 0.1 overshoots into rejecting legitimate single-click moves.

https://github.com/Kulitorum/Decenza/blob/00fcf7868f453d36920b37b703b977a9b549a53b/src/ai/dialing_blocks.cpp#L353-L361

7. `tasks.md` items 13 (`recentAdviceParityBetweenInAppAndMcp`) and 14 (`byteStabilityRecentAdvice`) are checked off, but neither test exists in the codebase. Same CLAUDE.md violation pattern flagged on PR 1057 ("Implement tasks sequentially; confirm every item is done before updating statuses.").

https://github.com/Kulitorum/Decenza/blob/00fcf7868f453d36920b37b703b977a9b549a53b/openspec/changes/add-closed-loop-advisor/tasks.md#L57-L60

8. The `enrichUserPromptObject` test pair (`mergesAllFourBlocks` / `suppressesEmptyBlocks`) was the established pattern for the four prior dialing blocks but the new `recentAdvice` argument has no equivalent coverage.

https://github.com/Kulitorum/Decenza/blob/00fcf7868f453d36920b37b703b977a9b549a53b/src/ai/aimanager.cpp#L487-L514

<details>

- [Worth a look] Confusing `++turnsAgo`/`--turnsAgo` net-no-op pair in `buildRecentAdviceBlock` left from a refactor — clean up to a single increment at the qualify-success site (score: 70).
- [Worth a look] `tst_dialing_blocks::initTestCase` uses an exact-string `QTest::ignoreMessage` — TESTING.md prefers `QRegularExpression` (score: 60).
- [Worth a look] `excludeId` (= `lastSavedShotId()`) is repurposed as "the shot under analysis" — wrong when the user invokes the advisor on a historical shot. Pre-existing pattern but newly elevated to identify the conversation's anchor; would be cleaner to derive from the resolved `ShotProjection` (score: 65).
- [OK to skip] Doc-comment on `recentAssistantTurns` lists two skip conditions but the impl has three (the structuredNext-not-an-object branch). Cosmetic doc gap (score: 55).

</details>

PR #1059, opened 2026-05-02
feat(ai): capture shot ratings — three escalating layers (#1055)

### LINKED ISSUE #1055: AI advisor: bestRecentShot is starving — most users never rate, anchor never fires
## Context

`dialing_get_context.bestRecentShot` is a deliberately powerful feature: highest-rated past shot on the same profile within 90 days, with a `changeFromBest` diff. The intent — "anchor advice on what success looked like, not just what changed last shot" — is exactly right.

The problem: in the May-2026 advisor testing run on a real user's database (199 shots, weeks of dial-in iteration), `bestRecentShot` was omitted from the envelope on every call because **none of the 199 shots had `enjoyment0to100` set**. The 90-day window had zero rated shots in it. The block is permanently dark for this user, and likely for most users — rating a shot is a deliberate action that's easy to skip.

The `tastingFeedback.hasEnjoymentScore` boolean is `false` on these shots too. The advisor reads that and dutifully asks "how did it taste?" before suggesting changes — correct behavior, but it doesn't write the rating back when the user replies. The conversational answer doesn't propagate to `ShotProjection.enjoyment0to100`.

## Why this matters

The advisor's most valuable mode is "walk back toward your best shot." Without ratings, that mode never engages. The current envelope's smarts (grinder context, SAW prediction, dial-in sessions) are all secondary supports — anchoring on a known-good prior is the headline feature, and it's silent.

This is a UX problem, not an envelope problem. The data plumbing is correct.

## Suggested approach

Three layers, escalating:

**1. Ask once, persist always.** When the advisor asks "how did it taste?" and the user replies in the conversation overlay with a score and notes, parse the reply and write back to `ShotProjection.enjoyment0to100` + `.espressoNotes`. The conversational rating is currently lost — only the manual rating UI persists. Closing this loop alone would 10× rating density without any new UI.

**2. One-tap rating prompt.** After every shot ends, a non-modal prompt ("Rate this shot: 1–100 / quick notes / dismiss") on the post-shot screen. The friction today is: open shot detail page → find rating field → type. Reduce to: tap one of three cup-rating icons (good/ok/bad → 80/60/40 default), optional notes via voice or short text. Dismissable. The current rating UI stays for users who want precision.

**3. Auto-rate inferred-good shots.** When the deterministic detectors fire `grind_clean` + `channeling_none` + on-target ratio (within 0.1) + on-target duration (within profile expectation) AND the user didn't manually rate, the advisor can treat that as a *provisional* anchor — surfacing it as `bestRecentShot` with a `confidence: "inferred"` flag. The system prompt teaches the LLM "treat inferred bestRecentShot as a hint, not ground truth — confirm with the user before strongly anchoring on it."

## Why not earlier

Pre-#1041, the in-app advisor didn't have `bestRecentShot` at all (only the MCP `dialing_get_context` tool did). The starvation problem only became visible once the in-app surface started shipping the block and we noticed it was always omitted.

## Acceptance criteria

- (Layer 1) When the advisor's response is followed by a user reply containing a numeric score (e.g., "75" or "75 out of 100"), parse it and write to `ShotProjection.enjoyment0to100`. Notes get the surrounding sentence(s). The shot id is the most recent shot in the conversation context.
- (Layer 1) Test pins this: simulate a turn where the advisor asks for taste, the user replies "82, balanced and sweet", and the persisted shot's enjoyment is now 82 with notes "balanced and sweet".
- (Layer 2) New post-shot rating prompt component. Three-tap rating + optional notes. Dismiss persists "user dismissed" so we don't re-prompt the same shot.
- (Layer 3) `dialing_get_context.bestRecentShot` (and the in-app envelope) gain an optional `confidence: "user_rated" | "inferred"` field. System prompt teaches the model to treat them differently.

## References

- #1041 — added bestRecentShot to the in-app envelope
- #1037 — detector observations the inferred-good auto-rating reads from
- May-2026 advisor testing report — concrete observation that no rated shots existed in a 199-shot DB

skialpine:
## Summary

Builds on #1058 (#1053 closed-loop). Stack base is `fix/1053-closed-loop-advisor`; merging that first will rebase this onto `main`.

`bestRecentShot` was permanently dark for the May-2026 testing user (199 shots, 0 with `enjoyment0to100` set). Three layers close the starvation gap:

### Layer 1 — Conversational rating capture
- `AIManager::parseUserRatingReply` — pure parser. Bare integer 1-100, optional `/100`/`out of 100`/`%` suffix, decimal rounding, rejects non-numeric and out-of-range, picks first in-range token. Negative numbers rejected.
- `AIManager::maybePersistRatingFromReply` — writes enjoyment + remaining-text notes via `ShotHistoryStorage::requestUpdateShotMetadata` when the prior assistant asked about taste AND the turn pair has a non-zero `shotId` (per #1053).
- `AIConversation::followUp` calls the helper before `sendRequest` so a network failure doesn't lose the rating.

### Layer 2 — `QuickRatingRow` QML component
- New `qml/components/QuickRatingRow.qml`: three emoji-icon buttons (😊/😐/😞 → 80/60/40) + dismiss `×`. `AccessibleButton`-based, `Theme`-styled, fully translated.
- `PostShotReviewPage` instantiates the row above the metadata fold; visible only when `editEnjoyment == 0` AND no per-shot dismiss flag. Tap persists immediately via `saveEditedShot`. Dismiss persists in `Settings(shotRatingDismissed/<shotId>)` — a UI nudge, not a shot attribute.
- **Manual QA only** — the project has no QML test harness; please open the page on an unrated shot and exercise the row before merging.

### Layer 3 — Inferred-good auto-rating + `confidence` flag
- Migration 14: `enjoyment_source TEXT NOT NULL DEFAULT 'none'`. Existing rated rows back-filled to `'user'`. Idempotency test pins re-initialize cleanly.
- `ShotProjection` / `ShotRecord` / `ShotSaveData` / `ShotSummary` gain `enjoymentSource` ("none"/"user"/"inferred").
- `updateShotMetadataStatic` defaults source to `"user"` on any explicit write that carries an `enjoyment` value.
- `saveShot` post-analyze: unrated + verdict clean + channeling none + grindDirection onTarget → enjoyment = 75, source = "inferred".
- `buildBestRecentShotBlock` ORDER BY prefers `user_rated` over `inferred` regardless of score; emits `confidence`.
- `buildCurrentBeanBlock` and `dialInSessions[].shots[]` emit `enjoymentSource: "inferred"` only (sparse signal). System prompt teaches LLM to treat inferred ratings as hints.

OpenSpec proposal at `openspec/changes/add-shot-rating-capture/` (validates strict).

Closes #1055.

## Test plan

- [x] 9 new `parseUserRatingReply_*` cases (bare numbers, suffixes, decimals, out-of-range, negatives, non-numeric, multi-token, empty)
- [x] 3 new `bestRecentShot_*` cases (user-over-inferred preference, inferred fallback, empty omission)
- [x] 1 new migration idempotency test (`v14_idempotentReapply`) — schema_version=14 + column survives re-init
- [x] All 2003 tests pass clean, zero warnings
- [ ] Manual QA on `PostShotReviewPage`: open an unrated shot, tap each face, verify persist + collapse to "Rated N" pill; tap dismiss, verify the row hides; reopen the same shot, verify the row stays hidden

## Notes for reviewers

- The "did the prior assistant ask about taste?" gate is a heuristic on phrase markers (`"how did"`, `"score"`, `"taste"`, etc.). False negatives are acceptable — the user can still rate via the editor or `QuickRatingRow`. False positive risk is bounded by the `shotId` binding so any writeback hits the conversation's anchored shot, not a random one.
- `pressure` field on `outcomeInPredictedRange` (#1053) is still unconditionally omitted; `ShotProjection` doesn't carry peak pressure. Out of scope for this PR.
- The system prompt's `confidence` teaching tells the LLM to confirm with the user before strongly anchoring on an `inferred` candidate.


skialpine:
### Code review

Found 7 issues:

1. `QuickRatingRow` dismiss tap does not hide the row. The `visible` binding reads `Settings.value("shotRatingDismissed/" + editShotId, false)`; that's not a notifiable property, and the `editShotData = Object.assign(...)` workaround doesn't help because `editShotId` isn't derived from `editShotData`. Fix: track dismiss in a local QML property that `visible` can bind to reactively.

https://github.com/Kulitorum/Decenza/blob/2168089624f54e98532ccabec7a6963b990bd267/qml/pages/PostShotReviewPage.qml#L677-L695

2. Inferred-good evaluator implements 3 of the 5 spec criteria. The spec requires `verdictCategory == "clean"` AND `channelingSeverity == "none"` AND `grindDirection == "onTarget"` AND ratio-within-0.1 (or yield within 0.5g of target) AND duration within ±15% of profile median. The code skips ratio and duration gates entirely — a clean-curve shot with a wildly off-target yield can still get auto-rated 75.

https://github.com/Kulitorum/Decenza/blob/2168089624f54e98532ccabec7a6963b990bd267/src/history/shothistorystorage.cpp#L971-L987

3. `QuickRatingRow` is hidden for inferred-75 shots, even though they have no user rating. `visible: editEnjoyment === 0` treats the inferred 75 as "rated"; should also check `editShotData.enjoymentSource !== "user"` so the user still gets the prompt to give an explicit score.

https://github.com/Kulitorum/Decenza/blob/2168089624f54e98532ccabec7a6963b990bd267/qml/pages/PostShotReviewPage.qml#L676-L685

4. `collapsedPill.onClicked` emits `rateClicked(currentScore)` — re-saves the same score, doesn't reopen the icon row. The spec says the pill should reopen the three-icon state on tap. Fix: emit a separate `reviseClicked()` signal, or zero `currentScore` locally.

https://github.com/Kulitorum/Decenza/blob/2168089624f54e98532ccabec7a6963b990bd267/qml/components/QuickRatingRow.qml#L142-L150

5. Multiple boxes in `tasks.md` are checked `[x]` but the named tests don't exist. Same CLAUDE.md violation flagged on PRs 1057 and 1058 (`Implement tasks sequentially; confirm every item is done before updating statuses`). Missing: `conversationalRatingPersists`, `conversationalRatingNoOpWhen*`, `currentBeanCarriesEnjoymentSourceForInferred`, `dialInSessionsSparseEnjoymentSource`, `inferredGoodTriggers` family, `schemaMigrationBackfill`.

https://github.com/Kulitorum/Decenza/blob/2168089624f54e98532ccabec7a6963b990bd267/openspec/changes/add-shot-rating-capture/tasks.md#L8-L86

6. `v14_idempotentReapply` opens a raw `QSqlDatabase::addDatabase`/`removeDatabase` pair instead of using the file-local `withRawDb` RAII helper that every other test in `tst_dbmigration.cpp` uses. A `QVERIFY` failure inside the block leaks the connection.

https://github.com/Kulitorum/Decenza/blob/2168089624f54e98532ccabec7a6963b990bd267/tests/tst_dbmigration.cpp#L555-L585

7. `maybePersistRatingFromReply` doc-comment lists 3 no-op conditions; the impl has 4 — the `priorAssistantAskedAboutTaste` heuristic guard is undocumented in the header.

https://github.com/Kulitorum/Decenza/blob/2168089624f54e98532ccabec7a6963b990bd267/src/ai/aimanager.h#L195-L208

<details>

- [Worth a look] `QuickRatingRow` Text prompt missing `Accessible.ignored: true` — TalkBack will discover it as an extra swipe target alongside the three buttons (score: 65).
- [Worth a look] Conversational rating capture works for the first followUp turn after `ask()` (where the prior assistant carries the latched shotId) but degrades to no-op for subsequent turns since followUp doesn't re-latch. Acceptable for the most common flow but worth documenting (score: 60).
- [OK to skip] Migration 14 retains a `hasColumn` guard while migrations 9-13 dropped it. Stylistic mismatch only — the guard makes the migration safer, not less correct (score: 55).

</details>

PR #1065, opened 2026-05-02
feat(mcp): surface enjoymentSource on shots_list / shots_get_detail / shots_update / recent resource

skialpine:
## Summary

Closes the parity gap between the advisor's dialing-context (which already emits `enjoymentSource` on `currentBean` and `dialInSessions[].shots[]`, plus `confidence` on `bestRecentShot`) and the generic `shots_*` MCP tools that scripts and the in-app overlay use. Without these, an MCP consumer browsing the shot library couldn't distinguish a user-rated 75 from an app-inferred 75.

- **`shots_list`** — emits `enjoymentSource` per row; accepts a new `enjoymentSource` filter (`"user"` / `"inferred"` / `"none"`), composable with the existing rating filters.
- **`shots_get_detail`** — `ShotProjection::toVariantMap` / `fromVariantMap` now round-trip the field so the projection that backs the detail response includes it.
- **`shots_update`** — accepts `enjoymentSource` with strict validation. Lets a caller downgrade an inferred rating to `"none"` or convert one to `"user"` without touching the numeric score.
- **`decenza://shots/recent`** resource — includes `enjoymentSource` per row.

## Test plan

- [x] All 2013 existing tests pass clean
- [x] Live-verified through MCP on the running app:
  - `shots_list` emits `enjoymentSource` on every row (`"none"` for unrated)
  - `shots_list { enjoymentSource: "user" }` narrows 880 shots to 42 user-rated shots
  - `shots_get_detail` returns `enjoymentSource: "user"` on shot 802
  - `shots_update { enjoymentSource: "bogus" }` rejected with `enjoymentSource must be "user", "inferred", or "none"; got "bogus"`

### DIFF
--- a/src/history/shotprojection.cpp
+++ b/src/history/shotprojection.cpp
@@ -17,6 +17,7 @@ QVariantMap ShotProjection::toVariantMap() const
     m["beanBrand"] = beanBrand;
     m["beanType"] = beanType;
     m["enjoyment0to100"] = enjoyment0to100;
+    m["enjoymentSource"] = enjoymentSource;
     m["hasVisualizerUpload"] = hasVisualizerUpload;
     m["beverageType"] = beverageType;
     m["roastDate"] = roastDate;
@@ -87,6 +88,8 @@ ShotProjection ShotProjection::fromVariantMap(const QVariantMap& m)
     p.beanBrand = m.value("beanBrand").toString();
     p.beanType = m.value("beanType").toString();
     p.enjoyment0to100 = m.value("enjoyment0to100").toInt();
+    p.enjoymentSource = m.value("enjoymentSource").toString();
+    if (p.enjoymentSource.isEmpty()) p.enjoymentSource = QStringLiteral("none");
     p.hasVisualizerUpload = m.value("hasVisualizerUpload").toBool();
     p.beverageType = m.value("beverageType").toString();
     p.roastDate = m.value("roastDate").toString();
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Found 1 issue:

1. `shots_list` accepts any string for the `enjoymentSource` filter and silently returns zero rows on a typo. `shots_update` already validates the same field strictly (rejecting anything outside `{"user", "inferred", "none"}`). The asymmetry between the read and write path means a misspelled filter (e.g. `"User"`) gets `{shots: [], total: 0}` with no diagnostic — the LLM has no signal to correct. CLAUDE.md MCP-conventions section emphasizes unambiguous behavior for AI consumers; the precedent in `shots_update` (mcptools_write.cpp:91-97) is the right pattern.

https://github.com/Kulitorum/Decenza/blob/acaa5b3a399b359a2f0395f66fe32cd30f63e3da/src/mcp/mcptools_shots.cpp#L131-L137

<details>

- [Worth a look] `enjoymentSource: "none"` filter compiles to `enjoyment_source = 'none'`, which doesn't match legacy rows that might be NULL or empty. Migration 14 backfilled all rows so this is currently unreachable, but a defensive `OR enjoyment_source IS NULL OR enjoyment_source = ''` would future-proof (score: 60).
- [Worth a look] `shots_list`'s filter description lists `"none" = unrated` but doesn't lead with the full enum — an LLM reading the description may not realize `"none"` is a valid value (score: 55).
- [Worth a look] `shots_update {enjoymentSource: "none"}` without an `enjoyment` write leaves the `enjoyment` column intact, producing a row with non-zero enjoyment + source `"none"` — contradicts the semantic. Either zero `enjoyment` automatically, or document that the caller must pass both. The description currently implies a clean "downgrade" workflow (score: 55).
- [OK to skip] `nullifyUnratedFields` doesn't enforce that `enjoymentSource = "none"` whenever it nulls `enjoyment0to100`. Cannot happen with current data invariants (inferred always sets 75; backfill set rated→user) but a paired-invariant guard is defensible defensively (score: 50).

</details>

PR #1068, opened 2026-05-02
fix(relay): tear down screen capture when phone goes silent

Kulitorum:
## Summary
- Adds a 30s activity watchdog in `RelayClient` so that `ScreenCaptureService` is torn down when the phone is OS-killed or loses network — the existing teardown paths (`stop_remote` and `onDisconnected`) are unreachable in that scenario because the tablet's own WS to AWS stays alive (5-min ping prevents idle timeout).
- Recognizes a `keepalive` relay_command as a no-op so DecenzaPocket can hold the session open while the user is just viewing without touching.
- Adds layered logs (`relay_command received: <cmd>`, `start_remote — creating ScreenCaptureService` / `…ignored — capture already active` / `…ignored — window not yet wired`, `stop_remote — wasActive=…`, `WebSocket disconnected — wasCapturing=…`) so the full message path is observable if this ever regresses.

## Why
Tablet logs showed `ScreenCaptureService: sent 1 tiles, 441 bytes` continuously after DecenzaPocket was backgrounded or killed. Root cause: tablet ↔ AWS WebSocket stays open independently of phone ↔ AWS, so neither `stop_remote` (phone may be killed before sending) nor `onDisconnected` (tablet's own WS still alive) fires.

The watchdog is a defense-in-depth backstop — `stop_remote` and `onDisconnected` still tear down capture immediately when reachable; the watchdog only fires when both paths fail.

## Phone-side contract
DecenzaPocket should send `relay_command: "keepalive"` at <30s cadence while the remote-control screen is foregrounded. Until that lands, idle viewing without touches will be cut off after 30s. Tunable at `relayclient.cpp:21` (`kRemoteActivityTimeoutMs`).

## Test plan
- [x] Pair phone, start remote, then graceful background → tablet logs `stop_remote — wasActive= true` immediately, `ScreenCaptureService: stopped`, no further `sent N tiles`. (Confirmed on device.)
- [ ] Pair phone, start remote, then **force-stop** DecenzaPocket from Android settings (no graceful background) → tablet should log `RelayClient: no remote activity for 30000 ms — tearing down ScreenCaptureService` followed by `ScreenCaptureService: stopped` within ~30s.
- [ ] Pair phone, start remote, drop tablet WiFi → `WebSocket disconnected — wasCapturing= true`, capture stops.
- [ ] Phone reconnects after a kill → no auto-resume; fresh `start_remote` required (existing behavior, unchanged).

### DIFF
--- a/src/network/relayclient.cpp
+++ b/src/network/relayclient.cpp
@@ -14,6 +14,11 @@ static constexpr int kReconnectBaseMs = 5000;
 static constexpr int kReconnectMaxMs = 60000;
 static constexpr int kPingIntervalMs = 5 * 60 * 1000;  // 5 minutes
 static constexpr int kStatusPushIntervalMs = 5000;      // 5 seconds
+// Tear down screen capture if the phone goes silent for this long. The phone
+// is expected to send a "keepalive" relay_command at a faster cadence (~10s)
+// while the remote-control screen is foregrounded; explicit touches and
+// other commands also count as activity.
+static constexpr int kRemoteActivityTimeoutMs = 30000;
 static const QString kRelayUrl = QStringLiteral("wss://ws.decenza.coffee");
 
 RelayClient::RelayClient(DE1Device* device, MachineState* machineState,
@@ -34,6 +39,10 @@ RelayClient::RelayClient(DE1Device* device, MachineState* machineState,
     connect(&m_pingTimer, &QTimer::timeout, this, &RelayClient::onPingTimer);
 
     connect(&m_statusPushTimer, &QTimer::timeout, this, &RelayClient::pushStatus);
+
+    m_remoteActivityTimer.setSingleShot(true);
+    connect(&m_remoteActivityTimer, &QTimer::timeout,
+            this, &RelayClient::onRemoteActivityTimeout);
 
     // Trigger immediate status push on state changes
     if (m_device) {
@@ -70,6 +79,7 @@ void RelayClient::setEnabled(bool enabled)
         m_reconnectTimer.stop();
         m_pingTimer.stop();
         m_statusPushTimer.stop();
+        m_remoteActivityTimer.stop();
         m_reconnectAttempts = 0;
         // Destroy capture service synchronously BEFORE closing the socket.
         // m_socket.close() is async — onDisconnected() fires later, but by then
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

Found 2 issues:

1. Timer used as a guard/workaround — explicit CLAUDE.md violation ([CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/main/CLAUDE.md) says: "Never use timers as guards/workarounds. Timers are fragile heuristics that break on slow devices and hide the real problem. Use event-based flags and conditions instead. [...] Everything else — including debounce — should use event-based flags.")

`m_remoteActivityTimer` is a receiver-side liveness watchdog — it guards the condition "the phone has disconnected" using a heuristic timeout rather than an event. The correct fix is a relay-server-side `peer_disconnected` message that fires when the phone's WebSocket closes, then tearing down `ScreenCaptureService` in response to that event in `onTextMessageReceived`. If the relay server doesn't support that yet, this PR should at minimum add a comment near the timer declaration explaining why the event-based path (`onDisconnected`) is unreachable here, so future readers don't apply the CLAUDE.md rule to remove the watchdog without understanding the constraint.

https://github.com/Kulitorum/Decenza/blob/0eb89398121049680f45ef6d333d18e249c2f7fd/src/network/relayclient.cpp#L338-L345

2. `kRemoteActivityTimeoutMs` comment presents unimplemented phone-side `keepalive` as a current contract. The PR description says "Until that lands, idle viewing without touches will be cut off after 30s" — confirming the ~10s keepalive cadence is not yet shipped. The comment as written will mislead future maintainers into thinking the phone already sends this.

https://github.com/Kulitorum/Decenza/blob/0eb89398121049680f45ef6d333d18e249c2f7fd/src/network/relayclient.cpp#L16-L22

---

<details>

- [OK to skip] `noteRemoteActivity()` only resets watchdog for `type == 0x02` binary frames — latent gap if protocol gains new binary message types, but only one type exists today (score: 25)
- [Worth a look] `binary_relay` comment says "noteRemoteActivity() runs inside onBinaryMessageReceived after the type byte is checked" — understates the guard: `m_captureService` must also be non-null (score: 25)
- [Worth a look] Double-logging: new `relay_command received:` log in `onTextMessageReceived` duplicates the pre-existing `Handling command:` log inside `handleCommand` (score: 35)

</details>

Issue #500, opened 2026-03-19
[FR] Relative Grind Size Database for quicker dialling-in & AI suggestions


fredphoesh:
Hi @skialpine Jeff and @Kulitorum Michael

For a very long time I've been wanting to have a "Relative Grind Size Database" so I and other users could take an informed guess at how to grind when changing Profiles from one to another. Obviously it's a ball-park grind size, but it's better than nothing. I had started working on a chart of the (then) most popular DE-1 Profiles, doing shots and comparing my grind sizes. Sadly I too was a beginner and gave up on my project, it wasn't supported by certain (snobs) online who would rather find fault in the idea than see how useful it could be. [HERE is the early and very unfinished data collection for my 2 Year old project.](https://docs.google.com/spreadsheets/d/1Uk405_6vf8ifvHaPotZLS49gUKiKDoAhZJgez8fcBm0/edit?usp=sharing)

AI fits VERY well into this realm of possibility, and the general idea is to have a way to do two things. Firstly to know how many STEPS finer or coarser you need to go on your grinder when changing from one Profile to another, and secondly, to be able to "calibrate" your personal grind settings with the "Relative Grind Size Database" (RGSD) so that the AI does not have to make it's own numerical guesses for grind setting changes needed for your future shots. 

When you change beans, the grind size needed on your grinder will often be very different even when using the same Profile, so the “steps” have to be **relative** amounts that one turns the grinder dial for each of the 9 “steps - from the finest grind needed on any DE-1 profile, to the coarsest grind needed, maybe for some gushing turbo shot.
 
To do this, a system has to be created, and once that is done, the user takes a couple of easy actions to find their relative “steps”. There are probably several ways to create the system, this is one way: 

**### Creating the system**
1/ We identify the DE-1 profile which requires the finest grind. We call this the profile for step 1
2/ Same for the coarsest grind, perhaps a gushing TT, and we call this the profile for step 9
3/ We count how many grinder dial values exist between the setting used for step 1 and step 9. This number is then divided by 8 to give the incremental value between each of the “steps”
4/ Then we do loads of shots using many different profiles, taking note of the grind setting that best showcases that profile. Preferably we end with at least one Profile per step. When we do this, it is important that we research well what the creator of the Profile, and users who know it well, consider to be a “shot on target”. Once we know that, we can keep the relevant markers as part of the Profile info and RGSD for AI data use.

**### What the user does**
[...]

PR #1067, opened 2026-05-02
feat(ai): bean-metadata capture + profile family rule + anti-hallucination

skialpine:
## Summary

Closes three gaps surfaced by a real advisor session that recommended LRv2 as an alternative to D-Flow for a dark roast (both are lever-decline — switching is a parameter tweak in disguise) and invented specific temperatures for profiles it had no recipe for.

- **Conversational bean-metadata write-back.** When the user clarifies a bean field mid-conversation (\"actually it's really dark\", \"the bean is from Sey\", \"roasted 2026-04-15\"), `AIManager` parses it and writes back to the anchored shot's metadata via the existing `ShotHistoryStorage::requestUpdateShotMetadata` background path. Mirrors the rating-capture mechanism from `shot-rating-capture`. Bean-identity fields only — physical recordings (dose/yield/grind/duration/curves) stay untouched.
- **Profile family field.** Every entry in `resources/ai/profile_knowledge.md` carries `Family: <name>` (one of `lever-decline`, `pressure-ramp-flow`, `flow-adaptive`, `blooming`, `flat-pressure`, `turbo`, `filter`, `allonge`, `manual`, `volume-based`, `gentle-long-preinfusion`, `tea`, `maintenance`). `buildProfileCatalog` surfaces it as `[family: <name>]` per catalog one-liner. System prompt teaches the model NOT to recommend within-family switches as meaningful profile changes unless the alternative encodes a constraint that can't be replicated by parameter tweaks on the current profile.
- **Other-profile parameter discipline.** System prompt teaches: full recipe data exists only for the current shot's profile (`result.profile.recipe`). For other profiles, only the one-line description is available. Don't invent setpoints — describe differences qualitatively (\"lower temperature regime\", \"higher peak pressure\") and let the user pull a reference shot for actual numbers.

OpenSpec proposal: `openspec/changes/add-shot-metadata-capture/`. Validated `--strict --no-interactive`.

## Test plan

- [x] Build clean (Qt Creator, 0 errors, 0 warnings)
- [x] Full test suite: 2025 passed, 0 failed, 0 warnings
- [x] New tests cover roast-level extraction (with canonicalization), compound-phrase rejection, brand/date/multi-field parsing, gating short-circuits (no shotId, no DB, no parser match), and that the system prompt contains `[family: ...]` tags + the families/anti-hallucination/bean-correction teaching blocks
[...]

PR #1069, opened 2026-05-02
Cross-profile UGS grind ordering + open MCP guidance to all clients

skialpine:
## Summary

- Add a **Cross-Profile Grind Ordering** section to the profile knowledge base, sourced from videoblurb.com/UGS (Mark Renowden's UGS calculator, GitHub issue #500): 16 canonical profiles with UGS positions, inferred positions for off-chart profiles, roast-level transition rules, and an explicit "UGS is not grinder clicks" caveat.
- Add per-profile \`UGS:\` lines on 21 profile sections so the AI sees the value adjacent to other facts when analyzing a shot, plus a \`Skip-Catalog: true\` mechanism for cross-cutting reference sections to keep the one-liner profile catalog clean.
- Add a **Grinder Adjustment Procedure** in \`espressoSystemPrompt()\` that gates magnitude advice through shot history (advisor uses \`dialInSessions\`, MCP clients use \`shots_list\`) and blocks UGS-to-clicks arithmetic when no anchor exists.
- Open \`get_agent_file\` and the \`decenza://dialing/current_context\` resource to all MCP clients (Claude Desktop is the primary target). Restructure \`claude_agent.md\` so universal principles apply to any client and Claude Code Remote Control's filesystem workflow is its own clearly-scoped section.
- Wire the cross-profile reference content into both the in-app advisor system prompt and \`dialing_get_context\` (default mode), so both surfaces see the data on every relevant call.

## Why

Failure mode from a real dialing conversation: the AI advised grinding *finer* when transitioning from a dark roast to a medium roast on the same profile — when *coarser* is correct. The agent file's existing rule ("ground grind changes in shot data, not generic advice") was right but only reached Claude Code Remote Control sessions. This PR makes the rule reach every MCP client and adds the cross-profile data + procedural rule needed to ground both directional and magnitude advice.

## Test plan

- [x] Build clean in Qt Creator (no errors, no warnings).
- [x] \`dialing_get_context\` default mode returns per-profile UGS + cross-profile reference appended to \`profileKnowledge\`.
- [x] \`dialing_get_context\` with \`includeFullKnowledge: true\` returns the full system prompt with cross-profile section injected and Skip-Catalog excluding it from the one-liner catalog.
- [x] Live \`ai_advisor_invoke\` on a real shot produces history-anchored advice (\`grinderSetting: "4.75"\` based on actual \`dialInSessions\` data, no UGS-to-clicks fabrication).
- [x] Cross-profile transition test surfaces a known weakness: the in-app advisor still slips into UGS-to-clicks arithmetic when there's no anchor in \`dialInSessions\`. **Followup planned** (OpenSpec proposal): pre-compute per-profile \`personalGrinderAnchors\` from shot history and inject them into the prompt so the advisor has empirical anchors to use instead of UGS arithmetic.

### DIFF
--- a/docs/CLAUDE_MD/MCP_SERVER.md
+++ b/docs/CLAUDE_MD/MCP_SERVER.md
@@ -223,7 +223,7 @@ This avoids holding HTTP connections and works naturally with the conversational
 | Tool | Description | Category |
 |------|-------------|----------|
 | `debug_get_log` | Read the persisted app debug log. Three modes: (1) `sessions=true` lists all sessions with index/start line/timestamp/line count; (2) `session=N` returns lines from that session (-1 = most recent); (3) default — raw line-based pagination via `offset`/`limit` (1–2000 lines). | read |
-| `get_agent_file` | Returns the current Decenza `claude_agent.md` system-prompt content + the app's version string. Used by Claude Code Remote Control sessions to self-update their CLAUDE.md so agent instructions evolve with app updates without manual user intervention. | read |
+| `get_agent_file` | Returns the current Decenza `claude_agent.md` system-prompt content + the app's version string. Any MCP client (Claude Desktop, Claude mobile, Claude Code, etc.) should call this at session start to load behavioral guidance for dialing assistance. Clients with filesystem access (Claude Code Remote Control) additionally use the version to self-update a local CLAUDE.md. | read |
 
 ### AI Dial-In Conversation (key feature)
 
--- a/docs/PROFILE_KNOWLEDGE_BASE.md
+++ b/docs/PROFILE_KNOWLEDGE_BASE.md
@@ -86,6 +86,7 @@ All other lines are free-form `Field: value` pairs consumed verbatim by the AI s
 | `DO NOT flag ...` | Explicit suppression instruction for the AI (plain sentence, no colon required) |
 | `Note:` | Clarifications or disambiguation |
 | `AnalysisFlags:` | Parsed by code — see above |
+| `Skip-Catalog:` | When `true`, the section is excluded from `buildProfileCatalog()`'s one-liner list and is also lifted out of normal profile-match injection. Used for cross-cutting reference sections (e.g. `## Cross-Profile Grind Ordering`) that aren't actual profiles but live in the KB so they're parseable and editable alongside profile entries. |
 
 The display renderer (`ProfileSelectorPage.qml`) bolds lines matching `Label: value` where the label is ≤35 characters and the line does not start with `-` (bullet lines are never bolded), italicizes `DO NOT` lines, and hides `Also matches:` and `AnalysisFlags:` lines entirely. All KB text is HTML-escaped before rendering.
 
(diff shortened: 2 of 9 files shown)


skialpine:
### Code review

Found 5 issues:

1. The MCP `dialing_get_context` default-mode path appends `crossProfileReferenceContent()` to `profileKnowledge` but omits the "Profile families" and "Other-profile parameter discipline" framing blocks that the in-app system prompt includes alongside the same cross-profile content. Those framing blocks were added in PR #1067 specifically to prevent the LLM from hallucinating numeric setpoints for non-current profiles. Default-mode MCP clients (Claude Desktop) get the cross-profile grind table without that guardrail.

https://github.com/Kulitorum/Decenza/blob/6685aaa90c9074ecf3a7e566cc2d2d0121712692/src/mcp/mcptools_dialing.cpp#L214-L237

2. The cross-profile injection in the MCP default-mode path has no `beverageType` guard. The in-app advisor wraps the same content in `if (beverageType.toLower() != "filter" && beverageType.toLower() != "pourover")` at `src/ai/shotsummarizer.cpp:1342`. For a filter or pour-over shot resolved by `dialing_get_context`, the response will include espresso-specific UGS content (Cremina/Rao Allongé table, dark→medium-roast espresso transition rules) — irrelevant for filter brewing. `bevType` is already in scope at the injection point.

https://github.com/Kulitorum/Decenza/blob/6685aaa90c9074ecf3a7e566cc2d2d0121712692/src/mcp/mcptools_dialing.cpp#L227-L235

3. The `skipCatalog` field doc comment says sections are "excluded from `buildProfileCatalog()` and from per-profile injection." The first is enforced; the second is not — `profileKnowledgeForKbId()` and `findProfileSection()` return `pk.content` without checking the flag. In practice "cross-profile grind ordering" is unlikely to fuzzy-match a real profile title, but the comment documents a contract the code does not enforce.

https://github.com/Kulitorum/Decenza/blob/6685aaa90c9074ecf3a7e566cc2d2d0121712692/src/ai/shotsummarizer.h#L359-L366

4. Two new comments describe the cross-profile content as "injected unconditionally," but injection in `shotAnalysisSystemPrompt()` is gated by `if (beverageType.toLower() != "filter" && ...)` at line 1342. For filter and pour-over beverages no injection happens. "Unconditionally" is only true within the espresso path.

https://github.com/Kulitorum/Decenza/blob/6685aaa90c9074ecf3a7e566cc2d2d0121712692/src/ai/shotsummarizer.h#L249-L258

https://github.com/Kulitorum/Decenza/blob/6685aaa90c9074ecf3a7e566cc2d2d0121712692/src/ai/shotsummarizer.cpp#L1389-L1400

5. `crossProfileReferenceContent()` rebuilds its joined string on every call instead of caching, unlike the neighboring `s_profileCatalog` (built once in `buildProfileCatalog()`) and `s_dialInReference` (built once in `loadDialInReference()`). With ~20 map entries the perf impact is trivial, but it's inconsistent with the established pattern in the same file.

https://github.com/Kulitorum/Decenza/blob/6685aaa90c9074ecf3a7e566cc2d2d0121712692/src/ai/shotsummarizer.cpp#L1564-L1583

<details>

- [OK to skip] `mcptools_dialing.cpp:225-228` comment "without having to ask for the full system prompt" — phrasing nit; clients avoid the parameter cost rather than actively asking (score: 50)

</details>

PR #1070, opened 2026-05-02
feat(accessibility): include phase name and units in shot graph TalkBack announcement

skialpine:
## Summary

- TalkBack users tapping the history shot graph now hear the active extraction phase name alongside the time and curve values
- Units added to every curve value in the announcement (bar, mL/s, °C, g, g/s)
- New `getPhaseAtTime()` helper looks up the phase marker active at the tapped time from the existing `phaseMarkers` array

**Before:** `"10.5. Pressure 9.0. Flow 2.5. Weight 18.0."`
**After:** `"At 10.5 seconds, Rise and Hold phase. Pressure 9.0 bar. Flow 2.5 mL/s. Weight 18.0 g."`

## Test plan

- [ ] Open a saved shot in the history detail page
- [ ] Enable TalkBack on Android
- [ ] Tap different points along the shot graph
- [ ] Verify announcement includes the phase name (e.g. "Pre-infusion", "Rise and Hold", "Decline") at the correct time boundaries
- [ ] Verify units are read aloud for each curve value
- [ ] Verify behavior is unchanged when no phase markers are present (announcement omits the phase clause)

### DIFF
--- a/qml/components/HistoryShotGraph.qml
+++ b/qml/components/HistoryShotGraph.qml
@@ -242,14 +242,18 @@ ChartView {
         return minDist < 1.0 ? closest.y : null
     }
 
-    // Return the phase label active at the given time, or empty string if none
+    // Return the phase label active at the given time, or empty string if none.
+    // Skips "Start"/"End" sentinels — they are structural markers, not user-facing phases.
     function getPhaseAtTime(time) {
         var label = ""
         for (var i = 0; i < phaseMarkers.length; i++) {
-            if (phaseMarkers[i].time <= time)
-                label = phaseMarkers[i].label
-            else
+            var m = phaseMarkers[i]
+            if (m.time <= time) {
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 1 issue:

1. `getPhaseAtTime()` returns the raw `"Start"` and `"End"` sentinel strings as phase names, so tapping at the beginning of a shot announces **"At 0.0 seconds, Start phase"** — meaningless to a TalkBack user. Every other place in the file that consumes `phaseMarkers` filters these sentinels out: `loadMarkers()` skips `"End"` and applies special styling to `"Start"`, and the pump-mode `Repeater` explicitly hides delegates where `modelData.label === "Start"` or `"End"`. The new helper should apply the same filter.

https://github.com/Kulitorum/Decenza/blob/1117fa6d488ff06859eae5d86bfae1f85acbd71b/qml/components/HistoryShotGraph.qml#L245-L255

PR #1082, opened 2026-05-03
fix(ui): remove per-shot dismiss from QuickRatingRow

skialpine:
## Summary

- Removes the × dismiss button from `QuickRatingRow` — the prompt always shows for unrated shots when in edit mode
- Drops `dismissed` property, `dismissedClicked` signal, `_ratingPromptDismissed` page property, and `onEditShotIdChanged` QSettings cache
- Removes the collapsed "Rated N — tap to revise" pill and its dead `currentScore`/`reviseClicked` API — the row is visible only when unrated so the pill was never reachable
- Fixes visibility condition to `enjoymentSource !== "user"`: inferred scores are internal AI signals; users see 0 and are prompted to rate
- Fixes C++ invariant: `enjoyment=0` now saves `enjoymentSource="none"`, never `"user"`

## Why

The dismiss suppressed the rating row permanently per shot via a QSettings key. If a user went back to edit an unrated shot, the row would never reappear.

Inferred scores (Layer 3 auto-rater) were leaking into the UI — the slider showed 75 for shots the user had never rated.

Orphaned `shotRatingDismissed/*` QSettings keys on existing installs are harmless dead data.

## Test plan

- [ ] Open post-shot review for an unrated shot — QuickRatingRow shows three emoji icons, no × button
- [ ] Open post-shot review for an AI-inferred shot (`enjoymentSource="inferred"`) — row visible, slider shows 0
- [ ] Tap a face — score saves, row hides; re-open shot — row stays hidden
- [ ] Navigate away and back to the same unrated shot — row is still visible

### DIFF
--- a/qml/components/QuickRatingRow.qml
+++ b/qml/components/QuickRatingRow.qml
@@ -6,13 +6,9 @@ import Decenza
 // QuickRatingRow — issue #1055 Layer 2.
 //
 // Low-friction one-tap rating row shown above the metadata fold on the
-// post-shot review page when the shot is unrated. Three icon buttons map
-// to default scores 80 / 60 / 40 (good / okay / bad). Tapping persists
-// the score immediately via the existing saveEditedShot path.
-//
-// Once the shot is rated (currentScore > 0) the row collapses into a
-// compact "Rated N — tap to revise" pill that reopens the three-icon
-// state on tap, so the row is also a fast revision surface.
+// post-shot review page for unrated shots (enjoymentSource != "user").
+// Three icon buttons map to scores 80 / 60 / 40 (good / okay / bad).
+// Tapping persists the score immediately via saveEditedShot.
 //
 // Translation: all visible text routes through TranslationManager so
 // non-English locales work. Accessibility: each interactive element is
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 2 issues:

1. The collapsed \"Rated N — tap to revise\" pill and `reviseClicked` signal are unreachable dead code. The outer `QuickRatingRow` has `visible: postShotReviewPage.isEditMode && editEnjoyment === 0`, but the pill inside has `visible: root.currentScore > 0`. These are mutually exclusive: the pill only renders when `currentScore > 0`, but the outer component is hidden when `editEnjoyment > 0`. The pill can never be shown, and `onReviseClicked` can never fire.

https://github.com/Kulitorum/Decenza/blob/da91aca1827f5735184f0d60abc4b0d79b0449e6/qml/pages/PostShotReviewPage.qml#L674-L680

https://github.com/Kulitorum/Decenza/blob/da91aca1827f5735184f0d60abc4b0d79b0449e6/qml/components/QuickRatingRow.qml#L116-L126

2. The new `editEnjoyment === 0` visibility condition hides `QuickRatingRow` for inferred-75 shots. The Layer 3 auto-rater (`shothistorystorage.cpp:999-1005`) saves `enjoyment = 75` / `enjoymentSource = "inferred"` for clean shots. `onShotReady` loads this into `editEnjoyment` via `editShotData.enjoyment0to100 ?? 0`, so `editEnjoyment` is 75, not 0, and the row is hidden. The old condition (`enjoymentSource !== "user"`) let inferred shots through so users could confirm or override. The new condition silently skips that confirmation path.

https://github.com/Kulitorum/Decenza/blob/da91aca1827f5735184f0d60abc4b0d79b0449e6/qml/pages/PostShotReviewPage.qml#L136-L143

PR #1084, opened 2026-05-04
perf(saw): faster convergence — N=5→3, graduation 2→1, read window 5→3

skialpine:
## Summary

- **Batch size N=5→3**: SAW has no feedback loop (a stop command does not alter pump pressure or flow dynamics for the next shot). The conservative N=5 from flow calibration is unnecessary here — N=3 converges 2× faster with equal or better accuracy.
- **Graduation threshold 2→1**: A single confirmed 3-shot batch is sufficient signal. Profiles graduate after 3 shots instead of 10.
- **Read window 5→3**: Consistent with the new batch and graduation sizes; averages last 3 committed medians instead of 5.

## Simulation Results (Apr 2026 real shot data)

**D-Flow/Q** (9 shots, true converged lag ~0.637 s):
- N=5: never graduates within 9 shots; systematic undershoot every shot using stale globalPool=0.369 s
- N=3: outlier at shot 3 (lag=1.681, likely channeling) → isolated by deviation gate, batch dropped → clean restart → graduates at shot 6 (lag=0.734 s) → converges to exactly 0.637 s at shot 9

**80's Espresso** (15 shots, new-regime true lag ~0.092 s after dose change):
- N=5: graduates at shot 10
- N=3: graduates at shot 3

## Test plan
- [x] All 17 `tst_SawSettings` tests pass
- [x] `commitBatch` helper updated to N=3
- [x] `batchAccumulatesUntilThreeThenCommits` replaces the N=5 batch test
- [x] `batchRejectedWhenDispersionTooHigh` updated to 2-tight + 1-outlier batch
- [x] `coldStartFallsBackToScaleDefaultThenBootstrapThenPerProfile` step 3 requires only 1 `commitBatch`
- [x] `singleMedianStillFallsBackToBootstrap` → `zeroMediansStillFallsBackToBootstrap` (old invariant is now invalid; 1 median now graduates)
- [x] `SAW_LEARNING.md` updated throughout

### DIFF
--- a/docs/CLAUDE_MD/SAW_LEARNING.md
+++ b/docs/CLAUDE_MD/SAW_LEARNING.md
@@ -34,10 +34,10 @@ Shot 1054 is a separate point: even within a single profile, end-of-shot flow ca
 The same architecture that fixed the analogous problem in flow calibration ([AUTO_FLOW_CALIBRATION.md](AUTO_FLOW_CALIBRATION.md), issue [#739](https://github.com/Kulitorum/Decenza/issues/739)) applies cleanly to SAW:
 
 1. **Per-(profile, scale) history** — each pair learns its own drip dynamics. Switching profiles or scales does not contaminate the active pair's data.
-2. **Batched median commits** — accumulate 5 shots' worth of pending entries before changing the model. The median is robust to single-shot outliers (channeling, scale glitches, cup interaction). 5 was chosen to match flow cal.
-3. **Batch-level dispersion check** — if the 5 lags within a batch are too spread out (IQR > 1.0 s, or any single lag > 1.5 s from the batch median), the whole batch is dropped. Dispersion that high indicates the user changed conditions mid-batch (different beans, different grinder, manual stop).
+2. **Batched median commits** — accumulate 3 shots' worth of pending entries before changing the model. The median is robust to single-shot outliers (channeling, scale glitches, cup interaction). N=3 is chosen because SAW has no feedback loop (a stop command does not alter pump dynamics for the next shot), so a single confirmed 3-shot batch is sufficient signal — the conservative N=5 from flow calibration is not needed here.
+3. **Batch-level dispersion check** — if the 3 lags within a batch are too spread out (any single lag > 1.5 s from the batch median), the whole batch is dropped. Dispersion that high indicates the user changed conditions mid-batch (different beans, different grinder, manual stop). IQR gating requires ≥ 4 values; with N=3 the per-element deviation check handles all outlier cases.
 4. **Global bootstrap** — the median lag across all graduated `(profile, scale)` pairs on the same scale is published as `saw/globalBootstrapLag/<scaleType>`. New pairs use this as their first-shot default instead of the scale's hardware-only `sensorLag()`.
-5. **Read-path fallback chain** — `perProfile` (≥ `kSawMinMediansForGraduation` committed batches, currently 2 = 10 SAW shots) → `globalBootstrap` → `globalPool` (legacy entries) → `scaleDefault`. New users / new pairs degrade gracefully.
+5. **Read-path fallback chain** — `perProfile` (≥ `kSawMinMediansForGraduation` committed batches, currently 1 = 3 SAW shots minimum) → `globalBootstrap` → `globalPool` (legacy entries) → `scaleDefault`. New users / new pairs degrade gracefully.
 
 ### Why this is the right shape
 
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 9 issues:

1. Stale log format in docs — the logging table entry for batch rejection still says `[SAW] batch rejected — high dispersion median_lag=… iqr=…` but the actual `qWarning` emits `[SAW] batch rejected — outlier lag=… deviates …s > …s from median`. The test regex at `tests/tst_saw_settings.cpp:79` already matches the real format, confirming the doc is stale.

https://github.com/Kulitorum/Decenza/blob/7c85fd543fdc0b10f9e7fdad57cf844d6c05fc71/docs/CLAUDE_MD/SAW_LEARNING.md#L150-L157

2. Dead code: `kBatchMaxIqr` is unreachable at N=3 — `iqrOf()` explicitly returns `0.0` when `v.size() < 4`, so with `kBatchSize = 3` the IQR gate at line 814 (`if (lagIqr > kBatchMaxIqr)`) can never be true. The constant is now unused in any meaningful sense and the branch is permanently dead. If `kBatchSize` is ever raised above 3 in the future this gate would silently re-enable with no comment explaining its semantics.

https://github.com/Kulitorum/Decenza/blob/7c85fd543fdc0b10f9e7fdad57cf844d6c05fc71/src/core/settings_calibration.cpp#L22-L26

https://github.com/Kulitorum/Decenza/blob/7c85fd543fdc0b10f9e7fdad57cf844d6c05fc71/src/core/settings_calibration.cpp#L799-L816

3. Stale comment in `addSawLearningPoint` — the comment at line 464 says "The pending batch holds **5 shots**" (should be 3) and "provides outlier rejection via the median + **IQR check**" (the IQR gate is dead at N=3 as noted above).

https://github.com/Kulitorum/Decenza/blob/7c85fd543fdc0b10f9e7fdad57cf844d6c05fc71/src/core/settings_calibration.cpp#L462-L470

4. Stale comment in `getExpectedDripFor` — "is small (capped at 10)" refers to `kMaxPairHistory` as the binding limit, but line 725 now caps the fetch at `entries.size() < 3`. The storage cap of 10 is no longer the operative bound for this read path; a reader of this comment will not realize only 3 entries are ever fetched.

https://github.com/Kulitorum/Decenza/blob/7c85fd543fdc0b10f9e7fdad57cf844d6c05fc71/src/core/settings_calibration.cpp#L718-L727

5. SAW_LEARNING.md still says "5 shots" for cold-start — "A new profile starting from `sensorLag(scaleType) + 0.1 s` will be off until **5 shots** populate its history." With `kBatchSize = 3` and `kSawMinMediansForGraduation = 1` a new profile graduates after 3 shots (one batch), not 5.

https://github.com/Kulitorum/Decenza/blob/7c85fd543fdc0b10f9e7fdad57cf844d6c05fc71/docs/CLAUDE_MD/SAW_LEARNING.md#L131-L134

6. `kBatchSize` type is `constexpr int`, not `constexpr qsizetype` — [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/7c85fd543fdc0b10f9e7fdad57cf844d6c05fc71/CLAUDE.md) says "Use `qsizetype` (not `int`) for container sizes… Assigning to `int` causes `-Wshorten-64-to-32` warnings." `kBatchSize` is directly compared against `batch.size()` at line 772 where `batch` is a `QJsonArray` whose `size()` returns `qsizetype`. `kSawMinMediansForGraduation` on line 20 already correctly uses `constexpr qsizetype`; `kBatchSize` and `kMaxPairHistory` on lines 22-23 do not.

https://github.com/Kulitorum/Decenza/blob/7c85fd543fdc0b10f9e7fdad57cf844d6c05fc71/src/core/settings_calibration.cpp#L20-L25

7. Stale "5 real shots" in `recomputeGlobalSawBootstrap` — the comment says "a single committed batch median **(5 real shots)** is already more informative…" but a batch now contains 3 shots.

https://github.com/Kulitorum/Decenza/blob/7c85fd543fdc0b10f9e7fdad57cf844d6c05fc71/src/core/settings_calibration.cpp#L896-L902

8. Write-path pseudocode still shows IQR as an active condition — `if IQR > 1.0 s or any |lag - median_lag| > 1.5 s:` implies both branches fire, contradicting the note at line 129 ("IQR gating requires ≥ 4 values; N=3 relies exclusively on the per-element deviation check") in the same doc.

https://github.com/Kulitorum/Decenza/blob/7c85fd543fdc0b10f9e7fdad57cf844d6c05fc71/docs/CLAUDE_MD/SAW_LEARNING.md#L102-L107

9. `sawLearnedLag()` global-pool read window was not updated — the PR title says "read window 5→3" but `sawLearnedLag()` (the legacy global-pool fallback, used when no per-pair data exists) still reads `count < 5` at line 230. The two per-pair paths (`sawLearnedLagFor` and `getExpectedDripFor`) were updated; this fallback was not. If intentional, a comment should explain why the global pool uses a different window.

https://github.com/Kulitorum/Decenza/blob/7c85fd543fdc0b10f9e7fdad57cf844d6c05fc71/src/core/settings_calibration.cpp#L227-L233

---

Issue #1093, opened 2026-05-05
[Bug] BLE connectivity conflict between DE1 and Eureka Precisa scale


mcastaldelli:
### What happened?

Hi Jeff, unfortunately the new build didn't work either, and neither did downgrading the FW back to stable :-(

### Steps to reproduce

Open Decenza and let the DE1 come online. Try to change profiles and confirm working with purple light. Connect Precisa. Try to change profiles again. This time it won't flash purple.

### Decenza version

version 1.7.4 build 3359

### Platform

Android

### Device / OS version

"Android Pie (9.0)" arch: "arm64" kernel: "linux" "4.4.147" Decent P80X Tablet

### Scale (if applicable)

Eureka Precisa

### Debug log / shot debug log

[debug 4.log](https://github.com/user-attachments/files/27415366/debug.4.log)

skialpine:
Wish I had your scale to test with, but we will figure it out. thanks for your patience.

mcastaldelli:
Thank you, Jeff!

skialpine:
I added more logging to a new build. If you could try again, let me know what happens and attach the log I would appreciate it. There is a small chance it is fixed in the new build, but I really just hope to figure out what is happening.

mcastaldelli:
Good morning, Jeff. Here's the new log with the added data. Thanks for the perseverance! let me know if you want me to test any other scenario.

[debug 5.log](https://github.com/user-attachments/files/27433561/debug.5.log)

skialpine:
Was that from the latest BETA? could you check if there is a newer beta to collect the log from. I t looks like you are running build 3359 and I was hoping for build 3362. Could you check for another build and try again. Sorry this is so time consuming. I think we are getting close.

skialpine:
hold off if you have not started. I am going to add more logging and put out another build.

mcastaldelli:
Hi Jeff,

Yes, the log above (debug 5.log) was from the latest build (3362). 

I also noticed the problem was not there in a pretty old build I used (1.6.0 - build 3188). I rolled back to that version and it's working there. Let me know if you want a log from it as well.

/Michel

skialpine:
a log from a working system would be great. I also could use a log from the newest build 3364. After you grab the log from 1.6.0 could you upgrade to build 3365.

Appreciate you helping out!

mcastaldelli:
[...]

Issue #1091, opened 2026-05-05
[Bug] BLE connectivity conflict with Eureka Precisa scale


mcastaldelli:
### What happened?

Hi Jeff, unfortunately the new build doesn't fix the issue. The same thing happens: Decenza connects fine with the DE1 until the Precisa is connected. Then it loses connectivity with the DE1 and stays reporting the weight from Precisa.

### Steps to reproduce

Open Decenza and let the DE1 come online. Try to change profiles and confirm working with purple light. Connect Precisa. Try to change profiles again. This time it won't flash purple.

### Decenza version

1.7.4

### Platform

Android

### Device / OS version

Android DE1 tablet - P80X

### Scale (if applicable)

Eureka Precisa

### Debug log / shot debug log

[debug.zip](https://github.com/user-attachments/files/27412365/debug.zip)

skialpine:
a new and different bug... playing wack a mole. your scale is a little different then mine. New build soon.

mcastaldelli:
Thanks. A piece of extra info that may be pertinent or not. It seems the issues started after I updated the DE1 firmware to the latest available in the beta channel.. If the next release doesn't fix it, I may try rollimg back the firmware too.

skialpine:
Next fixes are done. please try updated build when you have time and again open up a new issue if still having problems.

mcastaldelli:
Will do. I saw the new source code in the assets section. Waiting for the new build to show up. Are you also on FW v1352? I have a hunch this may be causing the issue, and it may spread to more users as the FW moves to stable and more people load it into their DE1s.

Issue #1087, opened 2026-05-05
[Bug] BLE connectivity issue in recent versions


mcastaldelli:
### What happened?

Versions 1.7.3 and 1.7.4 are displaying BLE connectivity issues in my setup with an Eureka Precisa scale and the original P80X tablet from Decent.

The DE1 disconnects as soon as I connect the scale. Only one of the devices stay properly connected. See screenshots.

<img width="2024" height="2698" alt="Image" src="https://github.com/user-attachments/assets/e1aa0c32-5ad2-4bc5-a88c-a201ff92ff4b" />

<img width="3520" height="2640" alt="Image" src="https://github.com/user-attachments/assets/8cc1488a-f1d6-46e5-997f-00a217111b17" />

<img width="3520" height="2640" alt="Image" src="https://github.com/user-attachments/assets/3744ecd9-a5e2-4a48-ada5-711d4eb485f4" />

I went back to the Decent app, and the problem doesn't happen with it.

### Steps to reproduce

Open Decenza, wait for it to connect to the DE1. Connect the scale. The connection to the DE1 still shows as "connected", but commands no longer work and the connectivity errors in the screenshot show up in the log.

### Decenza version

1.7.4

### Platform

Android

### Device / OS version

Decent tablet P80X

### Scale (if applicable)

Eureka Precisa

skialpine:
Can you attach the system log so I can see what is happening on you system? Not reproducible on mine. You could  try and power cycle the de1 to see if it solves the issue.

skialpine:
To get at the log file you go into settings on Decenza and from the History and Data tab you enable the server. Once you enable the server from your desktop computer go to the URL shown under the enable server to get to the Decenza web interface. On the web interface go to the debug tools menu (top right menu icon) and once on the debug page click the download log button.

mcastaldelli:
Thanks for the help! Here's the log. I downgraded (uninstalled the app first) to 1.7.2 to check, but it still shows the same issue.

When I initially wake up the app, it connects to the DE1 with no issue. If I try to select a new profile, it gets sent to the machine, which blinks purple to acknowledge it. After I turn the scale on and get it connected, the DE1 still shows as connected, but loading new profiles do nothing (i.e., no purple light). The log also shows the write errors. 

I power cycled the machine, the tablet, reset the Bluetooth and Wireless stacks.. nothing worked, except going back to the Androwish TCL app.

[...]

PR #1094, opened 2026-05-05
fix(ble): drop redundant scale CCCD auto-enable + add cross-transport diagnostics

skialpine:
## Summary

Two-part change targeting the Precisa-on-P80X scale interference bug ([#1087](https://github.com/Kulitorum/Decenza/issues/1087), [#1091](https://github.com/Kulitorum/Decenza/issues/1091), [#1093](https://github.com/Kulitorum/Decenza/issues/1093)).

### 1. Fix experiment: drop the auto-CCCD-enable in `QtScaleBleTransport`

`QtScaleBleTransport::onServiceStateChanged` was auto-writing CCCDs for **every** notify-capable characteristic right after service discovery, in addition to each scale's own `enableNotifications()` call. That's two CCCD writes per characteristic, hitting the Android GATT pipeline at exactly the moment scale connection completes.

In all four user logs across 1.7.2 / 1.7.3 / 1.7.4, the failure pattern is the same: scale connects, then 5–10 s later DE1 writes start failing with `CharacteristicWriteError` on `0xa00f` / `0xa010`. The timing matches the auto-CCCD writes landing on the shared HCI pipeline.

The `CoreBluetoothScaleBleTransport` (iOS/macOS) already disabled this auto-enable, citing Bookoo issues. de1app never auto-writes CCCDs at all. We're matching that behavior on Qt's path.

If hypothesis correct → user's profile changes after scale connect start working again. If wrong → no harm done (each scale already writes its own CCCD), and the diagnostics below tell us the real cause.

### 2. Cross-transport diagnostics, in case the experiment fails

If the auto-enable wasn't the cause, the next log we get back will let us pin it down without another round-trip:

- Every scale `writeCharacteristic` call → `[BLE QtTransport] write <uuid> (<n> bytes, WithResponse)`
- Every scale CCCD enable → `[BLE QtTransport] write CCCD enable <uuid>`
- Every DE1 `writeCharacteristic` submission → `[BLE DE1] write submit <uuid> (<n> bytes)`
- Scale notify-rate summary every 5 s → `[BLE QtTransport] notify rate: 124 in last 5012 ms (last 0000fff1 = 9 bytes)`

Together these give us a timeline of every BLE operation on both transports on the shared Android GATT pipeline. When DE1 writes fail we'll see exactly what scale traffic was happening in the same window — descriptor writes, command writes, or notify bursts.

## Test plan

- [ ] User installs build and reproduces the original repro: connect DE1, change profiles (works), connect Precisa, change profiles
- [ ] If profile changes now flash purple after scale connect → fix worked, remove diagnostics in a follow-up PR
- [ ] If still broken → next debug log identifies the actual conflicting BLE operation

### DIFF
--- a/src/ble/bletransport.cpp
+++ b/src/ble/bletransport.cpp
@@ -663,6 +663,9 @@ void BleTransport::writeCharacteristic(const QBluetoothUuid& uuid, const QByteAr
     m_lastWriteUuid = uuidShort;
     m_lastWriteData = data;
     m_writeTimeoutTimer.start();
+    // Log every submission so we can correlate DE1 write failures with surrounding scale BLE
+    // activity (scale CCCD writes, scale notify bursts) on the same Android GATT pipeline.
+    log(QString("write submit %1 (%2 bytes)").arg(uuidShort).arg(data.size()));
     m_service->writeCharacteristic(m_characteristics[uuid], data);
 }
 
--- a/src/ble/transport/qtscalebletransport.cpp
+++ b/src/ble/transport/qtscalebletransport.cpp
@@ -1,6 +1,7 @@
 #include "qtscalebletransport.h"
 #include "../blecapability.h"
 #include "../blemanager.h"
+#include <QDateTime>
 #include <QDebug>
 #include <QTimer>
 #include <QLowEnergyConnectionParameters>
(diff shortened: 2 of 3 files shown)


skialpine:
### Code review

Found 1 issue:

1. The new comment claims the CoreBluetooth transport disabled the auto-enable "for the same reason," but CoreBluetooth's actual stated rationale is Bookoo-scale confusion from double-enabling — not Android GATT pipeline starvation. Reading either file together creates a contradictory record.

https://github.com/Kulitorum/Decenza/blob/b1ac38abb5ebe4b2a9ba1e54ea617a49856d6aef/src/ble/transport/qtscalebletransport.cpp#L362-L370

Suggested rephrase: "The CoreBluetooth transport already disabled this (its comment cites Bookoo-scale confusion from double-enabling; here we remove it to reduce Android GATT pipeline pressure)."

<details>

- [Worth a look] "with auto-enable removed and keepalive CCCD re-writes gone" at [qtscalebletransport.cpp#L183-L192](https://github.com/Kulitorum/Decenza/blob/b1ac38abb5ebe4b2a9ba1e54ea617a49856d6aef/src/ble/transport/qtscalebletransport.cpp#L183-L192) — the "keepalive CCCD re-writes" reference is opaque to a reader of only this file (those were removed in #1092 from per-scale `sendKeepAlive()` methods). Could either drop the parenthetical or anchor it to where to look. (score: 50)

</details>

PR #1101, opened 2026-05-06
feat(qt): upgrade to 6.11.1 + GPU-accelerated CupFillView

skialpine:
## Summary

Two changes shipped together because the second depends on the first:

1. **Qt 6.10.3 → 6.11.1** in all 6 CI workflow files (`jurplel/install-qt-action`), Windows sccache cache key, and version references in `CMakeLists.txt`, `CLAUDE.md`, `openspec/config.yaml`, `README.md`, and the iOS/build docs.
2. **GPU-accelerated CupFillView** via a small `QCanvasPainter` wrapper.

## CupFillView GPU acceleration

Adds `JsCanvasPainterItem` — a `Canvas`-like QML wrapper around Qt 6.11's new `QCanvasPainter`. The QML `onPaint(ctx)` handler records draw commands into a flat POD buffer on the main thread; the renderer atomic-swaps it in `synchronizeData()` and replays through `QCanvasPainter` on the scene-graph render thread. Rendering goes through Qt's RHI (Metal / D3D12 / Vulkan / OpenGL).

Wrapper scope = exactly what CupFillView needs today: `beginPath`, `moveTo/lineTo`, `arc`, `ellipse`, `fill/stroke`, `fillRect/clearRect`, `save/restore/reset`; properties `fillStyle`, `strokeStyle`, `lineWidth`, `lineCap`, `globalAlpha`; linear + radial gradients with `addColorStop`. `fillColor` defaults to transparent + alphaBlending enabled so it composites under the `cup_mask` + `cup_lighten` ShaderEffect layers.

The override hook is `synchronizeData()` rather than `synchronize()` — Qt 6.11.1 renamed it (QTBUG-145406); the old name still works via a back-compat shim in 6.11.x but is removed in 6.12, so we target the forward-compatible name now.

Registered via `qmlRegisterType` in `main.cpp` (matches the project's existing convention; `QML_NAMED_ELEMENT` auto-registration fails because qmltyperegistrar can't resolve parent-type headers from QtCanvasPainter / QtQuick).

Files:
- `src/ui/jscanvaspainteritem.{h,cpp}` — new
- `src/ui/jscanvascontext.{h,cpp}` — new
- `qml/components/CupFillView.qml` — `Canvas { renderStrategy: Canvas.Threaded; onPaint: { var ctx = getContext("2d"); ... } }` → `JsCanvasPainterItem { onPaint: function(ctx) { ... } }` (×2)
- `CMakeLists.txt` — `Qt6::CanvasPainter` linked, new sources/headers added
- `src/main.cpp` — `qmlRegisterType<JsCanvasPainterItem>`

## Windows CI workaround (temporary)

Qt 6.11.0 reorganised the Windows online-installer repo layout (`qt6_6110/qt6_6110/Updates.xml` → per-arch `qt6_6110_msvc2022_64/`, etc.). `aqtinstall 3.3.0` (the latest release, default in `install-qt-action@v4`) only knows the old layout. The fix landed on aqtinstall master via miurahr/aqtinstall#1000 but isn't released yet (PyPI is still at 3.3.0 as of 2026-05-13). `windows-release.yml` pins `aqtsource` to that merge commit. Marked with `TODO(qt-6.11):` for cleanup once aqtinstall 3.4.0 ships on PyPI. Other 5 platforms work with stock `3.3.0`.

## Test plan

- [x] macOS Qt Creator build passes
- [x] macOS smoke: launch, simulator extraction, cup graphic + coffee fill render correctly
- [x] All 6 CI builds green on this branch (Windows / macOS / iOS / Android / Linux x64 / Linux arm64)
- [x] Decent tablet build + visual parity vs. Canvas baseline (steady-state perf captured: ~3.5 ms recording + ~230 µs replay per frame, ~11% of one core, well within 30 fps budget; old Canvas software-rasterization thread is gone)
- [x] Decent tablet CPU profile during a live shot via `/proc/<pid>/task` snapshot
- [x] Local post-rebase run on Qt 6.11.1: JsCanvasPainterItem initialised on Metal RHI, full sim shot rendered with no slow-record warnings, shot saved to history cleanly

## Pre-merge checklist

- [ ] Re-evaluate Windows `aqtsource` pin: drop it if `aqtinstall >= 3.4.0` is on PyPI, otherwise keep until it lands (greppable via `TODO(qt-6.11)`)
- [ ] Squash-merge per project convention (`/merge-pr` or `gh pr merge 1101 --squash --delete-branch`)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


skialpine:
### Code review

Found 4 issues:

1. **All 6 CI workflows omit `qtcanvaspainter` from `jurplel/install-qt-action` `modules:`, but `CMakeLists.txt` requires `Qt6::CanvasPainter`** — `QtCanvasPainter` is a separate Qt 6.11 module repo (`github.com/qt/qtcanvaspainter`); it does not ship in qtbase. Every CI build will fail at the CMake configure step (`Could not find Qt6 component CanvasPainter`). Add `qtcanvaspainter` to each workflow's `modules:` line.

https://github.com/Kulitorum/Decenza/blob/3c26011e3c06972df63fe02bfcba5466e26f2911/.github/workflows/macos-release.yml#L55-L60
https://github.com/Kulitorum/Decenza/blob/3c26011e3c06972df63fe02bfcba5466e26f2911/CMakeLists.txt#L77-L79

2. **Radial gradient inner-circle center is silently discarded — visible regression for the crema highlight** — `applyBrush()` builds `QCanvasRadialGradient(s.x1, s.y1, s.r1, s.r0)`, ignoring the inner circle center `(s.x0, s.y0)`. The crema gradient at `CupFillView.qml:259` deliberately offsets the inner focal point by `cremaRx * 0.1` to produce the asymmetric "lit-from-one-side" highlight. With this approximation the crema renders with a centered focal point — the asymmetry is dropped. The other two radial gradients in CupFillView (splash glow, center spot) have matching centers so they're unaffected.

https://github.com/Kulitorum/Decenza/blob/3c26011e3c06972df63fe02bfcba5466e26f2911/src/ui/jscanvaspainteritem.cpp#L42-L57
https://github.com/Kulitorum/Decenza/blob/3c26011e3c06972df63fe02bfcba5466e26f2911/qml/components/CupFillView.qml#L256-L272

3. **Comment overstates necessity of the renderer-side `p->reset()`** — comment says \"without our own reset here, residual state from the previous swap would leak in,\" but both QML handlers always start with `ctx.reset()`, which is recorded as a `DrawCmd::Op::Reset` and replayed inside the same `paint()` loop. The renderer-side reset is currently redundant defensive code, not load-bearing as the comment implies. Reframe as defense-in-depth.

https://github.com/Kulitorum/Decenza/blob/3c26011e3c06972df63fe02bfcba5466e26f2911/src/ui/jscanvaspainteritem.cpp#L74-L79

4. **`m_loggedInit` and surrounding comment say \"first paint\" but the block runs in `requestPaint()`, before the actual paint** — the actual paint happens later in `JsCanvasPainterItemRenderer::paint()` on the render thread. The log fires on the first `requestPaint()` call where `window()` is non-null. \"First paint\" is misleading for anyone debugging render-thread timing.

https://github.com/Kulitorum/Decenza/blob/3c26011e3c06972df63fe02bfcba5466e26f2911/src/ui/jscanvaspainteritem.cpp#L160-L172
https://github.com/Kulitorum/Decenza/blob/3c26011e3c06972df63fe02bfcba5466e26f2911/src/ui/jscanvaspainteritem.h#L33-L37

<details>

- [Worth a look] `qml/components/CupFillView.qml:391` shifts `g.fillTopY += yOff` but the value is never read again in the effects handler — `effectiveFillTopY` is recomputed independently. Pre-existing dead line in modified file; one-line removal. (score: 50)
- [Worth a look] `jscanvaspainteritem.cpp:15-16` says \"if it exceeds half a 30 fps frame we'll drop a frame\" while line 173-174 in the same file correctly says \"at risk of dropping frames.\" Match the wording. (score: 50)

</details>

skialpine:
### Code review

Found 4 issues:

1. **`int` narrowing of `QVector::size()` violates the project's qsizetype rule.** `src/ui/jscanvascontext.cpp:117` explicitly casts a `qsizetype` return to `int` (`return int(m_brushes.size()) - 1;`), and the surrounding API (`int brushId` field, `int newBrush()`, `int brushAt(int id)`) uses `int` for `QVector` indices. CLAUDE.md says: *"Use `qsizetype` (not `int`) for container sizes — `QVector::size()`, `QList::size()`, `QString::size()` etc. return `qsizetype` (64-bit on iOS/macOS). Assigning to `int` causes `-Wshorten-64-to-32` warnings."* The `DrawCmd::brushId` POD field is defensible as a deliberate fixed-width SoA layout, but the unguarded narrowing cast at the `newBrush()` return is the unambiguous violation; at minimum make the downcast explicit (`static_cast<int>(…)`) with a comment, otherwise the iOS/macOS build will warn.

https://github.com/Kulitorum/Decenza/blob/ce33fe7dd2bb64ae91150cd1c2ea18c403b7635b/src/ui/jscanvascontext.cpp#L115-L119

2. **Stale `synchronize` reference in a comment that the rename commit (`ce33fe7`) missed.** `src/ui/jscanvascontext.h:105` still reads `// Buffer access for the renderer (called only on synchronize, main blocked)`, but the override was renamed in this PR to `synchronizeData()` (and the companion comment in `jscanvaspainteritem.h:21-22` was updated). One-word fix to keep the doc consistent with the code.

https://github.com/Kulitorum/Decenza/blob/ce33fe7dd2bb64ae91150cd1c2ea18c403b7635b/src/ui/jscanvascontext.h#L103-L107

3. **"Multiple calls per frame coalesce inside Qt's update mechanism — only the latest recording wins" misdescribes the actual mechanism.** What makes the latest recording win is `m_ctx.resetForNextFrame()` at the top of every `requestPaint()` — each call clears the previous recording before re-running the QML handler. Qt's `update()` coalesces *paint scheduling* (so `synchronizeData()` + `paint()` run at most once per vsync) but does not "pick" a recording. A maintainer reading the comment to add an accumulate-across-calls mode would draw the wrong conclusion.

https://github.com/Kulitorum/Decenza/blob/ce33fe7dd2bb64ae91150cd1c2ea18c403b7635b/src/ui/jscanvaspainteritem.cpp#L160-L172

4. **`docs/IOS_CI_FOR_CLAUDE.md:276` says "Minimum iOS: 14.0 (set in CMakeLists.txt)" — the actual value is 17.0.** `CMakeLists.txt:135-136` sets `CMAKE_OSX_DEPLOYMENT_TARGET "17.0"` with the comment "Qt 6.11.1 was built for iOS 17.0". The PR touched the adjacent line 277 (Qt Version) in this file, so a stale value can be corrected in the same change.

https://github.com/Kulitorum/Decenza/blob/ce33fe7dd2bb64ae91150cd1c2ea18c403b7635b/docs/IOS_CI_FOR_CLAUDE.md#L274-L278

<details>
<summary>1 below-threshold issue (50–74)</summary>

- [Worth a look] `kSlowRecordThresholdNs = 16'000'000` is described as "half a 30 fps frame budget", but half of 33.3 ms is 16.67 ms — 16 ms is closer to half a 60 fps frame. Either widen the constant to `16'666'666` or reword the comment. (score: 50)

</details>

Issue #1089, opened 2026-05-05
[Bug] New version download stalls when launched from wake-up prompt


N6GQ:
### What happened?

This was a bug before, some time ago, and seems to be back. The downloading of a new installable version stalls at some point, never to restart. Kill the app, restart, try again works about half the time.

### Steps to reproduce

1. Wake Decenza from sleep
2. prompt often appears to download and install new version
3. Accept prompt to download - app begins downloading new version
At that point it often stalls and stops downloading. Sometimes around 50%, sometimes more, but this happens more often than not. Nothing I can do to get it to move again once it stalls other than restarting the app and trying again - and in that case sometimes it will work on the 2nd attempt, sometimes it takes 2 or 3 more tries before it starts and finishes and completes.
This is on an A9+ and using the beta, regular updates. Its been happening for a few weeks now I think...

### Decenza version

1.7.3

### Platform

Android

### Device / OS version

Newest android on Samsung A9+

### Scale (if applicable)

HDS

skialpine:
can you please attach the system debug log to this issue. I do not see the issue so hope the log will provide some details on what you are seeing. 

To get at the log file you go into settings on Decenza and from the History and Data tab you enable the server. Once you enable the server from your desktop computer go to the URL shown under the enable server to get to the Decenza web interface. On the web interface go to the debug tools menu (top right menu icon) and once on the debug page click the download log button.

N6GQ:
Here’s my log. It happened this morning FWIW

[debug.zip](https://github.com/user-attachments/files/27415501/debug.zip)

skialpine:
Are you switching apps while the download is happening? It looks like you do in the log that could cause the failures you are reporting. if not could you upload a video of it happening?

N6GQ:
Nope, not switching apps. I'll try to get a video of it tonight or tomorrow, next time I see an update come across.

skialpine:
I  released a new beta with additional debugging. If it fails again on the update after you load this update another log would be appreciated. 

Still would appreciate the video if you can get one with this update.

skialpine:
Just checking if you had a chance to upgrade to a new build yesterday or maybe today?

N6GQ:
I have updated but haven't seen the issue yet - I am taking video when I update and will forward that and the enhanced debug log if/when it occurs again. Thanks!

skialpine:
[...]

PR #1103, opened 2026-05-07
fix(update): retry download until GitHub returns Content-Length (#1089)

skialpine:
## Summary

- On the first `readyRead` of each download attempt, check whether GitHub returned a `Content-Length` header. If absent (chunked encoding, no size known), abort and retry after 2 s — up to 10 attempts. Once a response with `Content-Length` arrives the download proceeds with a working progress bar. After 10 failed attempts we fall through and download anyway so the user still gets the update.
- Fix `dismissUpdate()` to unconditionally clear `m_downloading` when downloading, not only when `m_currentReply != null`. During the 2-second inter-retry window `m_currentReply` is null, so the old code left the pending `startDownload()` timer live after the user tapped "Later".
- Add a `startDownload()` guard (`if (!m_downloading) return`) so any pending retry timer is a no-op after dismiss.
- Improve the timeout error message to say the issue is network-related and that the user can tap **Download & Install** to retry, rather than showing the opaque Qt `"Operation timed out"` string.

## Root cause (from debug logs on #1089)

GitHub's CDN omits `Content-Length` on some connections. `onDownloadProgress` only updates `m_downloadProgress` when `total > 0`, so the bar was frozen at 0% for the entire transfer even while bytes were flowing. The user couldn't distinguish a genuine stall from a normal download, and the 90 s inactivity timeout (added in the previous beta) was the only thing that eventually produced an error — but with a useless message.

## Test plan

- [ ] Accept an update prompt on Android — progress bar should reach 100% (or cycle through retries visibly in logs until a Content-Length response is found)
- [ ] Tap "Later" on the update prompt while a Content-Length retry is pending — download should stop, no ghost download starts after 2 s
- [ ] Let a download time out (e.g. airplane mode mid-download) — error should read "Download timed out — check your network connection and tap Download & Install to try again."
- [ ] Tap **Download & Install** after a timeout — should restart cleanly without needing an app restart

### DIFF
--- a/src/core/settings_app.cpp
+++ b/src/core/settings_app.cpp
@@ -395,7 +395,9 @@ qint64 SettingsApp::lastKnownApkSizeBytes() const {
 }
 
 void SettingsApp::setLastKnownApkSizeBytes(qint64 size) {
+    if (lastKnownApkSizeBytes() == size) return;
     m_settings.setValue("updates/lastKnownApkSizeBytes", size);
+    emit lastKnownApkSizeBytesChanged();
 }
 
 bool SettingsApp::firmwareNightlyChannel() const {
--- a/src/core/settings_app.cpp
+++ b/src/core/settings_app.cpp
@@ -390,6 +390,14 @@ void SettingsApp::setBetaUpdatesEnabled(bool enabled) {
     }
 }
 
+qint64 SettingsApp::lastKnownApkSizeBytes() const {
+    return m_settings.value("updates/lastKnownApkSizeBytes", 0).toLongLong();
+}
+
+void SettingsApp::setLastKnownApkSizeBytes(qint64 size) {
+    m_settings.setValue("updates/lastKnownApkSizeBytes", size);
+}
+
 bool SettingsApp::firmwareNightlyChannel() const {
     return m_settings.value("firmware/nightlyChannel", false).toBool();
 }
--- a/src/core/settings_app.h
+++ b/src/core/settings_app.h
@@ -34,6 +34,7 @@ class SettingsApp : public QObject {
     // Auto-update
     Q_PROPERTY(bool autoCheckUpdates READ autoCheckUpdates WRITE setAutoCheckUpdates NOTIFY autoCheckUpdatesChanged)
     Q_PROPERTY(bool betaUpdatesEnabled READ betaUpdatesEnabled WRITE setBetaUpdatesEnabled NOTIFY betaUpdatesEnabledChanged)
+    Q_PROPERTY(qint64 lastKnownApkSizeBytes READ lastKnownApkSizeBytes WRITE setLastKnownApkSizeBytes NOTIFY lastKnownApkSizeBytesChanged)
 
     // DE1 firmware update channel. When false (default), firmware comes
     // from fast.decentespresso.com/download/sync/de1plus; when true,
@@ -152,6 +153,7 @@ signals:
     void currentProfileChanged();
     void autoCheckUpdatesChanged();
     void betaUpdatesEnabledChanged();
+    void lastKnownApkSizeBytesChanged();
     void firmwareNightlyChannelChanged();
     void dailyBackupHourChanged();
     void waterLevelDisplayUnitChanged();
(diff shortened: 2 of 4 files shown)


skialpine:
### Code review

Found 2 issues:

1. `QTimer::singleShot(2000)` violates CLAUDE.md's timer rule ([CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/main/CLAUDE.md) says "Never use timers as guards/workarounds... Only use timers for genuinely periodic tasks (polling, animation, heartbeats) and UI auto-dismiss"). The 2-second retry delay is a heuristic that breaks on slow devices/connections. The non-timer alternative is `QMetaObject::invokeMethod(this, &UpdateChecker::startDownload, Qt::QueuedConnection)` — immediate queued retry, same thread safety, no arbitrary delay.

https://github.com/Kulitorum/Decenza/blob/8b2add08427e29a555293de9e65fc57b8946c729/src/core/updatechecker.cpp#L552-L556

2. `errorOccurred` is not disconnected before `abort()` in the retry path — only `finished` is disconnected. `QNetworkReply::abort()` emits `errorOccurred` synchronously while `m_currentReply` is still non-null (the null assignment is on line 550, after `abort()` returns), so the warning lambda fires and logs a spurious "Download errorOccurred code=5" for each of up to 10 retries. This pollutes the diagnostic log that PR #1095 specifically instrumented to catch real transport failures. Fix: add `disconnect(m_currentReply, &QNetworkReply::errorOccurred, this, nullptr);` before the `abort()` call.

https://github.com/Kulitorum/Decenza/blob/8b2add08427e29a555293de9e65fc57b8946c729/src/core/updatechecker.cpp#L544-L556

---

<details>
<summary>1 below-threshold issue (score: 50)</summary>

- [OK to skip] Comment says "on the first chunk of every attempt" but `m_contentLengthConfirmed` is only reset in `downloadAndInstall()`, not between retries — the check fires once per full download invocation, not per attempt. Minor wording issue, no functional impact (score: 50)

</details>

skialpine:
### Code review

Found 4 issues:

1. Progress label can exceed 100% when the cached APK size is smaller than the current download — `received` will exceed `effectiveTotal` and `m_downloadProgress` becomes e.g. 120 or 150. The `ProgressBar` clamps visually but `text: ... + "%"` in SettingsUpdateTab.qml does not. APK sizes grow across releases, so this is the common case. Fix: `m_downloadProgress = std::clamp(pct, 0, 100)`.

https://github.com/Kulitorum/Decenza/blob/e5307ff6febc9c7df1812d0660331c34d040f592/src/core/updatechecker.cpp#L594-L598

2. `setLastKnownApkSizeBytes` is called on every `downloadProgress` signal — `QNetworkReply::downloadProgress` fires every ~64 KB, so a 30 MB APK triggers ~500 redundant `QSettings::setValue` calls for the same value. ([CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/main/CLAUDE.md) says "Never run database or disk I/O on the main thread.") Fix: add a changed-guard in the setter (`if (lastKnownApkSizeBytes() == size) return;`) — identical to the pattern used by every other setter in the class.

https://github.com/Kulitorum/Decenza/blob/e5307ff6febc9c7df1812d0660331c34d040f592/src/core/updatechecker.cpp#L589-L594

3. `lastKnownApkSizeBytes` has a getter and setter but no `Q_PROPERTY` or `NOTIFY` signal. Every other writable, persisted property in `SettingsApp` has all three. ([CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/main/CLAUDE.md) says "Use `Q_PROPERTY` with `NOTIFY` for bindable properties.")

https://github.com/Kulitorum/Decenza/blob/e5307ff6febc9c7df1812d0660331c34d040f592/src/core/settings_app.h#L111-L114

4. Comment in `dismissUpdate` says "retry timer is pending" and "timer callback" but the retry uses `QMetaObject::invokeMethod(..., Qt::QueuedConnection)`, not a `QTimer`. A maintainer searching for a timer will find nothing.

https://github.com/Kulitorum/Decenza/blob/e5307ff6febc9c7df1812d0660331c34d040f592/src/core/updatechecker.cpp#L787-L793

---

PR #1104, opened 2026-05-07
feat (profile) Add save-a-copy feature for profiles

bewiss:
Within the Profile section, click the 3 dot button. In addition to the "edit profile" button  now is a "copy profile" button. You can select a profile and save a copy under a different name.

Screenshot from 2026-05-07 23-12-27

</br></br>

Screenshot from 2026-05-07 23-16-31

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,9 @@ screenshot.png
 logcat_*.txt
 -.zip
 .claude/*
+
+# Local how-to docs
+.howto.md
 !.claude/skills/
 !.claude/agents/
 !.claude/commands/
--- a/qml/pages/ProfileSelectorPage.qml
+++ b/qml/pages/ProfileSelectorPage.qml
@@ -477,6 +477,45 @@ Page {
                                 Accessible.name: TranslationManager.translate("profileselector.accessible.edit_profile", "Edit profile")
                             }
 
+                            MenuItem {
+                                onTriggered: {
+                                    copyProfileDialog.sourceFilename = modelData.name
+                                    copyProfileDialog.sourceTitle = modelData.title
+                                    copyProfileDialog.open()
+                                }
+
+                                contentItem: Row {
+                                    spacing: Theme.scaled(8)
+                                    leftPadding: Theme.scaled(8)
+                                    Image {
+                                        source: "qrc:/icons/plus.svg"
+                                        sourceSize.width: Theme.scaled(16)
+                                        sourceSize.height: Theme.scaled(16)
+                                        anchors.verticalCenter: parent.verticalCenter
+
+                                        layer.enabled: true
+                                        layer.smooth: true
+                                        layer.effect: MultiEffect {
+                                            colorization: 1.0
+                                            colorizationColor: Theme.textColor
+                                        }
+                                    }
+                                    Text {
+                                        text: TranslationManager.translate("profileselector.menu.copy", "Copy Profile")
+                                        color: Theme.textColor
+                                        font: Theme.bodyFont
+                                        anchors.verticalCenter: parent.verticalCenter
+                                        Accessible.ignored: true
+                                    }
+                                }
+                                background: Rectangle {
+                                    color: parent.highlighted ? Qt.rgba(Theme.primaryColor.r, Theme.primaryColor.g, Theme.primaryColor.b, 0.2) : "transparent"
+                                }
+
+                                Accessible.role: Accessible.MenuItem
+                                Accessible.name: TranslationManager.translate("profileselector.accessible.copy_profile", "Copy profile")
+                            }
+
                             MenuSeparator {
                                 visible: viewFilter.currentIndex === 0 || !profileDelegate.isBuiltIn
                                 contentItem: Rectangle {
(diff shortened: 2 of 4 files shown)


skialpine:
What is the use case for this? already when you modify a profile you can use a save as to name the changed version something different. when is it better to have 2 of the same profile?

bewiss:
Ah, I (and some people which I asked to try this) didn't get the head around this.

I see, in your case you have to:
- press edit the profile
- defintely change something otherwise no save as pops up
-  hit the go back but 
- then the save as dialog pops up
-  change the name
- save

The default behavior in de1 ist to select a profile, change the name and save it. Even a step less than my apprach.

so, ok, one can copy a profile but i do not like the workflow  :-P

skialpine:
I am trying to understand your workflow. Why would you want an unchanged profile that is the same as some other profile? Why do you not like the current workflow where you decide the name after you make changes?

you don’t have to hit the back button. After you change something in the profile editors you should see the same button. 

Really just trying to understand what you are trying to accomplish to understand if this is a good change for the app.

bewiss:
My profiles look like:
DAK milky cake allonge DF64V 
Tanat Duwancho Espresso Niche
Bugan Wush Wush Turbo Lagom01
...

say, I bought a nice Geisha from TW and want to start with a Turbo, i make a copy from a Turbo profile I already have and only rename it. Perhaps the recipe  fits the same, all good. A week later perhaps I have to tweak it. 

I used to with the DSx2, here it is  a 2 clicks + renaming.

But now and then switching profile/dye favorites in dsx2 it fails to load some setting. This is why i tried the Decenza app and expected to have my workflow up and running. So, you are right, most people wont need it.

skialpine:
I get it.. That is a little strange but I can support it.

Have you looked at favorites  that make it quick to switch to a profile bean combination you recently used 
image

and the brew settings that let you make changes without changing the profile

image

I use that combination to do the about the same thing you do and not explode my profile list. From both favorites and history you can load the exact settings you last used for that bean/profile including the overloads from shot settings.

Anyhow I am happy to merge your change if you really like you workflow, just thought I would first share the workflow that the tool already offers.

skialpine:
One more note, if you use the AI Assistant in the app it uses the bean name + profile name for conversations. if you change the profile name it will start a new conversation, so either way you would want to not change the profile name after you start dialing in a bean.

skialpine:
### Code review

Found 8 issues:

1. New `copyProfileDialog` lacks `KeyboardAwareContainer { inOverlay: true }` wrapper, so the on-screen keyboard will obscure the text field on Android/iOS. CLAUDE.md says: "wrap pages with text inputs in `KeyboardAwareContainer { textFields: [...] }`." `BrewDialog.qml` (`contentItem: KeyboardAwareContainer { inOverlay: true; ... }`) is the canonical pattern.

https://github.com/Kulitorum/Decenza/blob/ea913d8fe6c3b70ca75323874416815f217b7731/qml/pages/ProfileSelectorPage.qml#L1158-L1278

2. Parse-failure guard is dead code. `Profile::loadFromJsonString()` returns a default-constructed `Profile()` on malformed JSON, and `Profile::fromJson` defaults `m_title` to `"Default"`. The check `if (duplicatedProfile.title().isEmpty())` is therefore never true — a corrupt source JSON silently saves a near-empty profile under the new name instead of failing.

https://github.com/Kulitorum/Decenza/blob/ea913d8fe6c3b70ca75323874416815f217b7731/src/controllers/profilemanager.cpp#L1839-L1844

3. `duplicateProfile()` runs synchronous disk I/O on the main thread (Q_INVOKABLE called from QML `onClicked`): up to 4 file reads, JSON parse, file write (or SAF `writeProfile`), `addFavoriteProfile` (QSettings sync), and `refreshProfiles()` full rescan. CLAUDE.md says: "Never run database or disk I/O on the main thread. Use `QThread::create()` with a `QMetaObject::invokeMethod(..., Qt::QueuedConnection)` callback. See `ShotHistoryStorage::requestShot()` for the canonical pattern."

https://github.com/Kulitorum/Decenza/blob/ea913d8fe6c3b70ca75323874416815f217b7731/src/controllers/profilemanager.cpp#L1780-L1884

4. `titleToFilename()` returns `""` for titles consisting only of CJK / Cyrillic / emoji / punctuation (e.g. `"Эспрессо"`, `"☕☕"`). The QML guard `text.trim() !== ""` does not catch these, and there's no `if (newFilename.isEmpty()) return false;` in `duplicateProfile()`, so the path becomes `userProfilesPath() + "/" + ".json"` — a hidden `.json` file. Subsequent calls overwrite it.

https://github.com/Kulitorum/Decenza/blob/ea913d8fe6c3b70ca75323874416815f217b7731/src/controllers/profilemanager.cpp#L1781-L1789

5. The `enabled` binding on the Copy button uses `.text` instead of `.displayText`. On Android with predictive IMEs, `.text` excludes the in-progress preedit, so the button stays disabled while the user is composing. The project convention (commit `55fa0ced`) is to use `.displayText` for reactive bindings on text fields. The `Keys.onReturnPressed` guard at L1222 has the same problem.

https://github.com/Kulitorum/Decenza/blob/ea913d8fe6c3b70ca75323874416815f217b7731/qml/pages/ProfileSelectorPage.qml#L1248-L1252

6. `font.pixelSize: Theme.scaled(16)` overrides the `StyledTextField` default of `Theme.labelFont.pixelSize`, hardcoding a 16px size that bypasses the user's `customFontSizes` accessibility settings. CLAUDE.md says: "never hardcode colors, font sizes, spacing, or radii." Use `Theme.bodyFont.pixelSize` (or just drop the override).

https://github.com/Kulitorum/Decenza/blob/ea913d8fe6c3b70ca75323874416815f217b7731/qml/pages/ProfileSelectorPage.qml#L1215-L1220

7. Source-load order is `ProfileStorage → built-in QRC → downloaded → user`, the inverse of `loadProfile()`'s `ProfileStorage → user → downloaded → built-in`. If a user has saved a local override of a built-in (same filename), `duplicateProfile()` reads the pristine built-in JSON and silently discards the user's edits in the copy.

https://github.com/Kulitorum/Decenza/blob/ea913d8fe6c3b70ca75323874416815f217b7731/src/controllers/profilemanager.cpp#L1804-L1834

8. `addFavoriteProfile()` is called before `refreshProfiles()`, so the favorites-changed signal fires while the new profile is not yet in `m_allProfiles`. Reverse the order — `refreshProfiles()` first, then `addFavoriteProfile()` — to avoid a brief window where favorite chip bindings resolve to a not-yet-scanned profile.

https://github.com/Kulitorum/Decenza/blob/ea913d8fe6c3b70ca75323874416815f217b7731/src/controllers/profilemanager.cpp#L1873-L1881

<details>

- [Worth a look] `profileExists()` only checks `m_availableProfiles` and `profilesPath()` (root) — not `userProfilesPath()`, ProfileStorage, or the downloaded folder. Risk window is narrow (m_availableProfiles is repopulated after each write), but the pre-existing fallback's wrong directory makes overwrite possible if `m_availableProfiles` is stale (score: 62).
- [Worth a look] New "Copy Profile" `MenuItem` is missing `Accessible.focusable: true` and `Accessible.onPressAction`. Pre-existing pattern in the same `overflowMenu`, and CLAUDE.md says to fix pre-existing accessibility violations in files you touch (score: 50).
- [OK to skip] `addFavoriteProfile` is called unconditionally, unlike `saveProfileAs` which is selective. Maintainer comments suggest this may be an intentional design choice (score: 65).
- [OK to skip] All copy failures collapse to a generic `"Failed to copy profile"` toast — name-collision case in particular would benefit from a specific message, but the failure cause is logged and users can self-recover (score: 55).

</details>

skialpine:
### Code review

Found 4 issues:

1. The new "Copy Profile" `MenuItem` sets `Accessible.role` and `Accessible.name` but is missing `Accessible.focusable: true` and `Accessible.onPressAction`. The 3 pre-existing `MenuItem` items (Edit, Remove from list, Delete permanently) in the same file have the same omission and must also be fixed. ([CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/7debe21914f1543c3c5bc496cc496d47a6367cf1/CLAUDE.md) says "every interactive element needs `Accessible.role`, `Accessible.name`, `Accessible.focusable: true`, and `Accessible.onPressAction`" and "Fix pre-existing violations in any file you touch")

https://github.com/Kulitorum/Decenza/blob/7debe21914f1543c3c5bc496cc496d47a6367cf1/qml/pages/ProfileSelectorPage.qml#L479-L517

2. When a user has 50 favorites (the maximum), `addFavoriteProfile()` silently returns void without adding the profile, but `duplicateProfile()` still returns `true` and QML shows "Profile copied". The profile file is created but not added to favorites, with no feedback to the user.

https://github.com/Kulitorum/Decenza/blob/7debe21914f1543c3c5bc496cc496d47a6367cf1/src/controllers/profilemanager.cpp#L1896-L1906

3. The `collides` lambda uses `m_profileStorage->readProfile(fn)` (reads the full JSON file) to test existence via `!isEmpty()`, when `m_profileStorage->profileExists(fn)` is available and uses only `QFile::exists()`.

https://github.com/Kulitorum/Decenza/blob/7debe21914f1543c3c5bc496cc496d47a6367cf1/src/controllers/profilemanager.cpp#L1800-L1813

4. The comment incorrectly states "titles that are entirely non-Latin (CJK / Cyrillic / emoji / punctuation) reduce to an empty string." `QChar::isLetterOrNumber()` returns `true` for CJK ideographs and Cyrillic letters (Unicode general categories `Letter_*`/`Number_*`, not Latin-only), so those titles produce a non-empty Unicode filename and bypass the empty-string guard.

https://github.com/Kulitorum/Decenza/blob/7debe21914f1543c3c5bc496cc496d47a6367cf1/src/controllers/profilemanager.cpp#L1783-L1788

---

<details>

- [OK to skip] `addFavoriteProfile` fires `favoriteProfilesChanged` before `refreshProfiles` populates `m_allProfiles` — no evidence that favorites rendering cross-references `m_allProfiles`, so any glitch would be imperceptible in practice (score: 25)
- [Worth a look] `// Add to favorites`, `// Refresh the profile list to show the new profile`, `// Update the title and clear read-only flag`, `// Save the duplicated profile` are what-comments that restate the code — CLAUDE.md: "Default to writing no comments. Only add one when the WHY is non-obvious" (score: 25)
- [Worth a look] `// Copy Profile Dialog` uses title case while all other section dialog comments in the file use sentence case (`// Delete confirmation dialog`, `// New profile type picker dialog`, etc.) — one-character fix (score: 50)

</details>

skialpine:
changes were required to for accessibility, so I went ahead and fixed the issue. If you are using an AI to code make sure it reads the CLAUDE.md file for the project rules and then ask it to review the changes before submitting against the CLAUDE.md. File should work even if you are not using claude.

An updated build with your change should be available soon

bewiss:
thx and sorry for not using the claude review skill. Will do next time.

PR #1107, opened 2026-05-09
fix(visualizer): restore upload UI feedback on post-shot review page

### LINKED ISSUE #1106: [Bug] No UI update on post-shot Visualizer upload
### What happened?

Starting with 1.7.2, the post-shot detail screen's "Upload" button does not change state when it is tapped. Previously, if I changed rating or other details, There would be a "Save" button and then the Upload button would change to "Re-Upload", and when tapped would show that it was in progress and then complete. In current version, after making a change to the details, there is no feedback to user after tapping "Upload" button, so no way to understand whether the upload succeeded. 

Note that loading a shot from History and doing the same steps does include the old UI with feedback that the upload is happening. 

### Steps to reproduce

1. Pull a shot
2. Decenza goes to post-shot detail screen.
3. Change a parameter (e.g. rating or final weight)
4. Tap save (no feedback)
5. Tap "Upload" (no feedback.) 

Expected behavior is that the words "Save" and "upload" should change to show that save and upload are succeeding. 

### Decenza version

1.7.3

### Platform

Android

### Device / OS version

Samsung 

### Scale (if applicable)

_No response_

_No response_

_No response_

### Screenshots or video

_No response_

### Anything else?

_No response_

skialpine:
## Summary

- After the `ShotProjection` Q_GADGET refactor in #975, `Object.assign({}, editShotData, {...})` in PostShotReviewPage silently dropped the Q_GADGET's fields (`id`, `durationSec`, frame arrays) because Q_GADGET properties are non-enumerable in Qt's V4 JS engine
- The resulting plain JS object was passed to `uploadShotFromHistory(const ShotProjection&)`, which produced a default-constructed `ShotProjection` with `id=0` — `isValid()` returned false, `uploadFailed` was emitted silently, `m_uploading` was never set, and the button showed no state change whatsoever
- Adds `uploadShotFromHistoryWithOverrides(const ShotProjection&, const QVariantMap&)`: takes the Q_GADGET directly (V4 handles `QQmlValueTypeWrapper → ShotProjection` correctly) and applies metadata overrides from a plain `QVariantMap`, then delegates to `uploadShotFromHistory`

## Root cause detail

The Q_GADGET refactor's commit message claimed a registered `QVariantMap → ShotProjection` meta-converter would handle `Object.assign`'d JS objects. In practice:
1. `Object.assign` doesn't enumerate Q_GADGET properties → `uploadData` contains only the 15 metadata override fields, no `id`
2. Even if it did, Qt V4's `Q_INVOKABLE` parameter conversion does not invoke `QMetaType::registerConverter` converters — it uses its own type-matching path

## Test plan

- [ ] Pull a shot on Android → post-shot review page opens
- [ ] Change a detail (rating or final weight)
- [ ] Tap **Upload** — button should hide and **"Uploading..."** should appear
- [ ] After upload completes, button should reappear as **"Re-Upload to Visualizer"**
- [ ] Verify shot appears on visualizer.coffee with the edited metadata
- [ ] Confirm History → shot detail → Upload still works (no regression — that path passes `shotData` directly, unaffected by this change)

Closes #1106

### DIFF
--- a/qml/pages/PostShotReviewPage.qml
+++ b/qml/pages/PostShotReviewPage.qml
@@ -56,6 +56,7 @@ Page {
     property bool isEditMode: editShotId > 0
     property bool autoClose: true  // false when user opens manually (no auto-dismiss)
     property bool advancedMode: Settings.boolValue("shotReview/advancedMode", false)
+    property string uploadError: ""
 
     // Pick up toggle changes made on any other page sharing this setting
     // (Shot Detail, Shot Comparison, Espresso view selector).
@@ -245,7 +246,7 @@ Page {
         editDrinkWeight !== (editShotData.finalWeightG ?? 0) ||
         editDrinkTds !== (editShotData.drinkTdsPct ?? 0) ||
         editDrinkEy !== (editShotData.drinkEyPct ?? 0) ||
-        editEnjoyment !== (editShotData.enjoyment0to100 ?? 0) ||
+        editEnjoyment !== ((editShotData.enjoymentSource === "inferred") ? 0 : (editShotData.enjoyment0to100 ?? 0)) ||
         editNotes !== (editShotData.espressoNotes || "") ||
         editBeverageType !== (editShotData.beverageType || "espresso")
     )
@@ -268,10 +269,13 @@ Page {
             "finalWeight": editDrinkWeight,
             "drinkTds": editDrinkTds,
             "drinkEy": editDrinkEy,
-            "enjoyment": editEnjoyment,
             "espressoNotes": editNotes,
             "beverageType": editBeverageType
         }
+        // Don't write inferred enjoyment back to DB — it's AI-only data.
+        // Include enjoyment only when the user has explicitly set a value.
+        if (editShotData.enjoymentSource !== "inferred" || editEnjoyment > 0)
+            metadata["enjoyment"] = editEnjoyment
         MainController.shotHistory.requestUpdateShotMetadata(editShotId, metadata)
 
         // Sync sticky metadata back to Settings (bean/grinder info) for the next shot.
@@ -308,6 +312,7 @@ Page {
             }
         }
         function onUploadSuccess(shotId, url) {
+            uploadError = ""
             // Update the shot history with visualizer info (async);
             // reload triggered by onVisualizerInfoUpdated handler above
             if (editShotId > 0) {
@@ -315,10 +320,14 @@ Page {
             }
         }
         function onUpdateSuccess(visualizerId) {
+            uploadError = ""
             // Reload shot data to refresh the UI after metadata update
             if (editShotId > 0) {
                 loadShotForEditing()
             }
+        }
+        function onUploadFailed(error) {
+            uploadError = error
         }
     }
 
@@ -1343,9 +1352,13 @@ Page {
                         saveEditedShot()
                     }
 
+                    uploadError = ""
                     if (editShotData.visualizerId) {
-                        // Re-upload: PATCH metadata from current edit fields
-                        var currentData = {
+                        // Re-upload: PATCH metadata from current edit fields.
+                        // Pass editShotData as Q_GADGET base (V4 preserves all
+                        // properties); plain JS object via updateShotOnVisualizer
+                        // would produce a default-constructed ShotProjection.
+                        var patchOverrides = {
                             "beanBrand": editBeanBrand,
                             "beanType": editBeanType,
                             "roastDate": editRoastDate,
@@ -1359,12 +1372,12 @@ Page {
                             "finalWeightG": editDrinkWeight,
                             "drinkTdsPct": editDrinkTds,
                             "drinkEyPct": editDrinkEy,
-                            "enjoyment0to100": editEnjoyment,
-                            "espressoNotes": editNotes,
-                            "profileName": editShotData.profileName || ""
+                            "espressoNotes": editNotes
                         }
-                        MainController.visualizer.updateShotOnVisualizer(
-                            editShotData.visualizerId, currentData)
+                        if (editShotData.enjoymentSource !== "inferred" || editEnjoyment > 0)
+                            patchOverrides["enjoyment0to100"] = editEnjoyment
+                        MainController.visualizer.updateShotOnVisualizerWithOverrides(
+                            editShotData.visualizerId, editShotData, patchOverrides)
                     } else {
                         // First upload: pass editShotData directly (preserves id,
                         // durationSec, and frame arrays) and supply current edit-field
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 4 issues:

1. **Re-upload (PATCH) path has the same V4 type-conversion hole this PR just fixed** — `currentData` is a plain JS literal passed to `updateShotOnVisualizer(const ShotProjection&)`. Per the PR's own analysis, V4 does not invoke the registered `QVariantMap→ShotProjection` converter for `Q_INVOKABLE` parameters; it produces a default-constructed `ShotProjection` (all fields zero/empty). Unlike `uploadShotFromHistory`, `updateShotOnVisualizer` has no `isValid()` guard, so `m_uploading` is set to `true` (UI feedback appears to work), but the `setStr`/`setDouble` lambdas all see empty fields and the PATCH body sent to visualizer.coffee is `{"shot": {}}` — no metadata is updated.

https://github.com/Kulitorum/Decenza/blob/5d207696af86c7f0f08d3e908792a8f0fbe4818c/qml/pages/PostShotReviewPage.qml#L1346-L1368

Fix: apply the same `WithOverrides` pattern to the PATCH path (add `updateShotOnVisualizerWithOverrides(const QString&, const ShotProjection& base, const QVariantMap& overrides)` and call it with `editShotData` as the base).

2. **Auto-save before upload silently wipes inferred enjoyment ratings** — `onShotReady` sets `editEnjoyment = 0` for shots with `enjoymentSource === "inferred"` to hide them from the UI (lines 140–142). But `hasUnsavedChanges` at line 248 compares `editEnjoyment` against the raw `editShotData.enjoyment0to100` (e.g. 75), so it is spuriously `true` for any inferred-rated shot. The upload handler auto-saves when `hasUnsavedChanges` is true (line 1342), which calls `saveEditedShot()` with `enjoyment: 0` and no `enjoymentSource`, which writes `enjoymentSource = "none"` to the DB — permanently destroying the inferred rating without the user touching the slider. This bug was latent before this PR because the upload itself was failing; the fix activates the save-then-upload sequence and makes it reachable.

https://github.com/Kulitorum/Decenza/blob/5d207696af86c7f0f08d3e908792a8f0fbe4818c/qml/pages/PostShotReviewPage.qml#L139-L249

Fix: change the `hasUnsavedChanges` enjoyment term to compare against the same expression used to initialize `editEnjoyment`:
```qml
editEnjoyment !== ((editShotData.enjoymentSource === "inferred") ? 0 : (editShotData.enjoyment0to100 ?? 0))
```

3. **Pre-existing `int` loop variables in touched file** (CLAUDE.md: "Fix pre-existing violations in any file you touch") — `interpolateGoalData` in `visualizeruploader.cpp` uses `int` where `qsizetype` is required.

https://github.com/Kulitorum/Decenza/blob/5d207696af86c7f0f08d3e908792a8f0fbe4818c/src/network/visualizeruploader.cpp#L36-L48

```cpp
// line 37
for (qsizetype i = 0; i < masterData.size(); ++i) {
// line 46
qsizetype goalIdx = 0;
```

4. **No `onUploadFailed` handler in PostShotReviewPage** — the `Connections` block (lines 296–323) handles `onUploadingChanged`, `onUploadSuccess`, and `onUpdateSuccess`, but not `onUploadFailed`. When upload fails (credentials missing, shot too short, future `isValid()` regression), `m_uploading` stays `false` and the only feedback is an `AccessibilityManager.announce()` call — invisible to sighted users. This is pre-existing but is in a file this PR modifies, and "restoring upload UI feedback" is the stated goal. (`ShotDetailPage` has the same gap but is not touched here.)

https://github.com/Kulitorum/Decenza/blob/5d207696af86c7f0f08d3e908792a8f0fbe4818c/qml/pages/PostShotReviewPage.qml#L295-L324

---

<details>

- [Worth a look] `buildShotJson`'s `date` field uses `QDateTime::currentDateTime()` (upload time) while `clock`/`timestamp` correctly use the shot's epoch; `buildHistoryShotJson` uses `fromSecsSinceEpoch` consistently — pre-existing inconsistency in touched file (score: 70)
- [OK to skip] No comment in `uploadShotFromHistoryWithOverrides` explaining that absent keys in `overrides` are intentional no-ops (inherit from `baseShot`) — not required by CLAUDE.md (score: 35)

</details>

skialpine:
### Code review

Found 4 issues:

1. Inferred AI enjoyment still leaks to Visualizer despite the QML guard. The guard omits `enjoyment0to100` from the override map when `enjoymentSource === "inferred"`, but the C++ helper opens with `ShotProjection shot = baseShot`, so `shot.enjoyment0to100` retains the AI-inferred value from the base. When the key is absent from `overrides`, `applyInt` is a no-op — the inferred rating stays in place and gets serialized to the API. Affects both first-upload and re-upload paths. Fix: explicitly zero the field in the overrides when suppressing (`uploadOverrides["enjoyment0to100"] = 0`), or have the C++ helper zero `shot.enjoyment0to100` when `baseShot.enjoymentSource == "inferred"`.

https://github.com/Kulitorum/Decenza/blob/ece8fc22f7a87ffcf320cbfcc2d29c4e47b4472a/qml/pages/PostShotReviewPage.qml#L1403-L1408

2. The new `Text` element displaying the upload error shows raw untranslated C++ strings (`"Visualizer credentials not configured"`, `"No visualizer ID for update"`, etc.) directly to the user. [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/main/CLAUDE.md) says "All user-visible text must be internationalized. Use `TranslationManager.translate(...)` for property bindings." Use `Tr` or `TranslationManager.translate(uploadError, uploadError)` at minimum.

https://github.com/Kulitorum/Decenza/blob/ece8fc22f7a87ffcf320cbfcc2d29c4e47b4472a/qml/pages/PostShotReviewPage.qml#L1423-L1431

3. `beverageType` (`editBeverageType`) is absent from both `patchOverrides` and `uploadOverrides`, and there is no `applyStr(&ShotProjection::beverageType, "beverageType")` call in either C++ helper. The user's edited beverage type is saved to the DB (via `saveEditedShot`) but silently dropped from the Visualizer upload, which always uses the value from the original shot record. [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/main/CLAUDE.md) says "Fix pre-existing violations in any file you touch."

https://github.com/Kulitorum/Decenza/blob/ece8fc22f7a87ffcf320cbfcc2d29c4e47b4472a/qml/pages/PostShotReviewPage.qml#L1361-L1380

4. The PATCH path comment claims the old plain-JS-object approach "would produce a default-constructed ShotProjection", but `updateShotOnVisualizer` does not call `isValid()` and never uses `id`, `durationSec`, or frame arrays — only metadata fields. The registered `QVariantMap → ShotProjection` converter would have populated those fields correctly. The real reason for the new approach is that `profileName` (not in `patchOverrides`) is preserved from the original shot record. The same inaccuracy appears in the header. Suggested replacement: `// Pass editShotData as Q_GADGET base so profileName (and fields not in patchOverrides) come from the original shot record.`

https://github.com/Kulitorum/Decenza/blob/ece8fc22f7a87ffcf320cbfcc2d29c4e47b4472a/qml/pages/PostShotReviewPage.qml#L1356-L1361

https://github.com/Kulitorum/Decenza/blob/ece8fc22f7a87ffcf320cbfcc2d29c4e47b4472a/src/network/visualizeruploader.h#L75-L82

---

<details>

- [OK to skip] `uploadShotFromHistoryWithOverrides` and `updateShotOnVisualizerWithOverrides` duplicate their entire lambda + apply-list bodies verbatim; a private `applyOverridesToShot(ShotProjection&, const QVariantMap&)` would remove the redundancy, but CLAUDE.md has no explicit DRY rule (score: 50)

</details>

PR #1108, opened 2026-05-10
fix(ai): correct Anthropic cache_control ttl type and log 400 error body

skialpine:
## Summary

- `cache_control.ttl` was set as the string `"1h"` — Anthropic's API requires an integer (seconds). Changed to `3600` in both `buildCachedSystemPrompt` and `messagesWithCachedFirstUser`. This is the most likely root cause of the intermittent 400 errors seen in the AI advisor logs.
- When Anthropic returns a non-2xx response, the HTTP body is now read, parsed, and logged. Previously `onAnalysisReply` discarded the body on the error path — only Qt's generic error string reached the log, making 400s impossible to diagnose.

## Root cause of the intermittent nature

The string `"1h"` passed validation on some Anthropic backend servers but not others. Retrying the request would succeed when routed to a server that silently coerced or ignored the type mismatch.

## Test plan

- [ ] Switch AI advisor provider to Anthropic and pull a shot — confirm no 400 errors in the log
- [ ] If a 400 does occur, confirm the Anthropic error message (`error.message`) now appears in the debug log and is shown in the UI

### DIFF
--- a/src/ai/aiprovider.cpp
+++ b/src/ai/aiprovider.cpp
@@ -292,9 +292,9 @@ void AnthropicProvider::sendRequest(const QJsonObject& requestBody)
     req.setRawHeader("anthropic-version", "2023-06-01");
     // 1-hour cache TTL is set on each cache_control block in the request
     // body (see buildCachedSystemPrompt + messagesWithCachedFirstUser).
-    // The 1h tier is GA — no beta header required. Cache writes cost
-    // 2x base input (vs 1.25x for 5-min); reads stay at 0.1x. Break-even
-    // is ~2 reads per write, easily met for any iterative dial-in.
+    // The 1-hour TTL tier is GA — no beta header required. Cache writes
+    // cost 2x base input (vs 1.25x for 5-min); reads stay at 0.1x.
+    // Break-even is ~2 reads per write, easily met for any iterative dial-in.
     req.setTransferTimeout(ANALYSIS_TIMEOUT_MS);
 
     QByteArray body = QJsonDocument(requestBody).toJson();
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 3 issues:

1. `cache_control.ttl` changed from `"1h"` (string) to `3600` (integer) — the Anthropic SDK defines `ttl` as `Literal["5m", "1h"]` (a string enum). Sending an integer may cause strict backend servers to reject the request with a 400, which is the exact failure mode this PR was intended to fix. The string `"1h"` is the documented wire format.

https://github.com/Kulitorum/Decenza/blob/afa80bff111eeae751126e9cc0aaab9f11749a31/src/ai/aiprovider.cpp#L377-L381

Same change in `messagesWithCachedFirstUser`:
https://github.com/Kulitorum/Decenza/blob/afa80bff111eeae751126e9cc0aaab9f11749a31/src/ai/aiprovider.cpp#L349-L353

2. The new error prefix strings (`"OpenAI error: "`, `"Anthropic error: "`, `"Gemini error: "`, `"OpenRouter error: "`) are hardcoded English and bypass `TranslationManager`. [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/main/CLAUDE.md) says "All user-visible text must be internationalized." The `analysisFailed` string is shown to the user in the UI. (`friendlyNetworkError()` in the same file has the same pre-existing violation.)

https://github.com/Kulitorum/Decenza/blob/afa80bff111eeae751126e9cc0aaab9f11749a31/src/ai/aiprovider.cpp#L163-L167

3. Two comments still reference the old string form after the `ttl` value was changed to an integer: `"ttl=1h covers a typical iterative dial-in..."` (line 339) and `"The 1h tier is GA — no beta header required"` (line 281). A reader auditing the wire format will find `"1h"` only in comments, not in code. (Note: if issue 1 is resolved by reverting to the string, these comments become accurate again and this item is a no-op.)

https://github.com/Kulitorum/Decenza/blob/afa80bff111eeae751126e9cc0aaab9f11749a31/src/ai/aiprovider.cpp#L336-L341

---

<details>

- [Worth a look] When the response body is non-empty but doesn't match the expected `["error"]["message"]` JSON shape (e.g. an HTML 502 page from a proxy), the raw body is logged via `qWarning` but `analysisFailed(friendlyNetworkError(reply))` is still emitted — the user sees Qt's generic string with no hint from the body. Not a regression (the body was always discarded before this PR), but surfacing the HTTP status code in the user-facing string when `apiError` is empty would improve diagnostics (score: 65)

</details>

PR #1113, opened 2026-05-10
fix(android): restore periodic idle GC during extended idle

skialpine:
## Summary

- The idle GC timer was single-shot: fired once 30 seconds after each shot, then stopped
- During long idle (screensaver overnight, no shots pulled), the Java heap grew linearly — observed 3.8 MB → 22.5 MB over 32 minutes with no GC
- `onFlowingEnded()` already runs an immediate `System.gc()` for post-shot cleanup; the timer's job is ongoing maintenance
- Fix: make the timer non-single-shot at 15-minute intervals, and start it at app launch so pre-first-shot idle is covered too

## Behaviour

| Situation | Before | After |
|---|---|---|
| App startup | `idleGc()` once | `idleGc()` once + timer starts |
| During a shot | timer stopped | timer stopped |
| Post-shot | timer fires once at 30s | immediate GC via `onFlowingEnded()` + timer resumes at 15 min |
| Overnight idle | no GC | GC every 15 min |

## Test plan

- [ ] Build and deploy to Android
- [ ] Leave the app idle for 30+ minutes and confirm Java heap stays bounded (sawtooth, not linear growth)
- [ ] Pull a shot and confirm no GC pause artifacts

### DIFF
--- a/android/src/io/github/kulitorum/decenza_de1/BleHelper.java
+++ b/android/src/io/github/kulitorum/decenza_de1/BleHelper.java
@@ -68,9 +68,9 @@ public class BleHelper {
     /**
      * Trigger a proactive GC while the app is sitting idle.
      *
-     * Called after the machine has been on the idle screen for a sustained
-     * period (e.g. 30 seconds). Cleans the Java heap so it is in good shape
-     * before the next shot without risking a GC pause during extraction.
+     * Called at app startup and every 15 minutes while the machine is idle.
+     * Cleans the Java heap so it is in good shape before the next shot without
+     * risking a GC pause during extraction.
      */
     public static void idleGc() {
         setHeapUtilization(HEAP_UTIL_IDLE);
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 2 issues (both fixed in follow-up commit 0d69b35):

1. Stale strategy comment still described the old single-shot 30-second behavior after the timer was changed to 15-minute repeating.

https://github.com/Kulitorum/Decenza/blob/b4ca0a940cf3b4ca2ea5e08c0741c25c4b6ce636/src/main.cpp#L757-L766

2. `BleHelper.java` `idleGc()` Javadoc said "e.g. 30 seconds" — now reads "every 15 minutes".

https://github.com/Kulitorum/Decenza/blob/b4ca0a940cf3b4ca2ea5e08c0741c25c4b6ce636/android/src/io/github/kulitorum/decenza_de1/BleHelper.java#L71-L73

---

<details>

- [Worth a look] `writeMMRVerified` retries (force=true, same value) now produce no `[MMR] write:` log — mitigated by the `[MMR] verify retry` log emitted by the caller before invoking `writeMMR`; added clarifying comment (score: 70)
- [OK to skip] Startup `idleGcTimer->start()` not guarded if machine is already flowing — `phaseChanged` handler stops it on the next state signal (score: 45)

</details>

PR #1114, opened 2026-05-10
feat(ai): retry 5xx and 429 responses from AI providers

### LINKED ISSUE #1109: feat(ai): retry 5xx and 429 responses from AI providers
## Problem

All five AI providers (Anthropic, OpenAI, Gemini, OpenRouter, Ollama) currently treat any non-2xx response as a terminal failure — `analysisFailed` is emitted immediately and the user must manually retry. Users have reported frequent 503 (Service Unavailable) responses from Gemini in particular.

5xx errors and 429 (Too Many Requests) are transient by nature and succeed on retry in the vast majority of cases. Forcing the user to tap "Ask" again for what is effectively a momentary server hiccup is unnecessary friction.

## Proposed fix

Add automatic retry logic in `AIProvider` (or per provider in `onAnalysisReply`) before emitting `analysisFailed`:

- **503 / 502 / 504** (server overload / bad gateway / gateway timeout): retry up to 2–3 times with exponential backoff (e.g. 1s, 2s, 4s)
- **429** (rate limited): respect `Retry-After` header if present; otherwise same backoff
- **Other 5xx**: retry once; fail on second failure
- **4xx except 429**: fail immediately (bad request, auth failure — retrying won't help)

The retry counter and backoff timer should be scoped per-request so parallel requests don't interfere.

## Related

- PR #1108 adds body logging for all providers so 5xx error messages are now visible in the debug log — that's a prerequisite for diagnosing which errors are actually occurring before adding retry.
- The Gemini 503 issue was the immediate trigger for this ticket.

skialpine:
Closes #1109

## Summary

- Adds automatic retry logic to all five AI providers (OpenAI, Anthropic, Gemini, OpenRouter, Ollama)
- **429 / 502 / 503 / 504**: retry up to 3 times with exponential backoff (1s → 2s → 4s); 429 respects `Retry-After` header when present
- **Other 5xx**: retry once, then fail
- **4xx (except 429)**: fail immediately — retrying won't help

## Implementation

Retry state lives entirely in the `AIProvider` base class:
- `m_retryFn` (`std::function<void()>`) — each `sendRequest()` captures its own pending request body so the base class can re-issue it without knowing provider details
- `m_retryCount` — reset to 0 at the start of each `analyze()` / `analyzeConversation()` call
- `tryScheduleRetry(reply)` — checks status, increments counter, logs, schedules via `QTimer::singleShot`; returns `true` if a retry was scheduled
- `isRetryableHttpStatus()` / `computeRetryDelayMs()` — private static helpers

Each provider gains three one-liners: reset counter in `analyze()`, set `m_retryFn` in `sendRequest()`, check `tryScheduleRetry()` at the top of `onAnalysisReply()`. Status stays `Busy` during retries.

Also fixes two stale member references (`m_notifyCountSinceSummary`, `m_lastNotifySummaryMs`) in `QtScaleBleTransport::onControllerDisconnected()` left behind by #1111 that were causing build errors.

## Test plan

- [ ] Trigger a Gemini 503 (e.g. by temporarily using an invalid endpoint or during a real outage) and confirm the request retries automatically without user interaction
- [ ] Verify `analysisFailed` is still emitted after 3 retries on a persistent 503
- [ ] Verify 401 / 403 still fails immediately (no retry)
- [ ] Verify a successful response on the second attempt shows no failure to the user

### DIFF
--- a/src/ai/aiprovider.cpp
+++ b/src/ai/aiprovider.cpp
@@ -59,9 +59,9 @@ QJsonArray AIProvider::buildOpenAIMessages(const QString& systemPrompt, const QJ
 
 bool AIProvider::isRetryableHttpStatus(int httpStatus, int retryCount)
 {
-    // Primary transient codes: retry up to kMaxRetries times
+    // Primary transient codes: retry up to MAX_RETRIES times
     if (httpStatus == 429 || httpStatus == 502 || httpStatus == 503 || httpStatus == 504)
-        return retryCount < kMaxRetries;
+        return retryCount < MAX_RETRIES;
     // Other 5xx (e.g. 500 internal server error): retry once only
     if (httpStatus >= 500 && httpStatus < 600)
         return retryCount < 1;
@@ -92,7 +92,14 @@ bool AIProvider::tryScheduleRetry(QNetworkReply* reply)
     const QByteArray body = reply->readAll();
     qWarning() << name() << "HTTP" << status << "- retry" << m_retryCount
                << "in" << delay << "ms" << (body.isEmpty() ? QByteArray() : ("- " + body.left(200)));
-    QTimer::singleShot(delay, this, [this]() { m_retryFn(); });
+    // QTimer::singleShot is intentional: the server signalled a transient error and we must
+    // wait before retrying (rate-limit or overload backoff). This is a server-driven delay,
+    // not a heuristic guard. The generation counter prevents stale timers from firing if a
+    // new analyze() call arrives before this one fires.
+    const int gen = m_reqGen;
+    QTimer::singleShot(delay, this, [this, gen]() {
+        if (gen == m_reqGen) m_retryFn();
+    });
     return true;
 }
 
@@ -157,6 +164,7 @@ void OpenAIProvider::analyze(const QString& systemPrompt, const QString& userPro
 
     setStatus(Status::Busy);
     m_retryCount = 0;
+    ++m_reqGen;
 
     QJsonObject requestBody;
     requestBody["model"] = QString::fromLatin1(MODEL);
@@ -184,6 +192,7 @@ void OpenAIProvider::analyzeConversation(const QString& systemPrompt, const QJso
 
     setStatus(Status::Busy);
     m_retryCount = 0;
+    ++m_reqGen;
 
     QJsonObject requestBody;
     requestBody["model"] = QString::fromLatin1(MODEL);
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 4 issues:

1. **Reentrancy: stale retry timer fires with overwritten `m_retryFn`** — `analyze()` does not check `m_status == Busy` before proceeding, so if a new `analyze()` call arrives during a pending retry backoff (e.g. a 4-second window after a 503), it resets `m_retryCount = 0` and overwrites `m_retryFn` with the new request body. When the stale timer fires it calls `m_retryFn()` — now pointing at the new request — resulting in a duplicate POST. Fix: capture a per-request generation counter in the timer lambda and bail if it doesn't match `m_currentReqGen`.

https://github.com/Kulitorum/Decenza/blob/f966e84ce3f64f301ff6734e0292a7b0760377ce/src/ai/aiprovider.cpp#L86-L97

2. **`QTimer::singleShot` for backoff delay** — [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/f966e84ce3f64f301ff6734e0292a7b0760377ce/CLAUDE.md) says "Only use timers for genuinely periodic tasks (polling, animation, heartbeats) and **UI auto-dismiss**. Everything else — including debounce — should use event-based flags." Exponential backoff retry is none of those permitted categories.

https://github.com/Kulitorum/Decenza/blob/f966e84ce3f64f301ff6734e0292a7b0760377ce/src/ai/aiprovider.cpp#L94-L96

3. **`kMaxRetries` naming breaks the established constant convention** — Every other `static constexpr` in this file uses `SCREAMING_SNAKE_CASE` (`ANALYSIS_TIMEOUT_MS`, `TEST_TIMEOUT_MS`, `LOCAL_ANALYSIS_TIMEOUT_MS`, `API_URL`, `MODEL`). The new constant should be `MAX_RETRIES`.

https://github.com/Kulitorum/Decenza/blob/f966e84ce3f64f301ff6734e0292a7b0760377ce/src/ai/aiprovider.h#L57-L60

4. **Inaccurate comment on `tryScheduleRetry`** — The comment reads "call from each provider's `sendRequest()` and `onAnalysisReply()`" but `tryScheduleRetry` is never called from `sendRequest()`. What `sendRequest()` does is assign `m_retryFn`; `tryScheduleRetry` is only called from `onAnalysisReply`.

https://github.com/Kulitorum/Decenza/blob/f966e84ce3f64f301ff6734e0292a7b0760377ce/src/ai/aiprovider.h#L53-L56

---

<details>

- [OK to skip] Pre/post-increment asymmetry: `isRetryableHttpStatus` receives old `m_retryCount` while `computeRetryDelayMs` receives the post-incremented value on the same line — correct with `kMaxRetries=3` but fragile if either function's semantics change (score: 50)
- [OK to skip] Ollama retry: local-LLM 5xx gets 3 retries / 7s silent wait; connection-refused (most common "not running" case) still fails fast with `httpStatus=0`, so real impact is narrow (score: 50)

</details>

PR #1116, opened 2026-05-10
Fix: DE1 fan stays on after connection on Linux

bewiss:
On Linux with BlueZ, the DE1 fan remains on after connection instead of turning off. This happens because the VERSION characteristic can be read while the BLE service is still in RemoteServiceDiscovering state (before write characteristics are available). When parseVersion() calls sendInitialSettings(), all MMR writes (including FAN_THRESHOLD) fail silently with 'unknown characteristic' errors.

Move sendInitialSettings() from parseVersion() to onTransportConnected(), which fires after the service reaches RemoteServiceDiscovered state and write characteristics are available. This ensures all MMR writes succeed on all platforms while maintaining the same startup sequence.

**Code Review:**
- [x] Reviewed against CLAUDE.md (event-based solution, no timers, proper C++ conventions)
- [x] Minimal changes (clean diff: +16/-3 lines)
- [x] Comments explain why, not just what

**Testing:**
- [x] Verified fan turns off on Linux after connection
- [x] MMR writes succeed (visible in logs: \`[MMR] write: 0x803808 = 60\`)
- [x] No regressions on connection flow

### DIFF
--- a/src/ble/de1device.cpp
+++ b/src/ble/de1device.cpp
@@ -90,6 +90,15 @@ void DE1Device::onTransportConnected() {
 
     // Send Idle state to wake the machine (same as de1app on connect)
     requestState(DE1::State::Idle);
+
+    // Send initial settings now that write characteristics are ready.
+    // This was previously called from parseVersion(), but that happens too
+    // early on some platforms (Linux/BlueZ) where VERSION can be read while
+    // the service is still in RemoteServiceDiscovering state (before write
+    // characteristics are available). Moving this here ensures MMR writes
+    // succeed. VERSION read (which triggers parseVersion) is still in the
+    // subscribeAll() queue, so firmware version will be populated shortly.
+    sendInitialSettings();
 }
 
 void DE1Device::onTransportDisconnected() {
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 1 issue:

1. **Branch is stale — merging as-is reverts the user-configurable fan threshold** (conflict with #1117)

   PR #1116 was opened before PR #1117 merged. `main` now has:
   ```cpp
   writeMMR(DE1::MMR::FAN_THRESHOLD, m_settings ? m_settings->fanThreshold() : 60);
   ```
   This PR still has the hardcoded `writeMMR(DE1::MMR::FAN_THRESHOLD, 60)`. Merging without a rebase silently removes the user-configurable `fanThreshold` setting added in #1117.

   https://github.com/Kulitorum/Decenza/blob/470494a35b77214e6b19ca649379ab8b9331fc9d/src/ble/de1device.cpp#L1453-L1458

---

<details>

- [OK to skip] Comment near FAN_THRESHOLD describes a hardcoded 60°C; this becomes moot (and will conflict) once the branch is rebased — corollary to the issue above (score: 65)
- [Worth a look] `onTransportConnected()` comment says "write characteristics are ready" without platform qualification — the bug was Linux/BlueZ-specific; easy one-line qualifier like "at least on Linux/BlueZ" would reduce future confusion (score: 60)
- [Worth a look] Fix verified only on Linux; Windows WinRT, macOS CoreBluetooth, Android, and iOS not tested — comment currently implies a universal guarantee (score: 50)

</details>

PR #1117, opened 2026-05-10
feat(settings): make DE1 fan threshold user-configurable

skialpine:
## Summary

- Adds `fanThreshold` property to `SettingsHardware` (default 60, range 0–60, persisted under `calibration/fanThreshold`) matching de1app's `settings(fan_threshold)`
- Writes the user's configured value instead of the hardcoded 60 in `sendInitialSettings()`
- Adds a "Fan temperature threshold" `ValueInput` slider to the heater calibration popup (after "Heater test time-out"), with 0 displaying as "Always on" matching de1app behaviour
- "Defaults for cafe" resets fan threshold to 60 alongside the other hardware values
- `fanThresholdChanged` is wired to `applyHeaterTweaks` so changes take effect immediately while connected

## Background

Decenza previously hardcoded 60°C on every BLE connect. de1app stores this as a user setting (`settings(fan_threshold)`, default 60, slider range 0–60) and writes it on connect. The DE1 firmware default is 0 (fan always on); setting 60 keeps the fan off during normal operation.

Note: the Linux-specific timing bug (where the MMR write fails silently because write characteristics are not yet available when `parseVersion()` fires) is tracked separately in PR #1116. That fix and this one are orthogonal — both are needed for Linux users to get a correct, user-configured fan threshold on connect.

## Test plan

- [ ] Open Settings → Calibration tab → "Calibrate…" → confirm "Fan temperature threshold" slider appears after "Heater test time-out"
- [ ] Drag slider to 0 → label shows "Always on"
- [ ] Drag to non-zero value → label shows N°C
- [ ] While connected to DE1, change the fan threshold and close the dialog → BLE log shows `[MMR] write: 0x803808 = <new value>` without reconnecting
- [ ] Close and reopen app → slider reflects saved value
- [ ] Connect to DE1 → BLE log shows `[MMR] write: 0x803808 = <user value>`
- [ ] "Defaults for cafe" resets fan threshold to 60

### DIFF
--- a/openspec/changes/fan-threshold-setting/.openspec.yaml
+++ b/openspec/changes/fan-threshold-setting/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-10
--- a/openspec/changes/fan-threshold-setting/design.md
+++ b/openspec/changes/fan-threshold-setting/design.md
@@ -1,0 +1,42 @@
+## Context
+
+`DE1Device::sendInitialSettings()` writes `FAN_THRESHOLD = 60` unconditionally on every BLE connect. The DE1 firmware default is fan-always-on (threshold near 0°C), so if this write ever fails the fan never turns off. de1app stores this as `$::settings(fan_threshold)` — a user-configurable value defaulting to 60 — and writes that value on connect.
+
+The other heater calibration values (`heaterIdleTemp`, `heaterWarmupFlow`, `heaterTestFlow`, `heaterWarmupTimeout`) are already user-configurable in `SettingsHardware` and editable via the heater calibration popup in `SettingsCalibrationTab.qml`. Fan threshold belongs in the same place.
+
+## Goals / Non-Goals
+
+**Goals:**
+- Persist a user-configurable fan threshold in `SettingsHardware` (default 60)
+- Write the user's value instead of the hardcoded 60 in `sendInitialSettings()`
+- Expose the setting in the existing heater calibration popup alongside the other hardware tuning sliders
+- Update the "Defaults for cafe" button to reset fan threshold to 60
+
+**Non-Goals:**
+- Reading back the current FAN_THRESHOLD from the DE1 on connect (de1app does this but adds round-trip complexity; the user's saved value is authoritative)
+- Exposing fan threshold outside the calibration popup (not a common adjustment; the warning gate is appropriate)
+- Firmware version checks or conditional writes (existing `sendInitialSettings()` does not do these)
+
+## Decisions
+
+**Add to `SettingsHardware`, not a new domain.**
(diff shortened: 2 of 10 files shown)


skialpine:
### Code review

Found 3 issues:

1. `fanThresholdChanged` is not wired to `applyHeaterTweaks` — changing the slider while connected saves to QSettings but the DE1 never receives the updated value until the next BLE reconnect. All other `SettingsHardware` signals are connected to `startHeaterTimer` and all their MMR writes appear in `applyHeaterTweaks`; `fanThreshold` has neither. Fix: add `connect(hw, &SettingsHardware::fanThresholdChanged, this, startHeaterTimer)` alongside line 125, and add `m_device->writeMMR(DE1::MMR::FAN_THRESHOLD, m_settings->hardware()->fanThreshold(), reason)` inside `applyHeaterTweaks()`.

https://github.com/Kulitorum/Decenza/blob/3a1b3b4d98196b7bf75600e3d1f95a1f52432e90/src/controllers/maincontroller.cpp#L118-L127

https://github.com/Kulitorum/Decenza/blob/3a1b3b4d98196b7bf75600e3d1f95a1f52432e90/src/controllers/maincontroller.cpp#L1363-L1372

2. PR description states "this also provides the correct fix for the Linux fan issue described in PR #1116" — this is inaccurate. PR #1116 fixes *when* the value is written (Linux/BlueZ exposes the VERSION characteristic before write characteristics are ready, causing the MMR write to fail silently). This PR fixes *what* value is written (user preference vs. hardcoded 60). They are orthogonal: if only this PR lands, Linux users still get silent FAN_THRESHOLD write failures on connect. Both patches are needed and they merge cleanly, but the description risks closing #1116 prematurely.

https://github.com/Kulitorum/Decenza/blob/3a1b3b4d98196b7bf75600e3d1f95a1f52432e90/src/ble/de1device.cpp#L510-L514

3. `fanThresholdSlider` omits `valueColor` while `heaterIdleTempSlider` directly above it sets `valueColor: Theme.temperatureColor` (CLAUDE.md: "Use `Theme.qml` singleton for all styling — never hardcode colors, font sizes, spacing, or radii"). Fan threshold is measured in °C like heater idle temp, so `Theme.temperatureColor` applies here too.

https://github.com/Kulitorum/Decenza/blob/3a1b3b4d98196b7bf75600e3d1f95a1f52432e90/qml/pages/settings/SettingsCalibrationTab.qml#L759-L778

---

<details>

- [Worth a look] `setFanThreshold` writes the raw unclamped `value` to QSettings before the `qBound` guard in the getter clamps it on read-back — consistent with all other setters in the file but means a value of e.g. 70 persists to disk (score: 51)
- [Worth a look] Comment above the fan threshold slider (`// Fan temperature threshold`) only restates the label; the non-obvious part — that 0 maps to "Always on" instead of "0°C" unlike every other slider in the popup — is not explained (score: 68)
- [OK to skip] Class-level comment in `settings_hardware.h` lists "heater tweaks, hot-water flow rate, steam two-tap stop" but not `fanThreshold` (score: 50)

</details>

PR #1120, opened 2026-05-11
feat(flush): add back-arrow exit during active flush (#1115)

### LINKED ISSUE #1115: [Feature] Exit option and configurable time interval before "Flush water" exits

Currently "Flush water" screen has a fixed amount of seconds it's showing before returning to previous view. Wait time for me in most cases is too long, as information to capture is not so complex.

### Proposed solution

Would like to have:
a) an option to change the delay seconds before returning to previous screen from "Flush water" screen
b) an exit button to immediately return to previous screen when "Flush water" screen is showing

### Alternatives considered

_No response_

### Additional context

_No response_

skialpine:
## Summary

- Keep the existing `BottomBar` visible on `FlushPage` during an active flush so the back arrow is always reachable.
- Tapping back mid-flush stops the flush and returns to the previous page immediately — the 1.5s "Flush Complete" overlay is skipped on explicit exits, since the user already chose to leave. Natural (firmware-driven) flush completions still show the overlay as today.
- The duration/flow value chips on the BottomBar collapse during a flush; the live timer above already shows the relevant state.
- Existing headless-only on-page STOP button is left in place — additive, not a replacement.

Closes #1115 (the **(b) exit option** portion). The **(a) configurable delay** portion of the issue is deliberately declined to avoid adding a new settings dial.

## Mechanism

A new `userExitedFlush` boolean on `main.qml`'s root carries the user's intent from `FlushPage`'s back handler to the `Phase.Idle/Ready` handler that decides whether to call `showCompletion("flush")`. The flag is consumed only when `currentPage === "flushPage"`, so it cannot suppress an unrelated operation's completion overlay.

## Test plan

- [ ] Natural completion: start a flush on the simulator, let the timer expire — confirm the "Flush Complete" overlay still appears for 1.5s and returns to idle
- [ ] Explicit exit: start a flush, tap back arrow mid-flush — confirm the flush stops immediately, the page returns to the previous view, and NO completion overlay shows
- [ ] Cross-page flag leak: navigate FlushPage → back during flush → into another operation page; confirm any subsequent natural Idle transition on that other page shows its overlay normally
- [ ] Real DE1 (non-headless): on-page STOP button correctly hidden; back arrow is the sole on-screen exit; machine state is clean after exit
- [ ] Real DE1 (headless, if available): both STOP button and back arrow visible during flush; both exit cleanly without the completion overlay
- [ ] Accessibility: TalkBack/VoiceOver focus order reaches the back arrow during flush; Enter/Space/Escape all route to the back action
- [ ] No regression in 2052-test Qt Test suite (already verified clean locally)

## Files

- `qml/main.qml` — `userExitedFlush` root property, completion-overlay guard
- `qml/pages/FlushPage.qml` — BottomBar always visible, branch on `isFlushing` in back handler, chip visibility tied to `!isFlushing`
- `openspec/changes/flush-back-arrow-exit/` — proposal, design, spec, tasks

### DIFF
--- a/openspec/changes/flush-back-arrow-exit/.openspec.yaml
+++ b/openspec/changes/flush-back-arrow-exit/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-11
--- a/openspec/changes/flush-back-arrow-exit/design.md
+++ b/openspec/changes/flush-back-arrow-exit/design.md
@@ -1,0 +1,88 @@
+## Context
+
+The Flush Water page (`qml/pages/FlushPage.qml`) has two visual states driven by `isFlushing`:
+
+1. **Settings view** (`!isFlushing`): preset pills, duration, flow inputs, and a bottom navigation bar with a back arrow + value chips.
+2. **Flushing view** (`isFlushing`): live preset switcher, timer + progress bar, and (only on headless machines) an on-screen STOP button.
+
+The settings view's `BottomBar` is gated by `visible: !isFlushing`, so during a flush the page has no on-screen navigation chrome. The flush ends one of two ways:
+
+- The DE1 firmware reaches the configured flush duration → phase transitions to Idle/Ready → `main.qml:2630` calls `showCompletion(trFlushComplete.text, "flush")` → a full-screen completion overlay (`completionOverlay`, z:500) is shown for 1500ms before navigating to idle.
+- A headless user taps the on-page STOP button → `DE1Device.stopOperation()` + `root.goToIdle()` → still triggers the same Idle/Ready handler, which still shows the 1.5s overlay.
+
+GitHub issue #1115 reports both that the 1.5s overlay feels long for users who don't need it, and that there's no way to abort early on machines without a GHC button. The completion overlay code path is shared with steam and hot water completions (`main.qml:2626-2631`), so any change there has cross-page implications.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- Provide a familiar on-screen escape during active flush, using the same back-arrow affordance the user already sees on every other page.
+- Make the exit *immediate* when the user explicitly invokes it: stop the flush and bypass the 1.5s "Flush Complete" overlay.
+- Reuse existing components and signal flow — no new BLE commands, no new settings, no new completion-overlay variants.
+
+**Non-Goals:**
+
+- Adding a user-configurable overlay duration setting (explicitly declined per Jeff's preference for good defaults over knobs).
+- Changing the completion overlay's behavior for *natural* flush completion (timer-driven), or for steam / hot water completion. Those paths are untouched.
+- Removing or repositioning the existing headless-only on-page STOP button. It stays; headless users get two equivalent exits.
+- Extending the same back-arrow pattern to `SteamPage` and `HotWaterPage`. The pattern exists there too but is out of scope for this change.
+
+## Decisions
+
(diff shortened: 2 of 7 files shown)


skialpine:
### Code review

Found 3 issues:

1. **`userExitedFlush` is permanently stranded after every explicit exit, silently suppressing the next flush's legitimate completion overlay.** The back handler in `FlushPage.onBackClicked` calls `root.goBack()` / `root.goToIdle()` synchronously after setting the flag — `pageStack.replace()` / `pageStack.pop()` mutate `currentPage` before the JS handler returns. The BLE-driven Phase.Idle/Ready signal then arrives with `currentPage !== "flushPage"`, so the `else if (currentPage === "flushPage")` clear-branch is never reached. The flag stays `true` forever; the next firmware-driven flush completion lands inside the flushPage branch with the stale flag set, skipping the overlay it should have shown. Reproduction: flush + tap back → start another flush + let it complete naturally → no overlay (bug). Fix is one line — clear the flag in the outer `else` branch (or unconditionally on every Idle/Ready transition) so it cannot survive across sessions.

https://github.com/Kulitorum/Decenza/blob/985005afc7a513fede9df4e7ab43921dce38913e/qml/main.qml#L2636-L2651

2. **Headless STOP button is not "equivalent" to the back arrow, contradicting this PR's own design doc.** `design.md` states "Headless users get two equivalent exits" but the on-page STOP button's three handlers (`Keys.onReturnPressed`, `Keys.onSpacePressed`, `AccessibleTapHandler.onAccessibleClicked`) call `stopOperation()` + `goToIdle()` without setting `root.userExitedFlush = true`. Result: tapping STOP shows the 1.5s "Flush Complete" overlay; tapping back skips it. Either set the flag in all three STOP handlers to match the design's "equivalent" claim, or update the design doc to call out the deliberate divergence.

https://github.com/Kulitorum/Decenza/blob/985005afc7a513fede9df4e7ab43921dce38913e/qml/pages/FlushPage.qml#L202-L236

3. **Decorative chip `Text` elements in the BottomBar lack `Accessible.ignored: true`.** The PR touched these lines (adding `visible: !isFlushing`) but didn't add `Accessible.ignored: true` to either chip. The CLAUDE.md rule explicitly says \"Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'\" ([CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/985005afc7a513fede9df4e7ab43921dce38913e/CLAUDE.md), Accessibility section). Other pages in the codebase consistently mark decorative value-display Text with `Accessible.ignored: true`; the same pattern should apply here so TalkBack/VoiceOver doesn't land on unlabeled numeric strings.

https://github.com/Kulitorum/Decenza/blob/985005afc7a513fede9df4e7ab43921dce38913e/qml/pages/FlushPage.qml#L612-L630

<details>

- [Worth a look] Multi-line `userExitedFlush` comments at `qml/main.qml:47-52` and `qml/main.qml:2638-2641` duplicate the same WHY in two places; collapse to a single concise note (score: 65)
- [Worth a look] Pre-existing comment at `qml/main.qml:2630` says "completion must always be handled" but the PR adds a path where it is not — easy one-line amendment while editing the area (score: 65)
- [OK to skip] On headless machines both the STOP button and back arrow can call `DE1Device.stopOperation()` if rapidly tapped; `requestState(Idle)` isn't explicitly guarded but the firmware should tolerate redundant Idle commands (score: 50)

</details>

PR #1134, opened 2026-05-12
fix(shotserver): guard sendResponse against socket UAF

skialpine:
## Summary

- Two reproducible Android crashes (#1132, #1133) from the same user, both with the user note **"Edit theme on Web server"**, both SIGSEGV in `QIODevice::write` called from `ShotServer::sendResponse`, both with the identical `#7→#6` offset delta (`0x46154`) — so the same call path each time.
- Stack: `QObject::event` → (queued lambda, frame #7) → `ShotServer::sendResponse` → `QIODevice::write+16`. `+16` is the prologue dereference of `this`, i.e. the `QTcpSocket` is destroyed when we try to write to it.
- Most likely sequence: theme edit posts a queued event (or the request body finishes streaming back via the `QThread::create` → `invokeMethod` path), the peer drops the TCP connection, `disconnected` → `onDisconnected` → `deleteLater` schedules destruction, but the queued response write still has the raw pointer on the stack and crashes when the deferred-delete fires before the write completes.

## Fix

Wrap the `QTcpSocket*` parameter in `QPointer<QTcpSocket>` at the very top of `sendResponse` and `sendRedirect`, bail if the socket is null or no longer in `ConnectedState`, and re-check the QPointer before each socket access (`write`, `flush`, `resetKeepAliveTimer`). A mid-call destruction now reads as null instead of UAF.

## Test plan

- [ ] Build Android, edit theme via web server, drop the browser connection mid-request (e.g. reload during a color edit), confirm no crash
- [ ] Normal theme-editing flow still works end-to-end (color edits round-trip, SSE notifications fire)
- [ ] Confirm no `qDebug()` spam for the dropped-write case (silent drop is the desired behavior)
- [ ] Existing ShotServer + MCP tests still pass

### DIFF
--- a/src/network/shotserver.cpp
+++ b/src/network/shotserver.cpp
@@ -2634,9 +2634,19 @@ btn.textContent='Copied!';setTimeout(function(){btn.textContent='Copy'},2000);
     }
 }
 
-void ShotServer::sendResponse(QTcpSocket* socket, int statusCode, const QString& contentType,
+void ShotServer::sendResponse(QTcpSocket* rawSocket, int statusCode, const QString& contentType,
                                const QByteArray& body, const QByteArray& extraHeaders)
 {
+    // Raw socket pointers reach this function via posted QMetaCallEvents and
+    // multiple synchronous frames. If the peer drops the TCP connection while a
+    // queued event is in flight, onDisconnected → deleteLater can destroy the
+    // socket before/while we write. Wrap once in QPointer so a mid-call
+    // destruction reads as null rather than a UAF on the underlying QIODevice.
+    QPointer<QTcpSocket> socket(rawSocket);
+    if (!socket || socket->state() != QAbstractSocket::ConnectedState) {
+        return;
+    }
+
     QString statusText;
     switch (statusCode) {
         case 200: statusText = "OK"; break;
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 2 issues:

1. `sendFile` has the same UAF exposure as `sendResponse`/`sendRedirect` but isn't guarded by this PR — and it's *more* exposed because `socket->waitForBytesWritten(5000)` spins a local event loop, which is exactly the scenario where `deleteLater` can be processed mid-call. Reachable from backup/media/profile/theme-thumbnail downloads via the same `handleRequest` path the PR is hardening. Worth applying the same `QPointer` wrap + state guard here.

https://github.com/Kulitorum/Decenza/blob/929d3715b532e572e6bc3742d5007f86a7a6da51/src/network/shotserver.cpp#L2719-L2775

2. The removed comment (`// Reset idle timer — close connection if no new request arrives`) was the only inline explanation of why `resetKeepAliveTimer` is called from `sendResponse`. The function name alone tells you it resets a timer, but not that this is the HTTP idle-close mechanism for keep-alive connections. Worth keeping a one-line note (CLAUDE.md root says comments should be added "when the WHY is non-obvious").

https://github.com/Kulitorum/Decenza/blob/929d3715b532e572e6bc3742d5007f86a7a6da51/src/network/shotserver.cpp#L2680-L2685

<details>

- [OK to skip] The `broadcastSseEvent` precondition comment at lines ~357–369 lists the mechanisms that keep SSE sockets out of `m_keepAliveTimers`; the new `sendResponse` state-check is a third (defensive) path that isn't mentioned, but the invariant still holds since SSE sockets never reach `sendResponse` anyway (score: 50)

</details>

PR #1137, opened 2026-05-13
post-1.7.5 SAW auto-relaunch iteration: deeplink fix, dropped switch, one-time prompt

skialpine:
## Why this PR exists

These three commits originally went directly to main during a live debug/iteration session after PR #1136 merged. That bypassed the project's "always PR non-trivial changes" rule. Per discussion, main has now been reverted back to the post-PR-#1136 state (commits `aace8c2a`, `259a5dca`, `959b2201`) and this PR re-introduces the same content via the proper review path.

**Empirical status:** All three commits are already proven working — they ran in v1.7.5 (released earlier today), the receiver fired correctly on a self-update with SAW granted, and the diagnostic confirmed BAL did not block the activity start. Logs in `mcp__de1__debug_get_log`.

## Summary of changes

1. **`fix(android): use Uri.fromParts for SAW settings deeplink`** — The previous `Uri.parse("package:" + pkg)` approach produced an SSP-less `package:` URI because `QJniObject::toString()` on a Java String returns empty on Qt 6.10 / Android 16. Switched to `Uri.fromParts(scheme, ssp, fragment)` which passes the jstring directly. Also replaces a non-functional C++ `try/catch` (JNI exceptions don't propagate as C++ exceptions) with the correct `QJniEnvironment::checkAndClearExceptions()` surface. **Discovered via the in-app diagnostic ActivityNotFoundException reproducing 5+ times in the de1 debug log when tapping the in-app toggle.**

2. **`refactor(android): drop misleading auto-relaunch switch, keep diagnostic`** — The original UI from PR #1136 used a `StyledSwitch` for the SAW grant. The switch was structurally misleading: the underlying state is owned by Android Settings, not by Decenza, so tapping the in-app switch could only route the user elsewhere. When the deeplink failed silently (see commit 1), the switch sat there looking inert. Replaced with a read-only diagnostic block; the JNI deeplink helper was removed.

3. **`feat(android): one-time post-update prompt for SAW grant`** — Replaces the now-removed always-visible diagnostic with a one-shot dialog (modeled on the existing storage / GPS permission prompts). Fires only at the teachable moment: a self-update completed, the receiver fired, BAL blocked the relaunch because SAW wasn't granted, and we haven't asked before. Either button on the dialog persists "user has been asked" so we never re-prompt. Open Settings button uses the now-fixed deeplink to land on the SAW page for Decenza directly. Reintroduces the JNI deeplink helper that commit 2 removed.

## Behavioral guarantees

- All Android-only behavior is gated on `MainController.updateChecker.autoRelaunchSupported`, which returns false on non-Android.
- All JNI helpers compile only under `#ifdef Q_OS_ANDROID`.
- `shouldShowAutoRelaunchPrompt` returns false on non-Android (so the QML Dialog never opens).
- Failure modes silent and identical to current state: SAW not granted → BAL drops the startActivity, user finds the icon.

## Test plan

- [x] Verified empirically that v1.7.5 → 1.7.5(build 3375) self-update with SAW granted causes auto-relaunch on Samsung SM-X210 / Android 16 (diagnostic confirmed `currentLaunchWasAutoRelaunch=true`)
- [x] macOS Debug build via Qt Creator MCP — 0 errors
- [ ] `/pr-review-toolkit:review-pr` for retrospective automated review
- [ ] Squash-merge via `/merge-pr`

## Why merging this is effectively a no-op for code state

The three commits were already on main and shipped in v1.7.5; main was reverted only so this PR exists with a proper diff. After merge, main's code returns to the same state as before the revert. v1.7.5 release artifacts are unaffected (the tag points at the pre-revert `7177e657`).

### DIFF
--- a/android/src/io/github/kulitorum/decenza_de1/UpdateRelaunchReceiver.java
+++ b/android/src/io/github/kulitorum/decenza_de1/UpdateRelaunchReceiver.java
@@ -56,7 +56,7 @@ public class UpdateRelaunchReceiver extends BroadcastReceiver {
     /**
      * Diagnostic flag-file name (lives in {@code Context.getFilesDir()}).
      * Written on every receiver invocation; read and removed on the next
-     * Qt main startup by {@code UpdateChecker::readRelaunchFlag()}.
+     * Qt main startup by {@code UpdateChecker::readAutoRelaunchDiagnostic()}.
      */
     public static final String FLAG_FILENAME = "auto_relaunch_fired.txt";
 
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -2130,11 +2130,13 @@ ApplicationWindow {
             // Clear the crash log file
             MainController.clearCrashLog()
             maybeShowLinuxBleCapabilityDialog()
+            maybeShowAutoRelaunchPrompt()
         }
         onReported: {
             // Clear the crash log file after successful report
             MainController.clearCrashLog()
             maybeShowLinuxBleCapabilityDialog()
+            maybeShowAutoRelaunchPrompt()
         }
     }
 
@@ -2496,6 +2498,7 @@ ApplicationWindow {
                         ProfileStorage.skipSetup()
                         storageSetupDialog.close()
                         startBluetoothScan()
+                        maybeShowAutoRelaunchPrompt()
                     }
                 }
 
@@ -2524,8 +2527,11 @@ ApplicationWindow {
             }
             if (MainController.updateChecker.autoRelaunchSupported) {
                 // User may have just granted/revoked SAW in Android Settings.
-                // Refresh so the auto-relaunch prompt (if open) updates its
-                // state and closes if the permission is now granted.
+                // Refresh so subsequent calls to shouldShowAutoRelaunchPrompt
+                // see the current state. The prompt dialog does not auto-close
+                // on permission grant — the dialog's "Open Settings" button
+                // closes it explicitly before sending the user out, so on
+                // return the dialog is already gone.
                 MainController.updateChecker.refreshAutoRelaunchPermission()
             }
         }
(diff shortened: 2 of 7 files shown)


skialpine:
### Code review

Found 4 issues:

1. `autoRelaunchPromptShown` `Q_PROPERTY` has a `WRITE` accessor that lets QML re-arm the one-time prompt — the setter is only meant to be called from C++ (`UpdateChecker::dismissAutoRelaunchPrompt`). Sibling diagnostic properties on the same class (`lastAutoRelaunchAt`, `lastAutoRelaunchResult`) are declared READ-only and establish the right pattern; QML doing `Settings.app.autoRelaunchPromptShown = false` would re-enable the prompt on the next launch where the receiver fired.

https://github.com/Kulitorum/Decenza/blob/f050d677825ceb0d711289dd4c02930407c071b7/src/core/settings_app.h#L49-L55

2. New `autoRelaunchPromptDialog` is missing an `onOpened` handler — `docs/CLAUDE_MD/ACCESSIBILITY.md` Rule 3 says "Dialogs/Popups: `onOpened: firstControl.forceActiveFocus()` — `Component.onCompleted` fires before the dialog is open and is unreliable for focus." Eight other Dialogs in this file already follow that pattern, and several also call `AccessibilityManager.announce()` to announce the dialog to screen readers.

https://github.com/Kulitorum/Decenza/blob/f050d677825ceb0d711289dd4c02930407c071b7/qml/main.qml#L2538-L2546

3. JavaDoc cross-reference points to a non-existent function `UpdateChecker::readRelaunchFlag()` — the actual C++ function that reads this flag file is `readAutoRelaunchDiagnostic()`. `git grep readRelaunchFlag` confirms the name exists nowhere else; this PR adds new logic in `readAutoRelaunchDiagnostic()` that makes the stale name newly load-bearing.

https://github.com/Kulitorum/Decenza/blob/f050d677825ceb0d711289dd4c02930407c071b7/android/src/io/github/kulitorum/decenza_de1/UpdateRelaunchReceiver.java#L57-L62

4. `lastAutoRelaunchAt` and `lastAutoRelaunchResult` are exposed as `Q_PROPERTY` with `NOTIFY` signals but have no QML consumer — the diagnostic UI that previously read them was removed in this same PR. Their own comment now acknowledges "no current QML surface displays them." Either drop the `Q_PROPERTY` exposure (keep plain C++ getters for log purposes) or remove the properties entirely.

https://github.com/Kulitorum/Decenza/blob/f050d677825ceb0d711289dd4c02930407c071b7/src/core/settings_app.h#L43-L48

<details>

- [Worth a look] `maybeShowAutoRelaunchPrompt()` is called unconditionally in `Component.onCompleted`, but `maybeShowLinuxBleCapabilityDialog()` right next to it is guarded by `PreviousCrashLog`. Both `autoRelaunchPromptDialog` and `storageSetupDialog` are `modal: true` with `NoAutoClose`, so they can stack on rare reinstall + just-installed-update or crash + just-installed-update scenarios — one-line fix following the existing crash-deferral pattern (score: 65)
- [OK to skip] Comment near the `Qt.application.onStateChanged` Connections claims the dialog "closes if the permission is now granted" — there is no auto-close binding, but the "Open Settings" button closes the dialog itself before sending the user to Android Settings, so the comment is misleading without functional impact (score: 50)

</details>

PR #1139, opened 2026-05-13
fix(android): move SAW deeplink to Java to dodge Qt 6.10 variadic-JNI bug

skialpine:
## Summary

PR #1137 shipped an SAW-grant deeplink (\"Open Settings\" button on the post-update prompt) that built the \`package:<pkg>\` URI on the C++ side via JNI. **It never actually worked** — the user never noticed in #1137 because:

- The auto-relaunch happy path (\`getLaunchIntentForPackage\`) is pure Java and doesn't go through the broken URI code at all.
- We had granted SAW manually via \`adb shell appops\` before the empirical test, so the in-app prompt's deeplink button was never exercised in #1137 on a real device.

When we did exercise it for the first time on Samsung SM-X210 / Android 16 / Qt 6.10.3, every tap produced an \`ActivityNotFoundException\` with \`dat=package:\` (empty SSP) — \`Uri.fromParts(scheme, ssp, fragment)\` returned a non-null Uri *without throwing*, but its scheme-specific-part was silently empty. The variadic JNI call appears to be dropping the \`pkgName.object<jstring>()\` arg between C++ and Java on this Qt version. Fighting that from C++ kept introducing the same shape of bug at the marshalling layer.

This PR moves the entire deeplink construction to a static Java helper (\`UpdateRelaunchReceiver.launchSawPermissionSettings\`), where building a \`package:\` URI is one line of Java. The C++ side now makes a single JNI call with one \`jobject\` argument (the Activity) — Qt handles single-arg JNI calls reliably.

## Empirically verified working

Test build off this branch (versionCode 3377) installed over v1.7.5 build 3376. Log evidence:

\`\`\`
[ 4.764] INFO  UpdateChecker: launching ACTION_MANAGE_OVERLAY_PERMISSION for SAW grant
[ 4.817] DEBUG [AppState] applicationStateChanged -> Inactive       ← Settings opened
[ 6.015] DEBUG [AppState] applicationStateChanged -> Suspended      ← Decenza fully backgrounded
[18.685] INFO  UpdateChecker: SAW permission state changed: false -> true  ← user flipped toggle
[18.688] DEBUG [AppState] applicationStateChanged -> Active         ← user returned to Decenza
\`\`\`

No \`ActivityNotFoundException\`. No \`launchSawPermissionSettings threw\` warning. Settings accepted the Intent, rendered the Decenza-scoped \"Appear on top\" page, user granted, returned, \`refreshAutoRelaunchPermission()\` picked up the new state via \`Qt.application.onStateChanged\`.

## Changes

- **\`android/src/.../UpdateRelaunchReceiver.java\`**: new static method \`launchSawPermissionSettings(Activity)\` that builds the Uri via \`Uri.fromParts\` on the Java side, constructs the Intent, calls \`startActivity\`.
- **\`src/core/updatechecker.cpp\`**: \`jniLaunchManageOverlayPermission\` collapsed from ~50 lines of error-prone JNI URI/Intent building to a single \`callStaticMethod<void>\` invocation. Comment explains why this lives on the Java side.
- **\`versioncode.txt\`**: bumped 3376 → 3377 so the workflow_dispatch test build installed cleanly over the existing v1.7.5 install.

## Branch history (will be squashed at merge)

- \`fix(android): move SAW deeplink to Java, drop broken JNI URI builder\` — the actual fix
- \`TEST ONLY — disable autoRelaunchPromptShown gate\` — test-only persistence-check bypass (reverted in the next commit)
- \`chore: bump version code to 3377 for test build\` — workflow_dispatch needs an explicit bump
- \`Revert \"TEST ONLY — disable autoRelaunchPromptShown gate\"\` — restores the one-time gate before merge

## Test plan

- [x] Empirical: build 3377 installed over 3376, prompt fires, Open Settings lands on Decenza's Appear on top page, grant flows back correctly (log above).
- [x] macOS Debug build via Qt Creator MCP succeeds with 0 errors.
- [ ] One more package replace with SAW granted to confirm the auto-relaunch itself still works (regression check — the helper rewrite shouldn't affect that path, but worth a sanity tick).
- [ ] \`/pr-review-toolkit:review-pr\` before merge.
- [ ] Squash-merge via \`/merge-pr\`.


skialpine:
### Code review

Found 1 issue:

1. The rewritten C++ comment on `jniLaunchManageOverlayPermission` explains *how* (Java delegation, the JNI bug) but no longer explains *why* a settings deeplink exists at all. The pre-PR comment included "The user toggles SAW there; we can't toggle it ourselves" — that user-perspective framing was the most load-bearing sentence for a future maintainer wondering why we're sending the user out to system Settings instead of calling an API. The "how" is well-documented; consider re-adding the "why" so the comment carries both.

https://github.com/Kulitorum/Decenza/blob/70d5e177b4ef9304e654802170a16e484ba2bd54/src/core/updatechecker.cpp#L130-L139

<details>

- [Worth a look] `versioncode.txt` bumped 3376 → 3377 deviates from the CI-owned version-code policy in [`docs/CLAUDE_MD/CI_CD.md`](https://github.com/Kulitorum/Decenza/blob/70d5e177b4ef9304e654802170a16e484ba2bd54/docs/CLAUDE_MD/CI_CD.md). Contextually justified for the workflow_dispatch test install (the prior APK at 3375 couldn't install over the device's 3376), and the next tag push will bump cleanly past 3377 — but worth confirming this is the intended end state before merge (score: 50)
- [Worth a look] `intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)` in `launchSawPermissionSettings` is the correct pattern when called from a non-Activity context but redundant (and on some OEMs can break back-navigation) when called from an Activity. The empirical Samsung test returned to Decenza successfully, so this is not hitting users in practice — but worth removing for tighter back-stack behavior (score: 50)
- [OK to skip] No threading contract documented on `launchSawPermissionSettings`. Activity methods aren't formally main-thread-safe; works fine empirically because Qt's main thread interacts with Binder safely (score: 50)
- [OK to skip] Class-level JavaDoc on `UpdateRelaunchReceiver` only describes its `BroadcastReceiver` role; the new static utility is undocumented at the class level (score: 50)
- [OK to skip] Both comments claim "Qt 6.10's variadic JNI marshalling silently drops jstring arguments" — the symptom (empty SSP) is empirically confirmed but the root-cause attribution is inferred, not from a minimal repro. Code fix is correct regardless; the certainty in the claim could be softened (score: 50)

</details>

PR #1144, opened 2026-05-14
feat(charts): add Qt Graphs alongside Qt Charts (migration Stage 0)

skialpine:
## Summary

Stage 0 of the [`migrate-charting-to-qt-graphs`](openspec/changes/migrate-charting-to-qt-graphs/proposal.md) OpenSpec change. Adds Qt Graphs to the build alongside Qt Charts (no Charts removal yet) and creates three reusable bridge QML components that close known Qt Graphs feature gaps used by Decenza's graphs.

This PR is **infrastructure-only**. No graph file is migrated. Every existing graph continues to render via Qt Charts unchanged.

## What's in

**Build wiring**
- `find_package(Qt6 COMPONENTS Graphs)` added; `Qt6::Graphs` linked alongside `Qt6::Charts`. Both coexist through Stages 1–3; Stage 4 removes Charts.

**Bridge components** under `qml/components/graphs/`
- `AutoRangingAxis.qml` — `ValueAxis` wrapper that derives `min`/`max` from attached `XYSeries` with configurable padding and floor/ceiling clamps. Replaces Qt Charts' implicit auto-range behavior (Qt Graphs requires explicit ranges).
- `CustomLegend.qml` — themed horizontal legend with tap-to-toggle visibility. Styled via `Theme.qml`. Replaces `ChartView.legend`.
- `DashedLineSeries.qml` — `ShapePath` overlay aligned to a `GraphsView`'s `plotArea`, mapping axis-space points to pixel space. Uses `visualMin`/`visualMax` (Qt 6.11+) when present. Replaces `LineSeries { style: Qt.DashLine }`.

**Docs**
- `openspec/config.yaml` Tech Stack now notes both libraries linked during the migration.
- `tasks.md` 0.1 build wiring + 0.3 bridge components ticked.

## Verification

- Local build via Qt Creator MCP: succeeded, 0 errors, 1 warning. The warning is `Multiple C++ types called QAbstractAxis found!` — expected during coexistence (both `QtCharts::QAbstractAxis` and `QtGraphs::QAbstractAxis` are visible). Goes away after Stage 4 removes `Qt6::Charts`.

## Out of scope (separate work items)

- **Stage 0.2 spike** — verify QTBUG-142046 fix, `QXYSeries::replace()`, `plotArea` clipping behavior, `GraphsTheme` integration strategy. Requires a dev machine running the migrated graphs.
[...]

PR #1146, opened 2026-05-14
feat: migrate charting from Qt Charts to Qt Graphs (Stages 2–4)

skialpine:
## Summary

Completes the `migrate-charting-to-qt-graphs` OpenSpec change. After this PR Decenza renders every graph through Qt Graphs and no longer links Qt Charts at all. Builds on top of Stage 0 + 1 already on `main` (PR #1144).

Eight commits, one per stage so reviewers can read the migration in the order it was actually performed:

| Commit | Stage | What |
|---|---|---|
| `d3529c4f` | 2 | `SteamGraph` + `SteamDataModel` |
| `8c6cff33` | 2 fix | Wrap `GraphsView` in outer `Item` so overlays render (sibling vs child z-order) |
| `219d0110` | 3a | `ShotGraph` + `ShotDataModel` (live extraction view) |
| `f6e6e31b` | 3a fix | Time axis tracks `rawTime` continuously, dynamic `tickInterval` |
| `34520ffc` | 3b | `HistoryShotGraph` (shot detail / post-shot review / bean info / last-shot widget) |
| `ec172eac` | 3c | `ComparisonGraph` + `ShotComparisonModel` (3-shot side-by-side) |
| `bdecced4` | 3d | `ProfileGraph` (profile / recipe editor preview) |
| `47efa205` | 4 | Drop Qt Charts from `CMakeLists.txt`, tests, all 6 CI workflows |
| `5b461d22` | doc | `docs/CLAUDE_MD/PERFORMANCE_BASELINE.md` measurement protocol |

## Key architectural notes

- **Outer `Item` wraps every `GraphsView`** so `FastLineRenderer` traces, dashed goal curves, frame markers, and the crosshair render as siblings on top of the chart. Direct children of `GraphsView` get swallowed by its scene-graph paint — discovered the hard way on Stage 2 (lines were invisible until restructured).
- **Three bridge components in `qml/components/graphs/`** close the gaps Qt Graphs has vs. Qt Charts: `AutoRangingAxis`, `CustomLegend`, `DashedLineSeries` (now also supports solid stroke via a `dashed` property — used as the right-axis overlay on graphs with multiple Y axes).
- **Right-axis traces use `QtObject` value holders, not real `ValueAxis`** — Qt Graphs has no sanctioned `axisYRight` here on 6.11. `DashedLineSeries` reads `min`/`max` for its data→pixel mapping, so a min/max QObject is enough.
- **Data models lose their `QXYSeries*` handshakes.** Goal segments and right-axis traces are exposed as `QVariantList` Q_PROPERTYs that `DashedLineSeries` binds to directly. Same 33 ms flush batching is preserved.
- **`grep -r "QtCharts" qml/ src/` returns zero** after this PR. The "Multiple QAbstractAxis types" warning that flagged the coexistence period is gone.

## Performance — P.5 measurement on macOS Debug

Captured `QSG_RENDER_TIMING=1` per-frame totals on the same Mac with the same simulated A-Flow shot (~25–27 s extraction) on both builds. Pre-migration build is `f8eb3bbf` (commit immediately before Stage 0). Debug-vs-Debug, so absolute numbers will look very different on Release / tablet — the ratio is the signal.

**Shot extraction window (during the pour):**

| Metric | Qt Charts | Qt Graphs | Δ |
|---|---|---|---|
| Frames captured | 2851 | 3048 | +7 % |
| Mean frame time | **5.63 ms** | **4.31 ms** | **−23 %** |
| p95 frame time | 8.0 ms | 8.0 ms | flat (display refresh ceiling) |
| p99 frame time | 8.5 ms | 9.0 ms | +0.5 ms |
| Drops > 16 ms | 0 | 0 | none |
| Max frame | 16 ms | 24 ms | single outlier (goal-curve flush) |

**Full run (incl. idle, shot, history scroll, comparison):**

| Metric | Qt Charts | Qt Graphs | Δ |
|---|---|---|---|
| Mean frame time | 6.32 ms | 4.69 ms | **−26 %** |

**Tablet (Decent SM-X210, Android, Release APK from this PR's CI build):**

Sampled `/proc/<pid>/stat` at 2 Hz across a simulated shot.

| Metric | Qt Charts | Qt Graphs | Δ |
|---|---|---|---|
| Shot-window mean CPU (samples ≥ 40 %) | 68.0 % | 56.7 % | **−16.7 %** (−11.4 pp of one core) |
| Whole-run mean CPU | 61.3 % | 41.2 % | **−32.8 %** (−20.1 pp) |
| Whole-run p95 CPU | 78.4 % | 75.4 % | −3 pp |
| Max CPU | 104.5 % | 77.5 % | −27 pp |
| RSS max | 280.5 MB | 245.0 MB | −35 MB |

11 percentage points of one core off the shot-window mean is real heat / battery savings on the slower tablet — the absolutely-bigger win that the proposal predicted ("the larger remaining CPU cost on extraction-time UI"). The 33 % whole-run improvement is bigger because Qt Graphs also costs less when idling, scrolling history, and on the compare page.

P.5 verdict: **continue.** The proposal called for "measurable CPU drop on the live-shot metric → schedule Stages 2 + 3 immediately." 23 % mean reduction is well past noise; on the Decent tablet (where the 16 ms budget actually bites) the saved per-frame cost should keep more frames inside budget. Stages 2 + 3 + 4 are already in this PR; Qt 6.12 `useCanvasPainter` polish is tracked in `charts-qt-6-12-polish`.

Raw data: `/tmp/decenza-charts.log` (Qt Charts run), `/tmp/decenza-graphs.log` (Qt Graphs run), parser `/tmp/parse_render.py` — happy to re-run on any other commit if reviewers want.

## What's deferred

`openspec/changes/charts-qt-6-12-polish` tracks the items that need Qt 6.12: tick-mark length, leftmost X label alignment, `useCanvasPainter: true` flip per migrated `GraphsView`, declarative `XYSeries.data` for static series, `labelPostFormat`. None of these block this PR.

Tablet measurement is still a follow-up — Mac Debug confirms the direction; tablet Release will confirm the magnitude.

## Test plan

- [ ] On-device smoke test on the Decent tablet (Samsung SM-X210)
- [ ] Live espresso shot — pressure / flow / temperature / weight traces, dashed goal segments across pump-mode transitions, frame markers, "Start" / "End" markers, rotated marker labels with `[W]`/`[P]`/`[F]`/`[T]` suffixes, pump-mode color bars
- [ ] Live steam session — pressure / flow / temperature trace, dashed flow-goal curve, legend at top-left, right-axis temperature labels
- [ ] Shot detail screen — all traces render, crosshair + value tooltip on tap, right-axis margin tap toggles weight ↔ temperature
- [ ] Post-shot review — same checks
- [ ] Comparison page with 2–3 shots — solid (shot 0), dashed (shot 1), dash-dot (shot 2); per-shot Canvas phase markers; default-visible second-to-last phase seeds the crosshair
- [ ] Profile editor / Recipe editor / Simple profile editor / Profile preview popup / Profile info — pressure + flow + temperature preview live as you tap through frames
- [ ] Bean info page / Last-shot layout widget — `HistoryShotGraph` renders correctly
- [ ] Clean build on Windows, macOS, iOS, Android via CI
- [x] Mac Debug perf comparison captured (see "Performance" section above)
- [x] Tablet Release perf comparison captured (see "Performance" section above)

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -65,7 +65,7 @@ jobs:
           version: '6.11.1'
           target: 'android'
           arch: 'android_arm64_v8a'
-          modules: 'qtquick3d qtshadertools qtcharts qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtimageformats qtcanvaspainter'
+          modules: 'qtquick3d qtshadertools qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtimageformats qtcanvaspainter'
           install-deps: false  # Skip apt-get of 61 desktop packages — only libegl1/libgl1 needed for qsb shader compiler
           cache: true
           cache-key-prefix: 'qt-android-v1'
--- a/.github/workflows/ios-release.yml
+++ b/.github/workflows/ios-release.yml
@@ -75,7 +75,7 @@ jobs:
           version: '6.11.1'
           target: 'ios'
           arch: 'ios'
-          modules: 'qtquick3d qtshadertools qtcharts qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtimageformats qtcanvaspainter'
+          modules: 'qtquick3d qtshadertools qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtimageformats qtcanvaspainter'
           cache: true
           cache-key-prefix: 'qt-ios-v4'
 
--- a/.github/workflows/linux-arm64-release.yml
+++ b/.github/workflows/linux-arm64-release.yml
@@ -73,7 +73,7 @@ jobs:
           version: '6.11.1'
           target: 'desktop'
           arch: 'linux_gcc_arm64'
-          modules: 'qtshadertools qtcharts qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtserialport qtimageformats qtcanvaspainter'
+          modules: 'qtshadertools qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtserialport qtimageformats qtcanvaspainter'
           cache: true
           cache-key-prefix: 'qt-linux-arm64-v1'
 
--- a/.github/workflows/linux-release.yml
+++ b/.github/workflows/linux-release.yml
@@ -73,7 +73,7 @@ jobs:
           version: '6.11.1'
           target: 'desktop'
           arch: 'linux_gcc_64'
-          modules: 'qtquick3d qtshadertools qtcharts qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtserialport qtimageformats qtcanvaspainter'
+          modules: 'qtquick3d qtshadertools qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtserialport qtimageformats qtcanvaspainter'
           cache: true
           cache-key-prefix: 'qt-linux-v4'
 
--- a/.github/workflows/macos-release.yml
+++ b/.github/workflows/macos-release.yml
@@ -56,7 +56,7 @@ jobs:
           version: '6.11.1'
           target: 'desktop'
           arch: 'clang_64'
-          modules: 'qtquick3d qtshadertools qtcharts qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtwebengine qtserialport qtimageformats qtcanvaspainter'
+          modules: 'qtquick3d qtshadertools qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtwebengine qtserialport qtimageformats qtcanvaspainter'
           cache: true
           cache-key-prefix: 'qt-macos-v2'
 
--- a/.github/workflows/windows-release.yml
+++ b/.github/workflows/windows-release.yml
@@ -58,7 +58,7 @@ jobs:
           version: '6.11.1'
           target: 'desktop'
           arch: 'win64_msvc2022_64'
-          modules: 'qtquick3d qtshadertools qtcharts qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtserialport qtimageformats qtcanvaspainter'
+          modules: 'qtquick3d qtshadertools qtgraphs qtconnectivity qtmultimedia qtpositioning qtspeech qtwebsockets qtserialport qtimageformats qtcanvaspainter'
           tools: 'tools_ninja'
           cache: true
           cache-key-prefix: 'qt-windows-v3-aqt-master'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,6 @@ find_package(Qt6 REQUIRED COMPONENTS
     Quick
     QuickControls2
     Bluetooth
-    Charts
     Graphs
     Svg
     Multimedia
@@ -552,7 +551,6 @@ target_link_libraries(Decenza PRIVATE
     Qt6::Quick
     Qt6::QuickControls2
     Qt6::Bluetooth
-    Qt6::Charts
     Qt6::Graphs
     Qt6::Svg
     Qt6::Multimedia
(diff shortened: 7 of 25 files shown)


skialpine:
### Code review

Found 4 issues:

1. Weight curve in `ComparisonGraph` is double-scaled — `_curvePoints()` divides Y by 5, and `weightAxis.max` is also 60, so a 60 g sample renders at 12/60 = 20 % of plot height (not 100 % as the adjacent comment intends). Fix is to make either the division xor the axis range — not both. The comment at L374-375 explicitly states the desired invariant.

https://github.com/Kulitorum/Decenza/blob/73fc4053a15ac1700c76387dcfa210df81a1c580/qml/components/ComparisonGraph.qml#L368-L378

2. `_dataVersion` is bound to `comparisonModel.shotCount`, but `shotCount` returns `m_displayShots.size()` which stays constant (3) when the user shifts the comparison window between two windows that both contain 3 shots. The points binding reads `var _ = _dataVersion` as its only change-trigger, so it does not re-fire on window shifts. `_refreshAll()` only updates axes / phase data, not points. Suggest exposing a monotonic load-serial Q_PROPERTY (the model already has `m_loadSerial`) instead.

https://github.com/Kulitorum/Decenza/blob/73fc4053a15ac1700c76387dcfa210df81a1c580/qml/components/ComparisonGraph.qml#L138-L142

3. `_curves` comment references a non-existent \`fnY\` field — \"fnY is optional: if set, the y values from the model are transformed before the dashed overlay reads them (currently used for the /5 weight rescale)\". No curve in the array has an `fnY` field; the rescale lives in `_curvePoints()` as a `key === \"weight\"` branch. The comment would mislead anyone adding a new curve transform.

https://github.com/Kulitorum/Decenza/blob/73fc4053a15ac1700c76387dcfa210df81a1c580/qml/components/ComparisonGraph.qml#L62-L86

4. `HistoryShotGraph.updateTimeAxis()` early-returns on `pressureData.length === 0`, and the migration removed the `onTemperatureDataChanged` / `onWeightDataChanged` / `onGoalDataChanged` reload handlers. If a shot has temperature/weight data but no pressure (corrupt or partial row), the time axis stays at its default while the dashed overlays draw against it — leaving the trace misaligned. Either restore the data-change handlers, or drop the early return.

https://github.com/Kulitorum/Decenza/blob/73fc4053a15ac1700c76387dcfa210df81a1c580/qml/components/HistoryShotGraph.qml#L106-L125

<details>

- [Worth a look] `HistoryShotGraph.doReload()` header comment says Qt.callLater coalescing is intrinsic to the function, but the actual mechanism is in the `onXxxDataChanged: Qt.callLater(doReload)` callers — comment is misleading (score: 50)
- [OK to skip] `ShotGraph.recalcMax()` lacks the `_recalcInProgress` re-entry guard that `SteamGraph.recalcMax()` has; the value-equality check breaks the loop in practice, but the asymmetry is inelegant (score: 50)

</details>

skialpine:
### Code review

Found 3 issues:

1. New right-axis temperature labels in `ProfileGraph` use a hardcoded font size — `font.pixelSize: Theme.scaled(12)` instead of `font: Theme.captionFont`. The peer `ShotGraph` / `HistoryShotGraph` / `SteamGraph` right-axis labels all use `font: Theme.captionFont`, which also picks up `Settings.theme.customFontSizes.captionSize`. (CLAUDE.md says "Use `Theme.qml` singleton for all styling — never hardcode colors, font sizes, spacing, or radii".)

https://github.com/Kulitorum/Decenza/blob/6a1049b46152b5427d6fd67827e3519cefea71a0/qml/components/ProfileGraph.qml#L196-L207

2. `dCdtAxis` `QtObject` in `ComparisonGraph` has hardcoded `min: 0` / `max: 20` defaults that look static, but `_updateDCdtAxis()` rewrites both on every data load. `tempAxis` and `weightAxis` next door each have a why-comment; this one doesn't, so a future reader will mistake the initial values for the canonical range.

https://github.com/Kulitorum/Decenza/blob/6a1049b46152b5427d6fd67827e3519cefea71a0/qml/components/ComparisonGraph.qml#L387-L394

3. `DashedLineSeries` header docstring leads with "Place this *inside* a `GraphsView`" as the primary usage. After this PR every embedding (`ShotGraph` / `HistoryShotGraph` / `ComparisonGraph` / `SteamGraph` / `ProfileGraph`) places it as a sibling and passes `graphsView: chart.graphsViewRef`, because direct children of `GraphsView` get swallowed by its scene-graph paint. Swap the emphasis or future bridge users will reach for the deprecated pattern.

https://github.com/Kulitorum/Decenza/blob/6a1049b46152b5427d6fd67827e3519cefea71a0/qml/components/graphs/DashedLineSeries.qml#L1-L15

<details>

- [Worth a look] `_curves` comment in `ComparisonGraph` says "The delegate resolves the key to a real axis below" — actually it calls `chart._axisFor(curveDef.axisKey)` from the delegate; the comment hides the indirection (score: 60)
- [OK to skip] `ShotGraph.recalcMax()` lacks the explicit `_recalcInProgress` guard that `SteamGraph.recalcMax()` has — works in practice because the `if (newMax !== _lastAxisMax || tickInterval !== step)` check breaks the loop (score: 50)

</details>

PR #1155, opened 2026-05-15
fix: remove inferred-good shot auto-rating (#1150)

skialpine:
## Summary

Layer 3 of the May-4 \`shot-rating-capture\` change auto-stamped clean unrated shots with \`enjoyment=75\` and \`enjoymentSource="inferred"\` so the advisor's \`bestRecentShot\` block could stay warm even when the user hadn't rated anything. Two weeks of production data showed two failures:

1. **User-visible breakage** — issue #1150: the Visualizer "Default Shot Rating" setting (a long-standing user preference) was silently overwritten by the auto-rated 75.
2. **No measurable advisor benefit** — the inferred score was paired with \`confidence: "inferred"\` and a system-prompt instruction telling the LLM to treat it as a hint requiring user confirmation. Meanwhile the LLM already received the underlying detector signals (\`verdictCategory\`, \`channelingSeverity\`, \`grindDirection\`) directly.

This PR treats Layer 3 as a failed experiment and rolls it back. A shot's rating is now exclusively what the user provided (manually, via QuickRatingRow, or via conversational reply). Layers 1 and 2 are untouched.

## Key changes

- **Migration 16** drops \`shots.enjoyment_source\` in a transaction. Before the drop, rows with \`enjoyment_source = 'inferred'\` have their \`enjoyment\` reset to the user's configured default (\`QSettings shot/defaultRating\`, fallback 75) — **not** to 0, since most users have a non-zero default. Visualizer-uploaded affected rows are stashed in a pending-sync \`QSettings\` list.
- **MainController Visualizer back-sync** — \`processPendingVisualizerRatingSync\` drains the pending list serially via \`VisualizerUploader::updateShotOnVisualizer\` once credentials are available, so the cloud copy on visualizer.coffee also reflects the corrected rating. Failed entries persist for the next boot.
- **End-to-end strip** of \`enjoymentSource\` / \`bestRecentShot.confidence\` / "inferred" plumbing: \`ShotProjection\`, \`ShotData\` / \`ShotRecord\`, the INSERT/SELECT/UPDATE column maps, the MCP \`shots_list\` filter, the MCP \`updateShot\` validator, the dialing-block builders, \`ShotSummary\` fields, and the related system-prompt teaching.
- **QML cleanup** — \`PostShotReviewPage\` no longer treats inferred ratings specially; \`QuickRatingRow\` shows when \`editEnjoyment === 0\`.
[...]

PR #1156, opened 2026-05-15
fix: persist Visualizer shot id from controller + one-time reconciliation backfill

skialpine:
## Summary

Visualizer auto-upload succeeds server-side but the returned shot UUID was only written back to the local DB by a **transient QML page handler** (`PostShotReviewPage`/`ShotDetailPage` `onUploadSuccess`). When the post-shot review page is disabled (`visualizer/showAfterShot=false`), auto-closes, or is navigated away before the ~1 s upload round-trip resolves, the link is lost forever even though the shot is live on visualizer.coffee.

Confirmed on-device (v1.7.5 build 3381): DB shot 923 uploaded as [`e81fcb49-…`](https://visualizer.coffee/shots/e81fcb49-f9d1-4f7c-aaf1-bf07c19f2f26) (HTTP 200) yet `shots_get_detail(923)` reports `visualizerId:""`. Shots 901–923 are all orphaned this way. This also strands #1155's migration-16 inferred-rating back-sync, which filters on `visualizer_id != ''`.

## Component A — authoritative C++ writeback

- Thread the originating `shots.id` through the uploader; new signal `uploadSucceededForShot(dbShotId, visualizerId, url)`.
- `MainController` connects it to `requestUpdateVisualizerInfo`, independent of any UI page.
- Shot-end auto-upload now dispatched from the `shotSaved` callback (id known) instead of before save, so it can no longer orphan.
- QML `onUploadSuccess` persistence removed; UI refresh via `visualizerInfoUpdated` retained.

## Component B — one-time reconciliation backfill (recovers existing orphans)

- `MainController::processVisualizerReconciliation`: run-once (`QSettings visualizerBackfill/doneV1`), credentials-gated (absent → skip without consuming the flag), bounded to 60 days.
- `ShotHistoryStorage::reconcileVisualizerLinksStatic`: matches empty-`visualizer_id` rows to cloud shots by start time (`clock` ±2 s), strict 1:1, no reuse of an id already on a row, ambiguous skipped.
- Each linked row is queued onto the existing migration-16 serial drain to push the now-authoritative local rating up (cleared → JSON `null`). **Order-independent** of the migration-16 back-sync — repairs the orphaned-and-stranded cohort on its own.

Confirmed `GET /api/shots` returns `ShotSummary {id,clock,updated_at}` only (OpenAPI 1.8.2) — no `espresso_enjoyment` — so the self-correct PATCH is unconditional per linked row (idempotent over the bounded set) rather than adding a per-shot detail fetch just to compare.

OpenSpec change: `persist-visualizer-id-in-controller`.

## Test plan

- [x] `mcp__qtcreator__build` clean; full suite **2052 passed, 0 failed**.
- [x] New unit test `tst_dbmigration.cpp::reconcileVisualizerLinks_matchingContract` — in-window match, already-linked not reused, out-of-window skipped, ambiguous skipped, idempotent re-run.
- [ ] Manual (device): fresh shot with review page disabled records `visualizer_id`; reconciliation relinks orphans 901–923 and corrects their stranded `espresso_enjoyment=75` to unrated (`defaultShotRating=0`); reboot does not re-run.
- [ ] Sequencing: land before relying on #1155's migration-16 back-sync on-device — this carries the relink + self-correct so the stranded 75s are actually fixed.


skialpine:
### Code review (4 agents: general/CLAUDE.md, silent-failure, tests, comments)

One architectural root cause + supporting issues found and **fixed in 0fc4e2f4** (review ran on f58e42ac):

**Critical**
1. DB open / SELECT failure during reconciliation emitted an empty list indistinguishable from "0 matched"; the handler set the **irreversible** `visualizerBackfill/doneV1` run-once flag on *any* emission — one transient DB-locked-at-boot hiccup permanently killed the backfill for that device. Fixed: `reconcileVisualizerLinksStatic` now returns `bool`, the threaded wrapper ANDs `withTempDb` open success, the signal carries `ok`, and the flag is set only on a genuinely completed pass.

https://github.com/Kulitorum/Decenza/blob/0fc4e2f4fef2944238d0f85d57c19a3982451ba8/src/controllers/maincontroller.cpp#L2710-L2723

2. `reply->errorString()` was read after `reply->deleteLater()` — unreliable diagnostic on the only failure surface. Captured before.

**High**
3. Hitting the `kMaxPages` page ceiling before the real end emitted `shotListFetched` with truncated data → permanent partial backfill. Now emits `shotListFailed` (retry next boot) on the abnormal exit.

https://github.com/Kulitorum/Decenza/blob/0fc4e2f4fef2944238d0f85d57c19a3982451ba8/src/network/visualizeruploader.cpp#L582-L592

**Medium**
4. A 200 lacking a valid `paging` object (expired-session/error envelope) aliased "empty account" and burned the run-once flag — now fails safe; parse errors include `QJsonParseError::errorString()`.
5. `usedIds` seed-SELECT failure was swallowed, silently disabling the duplicate-link guard — now logs and returns `false`.

**Comments**
6. The "uploads are strictly serial (`m_uploading` guard)" claim was false (`m_uploading` is UI-only, never checked) — corrected in the header, cpp, and design.md D1 to state the real invariant (caller discipline). Sort-order over-attribution to OpenAPI softened.

**Tests added** (per the test-coverage review): `reconcileVisualizerLinks_toleranceBoundary` (±2 s inclusive / +3 s excluded — the load-bearing skew line) and `reconcileVisualizerLinks_multiRowAndEmpty` (the real many-orphan production scenario + cross-row no-reuse + empty-cloud no-op). Full suite **2054 passed**.

H-2 (`uploadShot` `=0` default): verified only two callers, both pass `dbShotId`; the `=0` default can't be cleanly dropped (it follows other defaulted params) and the `qWarning`-and-skip guard is adequate given no orphaning caller exists. `shouldStopPaging` extraction noted as a non-blocking follow-up (task 4.6).

Not auto-merging — manual on-device verification (§6.1–6.3: relink the known orphans 901–923, confirm Visualizer ratings corrected) is the next step per plan.

PR #1157, opened 2026-05-15
fix: QSettings store-scope mismatch (migration 16, reconciliation, accessibility)

skialpine:
## Summary

On-device verification of the v1.7.5 build carrying #1155 + #1156 surfaced a **real production bug** that the review agents and unit tests all missed.

The app persists structured settings to `QSettings("DecentEspresso","DE1Qt")` (the `Settings` facade + all 12 `settings_*.cpp` domain classes). But `main.cpp:387` sets `applicationName` to `"Decenza"`, so a **bare `QSettings()`** resolves to `DecentEspresso/Decenza` — the *secondary* store (documented at `settings.cpp:567`), which does not hold the primary keys.

Every settings access added by #1155/#1156 used a bare `QSettings()`. Observed on the build-3383 boot log:

- **Migration 16** read `shot/defaultRating` via bare `QSettings` → empty → `75` fallback (not the user's `0`) → the inferred-rating reset was a silent `75→75` no-op.
- **`processVisualizerReconciliation` / `processPendingVisualizerRatingSync`** read Visualizer credentials via bare `QSettings` → empty → `"skipped (no credentials)"` → the reconciliation never ran.

This is why the migration-16 unit test passed (its harness `QSettings` org/app happened to match the bare default) while production silently failed.

## Fixes

1. **Scope fix:** migration 16, the 6 `MainController` reconciliation/pending-sync sites, and the migration-16 regression test now construct `QSettings("DecentEspresso","DE1Qt")` (matching the existing `main.cpp:259` precedent). The test is now a genuine guard (it failed in the wrong-scope world only because test + production were wrong identically).

2. **Same bug class, found by a full audit (`accessibilitymanager`):** `AccessibilityManager` used `QSettings("Decenza","DE1")` — an isolated *third* store — so accessibility **backup captured defaults** and **restore was a silent no-op** (`shotserver_backup.cpp` read `accessibility/*` off the bare secondary store). `AccessibilityManager` now uses the primary store like every other domain, with a guarded, idempotent one-time `migrateLegacyStore()` that carries existing values over; `shotserver_backup.cpp` reads/writes `accessibility/*` on the primary store and stamps the migration guard on restore so a restored value isn't re-clobbered from the legacy store.

## Audit result

Every other bare-`QSettings` site (AI conversations, `shotMap/*`, `storage/*`, datamigration tokens, web-auth sessions, app-name migration, icon-version) is a **self-consistent secondary-store user by design** — no further occurrences of this bug.

## Deployment note

Makes #1155/#1156 actually work for users not yet on the buggy build 3383. Devices that already ran buggy migration 16 (schema_version=16, column dropped) won't auto-re-reset the formerly-inferred shots — handled separately (manual per-device, post-verification).

## Test plan

- [x] `mcp__qtcreator__build` clean; full suite **2054 passed**.
- [x] Migration-16 regression test now seeds/reads `QSettings("DecentEspresso","DE1Qt")` — fails if production reverts to bare scope.
- [ ] Manual: fresh device on the fixed build → migration 16 resets inferred shots to the *configured* default; reconciliation runs (credentials found) and relinks/corrects orphaned uploads.
- [ ] Manual: backup on the fixed build captures real accessibility settings; restore applies them and `AccessibilityManager` reads them.

### DIFF
--- a/src/controllers/maincontroller.cpp
+++ b/src/controllers/maincontroller.cpp
@@ -220,7 +220,13 @@ MainController::MainController(QNetworkAccessManager* networkManager,
             || visualizerId != m_migration16InFlightVisualizerId)
             return;
 
-        QSettings s;
+        // Must match the app's settings scope (Settings owns
+        // QSettings("DecentEspresso","DE1Qt")); a bare QSettings()
+        // resolves to a different, empty store (main.cpp sets
+        // applicationName "Decenza"). Same fix applies to every
+        // QSettings construction in this file's reconciliation /
+        // pending-sync code.
+        QSettings s(QStringLiteral("DecentEspresso"), QStringLiteral("DE1Qt"));
         QJsonArray pending = QJsonDocument::fromJson(
             s.value(QStringLiteral("migration16/pendingVisualizerSync")).toByteArray()).array();
         for (qsizetype i = 0; i < pending.size(); ++i) {
@@ -2581,7 +2587,7 @@ bool MainController::isSawSettling() const {
 
 void MainController::processPendingVisualizerRatingSync()
 {
-    QSettings s;
+    QSettings s(QStringLiteral("DecentEspresso"), QStringLiteral("DE1Qt"));
     if (!s.contains(QStringLiteral("migration16/pendingVisualizerSync")))
         return;
 
@@ -2602,7 +2608,7 @@ void MainController::dispatchNextPendingVisualizerSync()
     if (!m_migration16InFlightVisualizerId.isEmpty()) return;
     if (!m_visualizer || !m_shotHistory) return;
 
-    QSettings s;
+    QSettings s(QStringLiteral("DecentEspresso"), QStringLiteral("DE1Qt"));
     const QByteArray raw = s.value(
         QStringLiteral("migration16/pendingVisualizerSync")).toByteArray();
     if (raw.isEmpty()) return;
@@ -2650,7 +2656,7 @@ void MainController::dispatchNextPendingVisualizerSync()
         if (!shot.isValid()) {
             qWarning() << "MainController: migration16 sync — shot" << shotId << "no longer exists; dropping";
             // Pop the bad entry and continue.
-            QSettings ss;
+            QSettings ss(QStringLiteral("DecentEspresso"), QStringLiteral("DE1Qt"));
             QJsonArray remain = QJsonDocument::fromJson(
                 ss.value(QStringLiteral("migration16/pendingVisualizerSync")).toByteArray()).array();
             if (!remain.isEmpty()) remain.removeFirst();
@@ -2674,7 +2680,7 @@ void MainController::processVisualizerReconciliation()
 {
     if (!m_visualizer || !m_shotHistory) return;
 
-    QSettings s;
+    QSettings s(QStringLiteral("DecentEspresso"), QStringLiteral("DE1Qt"));
     if (s.value(QStringLiteral("visualizerBackfill/doneV1"), false).toBool())
         return;  // already reconciled on this device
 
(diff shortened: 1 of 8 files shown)


skialpine:
### Code review (2 agents: general/CLAUDE.md, silent-failure)

Review ran on `2d0a9a4b`; fixes in **caba4c6f**.

**Verdict:** scope fix verified **fully correct** — every moved read (`shot/defaultRating`, `visualizer/username|password`, `visualizerBackfill/doneV1`, `migration16/pendingVisualizerSync`, `accessibility/*`) confirmed to have always been written to the store it's now read from; no data orphaned. Migration design (guarded, copy-if-absent, legacy left intact for downgrade) sound. CLAUDE.md clean (no timers, threading unchanged, `qsizetype`/`m_` ok).

**HIGH — fixed:** `AccessibilityManager::migrateLegacyStore()` stamped the irreversible one-shot guard even when the legacy-store read provably failed (corrupt INI / access error) → permanent silent loss of a user's accessibility settings on a transient unreadable store, with a log line (`"migrated 0…"`) indistinguishable from the benign new-user case. Now `sync()`+`status()`-checks the legacy store; on a provable failure it `qWarning`s and returns **without** stamping the guard (retries next launch). Success log → `qInfo` with legacy key count so "nothing to migrate" vs "all already present" is distinguishable in support logs.

https://github.com/Kulitorum/Decenza/blob/caba4c6f915abc70ba687646df06d8d6f175510a/src/core/accessibilitymanager.cpp#L103-L131

**Hardenings also applied:**
- `Settings::factoryReset()` now clears the legacy `QSettings("Decenza","DE1")` store too — the migration guard lives in the primary store (wiped by reset), so without this a post-reset launch would resurrect old accessibility settings. Factory reset now actually resets accessibility. (general-agent observation; the migration this PR adds makes it relevant.)
- Restore: the migration guard is now stamped on **any** successful restore (not only when an `accessibility` block is present) — a restored profile is authoritative and must not later be merged with this device's stale legacy store (silent-failure Concern #3). Added a post-sync `status() != NoError` warning so a failed accessibility persist on restore is observable instead of silently reported as success (Concern LOW).

Concern #4 (scope move could orphan data) — agent confirmed **NONE**: reads and writes move together in the same commit; transient/idempotent where not.

Full suite **2054 passed**, build clean.

skialpine:
### Follow-up review (test-coverage + comment-accuracy agents)

Ran the two agents skipped on the first pass. Both verdicts positive; all real items addressed in **4eaccf67**.

**Test-coverage:** the extracted `migrateAccessibilityLegacyStore` static is the right seam and the irreversible/most-dangerous logic (guard burn, copy-if-absent, status-gated defer, idempotency, empty-vs-all-present discrimination) is well covered. Confirmed the migration-16 scope fix is genuinely regression-tested (`tst_dbmigration v16_resetsInferredAndDropsColumn` asserts `enjoy==50` not the 75 fallback — fails if production reverts to bare scope). Two flagged gaps:
- **crit-7 — added:** `legacyMigration_restoreStampedGuardSuppressesLaterLegacyMerge` pins that a restore-stamped guard suppresses a later legacy merge (the data-loss scenario the unconditional restore stamp exists to prevent).
- **crit-6 — documented, not flaky-tested:** the copy-before-status ordering can't be black-box forced — Qt 6.11's IniFormat parser is all-or-nothing for our fixtures (parses fully, or fails wholesale with zero keys). Rather than ship a Qt-version-fragile test, the invariant is preserved structurally and its observable contract (deferred read → no guard stamp → later run completes) is already deterministically covered by `_defersWithoutStampingGuardOnUnreadableLegacy`. Added a fixture-sanity `QVERIFY(status != NoError)` so a future tolerant Qt fails loudly instead of testing the wrong path.
- 6-site MainController repoint + socket-level backup/restore: honestly deferred (no harness; mechanical scope substitution anchored by the migration-16 regression test).

**Comment-accuracy:** no critical issues, no stale `QSettings(Decenza,DE1)` claims. 3 wording imprecisions fixed: (1) "legacyKeyCount is the best available signal" overstated mitigation → now "post-hoc breadcrumb, not a mitigation" + partial-copy-before-status made explicit; (2) header doc now states copies precede the status check; (3) "ANY successful restore" → "whenever extra_settings.json is restored". factoryReset 2b rationale and the `reconcileVisualizerLinksStatic` symbol reference verified accurate end-to-end.

Full suite **2060 passed**, build clean. Nothing from any of the 4 agents (2 first pass + 2 follow-up) remains open.

Issue #1161, opened 2026-05-15
Manually-stopped shots surfaced as dial-in data with no stop-reason flag


skialpine:
Parent: #1147. Overlaps with the `stoppedBy` field proposed in #1158.

## Problem

Shots ended manually have arbitrary yield. The MCP surfaces their yields as if they were dial-in outcomes, so the LLM diagnoses "inconsistent yield" / grind problems that are actually just the user hitting stop.

## Evidence (from #1147)

Rao Allongé shots at 71g and 89g on a 90g target. Reporter: *"I was ending these shots manually, so basing insight on that arbitrary data is not helpful."*

## Proposed direction (for discussion)

- Surface stop-reason in the MCP shot payload — this is the same `stoppedBy: "weight" | "volume" | "time" | "manual" | "frameLimit"` field proposed in #1158; this issue is the consumer side (discounting), #1158 is the producer side (provenance).
- Discount yield-derived diagnostics for manually-stopped shots in the dial-in blocks and the shared system prompt.
- First step: verify whether stop-reason is already persisted on the shot record; if not, that capture is a prerequisite.

## Relevant code

`src/history/shothistory_types.h`, `src/mcp/mcptools_shots.cpp`, `src/ai/dialing_blocks.{h,cpp}`.

skialpine:
cc @fredphoesh — broken out from #1147 (your report).

Issue #1158, opened 2026-05-15
MCP shot context lacks control-mode provenance → AI infers false causation on stop-at-weight / flow profiles


skialpine:
Parent: #1147 (AI bad advice via MCP).

## Problem

When an external LLM uses the Decenza MCP (`dialing_get_context`, `shots_*`), the per-shot payload exposes yield, duration, grind, etc. but does **not** state which variable was *controlled* vs *passive* for that shot's profile. The LLM therefore reads passive outputs as dial-in feedback and asserts physically-wrong causation.

## Evidence (from #1147, a D-Flow shot: stop-at-weight + flow-targeted pour)

- AI: *"you stepped finer to 22.5 and the next two both landed right on target (42.2–42.4g). Good instinct."* — yield landed on target because the profile **stops at weight**, not because of the grind change.
- AI: *"The concern is duration — you're pulling in 32–34s, notably faster…"* — for a flow-targeted pour that stops at weight, duration ≈ stop_weight ÷ flow_target; grind barely moves it unless the puck is choked or free-flowing. Duration is **not** a dial-in signal here.

Reporter: *"yield has nothing to do with grind setting with that profile, it will stop at 42g regardless"* / *"1.7g/s will complete a 42g shot in the same time, regardless of the pressure it takes to get that flow."*

Same failure class as the "passive signal read as goal" lessons in `docs/CLAUDE_MD/AI_ADVISOR.md`.

## Proposed direction (for discussion)

- Add explicit control-mode provenance to the per-shot MCP block and `dialing_get_context`, e.g. `pourControl: "flow" | "pressure"` and `stoppedBy: "weight" | "volume" | "time" | "manual" | "frameLimit"`.
- Mark yield/duration as non-diagnostic when stop-at-weight + flow-targeted.
- Reflect this in the shared shot-analysis system prompt so the in-app advisor benefits too.

## Relevant code

`src/mcp/mcptools_dialing_blocks.h`, `src/ai/dialing_blocks.{h,cpp}`, `src/ai/shotsummarizer.cpp`, `src/mcp/mcptools_shots.cpp`.

skialpine:
cc @fredphoesh — broken out from #1147 (your report).

skialpine:
Resolved on main:

- Control-mode provenance (`pourControl`) — #1163 (commit `e3fe66d3`)
- Manual-stop / `stoppedBy` capture + dial-in discounting — #1161 via #1167 (commit `63d8448a`)

[...]

Issue #1147, opened 2026-05-14
[Bug] AI Bad advice


fredphoesh:
### What happened?

**My writing/comments bold to distinguish.
Question: what did you think of the last few shots**

Claude responded: Here's what I'm seeing across your last few shots:
Analyzed recent espresso shots, identified grind adjustments, noted extraction speed concerns
Today's session — new bag (Mocha Java 160426-1)
You dialled in nicely. The first shot at 23 came up a bit short on yield (39.3g), so you stepped finer to 22.5 and the next two both landed right on target (42.2–42.4g). Good instinct.

**Not true! yield has nothing to do with grind setting with that profile, it will stop at 42g regardless.**

The concern is duration — you're pulling in 32–34 seconds, which is notably faster than your best-rated shot on a similar Mocha Java (shot 5188, back in March, which ran 49 seconds at setting 22 and scored 84). That's a meaningful gap.

**shot 5188? there is no such number on my decenza shot history on the web app…  Claude and Shot History on the web should give date/time. More importantly, I would likely have had a different bean back then, so the comparison isn’t meaningful and basing things on my taste score back then is flawed. It seems at some point, Decenza began rating unrated shots 75%**

What the data suggests
The new bag seems to want a finer grind than the previous one. Your April sessions on "Mocha Java 160426-2" were running 28–32s at 24.5, so you've already come in significantly — but there may still be more to explore on the finer side. The puck looks clean (no channeling), so there's no obstacle to going finer.
A thought for the next shot
Worth trying 22 or 21.5 to see if you can stretch the pour toward 40+ seconds and whether that brings out more complexity. The inferred 75s are solid, but your best Mocha Java shot tells you there's more on the table — and that shot ran much longer.

**I’m concerned about the data it’s referring to. Also there is no mention of the profile used. Looking back to when I wrote to Claude, it was likely a D-Flow. With D-Flow the target is flow, so unless you’re choking the shot, the duration should not vary much, and is not a criterion for judging a shot… as 1.7g/s will complete a 42g shot in the same time, regardless of the pressure it takes to get that flow.**

Have you noticed any flavour difference between the two batches? Adding a tasting note would help confirm whether the faster pour is actually affecting the cup, or whether the shorter time is working fine for this particular bean.

**Again, time is not relevant for Dflow (unless you’re chocking the shot or have no resistance at all and it’s flowing faster than the target)**

[...]

PR #1163, opened 2026-05-15
AI: pour control-mode provenance so SAW yield/duration isn't read as dial-in feedback

skialpine:
Addresses #1158 (broken out from #1147). Plain reference, not a closing keyword — the reporter hasn't confirmed the fix yet.

## Problem

External Claude (via the Decenza MCP `dialing_get_context`) and the in-app advisor were asserting physically-wrong causation on stop-at-weight + flow-controlled profiles:

- *"you stepped finer to 22.5 and the next two both landed right on target … Good instinct."* — yield is pinned by the scale cutoff, not the grind.
- *"The concern is duration — you're pulling in 32–34s…"* — for a flow-controlled pour that stops at weight, duration ≈ stopWeight ÷ flowTarget, not a grind/quality signal.

Root cause: the `dialInSessions` / `bestRecentShot` blocks shipped only `yieldG` / `durationSec` for historical shots with **no control-mode provenance**, and the rendered recipe showed only frame *maximum* durations (e.g. "Pouring (127s)") with no indication the shot truncates at the weight target — so the model pattern-matched yield/duration as dial-in outcomes.

## Change

- `DialingBlocks::pourControlFromProfileJson(const QString& profileJson)` — inline, test-linkable (same rationale as `buildCurrentBeanBlock`). Parses the profile recipe's `steps[]` and returns the **longest-duration frame's `pump`** (`"flow"` / `"pressure"`); returns `""` when the recipe is empty / unparseable / step-less so the field stays sparse. (An earlier draft derived this from runtime phase markers; live MCP verification showed that misclassified a flow D-Flow/Q pour as `"pressure"`, so it was replaced with this recipe-based derivation — see commit history.)
- Emit `pourControl` on dial-in history: **hoisted to the session `context`** when every shot in a session shares it (the common case → zero per-shot repetition), emitted per-shot only when a session mixes flow/pressure variants. Also emit `pourControl` + `targetWeightG` on `bestRecentShot`.
- `DialingBlocks::withStopAtWeightNote(recipe, targetWeightG)` — shared inline helper that appends a one-line stop-at-weight clarification to the rendered recipe (frame durations are maximums; the shot truncates at the weight target). Called by **both** structured profile-block recipe-assembly sites — `dialing_get_context` (`mcptools_dialing.cpp`) and the in-app advisor (`ShotSummarizer::buildCurrentProfileBlock`) — gated on the analyzed shot's own target weight on both paths so they cannot drift.
- New rule in the shared `shotAnalysisSystemPrompt()` ("Reading the Recipe for Expected Behavior"): for a flow-controlled pour that stops at a weight target, yield landing on target is mechanical (don't credit grind) and duration ≈ stopWeight ÷ flowTarget (not a dial-in/quality signal); judge by pressure-at-flow, taste, TDS/EY, channeling.

Ships to **both** the MCP and the in-app advisor via the shared block builders + shared system prompt.

## Scope

- **No schema change** — all signals already in the data path.
- Manual-stop detection (the Rao Allongé 71g/89g case) is intentionally **out of scope** — that needs `ShotTimingController` plumbing and is tracked in #1161.

## Test plan

- [x] Worktree build green via Qt Creator (0 errors, 0 warnings)
- [x] `pourControlFromProfileJson_derivesFromLongestFrame` — empty/malformed/no-steps → omitted; D-Flow/Q flow recipe → `flow`; pressure recipe → `pressure`; short trailing decline frame does not flip; string-valued `seconds`.
- [x] `withStopAtWeightNote_appendsNoteOnlyWhenWeightAndRecipePresent` — no-op on empty recipe / non-positive target; exactly one note appended otherwise.
- [x] DB-level: `dialInSessions_hoistsPourControlWhenUniform_elsePerShot` (uniform → hoisted to context, absent per-shot; mixed → per-shot, not hoisted) and `bestRecentShot_emitsPourControlAndTargetWeightFromRecipe`.
- [x] End-to-end verified live on the running MCP against real D-Flow/Q shots (correct `pourControl`, hoisting, and recipe stop-at-weight note in `dialing_get_context`).
- [ ] Reporter (@fredphoesh) to sanity-check advice quality on a D-Flow conversation via MCP

## Review follow-ups (addressed)

Applied from the automated code review on this PR:
- MCP path now gates the stop-at-weight note on the **shot's** stored target (`sd.targetWeightG`), matching the advisor — previously used the currently-loaded profile's target, which could drift for historical shots.
- Narrowed the `withStopAtWeightNote` "cannot drift" comment to the two structured profile-block surfaces (prose multi-shot history blocks intentionally render recipes without the note).
- Added the missing DB-level integration tests and the `withStopAtWeightNote` unit test (above).

### DIFF
--- a/src/ai/dialing_blocks.cpp
+++ b/src/ai/dialing_blocks.cpp
@@ -61,14 +61,13 @@ QJsonObject shotToJson(const ShotProjection& shot,
     h["doseG"] = shot.doseWeightG;
     h["yieldG"] = shot.finalWeightG;
     h["durationSec"] = shot.durationSec;
-    // Issue #1158: control-mode provenance so the LLM doesn't read
-    // yieldG / durationSec as dial-in feedback on stop-at-weight +
-    // flow-controlled shots. Paired with `targetWeightG` (added below
-    // when set) this lets shotAnalysisSystemPrompt()'s stop-at-weight
-    // rule fire.
-    const QString pourControl = DialingBlocks::pourControlFromProfileJson(shot.profileJson);
-    if (!pourControl.isEmpty())
-        h["pourControl"] = pourControl;
+    // Issue #1158: `pourControl` is NOT emitted here. It is a
+    // profile-family property, so buildDialInSessionsBlock hoists it to
+    // the session `context` when every shot in the session shares it
+    // (the common case → zero per-shot repetition) and only falls back
+    // to a per-shot field when a session genuinely mixes flow/pressure
+    // variants. Same dedup discipline as the grinder/bean identity
+    // hoist above.
     h["enjoyment0to100"] = shot.enjoyment0to100 > 0
         ? QJsonValue(shot.enjoyment0to100)
         : QJsonValue(QJsonValue::Null);
@@ -158,9 +157,27 @@ QJsonArray buildDialInSessionsBlock(QSqlDatabase& db,
         const DialingHelpers::HoistedSession hoisted =
             DialingHelpers::hoistSessionContext(identities);
 
+        // Issue #1158: pour control mode per shot, from each shot's own
+        // recipe. Hoist to session context when uniform (the common
+        // case — a session is usually one profile), emit per-shot only
+        // when a session mixes flow/pressure variants of the same kbId
+        // family (e.g. D-Flow/Q vs Damian's LM Leva). "" (no usable
+        // recipe) breaks uniformity so we never assert a value we
+        // didn't derive.
+        QStringList pourControls;
+        pourControls.reserve(ordered.size());
+        for (const ShotProjection& s : ordered)
+            pourControls.append(pourControlFromProfileJson(s.profileJson));
+        const bool pourControlUniform =
+            !pourControls.first().isEmpty()
+            && std::all_of(pourControls.cbegin(), pourControls.cend(),
+                           [&](const QString& p){ return p == pourControls.first(); });
+
         QJsonArray sessionShots;
         for (qsizetype i = 0; i < ordered.size(); ++i) {
             QJsonObject h = shotToJson(ordered[i], hoisted.perShotOverrides[i]);
+            if (!pourControlUniform && !pourControls[i].isEmpty())
+                h["pourControl"] = pourControls[i];
             if (i > 0) {
                 QJsonObject diff = changeFromPrev(ordered[i-1], ordered[i]);
                 h["changeFromPrev"] = diff.isEmpty() ? QJsonValue(QJsonValue::Null) : QJsonValue(diff);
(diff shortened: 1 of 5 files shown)


skialpine:
### Code review

Found 3 issues:

1. PR description is stale and misdescribes the implementation. The body's "## Change" / "## Test plan" still describe `DialingBlocks::pourControlFromPhases()` deriving from the pour phase marker's `isFlowMode`, and a test `pourControlFromPhases_derivesFromLastUsablePhase`. A later commit in this same PR replaced that with the recipe-based `pourControlFromProfileJson()` (test `pourControlFromProfileJson_derivesFromLongestFrame`). Update the PR body before merge so reviewers and git history aren't misled about the approach. (issue: PR description vs. merged code)

2. The shared `withStopAtWeightNote` helper still drifts between the MCP and advisor surfaces. The MCP path gates the stop-at-weight note on the *currently-loaded* profile's target weight (`profileManager->profileTargetWeight()`), while the advisor (`ShotSummarizer::buildCurrentProfileBlock`) uses the *analyzed shot's* target (`summary.targetWeight`). The recipe argument in both is the analyzed shot's stored profile JSON, so for a historical shot whose profile differs from the loaded one the two surfaces render different recipe text — defeating the helper's stated "cannot drift" purpose. Consider passing the shot's own stored target instead.

https://github.com/Kulitorum/Decenza/blob/272aa566cb6378031ae7a002dace6cfbeeb5e705/src/mcp/mcptools_dialing.cpp#L322-L330

3. No DB-level integration coverage for the new `pourControl` paths. The `ShotRow` fixture and `insertShot` never set `profile_json`, so every `buildDialInSessionsBlock` / `buildBestRecentShotBlock` test sees empty recipes; `pourControlFromProfileJson` returns "" and the new hoist-to-context, per-shot fallback, and `bestRecentShot.pourControl` / `targetWeightG` emission are never exercised. A regression in the hoist logic would pass CI. Add a `profile_json` column to the fixture and assert the emitted field.

https://github.com/Kulitorum/Decenza/blob/272aa566cb6378031ae7a002dace6cfbeeb5e705/tests/tst_dialing_blocks.cpp#L78-L91

<details>

- [Worth a look] The `withStopAtWeightNote` comment "the MCP and advisor surfaces cannot drift" overstates scope — two prose-history recipe sites (`aimanager.cpp` ~1210, `shotsummarizer.cpp` ~1026) render recipes via `describeFramesFromJson` without the note. Narrow the comment to the two structured profile-block surfaces it actually covers. (score: 68)
- [Worth a look] `withStopAtWeightNote` has no unit test, while its sibling `pourControlFromProfileJson` does; add no-op-when-empty, no-op-when-targetWeightG<=0, and positive-text cases next to the existing test. (score: 65)

</details>

Issue #1164, opened 2026-05-15
dialing_get_context payload bloat: full KB catalog sent unconditionally, un-hoisted repetition


skialpine:
Parent: #1147. Split out from the #1158 work (PR #1163) per review feedback: "don't just keep adding data, clean up what we already send."

## Problem

`dialing_get_context` ships a large, partly-redundant payload on every call. Token cost compounds across a multi-turn dial-in conversation. Concrete findings from live `dialing_get_context` responses (v1.7.5 / this branch, real shots):

### 1. Full cross-profile KB catalog sent even without `includeFullKnowledge` (dominant cost)

The tool description states the default returns "only the current profile's curated KB entry (~1 KB)" and that the full system prompt + reference tables + cross-profile catalog (~18 KB) are opt-in via `includeFullKnowledge: true` and "redundant on later turns."

Observed: every `dialing_get_context` call **without** `includeFullKnowledge` returned the entire `profileKnowledge` block — the full UGS chart, canonical + inferred UGS tables, "Cross-Profile Grind Ordering", "Profile families", "Other-profile parameter discipline", etc. (~18 KB). Either the gating is broken or the description is stale. This is by far the largest single contributor and is sent on every turn.

### 2. `grinderCalibration.profiles[]` — ~33 rows every call

The block lists every KB profile with `rgs`/`ugs`/`source` (~33 entries) on every call, regardless of which profile the conversation is about. Candidate for trimming to the relevant/nearby profiles or gating like the KB catalog.

### 3. Un-hoisted per-shot repetition in `dialInSessions`

`dialInSessions[].shots[]` repeats `profileName`, `targetWeightG`, `temperatureOverrideC` identically on every shot in a session, while grinder/bean identity (and, after #1158, `pourControl`) are hoisted to `session.context` and emitted per-shot only on change. The same `hoistSessionContext`-style dedup should cover these fields for consistency.

## Suggested direction

- Fix (or document + enforce) the `includeFullKnowledge` gate so the ~18 KB catalog is opt-in as designed; default to the current profile's KB entry only.
- Scope `grinderCalibration.profiles[]` to relevant profiles, or make it opt-in.
- Extend session-context hoisting to `profileName` / `targetWeightG` / `temperatureOverrideC`.

## Not in scope

`pourControl` provenance and the stop-at-weight recipe note (#1158 / PR #1163) — that work already follows the hoist discipline; this issue is about the pre-existing payload.

skialpine:
[...]

PR #1166, opened 2026-05-15
AI: refer to shots by date/time, never the numeric DB id (#1162)

### LINKED ISSUE #1162: Shots referenced by opaque numeric DB id instead of date/time (MCP responses + web Shot History)
Parent: #1147.

## Problem

The AI cited "shot 5188", which the reporter could not locate anywhere. CLAUDE.md's MCP conventions already mandate ISO timestamps over Unix time, but the per-shot *identifier* the LLM echoes is the numeric DB id, which has no user-facing counterpart in the web Shot History.

## Evidence (from #1147)

*"shot 5188? there is no such number on my decenza shot history on the web app… Claude and Shot History on the web should give date/time."*

## Proposed direction (for discussion)

- System prompt: instruct the model to refer to shots by local date + time, not the numeric id.
- Optionally include a human-readable `shotLabel` (date + time) alongside the id in MCP payloads so the model has an unambiguous handle to echo.
- Web Shot History UX: show date/time as the primary shot handle.

## Relevant code

`src/ai/shotsummarizer.cpp` (system prompt), `src/mcp/mcptools_shots.cpp`, `src/network/shotserver_shots.cpp` + web shot history UI.

skialpine:
## Summary

Fixes #1162 (parent #1147). An external AI cited *"shot 5188"*; the reporter couldn't find that number anywhere because Shot History and every user-facing surface key shots by **date/time** — the numeric `id` is an internal DB primary key with no user-facing counterpart. Worse, the shared system prompt actively taught the wrong behavior (*"...on shot 882 — start there"*).

Option A from discussion (prompt/instructions only — no new payload fields, no payload re-bloat after #1164). Lands on **both** surfaces:

- **`shotsummarizer.cpp`** — fixed the bad `shot 882` example and added an explicit standalone rule under `## How to Read Structured Fields`: cite shots by local date/time, use `id` only as an opaque tool argument. Reaches the **in-app advisor always**, and **MCP** via `includeFullKnowledge`.
- **`mcpserver.cpp`** — added the MCP server-level **`instructions`** string to the `initialize` result. The client retains it for the whole session and folds it into its system prompt. Since #1164 made the full system prompt opt-in, this handshake string is the **only always-present MCP carrier** — beverage-agnostic, every tool, zero per-call payload cost.
- **`MCP_SERVER.md`** — documented the date/time-not-id convention and why there are two carriers.

**No web Shot History change** — investigation showed `shotserver_shots.cpp` already renders `shot-date` (date/time) as the visible handle and never surfaces the numeric id. The third proposed direction in the issue was already satisfied; the bug was purely the AI echoing the id.

## Why two carriers (not drift)

The authoritative rule is one block of prose. The in-app advisor always gets the full system prompt (cache_control-wrapped). MCP clients get the same intent via the handshake `instructions` (always) and the full rule when they fetch `includeFullKnowledge` (the #1164 guardrail already directs them to do this once per session). Same wording, two delivery points — no behavioral divergence.

## Test plan

- [x] Builds clean via Qt Creator (0 errors / 0 warnings).
- [x] No code/test depends on the old `shot 882` example string.
- [ ] Live MCP: `initialize` result contains the `instructions` field with the date/time rule.
- [ ] Advisor dry-run: `systemPromptUsed` contains the new rule and no longer contains `shot 882`.
- [ ] `dialing_get_context` with `includeFullKnowledge: true`: profileKnowledge carries the rule.

cc @fredphoesh — your report (#1147).


skialpine:
### Code review

Found 2 issues:

1. The MCP `instructions` field is added to the `initialize` result unconditionally, with no protocol-version gate. `instructions` was introduced in spec revision **2025-03-26** and is absent from the `2024-11-05` `InitializeResult`. This file already establishes the version-gating pattern — `buildToolCallResponse` gates `structuredContent`/`resource_link` on `protocolVersion >= "2025-06-18"` with the explicit comment *"Strict 2024-11-05 clients reject the response"* — and `supportedProtocolVersions()` still negotiates `2024-11-05`. `negotiatedVersion` is already in scope here. Gate it: `if (negotiatedVersion >= QStringLiteral("2025-03-26")) result["instructions"] = ...`.

https://github.com/Kulitorum/Decenza/blob/d105a7c623ee3c0d071c25a01defc72d52770c47/src/mcp/mcpserver.cpp#L692-L710

Established precedent in the same file:

https://github.com/Kulitorum/Decenza/blob/d105a7c623ee3c0d071c25a01defc72d52770c47/src/mcp/mcpserver.cpp#L283-L288

2. No automated test asserts the new `instructions` field. `tst_mcpserver_protocol.cpp`'s `initializeNegotiatesRequestedVersion` only checks `protocolVersion`. Once the gate from issue 1 is added, extend an initialize test to assert `instructions` is present for `>= 2025-03-26` and absent for `2024-11-05`, matching this file's existing version-gate assertion pattern.

https://github.com/Kulitorum/Decenza/blob/d105a7c623ee3c0d071c25a01defc72d52770c47/tests/tst_mcpserver_protocol.cpp#L180-L196

Both issues are narrow and contained; the prose/system-prompt changes in `shotsummarizer.cpp` and the `MCP_SERVER.md` docs are correct and need no changes.

skialpine:
Addressed the review in ec450917:

1. **Version-gate (score 85)** — `result["instructions"]` is now wrapped in `if (negotiatedVersion >= QStringLiteral("2025-03-26"))`, mirroring the `buildToolCallResponse()` gate. Strict `2024-11-05` clients no longer receive the field.
2. **Missing test (score 75)** — added data-driven `initializeInstructionsGatedByVersion`: asserts `instructions` present for `2025-11-25` / `2025-06-18` / `2025-03-26` (and the unsupported→`2025-11-25` case) and **absent** for `2024-11-05`, plus that it carries the date/time rule. 34/34 `tst_McpServerProtocol` pass.

Also fixed two lower-scored-but-real items rather than leave them:

3. **Handshake instruction completeness (score 35)** — the `instructions` string now names where shots appear (`dialing_get_context` `dialInSessions`/`bestRecentShot`, `shots_list`), so the always-present carrier is self-complete and matches the fuller `shotAnalysisSystemPrompt` rule.
4. **Carrier wording precision (score 25)** — corrected the code comment and `MCP_SERVER.md`: `ai_advisor_invoke` always carries the full system prompt, so the accurate claim is that `instructions` is the only carrier for clients that call *neither* `ai_advisor_invoke` *nor* `dialing_get_context` with `includeFullKnowledge`.

The one remaining flagged item — the `MCP_SERVER.md` "(`## How to Read Structured Fields`)" parenthetical — was verified to be a **false positive**: the new paragraph is inside that section's string literal, so the reference is accurate. No change made there.

Builds clean via Qt Creator.

PR #1167, opened 2026-05-15
AI: capture + surface stoppedBy so manual stops don't pollute dial-in (#1161)

### LINKED ISSUE #1161: Manually-stopped shots surfaced as dial-in data with no stop-reason flag

skialpine:
Fixes #1161 (parent #1147); also delivers the producer-side `stoppedBy` that #1158 proposed but deferred.

## Problem

Manually-stopped shots have arbitrary yield. The MCP/advisor surfaced that yield as a dial-in outcome, so the AI diagnosed "inconsistent yield"/grind problems that were really just the user hitting Stop — the reporter's Rao Allongé at 71g and 89g on a 90g target.

## Producer (capture + persist)

- **Migration 17** adds `shots.stopped_by` — additive `TEXT DEFAULT ''`, NULL-tolerant (the dev/fake-shot path and any unknown reason bind a null `QString`; reads coalesce to `""`).
- `stoppedBy` threaded through `ShotSaveData` / `ShotRecord` / `ShotProjection`, the INSERT/SELECT/`convertShotRecord`/serialize paths, and `loadRecentShotsByKbIdStatic` (the dial-in loader).
- Classified in `MainController::onShotEnded`, precedence: **SAW** (`ShotTimingController::wasSawTriggered`) → `"weight"`; **SAV** (`MachineState::wasVolumeStopped`) → `"volume"`; else QML's already-authoritative `stopReason` → `"manual"`; else `"profileEnd"`.
- QML already resolves a comprehensive `stopReason` for its overlay across ~15 entry points. One new `onStopReasonChanged` handler forwards it to `MainController::reportShotStopReason` — **zero changes to the existing stop sites**.

### Honest limitation

The DE1 BLE protocol has no "who stopped me" byte: a stop from the DE1's **own hardware button** is indistinguishable from the profile completing naturally. Both collapse to `"profileEnd"`. The consumer rule covers this: an absent/`profileEnd` stop with `yieldG` well below `targetWeightG` is treated like `"manual"` — which is exactly the reporter's evidence.

## Consumer (discount the arbitrary yield)

- `dialInSessions[].shots[]`, `bestRecentShot`, and `shots_list` **sparsely** emit `stoppedBy` (`manual`/`weight`/`volume`; `profileEnd`/`""` omitted to stay lean per #1164). `shots_get_detail` carries it via the projection.
- `shotAnalysisSystemPrompt` gains a rule beside the #1158 SAW rule: a `"manual"` shot's yield/ratio/duration are user-chosen, **not** extraction outcomes — don't diagnose grind/ratio from them; absent `stoppedBy` + sub-target yield is treated the same; `weight`/`volume` yields are mechanical but pressure/duration/taste stay valid.
- `MCP_SERVER.md` data-conventions note added.

## Bug found + fixed during testing

The column was first `NOT NULL`; the dev/fake-shot path binds a null `QString`, which would have failed every simulated save (and surfaced as cascade test failures). Changed to NULL-tolerant.

## Test plan

- [x] `tst_dialing_blocks`: new sparse-emit case (manual emitted, profileEnd omitted, bestRecentShot manual) + fixture `stopped_by` column.
- [x] `tst_dbmigration`: bumped to schema **17**, pins the new column.
- [x] **2071/2071 tests pass**, clean Qt Creator build (0/0).
- [ ] Live MCP (needs app relaunched on this build): `dialing_get_context` / `shots_list` show `stoppedBy` for manual/weight shots and omit it for profile-completed ones; advisor no longer reads a manually-stopped yield as a dial-in signal.

cc @fredphoesh — your report (#1147).


skialpine:
### Code review

Found 5 issues:

1. **SAW shots are misclassified as `profileEnd` in back-to-back dial-in (the core #1161 scenario).** `wasSawTriggered()` returns `m_sawTriggeredThisShot`, which every SAW-completion path clears *before* emitting `shotProcessingReady` (`onSettlingComplete` clears at :489, emits via `qScopeGuard` at :501; the `startShot()` settling-cancel path clears at :78, emits at :83). The connection is direct/same-thread, so `onShotEnded` runs with the flag already false — the C++ "ground truth" guard is dead. In the single-shot flow this is masked because QML's `onTargetWeightReached` set `m_pendingStopReason="weight"`. But when the user starts the next shot while the previous SAW shot is still settling, `onEspressoCycleStarted` clears `m_pendingStopReason` (:1674) *and* `startShot()` clears the SAW flag, then the prior shot is saved — both signals gone → a genuine SAW shot is persisted as `profileEnd`, exactly the dial-in pollution this PR is meant to prevent. Confirmed by 3 independent review passes. Fix: snapshot the SAW reason at `endShot()`/`onSawTriggered` time into a flag that survives until the save consumes it.

https://github.com/Kulitorum/Decenza/blob/6a2431d7568f085a2210bc816ee9a890f4ddc16f/src/controllers/maincontroller.cpp#L2022-L2035

https://github.com/Kulitorum/Decenza/blob/6a2431d7568f085a2210bc816ee9a890f4ddc16f/src/controllers/shottimingcontroller.cpp#L486-L501

2. **`shots_get_detail` emits `stoppedBy: "profileEnd"` / `""`, contradicting the documented "omitted" contract.** `ShotProjection::toVariantMap()` writes `stoppedBy` unconditionally, so `shots_get_detail` (→ `toJsonObject`) returns `"profileEnd"`/`""` for completed/pre-migration shots — but `MCP_SERVER.md` and the new `shotAnalysisSystemPrompt` rule tell the AI `stoppedBy` is *absent* when a shot ran to completion (absence is the documented fallback trigger). Every other emit site sparse-guards `manual`/`weight`/`volume`. Apply the same guard in `toVariantMap()` (or document the raw value).

https://github.com/Kulitorum/Decenza/blob/6a2431d7568f085a2210bc816ee9a890f4ddc16f/src/history/shotprojection.cpp#L37-L40

3. **The `"volume"` sparse-emit arm is never tested.** `dialInSessions_and_bestRecentShot_surfaceStoppedBySparsely` exercises `manual`/`profileEnd`/`weight` but not `volume`, which appears in all three allowlists (`dialing_blocks.cpp:122`, `:337`, `mcptools_shots.cpp:230`). Removing it from any would pass CI — the exact emit-path-regression blind spot raised in the #1163 review. ~5-line addition to the existing test.

https://github.com/Kulitorum/Decenza/blob/6a2431d7568f085a2210bc816ee9a890f4ddc16f/tests/tst_dialing_blocks.cpp#L1668-L1672

4. **Classification comment omits the 4th branch.** The comment documents precedence as SAW → SAV → `manual` → else `profileEnd`, but the code has a 4th branch (`m_pendingStopReason == "weight"` → `weight`) immediately below — which (per issue 1) is actually the branch that does the real work in the normal SAW flow. The comment should document it.

https://github.com/Kulitorum/Decenza/blob/6a2431d7568f085a2210bc816ee9a890f4ddc16f/src/controllers/maincontroller.cpp#L2012-L2034

5. **Migration 17 comment: "DEFAULT '' back-fills existing rows" is mechanistically wrong for SQLite.** `ALTER TABLE ADD COLUMN ... DEFAULT ''` does not rewrite existing rows; SELECT returns the default *virtually*. The observable read (`""`) is correct, but "back-fills" implies a physical write and could mislead anyone reasoning about row storage / `IS NULL`.

https://github.com/Kulitorum/Decenza/blob/6a2431d7568f085a2210bc816ee9a890f4ddc16f/src/history/shothistorystorage.cpp#L912-L919

Issue 1 is the substantive one — it breaks the feature in its primary use case and is not caught by the unit test (which inserts `stopped_by` directly, bypassing the runtime classification).

<details>

- [OK to skip] `machinestate.h` `wasVolumeStopped()` comment says "reset at the next espresso cycle start" — actually resets at the Preinfusion transition, not `EspressoPreheating` entry where `espressoCycleStarted` fires; primary "valid at shot-end" claim is correct (score: 70)
- [Worth a look] `maincontroller.h` `reportShotStopReason` comment says "C++ SAW/SAV state is ground truth and overrides this" — undersells QML's `weight` path and is additionally misleading given issue 1; will need updating when issue 1 is fixed (score: 65)

</details>

skialpine:
Addressed all review findings in 8fcf0852:

1. **SAW misclassification (score 75) — fixed.** `wasSawTriggered()` now reads `m_stopAtWeightTriggered` (set in `onSawTriggered`, reset only in `startShot()` at line ~97, which runs *after* the settling-cancel save emit at line ~83) instead of `m_sawTriggeredThisShot` (cleared by `onSettlingComplete`/cup-removal/settling-cancel before `shotProcessingReady`). Valid at every `onShotEnded` save path. Added two `tst_settling` regression tests — normal-settle and settling-cancel/back-to-back — that fail on the old code.
2. **`shots_get_detail` contract (score 85) — fixed.** `ShotProjection::toVariantMap()` and `loadRecentShotsByKbIdStatic` now sparse-guard `manual`/`weight`/`volume`, matching every other emit site and the documented "omitted on completion" contract. The `profileEnd→""` round-trip collapse is harmless (consumers treat both as the fallback).
3. **`volume` arm untested (score 75) — fixed.** Added a `volume` shot to `dialInSessions_and_bestRecentShot_surfaceStoppedBySparsely`.
4. **Classification comment (score 80) — fixed.** Now documents all 5 precedence branches including `m_pendingStopReason == "weight"`.
5. **Migration-17 comment (score 75) — fixed.** Reworded: SQLite does not rewrite existing rows on `ADD COLUMN`; the `DEFAULT ''` is returned virtually by SELECT (no physical back-fill).

Below-threshold (also fixed): `wasVolumeStopped()` reset-timing comment clarified; `reportShotStopReason` precedence comment corrected.

2073/2073 tests pass (incl. the 2 new SAW regressions), clean Qt Creator build. Live MCP verification still pending an app relaunch on this build.

PR #1177, opened 2026-05-16
fix: preserve shot history scroll position when returning from a shot

skialpine:
## Summary

Reviewing shots one-by-one was broken: open a shot from the middle of Shot History, tap the back arrow, and the list jumped back to the top instead of staying where you were — so you couldn't easily look at the next shot.

Root cause: `ShotHistoryPage`'s `StackView.onActivated` fires on **every** return (including popping back from `ShotDetailPage`) and unconditionally called `loadShots()`, which both reset `contentY = 0` **and** truncated the model back to the first page (50 rows). So if you'd infinite-scrolled past row 50, even the raw scroll offset couldn't be restored.

## Changes (`qml/pages/ShotHistoryPage.qml`)

- Added `_initialized` / `_pendingRestoreContentY` page properties.
- `StackView.onActivated` now only does the full top-reset `loadShots()` on the **first** activation. Returning from a pushed child page calls a new `reloadPreservingScroll()`.
- `reloadPreservingScroll()` re-queries the **same number of rows** that were already loaded (not just the first page), so the model isn't truncated, then restores `contentY` via `Qt.callLater` (clamped to valid range) once the list has re-laid-out.
- Re-querying (rather than skipping the reload) keeps the list fresh when a rating/notes/badge was edited on the detail or post-shot-review page.

Filter / search / sort changes and deletes still reset to the top as before (they call `loadShots()` directly, unchanged).

## Test plan

- [ ] Open a shot from mid-list → back arrow → list stays where you were, next shot visible.
- [ ] Scroll deep (past the first 50 rows, infinite-scroll loaded more) → open a shot → back → position still preserved.
- [ ] Edit a rating/notes on the detail or post-shot-review page → back → the row reflects the change.
- [ ] Change filter/search/sort, and delete a shot → list still resets to the top.

### DIFF
--- a/qml/pages/ShotHistoryPage.qml
+++ b/qml/pages/ShotHistoryPage.qml
@@ -29,6 +29,12 @@ Page {
     property int filteredTotalCount: 0
     property bool _waitingForShotLoad: false
 
+    // First activation does a full reset-to-top load. Subsequent activations
+    // (returning from a pushed child page like Shot Detail) re-query the same
+    // window and restore the scroll position instead of jumping to the top.
+    property bool _initialized: false
+    property real _pendingRestoreContentY: -1
+
     // Wait for async loadShotWithMetadata to complete before popping
     Connections {
         target: MainController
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 1 issue:

1. `loadShots()` does not clear `_pendingRestoreContentY`, so a stale scroll-restore can be applied to an unrelated full refresh. After returning from a child page, `reloadPreservingScroll()` sets `_pendingRestoreContentY` and issues an async filtered request. If the user then changes the search/sort or clears the filter (or a batch delete fires `onShotsDeleted`) before that result arrives, `loadShots()` issues a new request and resets `contentY = 0`. The in-flight reload's result is dropped by the storage serial guard (`if (serial != m_filterSerial) return;` in `src/history/shothistorystorage_queries.cpp`), but `_pendingRestoreContentY` survives and is consumed by the `loadShots()` result in the `onShotsFilteredReady` non-append branch — jumping the freshly-filtered (often short) list to the old deep scroll offset instead of the top. Fix: add `_pendingRestoreContentY = -1` to `loadShots()` (it is the explicit reset-to-top entry point and should always win over an in-flight restore).

https://github.com/Kulitorum/Decenza/blob/31483bcd7eeaa1a1962b7bb68889751d5e90759d/qml/pages/ShotHistoryPage.qml#L104-L114

<details>

- [Worth a look] The pre-existing comment `// loadShots result (full refresh)` in the `onShotsFilteredReady` non-append branch is now also reached by `reloadPreservingScroll()` (and contains the new `_pendingRestoreContentY` restore block). Still technically accurate but no longer captures the scroll-preserving caller — a one-line tweak in the same handler the fix above touches (score: 55).

</details>

Issue #1183, opened 2026-05-17
KB as structured data (JSON): eliminate the prose-vs-code expert-band duplication


skialpine:
## Problem

Each cited expert-band fact currently lives in **two** places:

- as **prose** in the KB markdown for the LLM — e.g. `resources/ai/profile_knowledge.md` / `docs/PROFILE_KNOWLEDGE_BASE.md` ("grind for a pressure peak between 6 and 9 bar")
- as a **hardcoded C++ row** for the deterministic detector — `kBands` in `ShotSummarizer::expertBandForKbId` (`src/ai/shotsummarizer_kb.cpp`): `pressureBand(6.0, 9.0, "[SRC:...]", "high")`

There is no single source of truth for the band; the same cited number is authored twice and can drift. This will grow as Phase C/D and future lever rows are added.

## Why it's structured this way today

The `flag-off-expert-band-in-shot-summary` design deliberately kept bands in C++ and **out** of the KB, with the recorded rationale: *"seam-free — no `Expected*:` grammar; the table is shipped code, resolved fresh every call (recompute-on-load)."* That objection exists **only because the KB is markdown** — adding band data to markdown would mean bolting more hand-rolled grammar onto it. `loadProfileKnowledge()` already hand-parses `UGS:` / `AnalysisFlags:` / `Also matches:` / `Skip-Catalog:` from the markdown, which is fragile (citation/parsing subtleties have surfaced in prior PR reviews).

## Proposed direction to explore

Move the KB to **structured data (JSON)** — likely a hybrid: structured fields (identity/aliases, UGS, AnalysisFlags, **expert bands**) as JSON, with prose retained as a field for the LLM. This:

- dissolves the original "don't add grammar to markdown" rationale (JSON parsed by `QJsonDocument`, schema-checkable)
- makes `expertBandForKbId` read bands from the KB resource instead of a duplicated C++ table (single source; recompute-on-load preserved — the qrc resource ships with the binary just like code)
- replaces the fragile homegrown markdown structured-field parser

### Cautions / scope

- The KB's primary job is feeding **prose** to an LLM (`buildProfileCatalog`, dial-in / cross-profile reference blocks, AI prompt context). A naive md→json must not make prose worse to author/read — hence "hybrid", which is a real design exercise.
- Large, cross-cutting change touching every KB consumer. Needs its own `/opsx:explore` + openspec change — NOT a bolt-on.

## Note

[...]

PR #1182, opened 2026-05-17
feat(ai): Phase C — add Londinium expert-band row; Adaptive v2 gate-failed

skialpine:
## Summary

Phase C of `flag-off-expert-band-in-shot-summary`, processed in the authoritative C1 order, each arm behind its own community-shot gate (C2). All sources verified qualified: `decent-guide` = *Your First Decent Espresso* (official Decent guide), `*-video` = official Decent blog transcripts, `adaptive-adastra` = the Adaptive author's blog. No fabrication.

### Per-arm outcomes

| Arm | Gate | Result |
|---|---|---|
| **Adaptive v2** | FAIL | **Not added.** The `## Adaptive v2` KB section explicitly forbids flagging its by-design variable pressure (*"6.8 bar coarser grind … up to 9 fine … DO NOT flag variable pressure curves"*); the two authoritative sources even conflict (guide target 8–9 vs KB envelope 6.8–9). Community gate (20 shots, 12 users) showed no clean separation. The pre-registered "highest false-flag risk" arm, correctly left absent per C2. |
| **Londinium** | PASS | **Added.** `decent-guide`: *"9 bar peak with gradual decline; too fine: pressure over 9 bar; too coarse: pressure crash."* Non-adaptive. Community gate (20 shots, 5 users) clusters on the cited target (extraction-peak median 9.0); band **[8.0, 9.0]** partitions verbatim onto the guide — 9 silent on-target, 5 fire >9 (cited too-fine), 6 fire <8 (cited crash/coarse). |
| **Rao/Blooming Allongé** | BLOCKED | Its ~4.5 ml/s rail is a one-sided flow *floor* → must wait on the recorded Phase-B prerequisite (`ExpertBand` → `std::optional` + drop `None` reshape). Not gated yet. |
| **E61** | DEFERRED | KB describes it as flat-9-bar constant pressure, not a flow-target profile; the proposed "flow ~4" rail isn't a clean deterministic signal. |

### Changes

- `src/ai/shotsummarizer_kb.cpp`: one row — `{ "Londinium", PressurePeak 8–9, [SRC:decent-guide], medium }`. Canonical key is the standalone `## Londinium`; the KB Note keeps Damian's LRv2/LRv3 in the D-Flow family.
[...]

PR #1181, opened 2026-05-17
feat(ai): Phase B — add the A-Flow expert-band row (6–9, [SRC:aflow-repo])

skialpine:
## Summary

Adds the A-Flow expert-band row and **withdraws the earlier A-Flow STOP** (closed PR #1180). The STOP was wrong on two counts:

1. It framed the gate around user enjoyment ratings — not the criterion. Validity = qualified Decent/author guidance + faithful measurement.
2. It trusted one lenient community rater (Greg K — all 8 "rated 68–100" shots) to label limiter-pegged shots "good," making a correctly-behaved band look like it false-fired.

## Rating-independent validation (the real population)

The visualizer.coffee `A-Flow / default-medium` "rated 68–100" set was a single rater. The true population is the unfiltered set: **20 shots, 4 users (Mark Paulson, David Watson, Cris Cable, dniprodd), 0 rated**. The built-in's Flow-Extraction frame carries a **10.0 bar pressure limiter**. The 6–9 band partitions this population cleanly:

| Band outcome | Shots | Meaning |
|---|---|---|
| **SILENT 6–9** | 7/20 | on Janek's target, off the limiter — correctly left alone |
| **FIRE >9** | 11/20 | grind too fine → pegged the 10-bar limiter (9/11 confirmed); "10-bar A-Flow is normally terrible" — correctly fired |
| **FIRE <6** | 2/20 | too coarse — legitimate dial-in miss |

The limiter peg is the **mechanism that corroborates** why >9 is bad — exactly D1 (limiter corroborates, band primary), not a rival rail. Measurement-window is moot: pegged reads ~10 either way → fires; on-target ~6–7 → silent.

## Changes

- **B1** — single canonical `A-Flow` row: `PressurePeak 6–9`, `[SRC:aflow-repo]` (Janek's editor dial-in guidance step 1, verbatim, PROFILE_KNOWLEDGE_BASE.md:240), confidence `medium`. All shipped A-Flow profiles canonical-key to the one `## A-Flow` section (gold-pair-style dedup).
- **B2** — `tasks.md` records the corrected rationale; new slice test `expertBand_aflow_pegged10Fires_onTarget7Silent` pins the partition (~10 bar fires above-band/observation/no grind direction/`expertBandDeviation`; ~7 bar silent).

Jeff's own 53-shot library A-Flow batch is pre-fix/stale (predates #808/#1173) and was **not** used as the basis — the multi-user community population is the evidence of record.

## Test plan

[...]

PR #1184, opened 2026-05-17
refactor(ai): reshape ExpertBand — std::optional, validating factories, one-sided bands

skialpine:
**Stacked on #1182 (base = `feat/phase-c-londinium`).** Merge #1182 first; this diff shows only the reshape.

## Summary

Pays down the `ExpertBand` type debt flagged by the #1178 review (type-design-analyzer) and **unblocks Phase C's Allongé arm** (a one-sided flow floor that the old two-sided type literally could not express without fabricating a ceiling).

- **Absence → `std::optional<ExpertBand>`** (`std::nullopt` = strict no-op), never an in-band `Axis::None` sentinel. A constructed `ExpertBand` is always a valid cited band; `{None,5,8}` / `{PressurePeak,0,0}` garbage states are no longer representable.
- **`lo`/`hi` each `std::optional<double>`** → one-sided rails are first-class. `ExpertBand::flowFloor(min,…)` expresses Allongé's "reach ~4.5 ml/s, no ceiling" with no fabricated bound.
- **Validating factories** `pressureBand` / `flowBand` / `flowFloor` (Q_ASSERT invariants) are the only construction path.
- `Axis::None` dropped. File-scope struct + `using` alias kept deliberately (shared by ShotAnalysis + ShotSummarizer; default arg now `= std::nullopt`, no longer the NSDMI workaround).
- Detector emits **band-shape-aware prose**: two-sided keeps the EXACT prior wording; floor/ceiling name only the bound they have.
- `expertBandForKbId` → `std::optional`; the 4 KB rows migrate to the factory; `AnalysisInputs.expertBand` / `analyzeShot` / `generateSummary` / `shot_eval` take `std::optional`.

Behavior-identical for existing two-sided bands. The prose-vs-code band duplication is separately tracked in #1183 (out of scope here).

## Test plan

- [x] App build 0 errors / 0 warnings (Qt Creator)
- [x] Full Qt suite **2087 passed, 0 failed** (same count — behavior-preserving; existing `expertBand_*` + KB-resolution tests migrated to the optional/factory API)
- [x] `shot_eval` corpus **17/17**; Londinium community partition unchanged through the production path (9 silent / 5 fire>9 / 6 fire<8)
- [x] `openspec validate … --strict` clean


skialpine:
### Code review

Found 2 issues:

1. `flowFloor` — the entire motivation for this reshape (a one-sided floor that unblocks Allongé) — ships with **zero fire/silent tests**. The existing flow-axis test (`expertBand_flowAxis_firesAndSilent_perBand`) uses two-sided `flowBand`, not `flowFloor`; the one-sided detection path (`belowFloor` with `hi == nullopt`) and the `"below the ~X ml/s"` prose branch have no regression lock. This is the same class of gap the prior #1181 review flagged: the new thing the PR exists for is unexercised end-to-end. Recommend an `expertBand_flowFloor_firesAndSilent` test (observed < min−margin fires; observed ≥ min stays silent; text contains `"below the ~"` + `ml/s` and no grind direction).

https://github.com/Kulitorum/Decenza/blob/36378ec9e8417189ee6cfe41d30049673b9d80a3/tests/tst_shotanalysis.cpp#L1711-L1716

2. The comment over-claims enforcement the type does not provide. The header says *"Construct ONLY via the validating factories; the aggregate form is the factories' own implementation detail"* and *"Validating factories — the only supported construction path"* — but `ExpertBand` is an aggregate (no user-declared ctor), so `ExpertBand{ Axis::PressurePeak, 9.0, 6.0, "", "" }` is valid anywhere, and `Q_ASSERT` is compiled out in release, so the invariants are debug-only. So "only path" and "validating" are convention + debug-assert, not type-enforced. The *code* is fine for today's call sites (all compile-time-constant `kBands` rows; the deep-bug scan concurred), but the comment should either be softened to state the real guarantee, or the type tightened (private ctor / `class`, or factories return `std::optional`). (comment accuracy / overclaim — CLAUDE.md: comments must not mislead)

https://github.com/Kulitorum/Decenza/blob/36378ec9e8417189ee6cfe41d30049673b9d80a3/src/ai/shotanalysis.h#L24-L43

<details>

- [OK to skip] The ceiling-only `else` prose branch (`*band.hi` deref) is currently unreachable — no `flowCeiling` factory exists. Logically safe today; a one-line "no flowCeiling factory yet — placeholder" note would help a future implementer, but it's speculative until that factory lands (score: 50)

</details>

No functional bugs: the two-sided firing logic is provably byte-identical to pre-reshape for the gold pair / A-Flow / Londinium, the `flowFloor` path is UB-free (guards short-circuit before `*band.hi`), and the `constFind` rewrite correctly *fixes* a latent default-construction issue the `Axis::None` removal would otherwise have created. Behavior-preserving (suite 2087/0/0, corpus 17/17, Londinium partition unchanged). The #1178 type-design goals (drop `None`, optional-absence, one-sided) are delivered.

Issue #1176, opened 2026-05-16
[Bug] D-Flow error


GCDE-VER1:
### What happened?

Profile set to infuse until 4gms weight in cup. (Usually about 20secs into shot). Instead, moved on to pour at approx 6secs, reporting total weight of 9.6gms in cup. 

[debug.zip](https://github.com/user-attachments/files/27863032/debug.zip)

### Steps to reproduce

Pour D-Flow custom profile

### Decenza version

1.7.5 build 3385 Android

### Platform

Android

### Device / OS version

Samsung A9+ Android 14

### Scale (if applicable)

Half Decent Scale. Latest firmware

skialpine:
Thanks for the report. To make progress I really could use

the D-Flow/GCFlightMode profile json and the shot.json file. 

to get at those go to the app web interface and click into the shot you had issues with and click both of these buttons and attach both files

<img width="476" height="70" alt="Image" src="https://github.com/user-attachments/assets/c1fa3a63-259d-4f7d-94fa-cc9327257ed9" />


If easier, it is possible a visualizer link would be good enough.

skialpine:
actually, it looks like your scale is having BT issues with the tablet. I am going to make a change to the next beta to see if it helps. After the next update can you send another system debug file so I can see if it is working better?

GCDE-VER1:
Thanks. Will do. Can also send json later if that's helpful.

skialpine:
Probably tomorrow morning. Having some issues not breaking things with the possible fix.

skialpine:
New beta release ready to test. Can you please run for an hour and then add the current system debug log to this issue to help me verify the fix. also looking for your verification that things are working better.

Thanks for your help.

skialpine:
did you have a chance to update to the latest build and see if the behavior improved? I would love another debug log to verify the fix.

Thanks much!

GCDE-VER1:
Sorry about delay. Time differences I guess. Updated to 3388 and ran a shot. After about 5 secs into preinfusion screen pops up error message saying no scale and will estimate. Then at approx 12 secs weight appeared again until end of shot. I update beta often and haven't had this problem until last 2 betas.

[debug.zip](https://github.com/user-attachments/files/27936686/debug.zip)

skialpine:
[...]

PR #1185, opened 2026-05-17
fix(ble): adaptive scale connection-priority backoff for dual-HIGH contention

skialpine:
## Summary

On weak Android BT chipsets, requesting `CONNECTION_PRIORITY_HIGH` on **both** the DE1 and scale links starves one of them:

- **#1093** — DE1 `CharacteristicWriteError` cascade a few seconds after the scale (a Eureka Precisa) connects, on a Teclast P80X (Android 9).
- **#1176** — silent scale-notification stall mid-preinfusion that abandoned a D-Flow infuse frame at ~6 s / 9.6 g, on a Galaxy **Tab A8** (`SM-X200`, Unisoc T618, Android 14 — mislabelled "A9+" in the report).

The #1097 mitigation gates the scale-side HIGH request on Android `SDK < 30`, assuming weak chipset ⟺ old Android. The #1176 device disproves that (weak radio, SDK 34), so it still requests dual-HIGH. The SDK proxy is the wrong predictor; capability must be observed at runtime — and capable hardware (e.g. Tab A9+/Snapdragon) must keep its low-latency scale link.

## Approach

Runtime detection on the shared, **scale-agnostic** Qt scale transport (all 14 drivers; no `*scale.cpp` edits):

- **Primary (fast, pre-shot):** a DE1-link fault cluster shortly after the scale requests HIGH — `DE1Transport::de1LinkFault` (write-retry / write-failed / connection-teardown), forwarded through the stable `DE1Device`.
- **Backstop (in-shot):** `WeightProcessor::scaleFeedStalled`, evaluated on the DE1 sample cadence so it fires even while the scale is silent; gated on active+tared extraction so an idle/absent scale never trips it.
- **Recovery:** latch a session skip-HIGH flag and self-reconnect; the scale comes back at platform-default BALANCED via the **existing** skip path (proven good by #1097). Once per session — no reconnect loop.
- Session-only: **no persistence, no new settings, no live renegotiation**. The SDK<30 gate is retired. `ScaleFactory` registers the transport on the `ScaleDevice` base (single chokepoint). Decision logic isolated in a pure, header-only `BlePriorityDetector` with deterministic unit tests.

Capable hardware is unaffected — it never trips the detector and keeps scale HIGH/low-latency.

Full rationale, alternatives, and the (rejected) live-renegotiation/persistence/feed-freshness options are in `openspec/changes/scale-ble-priority-backoff/`.

## Test plan

- [x] Builds clean (Qt Creator); existing test suite green
- [x] New `tst_scaleblepriority` unit tests pass (threshold trigger, post-window ignore, stall backstop, never-armed/idle no-trigger, skip-flag loop guard, disarm)
- [ ] On-device: weak chipset (Tab A8-class) — verify DE1-cluster pre-shot backoff **and** the in-shot stall backstop both reconnect the scale at BALANCED; verify no reconnect loop; verify a fresh app run starts at HIGH again
- [ ] On-device: capable hardware (Tab A9+/Snapdragon) — verify scale stays HIGH, low-latency weight intact, no spurious backoff; idle scale / no scale attached never trips
- [ ] Reporter (#1176) confirmation before closing

Relates to #1093 and #1176 (not auto-closing — pending on-device + reporter confirmation).


skialpine:
### Code review

Found 3 high-confidence issues (all fixed in 6fd287e3):

1. **Critical — backoff was a permanent scale drop, not a reconnect-at-BALANCED** (bug). `triggerScaleBackoff()` → `disconnectFromDevice()` calls `m_controller->disconnect()` (severs controller signals) and, for a fully-discovered scale (`DiscoveredState`), skips the controller disconnect entirely, so `onControllerDisconnected()` never runs and the transport's `disconnected()` is never emitted. The main.cpp scale auto-reconnect is gated on `ScaleDevice::connectedChanged`, so the scale would have been dropped for the rest of the session with no reconnect — worse than the bug being fixed. Fix: emit `disconnected()` explicitly after teardown (single emission; controller signals are already severed).

https://github.com/Kulitorum/Decenza/blob/6fd287e3d1c179ed53f58a63affad76ff824704d/src/ble/transport/qtscalebletransport.cpp#L295-L319

2. **Re-arm starved the DE1-fault threshold on a flapping link** (bug, score 75). `BlePriorityDetector::armWindow()` reset the fault count/window on every (re)connect, so a weak link that flaps (fault → reconnect → fault) never accumulated to the threshold — defeating detection on exactly the hardware it targets; the in-shot stall backstop was the only thing left. Fix: faults now accumulate across reconnects; the window anchors to the first fault and slides, so isolated faults still never trip but a real cluster does.

https://github.com/Kulitorum/Decenza/blob/6fd287e3d1c179ed53f58a63affad76ff824704d/src/ble/transport/bleprioritydetector.h#L35-L60

3. **Counting every transient `write-retry` as a DE1 fault** (bug, score 75). `bletransport.cpp` emitted `de1LinkFault("write-retry")` on each retry, so two independent writes that each retry once then succeed (normal on capable hardware) could trip the threshold-2 backoff. This also diverged from design D1 ("FAILED after retries / ConnectionError"). Fix: only `write-failed` and `controller-error` count.

https://github.com/Kulitorum/Decenza/blob/6fd287e3d1c179ed53f58a63affad76ff824704d/src/ble/bletransport.cpp#L495-L503

A pre-existing duplicate comment block in `onControllerConnected()` (from incremental drafting) was also removed. The CLAUDE.md compliance pass found no rule violations (the `QElapsedTimer`/`setCurrentFrame` time checks are event-driven measurement, not timer-guards).

<details>
<summary>2 below-threshold issues (50-74) — also addressed</summary>

- [Worth a look] Cross-thread `WeightProcessor::scaleFeedStalled` → transport connection relied on AutoConnection; pinned to `Qt::QueuedConnection` explicitly so a future moveToThread reorder can't make it a direct cross-thread call into the BLE controller (score: 50). https://github.com/Kulitorum/Decenza/blob/6fd287e3d1c179ed53f58a63affad76ff824704d/src/main.cpp#L1383-L1386
- [OK to skip] Backoff log line's `%1×` count was always the threshold value (no diagnostic variability); dropped from the message (score: 50).

Dismissed (false positive): claimed scale-type-change race delivering signals to both old and new transport — the transport is `setParent(this)`'d to the scale driver, so it is destroyed synchronously with `physicalScale.reset()` (not via `deleteLater`), and its connections are removed before the new transport is wired.

</details>

skialpine:
Follow-up (commit 7dc83428): **scoped the skip-HIGH backoff to the whole app run, across all scales.**

Previously the skip-HIGH latch lived on the per-scale `BlePriorityDetector`, which is destroyed/recreated on a scale-type change. Since the dual-HIGH contention is a property of the device's BT radio + the DE1 link (not of any one scale — the fix is scale-agnostic; #1093 was a Eureka Precisa, #1176 a Decent), swapping scales mid-run would have wrongly re-paid the detection window on hardware already proven incapable that run.

The latch now lives on the process-lifetime `BLEManager` singleton (`scaleSkipHighPriority()` / `setScaleSkipHighPriority()`, in-memory only). `onControllerConnected()` seeds the detector from it; `triggerScaleBackoff()` sets it. Net behavior: the first scale to detect the conflict latches it for the entire app run — every scale afterward (including one connected after a scale-type change) skips HIGH and comes up BALANCED with no re-detection. Cleared only by an app restart; still never written to disk. openspec proposal/design/spec/tasks updated accordingly.

skialpine:
### Code review (round 2 — re-reviewed after commits 6fd287e3, 7dc83428, d870030b)

The first-round critical bug and the A/B/C/D findings were independently re-verified as **genuinely resolved** (single `disconnected()` emission; anti-starvation accumulate-across-reconnects; write-retry no longer counted; explicit `Qt::QueuedConnection`; app-run BLEManager latch correct and thread-safe — all touched only on the main thread). No new code bugs found. CLAUDE.md: clean.

Found 2 comment-accuracy issues (both fixed in a16db934):

1. The `triggerScaleBackoff()` comment presented the `DiscoveredState` controller-disconnect skip as part of the causal chain for "never routes through `onControllerDisconnected()`", when `m_controller->disconnect()` (signal sever) alone is sufficient in any state. Since this is the regression-prevention comment for the critical permanent-scale-drop fix, the imprecision risked a maintainer removing the explicit `emit disconnected()` for non-`DiscoveredState`. Rewritten to state signal-severing as the sufficient cause (score: 75).

https://github.com/Kulitorum/Decenza/blob/a16db934713aa532d678d76628da195211a0124e/src/ble/transport/qtscalebletransport.cpp#L329-L347

2. Stale "session" wording: after the app-run BLEManager latch (7dc83428) the authoritative scope is "once per app run, all scales", but several comments/logs still said "this session" (meaning per-transport lifetime). Clarified the detector header, `armWindow` doc (incl. why `nowMs` is unused), and the transport log/comment (score: 62).

https://github.com/Kulitorum/Decenza/blob/a16db934713aa532d678d76628da195211a0124e/src/ble/transport/bleprioritydetector.h#L10-L28

<details>
<summary>3 below-threshold issues (35-38) — deliberately not changed, rationale below</summary>

- [OK to skip] armWindow accumulates faults across reconnects → a theoretical false-positive if two genuine `write-failed`/`controller-error` faults straddle an unrelated reconnect within 20 s (score: 35). This is a deliberate anti-starvation tradeoff: resetting on every reconnect (the "fix") reintroduces the prior round's 75-scored starvation regression on flapping weak hardware, and the #1093/#1176 signature is exactly clustered faults across reconnects. False positive is non-sticky (one app run at BALANCED, self-corrects on restart).
- [OK to skip] `disconnectFromDevice()` skips the controller disconnect in `DiscoveredState` (potential deferred-teardown vs 5 s reconnect race) (score: 35). Pre-existing shared teardown used by every scale forget / type-change today; not introduced here; no evidence of GATT-133 from this path. Out of scope for this PR.
- [OK to skip] `de1LinkFault` connection left `AutoConnection` while the adjacent cross-thread `scaleFeedStalled` is pinned `Qt::QueuedConnection` (score: 38). Correct as-is (same-thread → Direct); forcing `Qt::DirectConnection` would be *less* safe (removes Qt's auto-Queued fallback if DE1 ever moves threads). Added a comment documenting the intentional asymmetry instead.

</details>

PR #1191, opened 2026-05-17
feat(shot-review): autosave edits with Undo, replacing manual Save

skialpine:
## Summary

Shot-review edits were frequently lost: the explicit **Save Changes** button only persisted on demand, so navigating away, auto-close, an R2 refractometer reading, or async bean-type auto-fill dropped the change. This replaces the save model with **autosave at every commit point** plus an **Undo** button.

- Every committed edit persists immediately. The in-memory baseline is advanced optimistically (no DB reload on save), so a concurrent edit in another field is never clobbered.
- R2 refractometer readings and async bean auto-fill persist the moment they land (previously lost unless the user manually hit Save).
- **Undo** button (replaces the Save button) reverts the last change, repeatable down to the loaded record.
  - Undo frames are **coalesced per control**: a continuous rating-slider drag (which emits `valueModified` ~10×) is a single Undo, not ten micro-steps.
  - `RatingInput` / dose / out imperatively assign their own `value` during interaction, severing the `value: editX` binding; `applyEditState` re-asserts those controls so Undo visibly restores the UI, not just the stored data.
- Edits are flushed on back / `StackView` deactivate / `Component.onDestruction`; the unsaved-changes confirmation dialog is removed (there is nothing to lose).
- Scope is `qml/pages/PostShotReviewPage.qml` only — the post-shot review / editor page is the sole place shot metadata is edited (ShotDetailPage is read-only and routes here).

## Test plan

- [ ] Edit any field (notes, bean/grinder, dose/out, rating, beverage type, roast level/date), navigate away or let auto-close fire → change is persisted on reopen.
- [ ] Drag the rating slider (e.g. 75→85), tap **Undo** once → rating jumps back to the original value in one step; tap again → no-op once back to loaded state.
- [ ] Adjust dose/out/TDS/EY steppers, then Undo → value visibly reverts (binding re-assert).
- [ ] Take an R2 reading on the review page, navigate away without touching Save → TDS/EY persisted.
- [ ] Pick a roaster whose bean types resolve async → auto-filled coffee field persists.
- [ ] Edit two different fields, Undo twice → each distinct control edit is its own undo frame.
- [ ] Regression: open editor from History → Edit (no crash); verify on macOS (threaded render loop) and one mobile/Android run.

### DIFF
--- a/qml/pages/PostShotReviewPage.qml
+++ b/qml/pages/PostShotReviewPage.qml
@@ -25,19 +25,24 @@ Page {
         }
     }
 
-    // Disconnect refractometer when leaving to save battery (requires physical wake to reconnect)
+    // Flush any pending edit and disconnect refractometer when the page is torn
+    // down. The destruction flush is the safety net that makes edits durable on
+    // *any* navigation path — not just the explicit back button.
     Component.onDestruction: {
+        autosave()
         if (Refractometer && Refractometer.connected) {
             Refractometer.disconnectFromDevice()
         }
     }
 
+    // Also flush whenever the page loses the foreground (a child page pushed on
+    // top, app backgrounded, etc.) so an in-progress text edit can't be lost.
+    StackView.onDeactivating: autosave()
+
     function handleBack() {
-        Qt.inputMethod.commit()
-        if (hasUnsavedChanges) {
-            unsavedChangesDialog.open()
-            return
-        }
+        // Every committed edit is already persisted; just flush a possible
+        // in-progress text field and leave — no confirmation prompt needed.
+        autosave()
         root.goBack()
     }
 
@@ -81,10 +86,8 @@ Page {
                  && postShotReviewPage.autoCloseTimeout < 31
                  && postShotReviewPage.StackView.status === StackView.Active
         onTriggered: {
-            // Auto-close: save silently and exit (user isn't here to answer a prompt)
-            if (postShotReviewPage.hasUnsavedChanges) {
-                postShotReviewPage.saveEditedShot()
-            }
+            // Auto-close: flush any pending edit and exit
+            postShotReviewPage.autosave()
             root.goBack()
         }
     }
@@ -146,6 +149,13 @@ Page {
                 // arrived via R2 before the shot data was ready, or where the DB already
                 // has a non-zero TDS from a previous session).
                 calculateEy()
+                // Establish the autosave baseline now that every edit field
+                // mirrors the loaded record. _editLoaded gates autosave so a
+                // pre-load flush can't write empty metadata over the shot
+                // (hasUnsavedChanges is transiently true before this point —
+                // empty baseline vs. dose defaulted from Settings).
+                _committedState = captureEditState()
+                _editLoaded = true
                 // Quality badges already arrived recomputed in `shot` via
                 // loadShotRecordStatic, which also persists drift to the DB
                 // and emits shotBadgesUpdated when it does. onShotBadgesUpdated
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 6 issues:

1. Undo permanently severs the TDS/EY `value` bindings, so R2 refractometer readings stop updating the UI after the first Undo (regresses this PR's own R2 feature). `applyEditState` imperatively assigns `tdsInput.value`/`eyInput.value`, but `ValueInput` never self-assigns `value` and the tds/ey `onValueModified` handlers don't either — those bindings are intact until Undo breaks them; afterward `editDrinkTds/editDrinkEy` (set by `onTdsChanged`/`calculateEy()`) no longer reach the fields. Fix: don't re-assert tds/ey (binding is live), or use `Qt.binding(...)`.

https://github.com/Kulitorum/Decenza/blob/68b33e29a09f50a1923797292448c5506d312fd5/qml/pages/PostShotReviewPage.qml#L335-L345

2. `autosave()` runs inside `Component.onDestruction` — it calls `Qt.inputMethod.commit()`, `requestUpdateShotMetadata()`, and singleton writes during QML teardown. `StackView.onDeactivating: autosave()` already covers normal exits, so the destruction flush is a redundant last-resort that adds teardown-time risk (notable given the intermittent `QSGRenderThread` fault observed on this branch). Consider dropping it or guarding tightly.

https://github.com/Kulitorum/Decenza/blob/68b33e29a09f50a1923797292448c5506d312fd5/qml/pages/PostShotReviewPage.qml#L30-L36

3. The Visualizer success handlers still `loadShotForEditing()`, which re-runs `onShotReady` and resets `_committedState` but never clears `_undoStack`/`_undoDepth`/`_lastEditKey` — an in-progress edit is clobbered and a later Undo can persist a stale state. This PR removed the identical reload from `onShotMetadataUpdated` for exactly this race; the two Visualizer paths were left in. Fix: update `editShotData.visualizerId` in place (like `onShotBadgesUpdated`) or clear undo state on reload.

https://github.com/Kulitorum/Decenza/blob/68b33e29a09f50a1923797292448c5506d312fd5/qml/pages/PostShotReviewPage.qml#L464-L470

https://github.com/Kulitorum/Decenza/blob/68b33e29a09f50a1923797292448c5506d312fd5/qml/pages/PostShotReviewPage.qml#L181-L186

4. Re-editing the same control after a commit is silently un-undoable. Coalescing keys on `_lastEditKey`: editing a field, blurring, then editing the same field again (no other control touched between) coalesces into one frame, so the intermediate committed value cannot be reverted — contradicting the "repeatable down to the loaded record" contract. Fix: reset `_lastEditKey` on focus-leave/commit so a new interaction opens a fresh frame.

https://github.com/Kulitorum/Decenza/blob/68b33e29a09f50a1923797292448c5506d312fd5/qml/pages/PostShotReviewPage.qml#L350-L363

5. A background-thread DB write is issued on every `onValueModified` of sliders/steppers. Undo frames coalesce, but `saveEditedShot()` (→ a per-call `QThread` SQLite write) is not coalesced — one rating-slider drag spawns ~10+ writes. CLAUDE.md says "Everything else — including debounce — should use event-based flags" (https://github.com/Kulitorum/Decenza/blob/68b33e29a09f50a1923797292448c5506d312fd5/CLAUDE.md). Persist on the coalescing boundary (key change / focus-leave), not per emission.

https://github.com/Kulitorum/Decenza/blob/68b33e29a09f50a1923797292448c5506d312fd5/qml/pages/PostShotReviewPage.qml#L869-L873

6. Comment/maintainability: the `autosave()` comment "A blank/absent key (lifecycle flush) always starts a fresh frame" omits that any differing key also pushes a frame; the `applyEditState` comment names only "RatingInput and the dose/out ValueInputs" though it re-asserts five (incl. tds/ey — which per #1 it shouldn't); and `_undoStack`/`_undoDepth` has no note that the int mirror must move in lockstep with every push/pop/shift.

https://github.com/Kulitorum/Decenza/blob/68b33e29a09f50a1923797292448c5506d312fd5/qml/pages/PostShotReviewPage.qml#L302-L305

<details>

- [Worth a look] After >50 coalesced edits in a session, `_undoStack.shift()` drops the loaded-record baseline, so Undo can no longer reach the original and the `undoLastChange` "Repeatable down to the loaded record" comment becomes false — keep index 0 un-shiftable or qualify the comment (score: 72)
- [OK to skip] New Undo button uses `Rectangle`+`MouseArea` instead of `AccessibleButton` (CLAUDE.md prefers `AccessibleButton`); required `Accessible.*` are all present and it matches the adjacent pre-existing uploadButton (score: 72)

</details>

PR #1196, opened 2026-05-17
fix(logging): suppress spurious Qt "window doesn't exist." message (QTBUG-146779)

skialpine:
## Summary
- Qt emits `window doesn't exist.` at **critical** severity during early startup and window/accessibility teardown ([QTBUG-146779](https://qt-project.atlassian.net/browse/QTBUG-146779)). It is harmless but floods the error log — ~34× in a single session — making real errors hard to spot in `debug_get_log`.
- Drop the exact message in `BtLogFilter`, which sits at the **top** of the `qInstallMessageHandler` chain (installed last, main.cpp:351), so it never reaches the web debug logger, async/persisted logger, or console.
- **Also fixes a latent bug:** `BtLogFilter::install()` was wrapped in `#if defined(Q_OS_LINUX) && !defined(Q_OS_ANDROID)`, making it a no-op on Android/macOS — exactly the platforms where this message appears. `install()` is now cross-platform; the CAP_NET_ADMIN rule stays Linux-gated by its own `#if`.

## Notes
- Exact `==` match (`QLatin1String("window doesn't exist.")`), not `contains`, so it cannot accidentally suppress a legitimate message.
- Remove once the upstream Qt fix lands.

## Test plan
- [ ] Launch on Android; confirm `window doesn't exist.` no longer appears in `de1 debug_get_log` (was ~8–34×/session)
- [ ] Confirm real ERROR/WARN lines still log normally
- [ ] Linux: CAP_NET_ADMIN false-alarm suppression still works (unchanged path)

### DIFF
--- a/src/core/btlogfilter.cpp
+++ b/src/core/btlogfilter.cpp
@@ -16,6 +16,11 @@ void messageHandler(QtMsgType type, const QMessageLogContext& context, const QSt
         return;  // our independent probe says caps are effective; drop the false alarm
     }
 #endif
+    // QTBUG-146779: Qt spuriously emits "window doesn't exist." during early
+    // startup and window/accessibility teardown. Harmless, but it lands at
+    // critical severity and floods the error log. Drop until the upstream fix.
+    if (msg == QLatin1String("window doesn't exist."))
+        return;
     if (g_previousHandler) g_previousHandler(type, context, msg);
 }
 } // namespace
@@ -24,10 +29,8 @@ namespace BtLogFilter {
 
 void install()
 {
-#if defined(Q_OS_LINUX) && !defined(Q_OS_ANDROID)
     if (g_previousHandler) return;
     g_previousHandler = qInstallMessageHandler(messageHandler);
-#endif
 }
 
 } // namespace BtLogFilter
--- a/src/core/btlogfilter.h
+++ b/src/core/btlogfilter.h
@@ -2,17 +2,18 @@
 
 #include <QtGlobal>
 
-// Chains into qInstallMessageHandler to drop Qt's "Missing CAP_NET_ADMIN
-// permission..." bluetooth warning when BleCapability::linuxMissing() is
-// false — Qt prints the warning under conditions that don't always mean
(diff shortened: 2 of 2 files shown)


skialpine:
### Code review

No issues found. Checked for bugs, CLAUDE.md compliance, comment accuracy, git-history regressions, and prior-PR reviewer concerns. The exact-match `==` suppression is correctly scoped (cannot over-suppress; cannot mask a `QtFatalMsg` abort), and broadening `install()` to all platforms is the right fix — the previous Linux-only gate was incidental scoping for the CAP_NET_ADMIN rule, not a deliberate exclusion, and the CAP rule retains its own `#if`. Handler-chain install ordering and header docs are accurate.

<details>

- [OK to skip] Lifecycle asymmetry: `BtLogFilter` is installed last (top of chain) but has no `uninstall()`, while `CrashHandler`/`AsyncLogger` restore at shutdown. Verified **not a reachable use-after-free**: `CrashHandler::uninstall()` swaps the active handler back to `AsyncLogger`, so the spliced-out `BtLogFilter` node is simply never invoked again — `qInstallMessageHandler` is a single pointer, not a live linked list. The pre-existing all-platform `WebDebugLogger` already exhibits the identical pattern and ships fine. Adding a symmetric `uninstall()` would be a tidy nicety, not a fix. (effective score: ~40 — one scorer emitted 75 but its own justification concluded "not a reachable use-after-free in normal operation")

- [OK to skip] `g_previousHandler` is a plain non-atomic pointer now read on all platforms. Set exactly once at startup (`main.cpp:351`, before `QApplication` and any logging worker threads exist) and never mutated; the same non-atomic pattern ships unchanged in `CrashHandler` and `WebDebugLogger`. ThreadSanitizer-pedantic, no real concurrent-access window. (score: 35)

</details>

[`src/core/btlogfilter.cpp` L30-L34](https://github.com/Kulitorum/Decenza/blob/f315f2c087061185cd0f8cffbf5d753b1cc2d392/src/core/btlogfilter.cpp#L30-L34)

Issue #1198, opened 2026-05-17
AI KB: bean-specific variant titles (e.g. "D-Flow / Q - Jeff") no longer inherit the parent recipe's knowledge


skialpine:
## Summary

After the #1192 KB restructure (`profile_knowledge.json` validated format), a user-renamed variant of a known recipe — e.g. **`D-Flow / Q - Jeff`**, `D-Flow / Q2`, `D-Flow / Q3`, `D-Flow / Q - <bean>` — no longer resolves to the specific KB entry it logically belongs to (`d-flow-q-variant`, with the 6-bar / 84°C / UGS 1.0 guidance). It now falls through to the **generic editor default** (`d-flow`, "D-Flow", UGS 0.5), losing the recipe-specific advice. In the dialing-context path some of these show up as `no kbId resolvable` and contribute nothing.

## Root cause

`ShotSummarizer::matchProfileKey` resolves a profile title by:
1. **Exact** normalized-title lookup against `s_aliasToId` (`displayName` + each `alsoMatches`, exact strings only). `normalizeProfileKey` only lowercases/trims (`é→e`, `" & "→" and "`) — it does **not** split on `/` or strip a ` - <suffix>`. So `"D-Flow / Q - Jeff"` ≠ `"D-Flow / Q"` → miss.
2. Editor-type fallback → `__editor_default__:dflow` → entry `d-flow` (generic).

The pre-restructure resolver had a greedy `startsWith`/`contains` scan that would prefix-match `"D-Flow / Q - Jeff"` → `"D-Flow / Q"`. That scan was **deliberately deleted** (`src/ai/shotsummarizer_kb.cpp:309-311`: "that order-dependent fallback silently mis-resolved real profiles — deleted, never reintroduced"). The new schema's `alsoMatches` is exact-only with no family/prefix concept, so the variant case was not carried over.

## Why a separate issue

This is a **resolver / KB-data-model** decision, distinct from the dialing_blocks identity-grouping fix (PR for that collapses the *exact* `D-Flow / Q` ≡ `Damian's Q` ≡ `D-Flow Q variant` — see that PR). Reintroducing prefix matching must be **deterministic and validator-gated**, not a silent re-add of the greedy scan the restructure intentionally removed.

## Proposed direction (for design discussion)

[...]

PR #1199, opened 2026-05-17
fix(ai): resolve grinder-calibration profiles by KB id, not displayName

skialpine:
## Summary

`buildGrinderCalibrationBlock` grouped shots by `canonicalNameForKbId()` — a **displayName string** — and rejected any group whose name != that displayName as a "variant". After the JSON KB restructure (#1192), `displayName` ("D-Flow Q variant") no longer equals the real profile title ("D-Flow / Q", which is now an `alsoMatches` alias). So the user's canonical "D-Flow / Q" usage was excluded from anchoring as a "variant of D-Flow Q variant", and legacy per-shot `profile_kb_id` strings split one logical profile into multiple groups.

This makes the dialing code consume the JSON by **stable `id` identity**, the way the new schema intends — not a patch on the string compare.

## Changes

- **`ShotSummarizer::resolveKbId()` (new public API)** — resolves a stored `profile_kb_id` (a current `id` OR a legacy normalized title/alias, D14a) to the canonical `id`. This is the identity primitive consumers should group/dedupe on; `canonicalNameForKbId` only returns a displayName and can't serve as a key. Thin wrapper over the existing private `resolveKbInput` (+ `loadProfileKnowledge`) — strictly additive, no behavior change for existing callers.
- **Group + anchor `buildGrinderCalibrationBlock` by resolved `id`.** Every declared-equivalent title (displayName + all `alsoMatches`: `D-Flow / Q` ≡ `Damian's Q` ≡ `D-Flow Q variant`) now collapses into one group regardless of which legacy kb-id each shot persisted. The string-equality "variant of" gate is **deleted** — a resolved id IS the canonical entry; custom/bean-specific recipes resolve to no id and are excluded by the existing empty-id check.
- **Cross-profile RGS dedup fixed (audit class-B site).** Coverage is now keyed by `id`, not displayName, so an `alsoMatches`-aliased history group can no longer split off and be re-emitted as a phantom standalone profile row.

A codebase-wide audit of every KB-data consumer (`canonicalNameForKbId`, `computeProfileKbId`, `ugsForKbId`, `getAnalysisFlags`, `expertBandForKbId`, `s_aliasToId`, the `profile_kb_id` write path, etc.) found exactly two class-B sites — both in `dialing_blocks.cpp`, both fixed here. All other consumers already key off `id` correctly or use displayName only for human-facing output.

## Out of scope — tracked in #1198

Bean-specific renamed variants like `D-Flow / Q - Jeff` / `D-Flow / Q3` still fall through to the generic editor default (they aren't in `alsoMatches`, and the restructure deliberately removed greedy prefix matching). Restoring that needs a deterministic, validator-gated prefix rule — designed properly in **#1198**, not bolted on here.

## Testing

- New regression test `TstDialingBlocks::calibrationBlock_aliasEquivalentTitlesCollapseToOneGroup`: shots stored under all three declared-equivalent identifiers (`d-flow-q-variant`, legacy `d-flow / q`, title-resolved `Damian's Q`) must collapse into **one** group with the pooled median, and the alias titles must not leak as phantom rows.
- Clean build: 0 compiler errors, 0 warnings (all targets).
- Authoritative real run via Qt Creator (588 ms) over the full reworked surface — all calibration paths (happy / badged / aborted / inferred-fallback / degenerate / fewer-than-two), the variant-distinct-position guards (`dflowVariantUgs_distinctPositions_1160`, `dflowLaPavoniVariant_distinctPosition`), the cross-consumer resolver contract (`resolver_unknownInput_isStrictNoOpAcrossConsumers`), KB-coverage/shipped-KB guards, and the new test: **36 passed, 0 failed, 0 warnings**.
- Caveat: the Qt Creator Autotest MCP returned cached snapshots (1–5 ms) for broad/repeated scopes this session, so a fresh *full-suite* run through it wasn't obtained. The Linux CI workflow runs the complete `ctest` suite on tag push for the full gate. (Flagging the caching behavior as QC-MCP feedback.)

## Test plan

- [ ] CI Linux full `ctest` suite green
- [ ] On device: a `D-Flow / Q` shot's AI dial-in context shows D-Flow/Q anchoring/median (no "skipped — variant of"), and "D-Flow / Q" / "Damian's Q" no longer appear as duplicate profile rows

### DIFF
--- a/src/ai/dialing_blocks.cpp
+++ b/src/ai/dialing_blocks.cpp
@@ -757,20 +757,27 @@ QJsonObject buildGrinderCalibrationBlock(QSqlDatabase& db,
         return QJsonObject();
     }
 
-    // Group numeric settings by canonical KB name.
-    // canonicalKey: pk.name if kbId found in KB, else profile_name.
+    // Group numeric settings by canonical KB *id* (identity), not by the
+    // displayName string. Resolving the stored profile_kb_id (or, for very
+    // old shots with an empty kb-id, the profile title) collapses every
+    // declared-equivalent title — displayName + all alsoMatches, e.g.
+    // "D-Flow / Q" ≡ "Damian's Q" ≡ "D-Flow Q variant" — into ONE group,
+    // regardless of which legacy kb-id string each shot persisted (D14a).
+    // Custom/bean-specific recipes that resolve to no id (e.g.
+    // "D-Flow / Q - Jeff") key off the profile title and are excluded from
+    // anchoring downstream by the empty-id check.
     struct ProfileGroup {
-        QString kbId;           // first non-empty kbId seen for this canonical name
-        QString canonicalName;
+        QString kbId;           // resolved canonical id ("" if unresolved)
+        QString canonicalName;  // displayName (resolved) or raw title — logs only
         QList<double> settings;
     };
-    QMap<QString, ProfileGroup> groups;  // key = canonicalName
+    QMap<QString, ProfileGroup> groups;  // key = canonical id, else profile title
 
     int rowCount = 0;
     int nonNumericCount = 0;
     while (q.next()) {
         ++rowCount;
-        const QString kbId = q.value(0).toString().trimmed();
+        const QString rawKbId = q.value(0).toString().trimmed();
         const QString profileName = q.value(1).toString().trimmed();
         const QString settingStr = q.value(2).toString().trimmed();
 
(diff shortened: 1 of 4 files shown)


skialpine:
## Automated review summary (4 agents)

**Verdict: no bugs. Identity model verified correct.** Follow-ups addressed in 1ba796c5.

### Findings

- **General / CLAUDE.md:** clean. Grouping precedence, `g.kbId`/`canonicalNameForKbId` non-emptiness, id-keyed dedup, deleted variant gate, NaN-UGS/empty-id handling, and `resolveKbId` conventions all verified correct. No changes required.
- **Deep bug scan (opus):** no significant bugs. Traced split/merge correctness (both resolution paths converge on the same id), NaN-UGS single-emit, no phantom/duplicate `customMedians` rows, ordering change is non-functional, `resolveKbId` compiles/links, `selectAnchors` unchanged and pointer-safe.
- **Test coverage:** new test solid for the happy path; flagged two criticality-8 gaps for the *new* history-emit branches → **both now covered** (see below).
- **Comments:** one real stale comment (`dialing_blocks.cpp:821` "exact KB name", invalidated by the rework) + one minor over-generalization → **both fixed**.

### Actions taken (1ba796c5)

1. `calibrationBlock_resolvedNoUgsEmittedAsHistoryRow` — resolved-but-NaN-UGS id emitted exactly once as a displayName-labelled history row.
2. `calibrationBlock_unresolvedCustomTitleEmittedAsRawHistoryRow` — unresolved custom title surfaces once via `customMedians` with the raw title; asserts no contamination of `d-flow-q-variant` (its cross-profile row stays non-"history"). First draft of this test had an over-strict assertion (it asserted the KB UGS row was absent; it legitimately appears as derived/extrapolated) — corrected; production code was always correct, confirmed by the debug trace.
3. Fixed the stale `:821` comment and tightened the dedup comment re: unresolved custom groups.

Real Qt Creator runs (per `Class::function`, the form that forces genuine execution): the full reworked surface + 3 new/fixed tests all green, **0 failed, 0 warnings**. Lower-criticality suggestions (title-only-as-sole-anchor; distinct-profiles end-to-end) left for the reviewer's call — resolution distinctness is already independently covered by `dflowVariantUgs_distinctPositions_1160`.

Note: the opus agent defensively flagged the project's own `decenza` MCP server instruction block as possible injected text — benign (it's our legitimate MCP instructions, not an external attack).

PR #1201, opened 2026-05-17
fix(ai): resolve renamed recipe variants via deterministic KB prefix (#1198)

### LINKED ISSUE #1198: AI KB: bean-specific variant titles (e.g. "D-Flow / Q - Jeff") no longer inherit the parent recipe's knowledge

skialpine:
## Summary

Fixes #1198. After the #1192 KB JSON restructure the profile→KB resolver was exact-alias-or-editor-default, so a user-renamed/numbered variant of a documented recipe (`D-Flow / Q - Jeff`, `D-Flow / Q2`, `Damian's Q - decaf`, `Adaptive v2 - Jeff`, …) missed the exact alias and fell through to the generic editor default — losing the recipe-specific band / UGS / analysisFlags.

- Adds a **deterministic recipe-alias longest-boundary-prefix** step to `matchProfileKey`, between exact-match (still wins first) and the editor-type default fallback (still last). Shared with `resolveKbInput` so legacy persisted variant-title kbIds heal under recompute-on-load.
- **Editors are namespaces, not recipes**: `defaultForEditorType` entries' aliases are excluded as prefix anchors (the bare `D-Flow` alias is *not* an anchor); the editor namespace is resolved only by the explicit editor-default step. Recipe aliases of **any** documented profile anchor — proven general (Adaptive v2 / Londinium / Allongé), not D-Flow-specific.
- **Longest-wins**, prefix-only, closed separator set (`/`, `-`, space, ASCII digit). This is *not* the order-dependent greedy `startsWith`/`contains` scan #1192 deleted — it is anchored, total, and test-gated. A genuine equal-length tie is impossible by construction (reduces to the existing alias-collision rejection).
- Built-ins always resolve by exact match (step 1) and never depend on the prefix step (D8 invariant test).
- Non-fatal `validate_kb.py` lint flags one recipe alias boundary-prefixing another with a different id (author awareness) — **0 hits** on the shipped KB.
- The pre-#1198 `calibrationBlock_unresolvedCustomTitleEmittedAsRawHistoryRow` test is **inverted** (renamed to `…_recipeVariantTitleResolvesToParentGroup_1198`): the issue explicitly anticipated that variants now collapse into the parent id group.

Design/rationale in `openspec/changes/resolve-kb-variant-title-prefix/`.

## Test plan

- [x] `openspec validate resolve-kb-variant-title-prefix --strict` → valid
- [x] `python3 tools/validate_kb.py` → OK, new lint adds 0 WARNs (10 pre-existing unchanged), 0 ERRORs, KB data unchanged
- [x] Qt Creator build clean (0 errors, 0 warnings)
- [x] Full Qt Test suite: **2108 passed, 0 failed, 0 skipped, 0 warnings** (incl. new `recipeVariantPrefixResolution_1198` + inverted dialing-blocks test)
- [ ] Post-merge: update the "save a profile" Manual wiki page (kept-prefix convention) — deferred (document shipped behavior)
- [ ] Post-deploy: live `dialing_get_context` spot-check on a real renamed-variant shot


skialpine:
### Code review

Found 3 issues:

1. This PR adds the deterministic `recipePrefixResolve` fallback, but several authoritative comments still assert the resolver is exact-match-only with no fuzzy fallback and were not updated (file header "exact-match-or-explicitly-unresolved … greedy startsWith/contains fallback is DELETED"; `normalizeProfileKey` "there is NO fuzzy fallback after a miss" at L58; the `resolveKbInput` block comment at L312-316; and the header declaration "Exact-match-or-unresolved, never fuzzy" in `src/ai/shotsummarizer.h#L431`). These now contradict the implemented behavior and mislead maintainers about the resolver contract.

https://github.com/Kulitorum/Decenza/blob/e0f10b23800944cdf090aefc8b942df0ba27ee24/src/ai/shotsummarizer_kb.cpp#L15-L18

2. `src/ai/dialing_blocks.cpp:766-768` (not in this PR's diff) and the new test's own inline comment both still say `"D-Flow / Q - Jeff"` "resolve(s) to no id" / "empty → resolves to no id" — now false: it resolves to `d-flow-q-variant` via the new prefix step. This stale comment sits directly in the grinder-calibration grouping logic developers consult when debugging why profiles group together.

https://github.com/Kulitorum/Decenza/blob/e0f10b23800944cdf090aefc8b942df0ba27ee24/src/ai/dialing_blocks.cpp#L765-L769

https://github.com/Kulitorum/Decenza/blob/e0f10b23800944cdf090aefc8b942df0ba27ee24/tests/tst_dialing_blocks.cpp#L1303-L1306

3. Inverting `calibrationBlock_unresolvedCustomTitleEmittedAsRawHistoryRow` removed the only integration-level assertion that a genuinely unresolved custom title (empty stored kbId, no recipe-prefix match) still surfaces as a raw `customMedians` history row in `buildGrinderCalibrationBlock`. `recipeVariantPrefixResolution_1198` covers the unit-level empty-resolution case but nothing re-covers the calibration-block emission path, which still exists and must stay correct.

https://github.com/Kulitorum/Decenza/blob/e0f10b23800944cdf090aefc8b942df0ba27ee24/tests/tst_dialing_blocks.cpp#L1262-L1270

<details>

- [OK to skip] `s_recipeAliases.append` is unconditional while `validate_kb.py` uses `setdefault`; the dangerous cross-entry-collision case is already a FATAL validator gate and same-id within-entry duplicates are harmless (longest-first returns the same id), so a dedup guard is disproportionate to the gated risk (score: 50)

</details>

PR #1202, opened 2026-05-18
Scale connection-priority: preheat liveness gate + startup probe + MCP read/reset

skialpine:
## Why

Builds on the merged [`scale-ble-priority-backoff`](https://github.com/Kulitorum/Decenza/pull/1185) (#1185), which is field-confirmed working on the weak Tab A8 (Unisoc T618). Build-3388 logs show two residual costs of the in-memory app-run skip-HIGH latch:

1. detection was gated to **active extraction**, so on the triggering shot the user sees a ~6–8 s "no scale — estimating" window *mid-shot* (the feed actually dies ~6 s earlier, during preheat prep, but we don't look until the pour);
2. the latch is in-memory, so a known-incapable device re-pays that window on the **first shot of every app run**.

## What changed (all three together; persistence deliberately parked — design D2)

**1. Preheat liveness gate — the guaranteed net.** `WeightProcessor` gains `m_preheatActive`/`m_probeActive`; the stall check is factored into `checkScaleFeedStall()`; the gate widens to `((m_active || m_preheatActive) && m_tareComplete) || m_probeActive`. `main.cpp` drives `setShotCycleActive()` from `MachineState::phaseChanged` (EspressoPreheating only) so the stall is caught during warm-up and the backoff + reconnect can begin **before the pour**. *Does not depend on the probe.*

**2. Startup connection-priority probe — the early win, strictly additive.** New `ScalePriorityProbe` runs a **bounded read-only** DE1 burst (`DE1Device::issueConnectionPriorityProbeReads()` — safe MMR block reads + reads of `a001/a00a/a00e/a011` only; **never** a write/DANGER characteristic) once the scale is **confirmed streaming at idle**, ticking `pollScaleFeedLiveness()` so a provoked stall drives the *existing* backoff → reconnect-at-BALANCED **before any shot**. No-op when it provokes nothing; correctness never depends on it; yields immediately if a shot starts; runs at most once per connect; WARN start/end logging (self-evidencing in one debug.log); `DECENZA_DISABLE_SCALE_PRIORITY_PROBE` kill-switch.

**3. MCP read + reset of the in-memory latch.** `BLEManager` carries the trigger kind + set-time + app-start; `devices_connection_status` reports the (non-persisted) `scaleConnectionPriority` state (latched?, kind, ISO-8601 set-time, human elapsed-since-app-start); new `devices_reset_scale_priority` control tool clears it (eventually-consistent, no forced teardown). `MCP_SERVER.md` updated (50 → 51 tools).

**Deferred (NOT in this change):** persisting the latch across restarts — parked per design D2 (would remove the non-sticky safety valve; revisit only if field data shows the per-run window is still painful).

## Safety

The probe stressor is **read-only by construction** — it cannot misconfigure the DE1, so no profile re-upload/verification is needed. Capable hardware sustains the read burst → no stall → never trips (covered by a unit test). The preheat/extraction liveness net is the guarantee whether or not the probe ever provokes.

## Testing

- Qt Creator build (Decenza-Desktop worktree): **0 errors / 0 warnings**.
- `tst_weightprocessor`: **29/29** — includes 5 new gate tests (preheat-stall fires; idle/no-cycle/no-probe does *not*; probe-window stall fires at idle; capable-HW sustained read burst does *not* trip; probe-deactivation clears the sticky stale flag).
- `tst_scaleblepriority`: **9/9** (unchanged).
- No persistence added on any path (in-memory `BLEManager` only).

## Field follow-up

A new field report (current shipped build, #1185 only) shows shot 1 still eats ~8 s mid-shot (this PR moves that detection into preheat / to the idle probe) and a milder **~3 s no-weight at the start of shot 2** at BALANCED (no backoff path — already BALANCED). The ~3 s is a separate, milder phenomenon the dual-HIGH backoff structurally cannot fix; **we still need to review the next reporter log** (ideally from a build with this change, so the probe/preheat WARN markers + the MCP read make a single debug.log self-diagnosing) to determine whether it is benign BALANCED first-sample latency or genuine residual contention.

Relates to #1093 / #1176 and the merged #1185. Plain links (no `Fixes`) until the reporter confirms.


skialpine:
## Multi-agent review applied (commit f74bc009) — one-and-done

Ran code / test / comment / silent-failure / type-design review agents. Disposition of every finding:

**Fixed (correctness):**
- **Probe never ran if streaming was confirmed while the machine was non-idle** (the headline gap — scale connected during warm-up → probe skipped the whole connect). `onMachinePhaseChanged()` now also tries the guarded/idempotent start on the become-idle transition.
- `~ScalePriorityProbe()` destructor → idempotent `endProbe()`, so `m_probeActive` can't strand `true` on the worker.
- `setProbeActive(false)` only clears `m_scaleFeedStale` when no other "weight expected" context is active (no concurrent-stall swallow).

**Fixed (observability / honesty):**
- `onScaleFeedStalled()` logs when the detector is disarmed — the single field debug.log never silently drops a stall.
- **H2 re-scoped by evidence:** `DE1Device::errorOccurred` has **no UI/QML consumer anywhere** (verified) — log-only. So no error-channel suppression (which would risk masking real faults); instead the probe counts DE1 link faults in its window and reports the count in the end-line.
- MCP reset reports `accepted/queued + appliesOnNextReconnect`, not a fake `success`.

**Fixed (type-design / invariant expression):**
- `BLEManager` latch → `ScaleSkipHighLatch` value type (kind+time set iff latched); `latchScaleSkipHighPriority()` takes a **mandatory** trigger kind (removes the silent-`"unknown"` footgun).
- `ScalePriorityProbe` bool soup → `enum class State {Idle,Confirming,Armed,Probing,Done}` — illegal states unrepresentable.

**Tests:** `tst_weightprocessor` 32/32 (+in-shot `m_active` regression, preheat-no-tare negative, probe-then-extraction); new `tst_scaleskiphighlatch` 7/7; `tst_scaleblepriority` 9/9. Build 0/0.

**Accepted as-is (rationale in design.md D7, not silently dropped):** aboutToQuit reorder (shutdown-only, benign); unlogged disconnected-transport probe no-op (end-line records outcome); a dedicated probe-controller test target (the `enum class State` makes illegal states unrepresentable; decision logic is unit-tested; a ≈half-the-app link target is higher risk than value; on-device validates integration).

skialpine:
### Code review

The two highest-impact candidate findings were verified **false positives**: a probe-provoked `m_scaleFeedStale` "carrying into preheat" and a stale-`m_lastWallClockMs` "false-positive preheat backoff" both cannot occur, because `MachineState::espressoCycleStarted` is emitted synchronously at EspressoPreheating entry and its queued `WeightProcessor::startExtraction()` (which zeroes both `m_scaleFeedStale` and `m_lastWallClockMs`) is enqueued on the worker before the deferred `phaseChanged → setShotCycleActive(true)`. A flagged "probe never re-runs after MCP reset" concern was also verified to be **correct, intentional behavior** (the proposed change would itself regress the MCP-reset re-probe contract). CLAUDE.md compliance (timer-as-guard rule, cross-thread QueuedConnection, MCP data conventions, qsizetype, settings domains): clean.

<details>
<summary>7 below-threshold issues (50-74)</summary>

- [Worth a look] `clearScaleSkipHighPriority()` early-return `if (!m_scaleSkipHigh.latched && m_scaleSkipHigh.triggerKind.isEmpty())` — the second clause is redundant given the new `ScaleSkipHighLatch` invariant; idiomatic guard is `if (!m_scaleSkipHigh.latched) return;` (one-line, in code this PR modifies; `src/ble/blemanager.cpp`) (score: 65)
- [Worth a look] `src/ble/de1device.h` probe comment says it "NEVER writes" but `issueConnectionPriorityProbeReads()` issues `m_transport->write(READ_FROM_MMR, …)` length-0 read-requests — the operative safety guarantee (no DANGER char, no state mutation) is correct; only the wording is imprecise (score: 50)
- [Worth a look] `src/ble/blemanager.h` comment "'latch without a reason' is a compile error, not a silent 'unknown'" — `latchScaleSkipHighPriority("")` compiles and silently yields "unknown"; only arg-omission is a compile error. All real callers pass a non-empty literal (score: 50)
- [OK to skip] MCP `detectionActive` boolean can read as "detection running now" vs "not latch-suppressed"; mitigated by clearer sibling fields (`latchedToBalanced`, `scaleLinkPriority`); not explicitly called out by CLAUDE.md (score: 50)
- [OK to skip] `src/ble/blemanager.h` calls `m_appStartTime` "(process start)"; it is BLEManager-construction time, so MCP `elapsedSinceAppStartSec` undercounts by the pre-BLEManager init (sub-second/few seconds) — negligible for coarse triage (score: 50)
- [OK to skip] `ScalePriorityProbe` `Armed` state-doc lists the start gate but omits the `probeDisabled()` kill-switch conjunct — documentation completeness only (score: 50)
- [OK to skip] Probe class-doc "runs at most once per scale connect" does not restate that the BLEManager app-run latch makes it effectively once per app run — accurate as written; the latch behaviour is documented at `BLEManager` (score: 50)

</details>

skialpine:
## D8 + D9 added (commit e9afec34) — the actual root-cause fix

Root cause established (research brief + #1176 logs + git archaeology): the dual-HIGH contention is **not** new traffic or #1185 — **PR #604 / v1.6.3 restored a working `CONNECTION_PRIORITY_HIGH` request** after Qt 6.10 silently broke the old reflection hack. The weak Tab A8 (Unisoc T618) ran fine for the whole window where HIGH was a no-op (effectively both-BALANCED); restoring real dual-HIGH brought the contention back. de1app requests no priority at all; Android docs say HIGH is burst-only. #1185's scale-only backoff is structurally insufficient — the #1176 shot-2 log shows a lone HIGH DE1 still starves even a BALANCED scale (no backoff event, ~3.7 s stall).

**D8 — DE1 also skips HIGH when latched** (reverses #1185 "DE1 Link Always Requests High Priority", latched case only). `BleTransport::onControllerConnected()` consults the same `BLEManager` latch the scale transport uses → both links BALANCED = the known-good config (= de1app, = Android guidance). Eventually-consistent (no live renegotiation), logged both branches (closes the DE1-priority observability gap), capable HW never latches → no regression. SAW-latency objection quantified: BALANCED adds ≲~35 ms to the stop write → <0.1 g at typical flow; de1app ships DE1 at BALANCED. On-device SAW validation on the Tab A8 is the acceptance gate.

**D9 — persist the classification, build-scoped** (the "-persist" in the change name, finally in scope). D2 parked persistence on "residual cost is a blip"; the #1176 build-3388 log overturned that (ruined first shot every restart — D2's own revisit-trigger). New internal `SettingsHardware` `connectionPriority/` record (no Q_PROPERTY/QML/UI). `BLEManager::setSettings()` loads it before any BLE connect: same `versionCode()` → seed (both links start BALANCED, **no detection window, no bad first shot**); different build → wipe (safety valve — every build re-detects). MCP reset + build-scoped auto-clear resolve the old "no recovery for mis-latched capable HW" objection; false-positive cost is "runs at BALANCED" = de1app's normal config.

OpenSpec: D2 superseded, D8/D9 added, spec MODIFIES #1185's "DE1 Link Always Requests High Priority" + "App-Run Scoped, Not Persisted" (latched case); `--strict` valid. Tests: `tst_scaleskiphighlatch` 10/10 (5 invariant + 5 D9 persisted/build-scope, QStandardPaths-isolated), `tst_weightprocessor` 32/32, `tst_scaleblepriority` 9/9; build 0/0.

Net end-state for the reporter: first run of a build re-detects (probe usually catches it at idle pre-shot); **every subsequent run of that build starts both links BALANCED with no detection window and no ruined first shot**. New build → re-detect (one-shot safety-valve cost). The single debug.log is now fully self-evidencing (probe markers + DE1 priority decision + persisted-load line).

Recommend a fresh `/review` over the D8/D9 delta before merge.

skialpine:
### Code review — D8/D9 delta (5 agents) + fixes applied (commit 8d2c0035)

Reviewed the D8/D9 delta with 5 specialized agents (type-design, tests, silent-failure, comments, + a CLAUDE.md/general pass). Strong cross-agent convergence; all must- and should-fix items are now fixed.

**Fixed**

1. **Rehydration broke the `ScaleSkipHighLatch` invariant** (3-agent consensus, HIGH). `setSettings()` seeded `latched=true` with `QDateTime::fromString(…ISODate)` and no validity check → a malformed/empty/partial-write persisted timestamp produced `latched=true` + invalid `setTime` (the state the value type forbids), and the MCP read then silently dropped the timestamp with no log. Added a sanitizing `ScaleSkipHighLatch::rehydrate(kind,time)` that preserves the original set-time (vs `set()` which re-stamps), coerces corrupt input so the invariant always holds, and returns `false` iff the time was substituted so `setSettings()` logs the anomaly. The classification is kept (load-bearing) rather than discarded over a bad diagnostic field.
2. **`"persisted": false` was a factual lie post-D9.** `devices_connection_status` now reports `persisted=true` + `persistenceScope="build"`; the stale `// by design (D2)` / "in-memory, app-run-scoped" comments are corrected; the `devices_reset_scale_priority` description now states the reset is durable (clears the persisted build-scoped record).
3. **Stale post-D9 doc comments** corrected: `blemanager.h` `ScaleSkipHighLatch` block + `clearScaleSkipHighPriority` doc; `bletransport.cpp` `"QTBUG: no connectionUpdated() signal"` reworded (the signal exists — Android never fires the underlying callback) and the "not linked" header terminology.
4. **Silent QSettings write failures** (Android read-only/full disk would look exactly like "still bad every restart" with no log — the diagnostic black hole this feature exists to avoid): `setConnectionPriorityLatch` now `sync()`s and `qWarning()`s on `QSettings::status() != NoError`. Verified CLAUDE.md-compliant (QSettings is the sanctioned settings system used identically by every existing `SettingsHardware` setter; the no-DB/disk-on-main-thread rule targets the sqlite/shot-DB layer).
5. **Tests** (Gap 1, now pure-testable): +3 `rehydrate` tests + 1 ISO round-trip test.

**Accepted as-is (multi-agent agreement + rationale):** D8 latched-skip branch untested (harness; `#else` keeps the prod path exercised), build-scope-wipe not e2e-tested (≈half-app heavy link — consistent with the recorded D7 rationale), 4 loose persisted keys (matches the settings pattern; single writer/reader; group-remove on clear), free `QString` trigger kind, `versionCode()` int scope key, silent-failure findings 5–8 (intended-benign).

**Verification:** build 0 errors / 0 warnings; `tst_scaleskiphighlatch` 14/14, `tst_weightprocessor` 32/32, `tst_scaleblepriority` 9/9; full suite **2129/2129**, `tests_with_warnings` empty.

Issue #1204, opened 2026-05-18
[Feature] Schedue longer than 8hrs


ChiSaw:
### What problem does this solve?

As a user, I set my Decenza to wake up at 6:10am, but I need it until 5pm. That is 9hrs. Currently, the limit is at 8hrs. I think there is no specific reason why 8hrs should be a max limit.

### Proposed solution

Allow stay on for longer than 8hrs, e.g. 24hrs.

skialpine:
Shipped in [#1214](https://github.com/Kulitorum/Decenza/pull/1214) (merged to `main`): the **Settings → Screensaver → Stay awake** duration cap is raised from 8 h to **12 h** (15-min steps). This covers a 6:10 am wake → ~5 pm ready window with margin.

Note: the limit was set to 12 h rather than 24 h. If you find you need it longer, comment here and we can extend it further.

Issue #1203, opened 2026-05-18
[Bug] Machine turns off even if schedule is set


ChiSaw:
### What happened?

In Screensaver, I set a schedule, M T W T F S S

- wake up: 6:10am
- stay awake 8hrs

The auto-turn off is set at 1hr (which I understand as if I use the maschine outside the schedule)

When I pull an espresso shot, the machine turns off after 1hr. In my intuition, a schedule always takes priority over an auto-off feature

### Steps to reproduce

1. Open Decenza
2. Set schedule
3. Set auto-turn off
4. Pull shot

### Decenza version

1.7.5

### Platform

Android

### Device / OS version

Galaxy Tab A9+ Android 14

### Scale (if applicable)

Bookoo

skialpine:
Fixed in [#1216](https://github.com/Kulitorum/Decenza/pull/1216) (squash-merged to `main`, commit 19ae62e7).

The scheduled stay-awake window is now evaluated continuously from the schedule + clock instead of being armed once at the wake instant, so it takes priority over the 1 h auto-off and keeps working even when the app wasn't running at the exact wake minute / the machine was woken manually / the tablet suspended the app — the scenarios that caused this report. Verified by unit tests and a clean Qt Creator build; behavior is documented in the wiki manual (Auto-Wake / Auto-Sleep).

PR #1216, opened 2026-05-18
fix(autowake): schedule stay-awake now overrides auto-off (#1203)

skialpine:
## Problem

[Kulitorum/Decenza#1203](https://github.com/Kulitorum/Decenza/issues/1203): with a schedule set (wake 6:10 am, stay awake 8 h) and auto-off at 1 h, the machine still turns off ~1 h after a shot. The reporter's expectation — *"a schedule always takes priority over an auto-off feature"* — was not met.

## Root cause

The "stay awake N hours" protection was a **one-shot countdown armed only at the exact instant `AutoWakeManager` fired `wakeRequested()`**. `main.qml`'s `sleepCountdownStayAwake` was set solely inside `onAutoWakeTriggered()`; the machine slept when **both** it and the 1 h inactivity counter reached 0.

That silently fails whenever the wake tick isn't observed at that precise minute — the normal case on a tablet not running Decenza overnight:

- **App cold-starts after the wake minute.** `scheduleNextWake()` skips any wake time already past today (`wakeDateTime <= now`) and reschedules for next week, so `onAutoWakeTriggered` never fires today, `sleepCountdownStayAwake` is never armed, and the generic 1 h auto-off wins ~1 h after the shot — exactly the report.
- Machine manually woken during the scheduled window.
- Process suspended across the wake minute (Android).

## Fix

Replace the fragile one-shot with a **continuously evaluated predicate**, `AutoWakeManager::isWithinStayAwakeWindow()`, derived from the schedule + wall clock on every sleep-countdown tick:

- Schedule now takes priority over auto-off and **survives app restarts / manual wakes / process suspension**.
- Also inspects the **previous day's** entry, so a window that began before midnight (e.g. wake 22:00 + 8 h, or a late wake under the new 12 h cap from #1204) keeps applying into the next day.
- Outside the window the 1 h auto-off behaves exactly as before. **No new settings.**
- Removes a timer-armed guard in favour of an event/state condition, per the CLAUDE.md design principle.

## Changes

- `src/core/autowakemanager.{h,cpp}`: `isWithinStayAwakeWindow()` + a testable `isWithinStayAwakeWindowAt(now)` core (clock injected; `friend class` under `DECENZA_TESTING`).
- `src/main.cpp`: expose `AutoWakeManager` as a QML context property.
- `qml/main.qml`: drop `sleepCountdownStayAwake` entirely; auto-sleep is suppressed whenever the live-evaluated window is active.
- `tests/tst_autowakewindow.cpp` (new): same-day window edges (inclusive start / exclusive end), midnight spill, 12 h duration, and the disabled / zero-duration / unscheduled-day guard cases.

## Testing

- ✅ **`tst_AutoWakeWindow` green** — 12/12 assertions pass (Qt Creator, Debug): same-day edges, midnight spill, 12 h duration, guard cases.
- ✅ **Full project build green** — 0 errors, 0 warnings (Qt Creator, Debug); confirms the `main.cpp` and `autowakemanager` changes compile in the app.
- ⚠️ `main.qml` changes are bundled/parsed but **not runtime-verified** here (only JS logic inside existing handlers + a removed property). Recommend a quick on-device sanity check on Android: schedule a wake, cold-start the app after the wake time, leave it idle past the 1 h auto-off, confirm it stays awake until the window ends.

Addresses Kulitorum/Decenza#1203

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -350,19 +350,22 @@ ApplicationWindow {
         return (val === undefined || val === null) ? 60 : parseInt(val)
     }
 
-    // Two-counter sleep system:
+    // Sleep system:
     // - sleepCountdownNormal: resets on user activity, counts down from autoSleepMinutes
-    // - sleepCountdownStayAwake: only set on auto-wake, never reset by activity
-    // Sleep when BOTH <= 0
+    // - the scheduled stay-awake window is evaluated continuously from the
+    //   schedule + wall clock via AutoWakeManager.isWithinStayAwakeWindow(),
+    //   so it survives app restarts / manual wakes / process suspension
+    //   (was previously a one-shot counter armed only on the wake event,
+    //   which silently failed whenever that exact tick wasn't observed —
+    //   #1203). Sleep only when the normal countdown has expired AND we are
+    //   not inside a scheduled stay-awake window.
     onAutoSleepMinutesChanged: {
         if (autoSleepMinutes > 0 && sleepCountdownNormal < 0) {
             sleepCountdownNormal = autoSleepMinutes
-            sleepCountdownStayAwake = 0
             console.log("[AutoSleep] Setting changed: normal=" + sleepCountdownNormal)
         }
     }
     property int sleepCountdownNormal: -1      // Minutes remaining (-1 = not started)
-    property int sleepCountdownStayAwake: -1   // Minutes remaining (-1 = already satisfied)
 
     // Active operation phases that should pause the sleep countdown
     property bool operationActive: {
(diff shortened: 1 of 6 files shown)


skialpine:
## Multi-agent review applied (commit 809a8eb2)

Ran code-review / test-coverage / comment-accuracy / silent-failure agents. Outcome:

**Fixed**
- **Test isolation** — moved the QSettings snapshot/restore from per-test `init()/cleanup()` to once-per-suite `initTestCase()/cleanupTestCase()`, so intermediate test mutations no longer churn the real shared store. Documented the one inert caveat: an originally-unset `autoWake/schedule` key restores as an all-days-disabled list, which is behaviorally identical to unset — verified no production code treats schedule presence as a sentinel (auto-wake on/off is the separate `autoWake/enabled` key).
- **Coverage** — added `yesterdayWindowDoesNotSpanMidnight`, `adjacentEnabledDaysOverlap`, `shortAndEmptySchedule`, `dayEntryMissingTimeKeysDefaultsToSevenAM`, a negative-duration assertion, and an inclusive-start check on the spill path. These pin the fragile today/yesterday loop and the guard branches against future refactors. **10/10 pass.**
- **Comment accuracy** — reworded the test comment that wrongly implied the predicate enforces the 12 h cap (it doesn't; the cap is the Settings slider bound from #1204 / [#1214](https://github.com/Kulitorum/Decenza/pull/1214)).
- **Stale class doc** — `autowakemanager.h` no longer claims "checks every 30 seconds / time-passed approach"; it now describes the actual single-shot timer + the new window predicate.

**Deliberately scoped out — known residual (not a blocker)**

The silent-failure agent correctly notes `isWithinStayAwakeWindowAt()` is *silent* on a structurally-corrupt/short schedule or out-of-range hour/minute: it returns `false` → the machine auto-sleeps through the window with no log, operationally indistinguishable from the #1203 symptom.

I am **intentionally not adding diagnostic logging in this PR**:
- No current code path produces such data — `setAutoWakeSchedule()` always writes a 7-entry list from the UI, and `setAutoWakeDayTime()` clamps `0–23 / 0–59`. The bad-data trigger requires storage corruption or a future bad import path that does not exist today.
- A useful warning must be rate-limited (the predicate is hit every 60 s once the inactivity counter expires), which means adding warn-state — disproportionate to defending against corruption that cannot currently occur, and contrary to this codebase's minimal-root-cause principle.
- The behavior is also *consistent* with the existing `scheduleNextWake()` defaulting.

**Recommended optional follow-up** (separate, only if a malformed-schedule path is ever introduced): add a one-shot `qWarning` distinguishing "stay-awake enabled but schedule list size != 7 / entry has invalid time" from the legitimate "not scheduled / disabled / zero-duration" guard returns, and/or validate `QTime::isValid()` before constructing the window.

Verified via Qt Creator (Debug): full build **0 errors / 0 warnings**; `tst_AutoWakeWindow` **10/10 pass**.

PR #1217, opened 2026-05-18
feat(about): add Manual button to the version card

skialpine:
## Summary

Adds a **Manual** button to the right of the Decenza / version / build stack on the About page (**Settings → Update**, the "Current Version" card). Tapping it opens the user manual — the GitHub wiki at <https://github.com/Kulitorum/Decenza/wiki> — in the default browser, giving users one-tap access to the docs.

## Details

- Placed inside the existing version card, `anchors.right` / `verticalCenter`.
- Declared **after** the existing 7-tap translation-upload easter-egg `AccessibleMouseArea` and given `z: 1`, so taps on the button open the manual while taps elsewhere on the card still feed the tap counter (no regression to the easter egg).
- Styled with the same `Theme` tokens and the accessible-button idiom as the adjacent Donate button in the same file (`Accessible.role/name/focusable/onPressAction`, `Qt.openUrlExternally`, pressed-state color).
- Label + accessible name internationalized via `TranslationManager.translate` (`about.manual`, `about.manualHint`), consistent with neighbouring strings.

## Testing

- Qt Creator (Debug) build: **0 errors, 0 warnings** — includes `qmlcachegen`, so the new QML is syntactically valid and bindings resolve.
- Visual placement not yet runtime-verified; recommend a quick look on the About page (the layout uses the same anchor pattern as the existing Donate button, low risk).

### DIFF
--- a/qml/pages/settings/SettingsUpdateTab.qml
+++ b/qml/pages/settings/SettingsUpdateTab.qml
@@ -82,9 +82,17 @@ Item {
                         }
                     }
 
-                    // Easter egg: tap 7 times to enable translation upload
+                    // Easter egg: tap 7 times to enable translation upload.
+                    // Bounds stop at the Manual button's left edge so the two
+                    // accessible elements never overlap — ACCESSIBILITY.md
+                    // forbids nesting accessible elements (TalkBack would
+                    // otherwise expose only one).
                     AccessibleMouseArea {
-                        anchors.fill: parent
+                        anchors.left: parent.left
+                        anchors.top: parent.top
+                        anchors.bottom: parent.bottom
+                        anchors.right: manualButton.left
+                        anchors.rightMargin: Theme.scaled(6)
                         accessibleName: TranslationManager.translate("update.versionBuild", "Version %1, Build %2").arg(AppVersion).arg(AppVersionCode)
                         onAccessibleClicked: {
                             var now = Date.now()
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 6 issues (all accessibility; the PR is otherwise functional — the "renders 0×0" concern was checked and is a false positive, since a bare QML item sizes to its `implicitWidth/Height` when `width/height` are unset):

1. Manual button is a raw `Rectangle` + plain `MouseArea` with hand-attached `Accessible.*`, instead of the project's `AccessibleButton`/`AccessibleMouseArea` — loses the announce-first TalkBack flow (CLAUDE.md says "Prefer `AccessibleButton` / `AccessibleMouseArea` over raw `Rectangle+MouseArea`").

https://github.com/Kulitorum/Decenza/blob/a40e919a3d7fd2df03ca10da8999077de1fc5f36/qml/pages/settings/SettingsUpdateTab.qml#L120-L155

2. The button's accessible bounds sit inside the easter-egg `AccessibleMouseArea` (`anchors.fill: parent`). `z:1` only fixes pointer routing, not the a11y tree — TalkBack/VoiceOver may expose only one element (docs/CLAUDE_MD/ACCESSIBILITY.md: "Never position accessible buttons inside another accessible element's bounds … TalkBack will only discover one element").

https://github.com/Kulitorum/Decenza/blob/a40e919a3d7fd2df03ca10da8999077de1fc5f36/qml/pages/settings/SettingsUpdateTab.qml#L85-L126

3. `Rectangle` sets `Accessible.focusable: true` but not `activeFocusOnTab: true`, so the button is unreachable by desktop keyboard Tab (docs/CLAUDE_MD/ACCESSIBILITY.md: "Missing `activeFocusOnTab: true` — Control is unreachable by keyboard Tab navigation even if it has `Accessible.focusable: true`. Both must be set together").

https://github.com/Kulitorum/Decenza/blob/a40e919a3d7fd2df03ca10da8999077de1fc5f36/qml/pages/settings/SettingsUpdateTab.qml#L132-L138

4. `Accessible.description` is used as a general hint on a primary-only button; ACCESSIBILITY.md reserves it for secondary-action hints ("Any interactive element with secondary actions (long-press, double-tap) must set `Accessible.description`").

https://github.com/Kulitorum/Decenza/blob/a40e919a3d7fd2df03ca10da8999077de1fc5f36/qml/pages/settings/SettingsUpdateTab.qml#L134-L136

5. Comment overstates the mechanism: in QML `z:1` alone (not declaration order) raises the button over the default-`z:0` easter-egg area; and the existing line-85 easter-egg comment is now incomplete (the button intercepts part of the card).

https://github.com/Kulitorum/Decenza/blob/a40e919a3d7fd2df03ca10da8999077de1fc5f36/qml/pages/settings/SettingsUpdateTab.qml#L116-L121

6. Color binds to `manualButtonArea.pressed`, but `Accessible.onPressAction: manualButtonArea.clicked(null)` never sets `pressed`, so screen-reader activation gives no visual press feedback (bug due to `color: manualButtonArea.pressed ? … : …` vs `clicked(null)`).

https://github.com/Kulitorum/Decenza/blob/a40e919a3d7fd2df03ca10da8999077de1fc5f36/qml/pages/settings/SettingsUpdateTab.qml#L127-L138

<details>

- [Worth a look] i18n keys use `about.*` while the version card uses `update.*`/`settings.update.*` — easy rename in code already being changed (score: 50)
- [Worth a look] Centered version `ColumnLayout` has no width constraint vs the right-anchored button — possible text/button overlap at large text scaling or long platform strings (score: 50)

</details>

skialpine:
Addressed all 6 findings in 06157802:

- Findings 1, 3, 6 (raw Rectangle+MouseArea / missing `activeFocusOnTab` / no SR press feedback) — replaced with `AccessibleButton`, the project-preferred Control: keyboard Tab focus, Return/Space, announce-first, press feedback all built in.
- Finding 2 (accessible-bounds overlap) — the easter-egg `AccessibleMouseArea` now anchors `left/top/bottom` + `right: manualButton.left` instead of `anchors.fill`, so the two accessible elements are disjoint (`z:1` hack removed).
- Finding 4 — dropped `Accessible.description`.
- Finding 5 — rewrote the inaccurate comment and extended the easter-egg comment.
- Below-threshold i18n note — switched to `settings.update.manual` / `settings.update.manualAccessible` to match the sibling AccessibleButtons in this column.

The remaining below-threshold item (text/button overlap at extreme text scaling) is no longer an a11y-tree issue now that bounds are disjoint; the centered version text is short so visual collision needs extreme scaling — left as-is.

Verified via Qt Creator (Debug): 0 errors / 0 warnings.

skialpine:
Final >49 item addressed in 885836cc: the version `ColumnLayout` now centers within the region left of the Manual button (not the whole card), and the variable-length platform/"SIMULATION MODE" line uses `fillWidth` + `elide: ElideRight`. Text and button can no longer collide under large text scaling or a long platform string.

All review items scoring 50+ are now resolved (six 75s in 06157802, both 50s — i18n namespace and text overlap — in 06157802/885836cc). Qt Creator Debug build: 0 errors / 0 warnings.

PR #1222, opened 2026-05-18
fix(ble): reject DiFluid R2 out-of-range TDS sentinel before it is saved

skialpine:
## Problem

A coffee's TDS was saved as **655.09 %** (EY **1342.9 %**) on a shot. Investigation via the device debug log showed the DiFluid R2 emits an **out-of-range error sentinel** in the TDS field when a measurement fails mid-flight:

```
[296.409] TDS: 10.39% (raw=1039)      ← valid
[300.770] TDS: 10.28% (raw=1028)      ← valid
[305.106] Temperature: prism=27.9°C tank=39.7°C   ← tank temp spikes
[305.121] TDS: 655.09% (raw=65509)    ← garbage sentinel (0xFFE5)
[308.916] WARN [BLE DiFluidR2] R2 error: class=0 code=2   ← device errored
```

The `/100` scaling is **correct** (1039 → 10.39%); the bug is that the well-formed-but-impossible sentinel packet (`raw=0xFFE5`) passed the checksum, was emitted as a real reading, and got autosaved onto the shot.

Why the app's "Read TDS" button looked fine while the physical R2 Start button didn't: the button does a clean single test that finishes before the failure; the physical-button stream included the failure sentinel. **Both paths share the same `case 2` parser**, and the consumer guard in `PostShotReviewPage.onTdsChanged` only had a *lower* bound (`kMinimumPlausibleTds = 3.0`), so the inflated value sailed through.

## Fix

- Route pack 2 (single test — the app **Read TDS** button *and* the physical **R2 Start** button) and pack 3 (average) through one shared `emitTdsResult()` helper that drops any TDS above the device's physical range (`MAX_PLAUSIBLE_TDS = 35 %`). One chokepoint → **identical behaviour on both ingestion paths**, which was the explicit requirement.
- Mirror it with a symmetric `kMaximumPlausibleTds` in `PostShotReviewPage.onTdsChanged` (the documented consumer-validation point), so a garbage value cannot be persisted by either the C++ or QML layer.
- Valid readings (including unusually strong but in-range ones) and the existing low/zero path are unchanged.

## Tests

`tests/tst_difluidr2.cpp`:
- `rejectsImplausiblyHighTds` — exact field sentinel (raw 65509) on pack 2: no `tdsChanged`/`measurementComplete`, `errorOccurred` emitted, not measuring, `m_tds` untouched.
- `rejectsImplausiblyHighAverageTds` — same for pack 3.
- `acceptsTdsAtTopOfDeviceRange` — 30 % still passes (no over-rejection).

Run the `tst_DiFluidR2` suite in Qt Creator (BUILD_TESTS=ON) to verify.

## Note on the streamed-intermediate behaviour

The physical-button path autosaves every streamed reading; with the sentinel rejected, the last *valid* reading (≈10.28 % here) stays — correct and within 0.1 % of the clean app-button re-read (10.38 %). A "wait for Test finished before accepting" gate was deliberately **not** added: we have no log of a *successful* device-Start's finish signal, so gating risks regressing the working physical-button workflow, and the range guard already fully eliminates the reported bug on both paths. Can be revisited if a successful device-Start log is captured.

The affected shot's stored value was separately corrected (TDS 10.38 %, EY 21.3 %) out-of-band.

### DIFF
--- a/qml/pages/PostShotReviewPage.qml
+++ b/qml/pages/PostShotReviewPage.qml
@@ -231,6 +231,13 @@ Page {
     // Real espresso TDS is 5–22%; below 3.0% is a calibration or empty cuvette.
     readonly property real kMinimumPlausibleTds: 3.0
 
+    // Above the R2's physical measurement range it's a device error sentinel,
+    // not a reading (the R2 emitted raw 0xFFE5 → 655.09% during a failed
+    // measurement and it was autosaved onto a shot). Symmetric with
+    // kMinimumPlausibleTds so the physical R2 Start button and the "Read TDS"
+    // button — both of which arrive via onTdsChanged — are gated identically.
+    readonly property real kMaximumPlausibleTds: 35.0
+
     // Gate by visibility so device-initiated R2 readings between shots don't
     // land on whichever shot happens to be loaded.
     //
@@ -253,6 +260,13 @@ Page {
             if (tds < postShotReviewPage.kMinimumPlausibleTds) {
                 console.debug("[PostShotReview] R2 tds", tds.toFixed(2),
                     "dropped: below threshold", postShotReviewPage.kMinimumPlausibleTds,
+                    "shotId=", editShotId,
+                    "wasMeasuring=", (typeof Refractometer !== "undefined" && Refractometer) ? Refractometer.measuring : false)
+                return
+            }
+            if (tds > postShotReviewPage.kMaximumPlausibleTds) {
+                console.debug("[PostShotReview] R2 tds", tds.toFixed(2),
+                    "dropped: above threshold", postShotReviewPage.kMaximumPlausibleTds,
                     "shotId=", editShotId,
                     "wasMeasuring=", (typeof Refractometer !== "undefined" && Refractometer) ? Refractometer.measuring : false)
                 return
--- a/src/ble/refractometers/difluidr2.cpp
+++ b/src/ble/refractometers/difluidr2.cpp
@@ -19,6 +19,14 @@
 static constexpr uint8_t PACKET_HEADER = 0xDF;
 static constexpr int PACKET_MIN_LENGTH = 6;  // header(2) + func(1) + cmd(1) + datalen(1) + checksum(1)
 
+// DiFluid R2 hardware measurement ceiling (TDS %). Real coffee never approaches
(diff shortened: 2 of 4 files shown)


skialpine:
Verified locally: Qt Creator Debug build succeeds (0 errors, 0 warnings). `tst_DiFluidR2` full suite **29 passed / 0 failed** — including the 3 new regression tests (`rejectsImplausiblyHighTds` reproduces the exact field sentinel raw 65509 → 655.09%, `rejectsImplausiblyHighAverageTds` for pack 3, `acceptsTdsAtTopOfDeviceRange` guards against over-rejection). No regression in the 24 pre-existing R2 tests.

skialpine:
### Code review

Found 1 issue:

1. The class doc-comment — rewritten in this PR — still claims the driver emits `tdsChanged` for "idle polls", but no polling timer or periodic-measurement path exists in `difluidr2.cpp` (`m_measurementTimer` is a timeout watchdog, `m_initTimer` is a connect delay, and `requestMeasurement()` is the only measurement trigger and is caller-invoked). The phrase predates this PR but the same docstring block was edited here. (Stale comment — describes no existing code path; CLAUDE.md: "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'.")

https://github.com/Kulitorum/Decenza/blob/5093a1c0ae72201a2191c04c99ce201909ed9251/src/ble/refractometers/difluidr2.h#L11-L15

---

<details>

- [Worth a look] The same rewritten docstring says physically-impossible readings are "dropped here" and "consumers must still gate by context", but doesn't note that the sub-3% lower-bound filter (`kMinimumPlausibleTds`) still lives only in the QML consumer — only the upper bound is C++-enforced. A future dev could assume both bounds are gated in C++. One-line clarification in the comment already being changed. (score: 65)

</details>

skialpine:
Addressed both surviving review items in `13f61a26` (docstring-only, no functional change):
- Dropped the stale "idle polls" phrase — no polling path exists in the driver; device-initiated readings come only from the physical R2 button.
- Clarified the validation ownership split: the driver gates only the impossible upper-range sentinel; the sub-3% lower bound and context gating remain the consumer's responsibility.

The four discarded findings (scores 25-40) were intentionally left as-is: pre-existing/by-design or accurate-enough wording, not worth speculative churn.

PR #1224, opened 2026-05-19
fix(scale): drive stall detector by sample arrival, not value change (#1176)

skialpine:
## Root cause of #1176 — found, and it's a one-line app bug

The scale-feed stall detector + SAW de-jitter were fed by `ScaleDevice::weightChanged`, which is **deduped**:

```cpp
void ScaleDevice::setWeight(double weight) {
    if (m_weight != weight) {        // unchanged value is silently dropped
        m_weight = weight;
        emit weightChanged(weight);
    }
}
```

The dedup is correct for what `weightChanged` backs — the `weight` Q_PROPERTY, QML bindings, MQTT — none of which should churn on a constant reading. But `WeightProcessor::processWeight` (the stall/liveness + SAW input) was wired to the **same** deduped signal.

So a perfectly healthy scale reporting a **constant** weight through the DE1 `EspressoPreheating` dwell (a static cup at `0.0 g`) emits **zero** `weightChanged` signals → `processWeight` never runs → `m_lastWallClockMs` goes stale → the DE1-tick stall evaluator (`setCurrentFrame → checkScaleFeedStall`) sees a growing gap → **SUSPECTED → CONFIRMED → mid-shot connection-priority backoff** (scale bounce / ruined shot, e.g. GCDE-VER1's `finalWeightG=0.3`). This is exactly #1176, and the symptom the #1185/#1219/#1220 backoff machinery was chasing.

### Why it's *here*, not Qt / Android / scale firmware (all checked)

- **Qt Android path:** `QtBluetoothLE.java` `handleOnCharacteristicChanged` → `QLowEnergyControllerPrivateAndroid::characteristicChanged` emits **unconditionally** — no value comparison, no dedup.
- **Android:** `BluetoothGattCallback.onCharacteristicChanged` fires per notification and does not coalesce identical GATT notifications.
- **Decent driver:** `DecentScale::parseWeightData` → `setWeight()` with no unchanged-value skip.
- `ScaleDevice::setWeight`'s `!=` guard is the **only** suppressor in the entire chain — and the reporter reflashed the scale firmware with **no change**, consistent with the suppression being app-side.

This also explains every prior observation: "always during preheating" (only long static-zero dwell), the drift-vs-static discriminator across the 4-month corpus + today's shots, independence from connection-priority era (no-HIGH / silently-broken-HIGH / dual-HIGH) and from read-after-write, and GCDE-VER1's latest build still failing with both links BALANCED from boot.

## The fix

- Add `ScaleDevice::weightSampleReceived(double)` — emitted **unconditionally** for every accepted sample.
- Repoint **all 16** `WeightProcessor::processWeight` connect/disconnect sites in `main.cpp` (physical / flow / USB / simulated) to it.
- Leave `weightChanged` deduped for the **13** `onScaleWeightChanged` / MQTT / UI sites — bindings still don't churn on a constant reading.

Minimal, root-cause, correct layer (`ScaleDevice` — covers all scale types); no new setting, no timer, no firmware change. Makes the spurious scale-feed-stall trigger effectively impossible, which retires the enforce/mid-shot-bounce path for this cause (it was treating a symptom of this `!=` filter).

## Tests

`tests/tst_scalefeedliveness.cpp`:
1. `sampleSignalFiresEvenWhenValueUnchanged` — the `setWeight` emit contract (sample fires every time; changed deduped).
2. `constantWeightOnSampleSignalDoesNotStall` — 15 s static `0.0` on the fixed wiring trips no SUSPECTED/CONFIRMED.
3. `constantWeightOnChangedSignalStillStalls` — same feed on the old wiring still confirms the #1176 false stall, so a regression to the wrong signal can't pass silently.

Not built here (project convention: builds run in Qt Creator). Needs `-DBUILD_TESTS=ON`; run `tst_scalefeedliveness`.

Relates to #1176, #1185.

### DIFF
--- a/src/ble/scaledevice.cpp
+++ b/src/ble/scaledevice.cpp
@@ -31,6 +31,10 @@ void ScaleDevice::setSimulationMode(bool enabled) {
         m_weight = 0.0;
         m_flowRate = 0.0;
         m_batteryLevel = 85;
+        // Keep the liveness contract intact: every weight value this layer
+        // publishes also goes out on weightSampleReceived (the stall detector
+        // and SAW path listen only to that, never weightChanged). #1176.
+        emit weightSampleReceived(m_weight);
         emit weightChanged(m_weight);
         emit flowRateChanged(m_flowRate);
         emit batteryLevelChanged(m_batteryLevel);
--- a/src/ble/scaledevice.cpp
+++ b/src/ble/scaledevice.cpp
@@ -83,6 +83,11 @@ void ScaleDevice::setConnected(bool connected) {
 }
 
 void ScaleDevice::setWeight(double weight) {
+    // Unconditional: a sample arrived. Drives the scale-feed stall detector and
+    // SAW de-jitter, which must track sample arrival, not value change (#1176).
+    emit weightSampleReceived(weight);
+    // Deduped: only on a genuine value change. Drives the `weight` Q_PROPERTY,
+    // QML bindings and MQTT — a constant reading must not churn those.
     if (m_weight != weight) {
         m_weight = weight;
         emit weightChanged(weight);
--- a/src/ble/scaledevice.cpp
+++ b/src/ble/scaledevice.cpp
@@ -86,8 +86,9 @@ void ScaleDevice::setWeight(double weight) {
     // Unconditional: a sample arrived. Drives the scale-feed stall detector and
     // SAW de-jitter, which must track sample arrival, not value change (#1176).
     emit weightSampleReceived(weight);
-    // Deduped: only on a genuine value change. Drives the `weight` Q_PROPERTY,
-    // QML bindings and MQTT — a constant reading must not churn those.
+    // Deduped: only on a genuine value change. Drives the `weight` Q_PROPERTY
+    // and QML bindings (and onScaleWeightChanged, which feeds MQTT) — a
+    // constant reading must not churn those.
     if (m_weight != weight) {
         m_weight = weight;
         emit weightChanged(weight);
--- a/src/ble/scaledevice.h
+++ b/src/ble/scaledevice.h
@@ -70,10 +70,14 @@ public slots:
 signals:
     void connectedChanged();
     void weightChanged(double weight);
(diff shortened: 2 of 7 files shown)


skialpine:
### Code review

Found 1 issue:

1. New test uses `QTEST_MAIN` instead of `QTEST_GUILESS_MAIN` (docs/CLAUDE_MD/TESTING.md says: *"Create `tests/tst_yourtest.cpp` with `QTEST_GUILESS_MAIN(tst_YourTest)`"*). `QTEST_MAIN` spins up a full GUI `QApplication`; this test exercises only signals + a fake clock. Every sibling does it right — `tst_weightprocessor.cpp:841`, `tst_saw.cpp:334`, `tst_settling.cpp:311` — including the one testing the same `WeightProcessor`.

https://github.com/Kulitorum/Decenza/blob/cdddc9aa2570e0e40845ce2c13db53c3f0107aed/tests/tst_scalefeedliveness.cpp#L113-L117

<details>

- [Worth a look] `weightSampleReceived` doc comment says it fires for "EVERY accepted weight sample", but `ScaleDevice::setSimulationMode()` emits `weightChanged` directly, bypassing `setWeight()` and the new signal — narrow the wording to "every sample that passes through `setWeight()`" (score: 50)
- [OK to skip] `setWeight` comment says the deduped `weightChanged` "Drives the `weight` Q_PROPERTY, QML bindings and MQTT" — MQTT is actually an indirect periodic read via `MainController::onScaleWeightChanged`/`MachineState`, not a direct slot (score: 50)
- [OK to skip] `tst_scalefeedliveness.cpp` comment lists what `startExtraction()` resets but omits `resetStallTracking()`, which is central to this stall test (score: 50)

</details>

Core change verified sound: all 16 `WeightProcessor::processWeight` connect/disconnect sites moved to `weightSampleReceived` symmetrically, the 13 `onScaleWeightChanged`/MQTT/UI sites correctly stay on the deduped `weightChanged`, no double-processing, emit-before-`m_weight`-update is safe (slot takes the value by argument), and the regression test pair is well-formed.

PR #1234, opened 2026-05-19
fix(ble): keep scale connection-priority machinery off the refractometer link

skialpine:
## Summary

The DiFluid R2 refractometer reuses `QtScaleBleTransport` but is **not** a scale. On every connect it was forced to `CONNECTION_PRIORITY_HIGH`, adding a *third* HIGH BLE link alongside the DE1 and scale. On the Android tablet the GATT scheduler then tore the weakest link (the R2) down mid service-discovery (`CONTROLLER ERROR: ConnectionError (state=Unconnected)`, ~6 s in), so the R2 never stayed connected and the shot-review button never flipped to **"Read TDS"** — even though auto-connect was firing correctly. Diagnosed from the on-device debug log: every R2 controller-connect was immediately followed by `Scale connection-priority: OBSERVE mode — forcing HIGH`, then a teardown.

Changes:
- New `setConnectionPriorityManaged(bool)` on the `ScaleBleTransport` interface — default **no-op**, so scales and the CoreBluetooth (macOS/iOS) path are untouched.
- `QtScaleBleTransport` gains `m_connectionPriorityManaged` (default `true`). When `false`, `onControllerConnected()` skips the connection-priority request (no forced HIGH) and arms no DE1-fault / feed-stall detection, while still emitting `connected()` so service discovery proceeds normally. The four scale-stall / DE1-fault slots also early-return, so that machinery can never touch a non-scale link.
- The refractometer marks its transport non-scale right after construction in `main.cpp`.
- Documents the single-scale invariant in `main.cpp` so a future *simultaneous* multi-scale change doesn't silently reintroduce the same >2 forced-HIGH link contention.

Scales are byte-for-byte unchanged (flag defaults `true`). +49/-1, additive.

## Test plan

- [ ] Build + deploy to the Android tablet (Samsung SM-X210).
- [ ] Power on the DiFluid R2 from the shot-review page; confirm log now shows `Connection-priority: unmanaged link (non-scale) — staying at platform default` instead of `Scale connection-priority: OBSERVE mode — forcing HIGH`.
- [ ] R2 reaches `[BLE DiFluidR2] Connected and ready for measurements` and **stays** (no recurring `CONTROLLER ERROR: ConnectionError`); shot-review button flips to "Read TDS" and remains.
- [ ] Take a TDS reading; confirm it lands on the shot.
- [ ] Regression: a normal espresso shot still logs `Requesting CONNECTION_PRIORITY_HIGH on scale` and SAW stops accurately (scale path unaffected).

### DIFF
--- a/src/ble/transport/qtscalebletransport.cpp
+++ b/src/ble/transport/qtscalebletransport.cpp
@@ -263,6 +263,17 @@ void QtScaleBleTransport::onControllerConnected() {
     QT_TRANSPORT_LOG("Controller connected!");
     m_connected = true;
 
+    if (!m_connectionPriorityManaged) {
+        // Non-scale link (refractometer): leave the connection at the
+        // platform-default interval and arm no DE1-fault / feed-stall
+        // detection. Forcing a third HIGH link here contends with the DE1 +
+        // scale and the GATT scheduler tears this (weakest) link down.
+        QT_TRANSPORT_LOG("Connection-priority: unmanaged link (non-scale) — "
+                         "staying at platform default, no detection armed");
+        emit connected();
+        return;
+    }
+
     // Connection-priority for the scale link (dual-HIGH BLE contention,
     // #1093/#1176). The DE1 always requests HIGH; the scale also requests
     // HIGH UNLESS a backoff was triggered earlier this app run, in which case
@@ -322,6 +333,7 @@ void QtScaleBleTransport::onControllerConnected() {
 }
 
 void QtScaleBleTransport::onDe1LinkFault(const QString& kind) {
+    if (!m_connectionPriorityManaged) return;  // non-scale link, no detection
     // Primary signal: a DE1-link fault clustered shortly after the scale
     // requested HIGH is the dual-HIGH contention signature (#1093/#1176).
     const bool fired = m_priority.onDe1Fault(nowMs());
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

The change is correctly additive: the scale path is byte-for-byte unchanged (the new `m_connectionPriorityManaged` flag defaults `true`), the non-scale early-return in `onControllerConnected()` still sets `m_connected` and emits `connected()` so the `DiFluidR2` discovery state machine proceeds normally, the refractometer gets its own transport instance so detector state cannot leak to a scale, and the raw `transport` pointer is safely reparented to the owning `DiFluidR2`. Consistent with the #1219/#1220/#1224/#1226 connection-priority invariants.

---

<details>

- [OK to skip] The new doc comment says "CoreBluetooth never forces HIGH" — strictly that is an implementation fact (the CoreBluetooth transport never requests connection priority) rather than a platform guarantee. Defensible as written; reword only if convenient. (score: 50)

  https://github.com/Kulitorum/Decenza/blob/e766e5fbc420662baf87a8470a9b8266d83e3aab/src/ble/transport/scalebletransport.h#L118-L122

- [Worth a look] Pre-existing `setSkipHighPriority` doc says "only QtScaleBleTransport — Android — implements it", but the new sibling doc at L120 correctly says "Android/desktop" (QtScaleBleTransport compiles for all non-Apple platforms — Android, Windows, Linux). Trivial one-line doc sync in a file this PR already modifies. (score: 55)

  https://github.com/Kulitorum/Decenza/blob/e766e5fbc420662baf87a8470a9b8266d83e3aab/src/ble/transport/scalebletransport.h#L105-L109

</details>

PR #1235, opened 2026-05-19
diag(ble): instrument R2 connect path to capture the real timeline

skialpine:
## Summary

Follow-up to #1234. The DiFluid R2 still will not hold a connection, and the two prior root-cause hypotheses (forced-HIGH BLE contention; missing keep-alive) were both wrong against the observed repro:

- When the R2 powers on, auto-connect does **not** make the shot-review button flip from "R2 Off" to "Read TDS".
- A **manual** button click connects the R2 and it **stays connected**.

That divergence points at the auto-connect path / UI-state wiring, not idle behaviour or connection priority. Rather than guess a third time, this PR adds targeted, event-driven `[R2-diag]`-tagged logging at every decision point so the next on-device capture *shows* the timeline:

- `refractometerDiscovered` lambda: entry (device, existing instance ptr, connected?), the **teardown** branch (only the *create* was logged before), and the new instance ptr — exposes DiFluidR2 create/destroy churn
- Reconnect-timer tick: attempt # + `isRefractometerConnected`
- `BLEManager::setRefractometerDevice`: old → new instance ptr
- `tryDirectConnectToRefractometer` / `scanForDevices`: which path ran, whether a scan/connect was already in flight (re-entrancy)
- `onDeviceDiscovered` R2 branch: dedup-list early-return vs which gate matched (savedMatch / userInitiated / ignored)
- `onScanFinished`: 15 s scan-cycle boundary + flag clear
- `DiFluidR2`: `connectedChanged` TRUE/FALSE transitions tagged with instance pointer + reason (ready / transport-disconnected / error)

This will definitively answer: does the auto path churn instances, does the UI-facing `refractometerConnected` ever transition, is a scan starting on top of an in-flight connect, and how the manual path differs.

Diagnostic-only — all added lines are logging. The two non-logging hunks are behaviour-preserving (named the existing `savedMatch` condition; brace-wrapped a one-line early return). No functional change; defaults and control flow unchanged.

## Test plan

- [ ] Build + deploy to the Android tablet (SM-X210), DE1 + Decent Scale + R2 all present.
- [ ] Power the R2 on and let auto-connect attempt; then manually click the button. Pull the de1 debug log.
- [ ] `grep [R2-diag]` and read the timeline: instance churn, `setRefractometerDevice` ptr vs the instance that hits `connectedChanged -> TRUE`, scan-vs-connect re-entrancy, auto vs manual path difference.

### DIFF
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -539,8 +539,11 @@ void BLEManager::onDeviceDiscovered(const QBluetoothDeviceInfo& device) {
         // Avoid duplicates
         for (const auto& existing : m_refractometerDevices) {
             if (getDeviceIdentifier(existing) == getDeviceIdentifier(device)) {
-                qDebug().noquote() << QString("[R2-diag] R2 advert dev=%1 ALREADY in discovered list — dedup return, NO refractometerDiscovered emitted")
-                    .arg(getDeviceIdentifier(device));
+                // No log here: onDeviceDiscovered fires on every BLE
+                // advertisement (~100-500 ms); the device is already logged
+                // once on first discovery below. Logging here would flood the
+                // diagnostic timeline this instrumentation exists to keep
+                // readable.
                 return;
             }
         }
@@ -556,8 +559,8 @@ void BLEManager::onDeviceDiscovered(const QBluetoothDeviceInfo& device) {
                  savedMatch ? QStringLiteral("true") : QStringLiteral("false"),
                  m_userInitiatedScaleScan ? QStringLiteral("true") : QStringLiteral("false"),
                  savedMatch ? QStringLiteral("emit refractometerDiscovered")
-                            : (m_userInitiatedScaleScan ? QStringLiteral("listing only (no auto-connect)")
-                                                         : QStringLiteral("ignored")));
+                            : (m_userInitiatedScaleScan ? QStringLiteral("listed only (no auto-connect)")
+                                                         : QStringLiteral("listed, skip auto-connect (not saved device)")));
         // Auto-connect if this is our saved refractometer
         if (savedMatch) {
             emit refractometerDiscovered(device);
@@ -605,7 +608,7 @@ void BLEManager::onDeviceDiscovered(const QBluetoothDeviceInfo& device) {
 }
 
 void BLEManager::onScanFinished() {
-    qDebug().noquote() << "[R2-diag] scan cycle finished — clearing scanningForScales/userInitiated flags";
+    qDebug().noquote() << "[R2-diag] scan cycle finished — clearing scanning/scanningForScales/userInitiated flags";
     m_scanning = false;
     m_scanningForScales = false;
     m_userInitiatedScaleScan = false;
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 4 issues:

1. Diagnostic log added inside the dedup-return of `onDeviceDiscovered` fires on every R2 BLE advertisement (~100-500 ms) for the rest of each 15 s scan window once the R2 is in the discovered list — 30-150 repeated lines per scan cycle, flooding the exact timeline this PR exists to make readable. The repo has previously removed per-sample BLE log spam for this reason (e.g. #1111/#1112). Log it once on first discovery, or gate per scan session.

https://github.com/Kulitorum/Decenza/blob/dd13470e687d6262e83aa75874d44e29c37960f4/src/ble/blemanager.cpp#L539-L545

2. `connectedChanged -> TRUE` is logged *after* `m_connected = true` and `emit connectedChanged()`, whereas both symmetric FALSE-transition logs are emitted *before* the mutation. In a PR whose sole purpose is an accurate chronological timeline, this inverts cause/effect for the TRUE case (slot side-effects appear before the line announcing the transition). Move the log above line 57.

https://github.com/Kulitorum/Decenza/blob/dd13470e687d6262e83aa75874d44e29c37960f4/src/ble/refractometers/difluidr2.cpp#L56-L61

3. The `[R2-diag] created DiFluidR2 ... connecting to` log is placed at the end of the `refractometerDiscovered` lambda — after `connectToDevice()` (L1653) and `setRefractometerDevice()` (L1659, which emits `refractometerConnectedChanged`). On a fast BLE connect this line can print *after* `transport connected`/`setRefractometerDevice`, inverting the apparent order in the captured timeline. Move it to immediately after `make_unique`.

https://github.com/Kulitorum/Decenza/blob/dd13470e687d6262e83aa75874d44e29c37960f4/src/main.cpp#L1671-L1675

4. `connectedChanged -> FALSE ... wasConnected=%2` is logged unconditionally in `onTransportDisconnected`/`onTransportError`, including when `m_connected` was never true (a transport error during service discovery, before "ready" — the dominant observed R2 failure shape). The leading `-> FALSE` reads as a disconnect when it is actually a failure-before-connect, distorting the very timeline being reconstructed. Gate the label on the prior connected state. Same pattern at line 201 (`onTransportError`).

https://github.com/Kulitorum/Decenza/blob/dd13470e687d6262e83aa75874d44e29c37960f4/src/ble/refractometers/difluidr2.cpp#L185-L189

---

<details>

- [Worth a look] `onDeviceDiscovered`'s log `-> ignored` arm is reached only after the device was already appended to `m_refractometerDevices` and `refractometersChanged` emitted, so "ignored" misstates the effect (it is listed, just not auto-connected); reword to "skip auto-connect" — one-line fix in code already changed (score: 50) — https://github.com/Kulitorum/Decenza/blob/dd13470e687d6262e83aa75874d44e29c37960f4/src/ble/blemanager.cpp#L553-L560
- [Worth a look] Minor log-label inconsistencies: `scanForDevices` labels one field `wasScanning=` but the sibling with the bare flag name; `onScanFinished` says "clearing scanningForScales/userInitiated flags" but also clears `m_scanning` (score: 50)

</details>

PR #1236, opened 2026-05-19
fix(ai): stop fabricated cross-profile grind advice (#1223)

### LINKED ISSUE #1223: [Bug] AI in app (Gemini) and Claude on my PC both giving bad advice
### What happened?

See video below
1/ Doing Dflow, which has a 42g SAW, the AI refers to grind setting in relation to yield (again) - so that's obvs wrong.
2/ I had asked it about doing a TurboTurbo or Rao Allonge shot, and the advice was based on a grind setting I had a while back with very different coffee, instead of the bean I had just used.

### Steps to reproduce

1/ did a dflow shot, well dialed in.
2/ Told AI the shot was dialled in  and asked it to recommend what to do for a TurboTurbo and rao allonge shot... 
 3/ Told AI it had made a mistake, that the grind setting for turboturbo should be closer to 28-30, and definitely not finer than the dflow shot.
4/ asked Claude MCP on my PC the exact same questions and got similarly bad answers.

### Decenza version

1.8.5 3391

### Platform

Android

### Device / OS version

12

### Scale (if applicable)

_No response_

_No response_

_No response_

### Screenshots or video

videos on ftp
https://videoblurb.com/mp/decenza/Screen_Recording_20260518_093229.mp4
https://videoblurb.com/mp/decenza/Screen_Recording_20260518_093547.mp4

### Anything else?

_No response_

skialpine:
## Summary

Phase 1 of openspec change `fix-grinder-calibration-cross-profile`. Fixes [#1223](https://github.com/Kulitorum/Decenza/issues/1223): the in-app (Gemini) and MCP (Claude) advisors both told a user to grind *finer* for TurboTurbo when it should be much coarser. Root cause: `buildGrinderCalibrationBlock` pooled all-coffee, all-time medians and picked a two-profile anchor pair → wrong-signed `conversionKey` (`-2.4` on real data), with out-of-range numbers labelled like measured data.

Rewritten model: `grind(profile, coffeeBatch) ≈ batchBaseline + UGS·conversionKey`.

- **Within-roast-batch paired Theil–Sen slope** — cancels the dominant per-batch baseline (same bean dialed 5 vs 11 across roast dates) that made the pooled slope wrong-signed.
- **Dimensionless spread gate** (`IQR ≤ ratio·|key|`) — grinder-portable; an absolute threshold is not (slope magnitude is grinder-specific).
- **Per-current-batch anchor** + **hard UGS extrapolation cap** — numbers only near validated data, anchored on a recent dialed-in shot of the current roast batch.
- **Directional fallback** — when no usable signal (the common case), emit only finer/coarser from KB ordering, never a fabricated number. `confidence` = `approximate` | `directional`.
- Advisor + `dialing_get_grinder_calibration` + KB prose enforce the same UGS-is-directional constraint. Legacy pooled/anchor-pair code deleted.

Constants tuned against a real 928-shot DB via `tools/calib_analysis.py`; only the literature-backed `cap` (Al-Shemmeri/Gagne) is grinder-independent and it alone defeats #1223. Survived two rounds of independent review (grinder-portable gate, batch keying, anchor-free direction).

## Test plan

- [x] Full suite green via Qt Creator: 2192 passed, 0 failed, 0 WARN lines
- [x] 10 new `tst_dialing_blocks` calibration tests + rewritten `tst_aimanager` rendering test
- [x] Live decenza MCP on the real DB: shot 961 `-2.4` → `confidence:directional`, TurboTurbo `coarser`/no number
- [ ] `/pr-review-toolkit:review-pr`
- [ ] Comment on #1223; request reporter DB for Phase 2 validation


skialpine:
### Code review

Found 5 issues:

1. The `dialing_get_grinder_calibration` MCP tool description still advertises the removed payload shape — `fineAnchor`/`coarseAnchor`, `source: "extrapolated"`, "unavailable until the user has pulled at least two KB-known profiles", and "a stable property of the grinder so one fetch per conversation is enough". The rewritten builder emits `confidence`, `coffeeAnchor`, `currentProfileUgsPlaced`, and `source: history|derived|directional`, and is coffee-batch-anchored (not stable across coffee changes). An LLM will look for fields that no longer exist and may treat actionable `directional` output as an error state.

https://github.com/Kulitorum/Decenza/blob/0c39a0ad8a8b71837ba992407ac415b23ff5411c/src/mcp/mcptools_dialing.cpp#L394-L407

2. `int` loop counters and count constants compared against `qsizetype` container sizes (CLAUDE.md: "Use `qsizetype` (not `int`) for container sizes — `QVector::size()`, `QList::size()` ... return `qsizetype` (64-bit on iOS/macOS). Assigning to `int` causes `-Wshorten-64-to-32` warnings"). Affects the constants below plus `for (int i = 0; i < rows.size(); ++i)` (L874) and `for (int i/j = ...; < ids.size(); ...)` (L926-927).

https://github.com/Kulitorum/Decenza/blob/0c39a0ad8a8b71837ba992407ac415b23ff5411c/src/ai/dialing_blocks.cpp#L717-L721

3. The `onTarget` dialed-in predicate derives the stop-at-weight target only from `json_extract(profile_json,'$.target_weight')` (L792), ignoring the dedicated `yield_override` column where native Decenza SAW shots persist it. The file already has `effectiveTargetWeightG()` (L58-74) as the single source of truth. The common SAW dial-in cohort (no rating, no refractometer, target in `yield_override`) fails all three predicates and is dropped, degrading `approximate` to `directional`/empty when real dial-in data exists.

https://github.com/Kulitorum/Decenza/blob/0c39a0ad8a8b71837ba992407ac415b23ff5411c/src/ai/dialing_blocks.cpp#L834-L840

4. Empty bean metadata re-pools unrelated coffees into one batch. When the resolved shot has no bean fields, `curBean == " / "`, and `normEq` matches every other metadata-less row; the undated path then reduces to the ±90-day window. A user who never fills bean fields and runs several coffees within 90 days gets them treated as one roast batch — reintroducing the exact cross-batch baseline confound (#1223) this PR removes. No empty-bean guard.

https://github.com/Kulitorum/Decenza/blob/0c39a0ad8a8b71837ba992407ac415b23ff5411c/src/ai/dialing_blocks.cpp#L769-L772

5. `tools/calib_analysis.py` defaults `--min-pair-span` to `1.0` (and the docstring usage shows `1.0`) while the shipped C++ constant `kCalibMinPairSpanUgs` is `0.75`. The tool is the documented provenance for the tuned constants (tasks.md task 1.2), so a no-flags run does not reproduce production behavior.

https://github.com/Kulitorum/Decenza/blob/0c39a0ad8a8b71837ba992407ac415b23ff5411c/tools/calib_analysis.py#L99-L101

<details>

- [Worth a look] The spread gate's `q1 = s[s.size()/4]`, `q3 = s[(3*s.size())/4]` is the full range at n<=4, not a true interquartile range — it fails safe (overstated spread -> more directional) and the constant was tuned with the same formula, but the "IQR" naming/comment can mislead; one-line comment or rename (score: 50)
- [Worth a look] `mcptools_dialing.cpp` sets `available: false` alongside a populated directional `grinderCalibration` block — spec-sanctioned and the `reason` mitigates, but the dual signal may cause some models to drop the actionable finer/coarser data (score: 50)
- [Worth a look] `tools/calib_analysis.py` `select_anchors` is annotated "faithful port of dialing_blocks.cpp selectAnchors" but this PR deletes that C++ symbol; comment is stale post-merge (score: 50)
- [OK to skip] `tools/calib_analysis.py` module docstring says it "faithfully replays the proposed within-coffee paired design" but only reproduces slope/gate logic, not the anchor/rgs/cap/directional assembly — dev-tool docstring overstatement (score: 50)
- [OK to skip] `resolvedShotId==0` now early-returns empty instead of the old "unused" contract — all three current call sites are guarded by validity checks and the header contract was updated, so risk is only theoretical for future callers (score: 50)

</details>

skialpine:
### Code review

Found 2 issues (the prior 5 are all resolved):

1. `tools/calib_analysis.py` `dialed_in` only reads `json_extract(profile_json,'$.target_weight')` and never queries `yield_override`, while the C++ side now correctly prefers `yield_override > 0 ? yield_override : json_target_weight`. The docstring states defaults "MUST stay in sync with the shipped C++ constants" but the eligibility predicate itself has diverged — the harness will silently classify the native SAW dial-in cohort (no rating, no TDS, target only in `yield_override`) as NOT dialed-in. Since this is the tool used to tune the calibration constants, the tuning was done on an incomplete view. Mirror the C++ precedence: add `COALESCE(yield_override,0)` to the SELECT and prefer it in `dialed_in`.

https://github.com/Kulitorum/Decenza/blob/ae6a42d6984cf93837d0077a7f4690b31460a2d3/tools/calib_analysis.py#L246-L257

2. `variableRpm` is declared on `GrinderEntry` and populated on ~20 variable-RPM grinders, but `buildGrinderCalibrationBlock` never consults it. `parseGrinderSetting` strips the RPM suffix from "24 1400rpm", so on a variable-RPM grinder, shots at the same dial but different RPM collapse into a single linear value — within-batch medians/slope can be computed over an RPM-conflated population and a numeric `rgs` may publish without an RPM caveat. The spread gate empirically catches this on Fred's DF83V (directional), but that is data-dependent; a different user with tight same-RPM-by-coincidence data could publish a number valid only at one RPM. At minimum, fold RPM into batch identity (so only same-RPM shots pair), or emit a `rpm` field on `coffeeAnchor` so the model knows the anchor is RPM-specific. Field comment ("the AI should know it is a second grind axis") promises behavior the code does not yet provide.

https://github.com/Kulitorum/Decenza/blob/ae6a42d6984cf93837d0077a7f4690b31460a2d3/src/ai/dialing_blocks.cpp#L869-L876

<details>

- [Worth a look] `variableRpm` doc comment in `src/core/grinderaliases.h` says "the AI should know it is a second grind axis" but the field is metadata-only today; reword as future-intent (score: 50)
- [Worth a look] `parseGrinderSetting` doc comment focuses on the `NumericWithSuffix` rejection of `"1+4"` / `"1 + 4"`; should also state that Compound grinders accept both `a+b` and `a + b` (the regex allows whitespace around the `+`) so readers don't misread the contract (score: 50)

</details>

PR #1241, opened 2026-05-20
feat(visualizer): auto-update shot metadata on edit

skialpine:
## Summary

- Adds `visualizerAutoUpdate` bool setting (default `true`, key `visualizer/autoUpdate`) to `SettingsVisualizer` with full backup/restore and MCP read/write support
- New **Auto-Update Shots** toggle in Settings → Visualizer, positioned below Auto-Upload and disabled (dimmed) when Auto-Upload is off
- `PostShotReviewPage`: tracks `pendingVisualizerUpdate` flag; fires a single visualizer PATCH in `Component.onDestruction` (true page exit only — not triggered by sub-pages opening on top). Clears the flag on any manual upload success so no double-PATCH occurs
- `shots_update` MCP tool: after a successful DB write, reads the shot's `visualizer_id`, then loads the full shot and PATCHes visualizer.coffee if `visualizerAutoUpdate` is on. Response now includes `visualizerUpdateTriggered: bool`

## Test plan

- [ ] Toggle is visible below Auto-Upload in Settings → Visualizer
- [ ] Toggle is non-interactive (dimmed) when Auto-Upload is off
- [ ] Edit a field on PostShotReviewPage, navigate to a sub-page and back — no PATCH fires mid-session
- [ ] Edit a field and tap Back — PATCH fires on exit (confirm via network log or visualizer.coffee)
- [ ] Edit a field and let auto-close timer dismiss the page — PATCH fires
- [ ] Manual upload clears the pending flag — closing afterward does not send a second PATCH
- [ ] Shot with no `visualizer_id` and auto-upload on: first upload triggers on close
- [ ] `visualizerAutoUpdate = false`: close after edit fires no PATCH
- [ ] MCP `shots_update` on a shot with `visualizer_id`: response includes `visualizerUpdateTriggered: true`, PATCH fires
- [ ] MCP `shots_update` on a shot with no `visualizer_id`: `visualizerUpdateTriggered: false`
- [ ] `settings_get category=visualizer` returns `visualizerAutoUpdate`
- [ ] `settings_set visualizerAutoUpdate=false` persists and survives restart

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,7 @@ logcat_*.txt
 !.claude/hooks/
 .DS_Store
 /.codex_tmp
+
+# Python bytecode (from tools/*.py)
+__pycache__/
+*.pyc
--- a/openspec/changes/auto-update-visualizer-shots/.openspec.yaml
+++ b/openspec/changes/auto-update-visualizer-shots/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-05-20
--- a/openspec/changes/auto-update-visualizer-shots/design.md
+++ b/openspec/changes/auto-update-visualizer-shots/design.md
@@ -1,0 +1,55 @@
+## Context
+
+Decenza already supports one-way auto-upload of fresh shots to visualizer.coffee via `VisualizerUploader` and the `visualizerAutoUpload` setting. A PATCH path (`updateShotOnVisualizerWithOverrides`) exists and is already wired in `PostShotReviewPage` for the manual "re-upload" button. The gap is that this manual step is required every time a user edits metadata — notes, rating, TDS, beans — either through the review page or via the MCP API.
+
+The feature adds a `visualizerAutoUpdate` setting and two automatic triggers: page-close on `PostShotReviewPage` and MCP shot-edit tools.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- Add `visualizerAutoUpdate` boolean setting (default `true`) to `SettingsVisualizer`.
+- Show an **Auto-Update Shots** toggle in `SettingsVisualizerTab`, subordinate to Auto-Upload.
+- On `PostShotReviewPage` close, if changes were made and the shot has a `visualizer_id`, fire `updateShotOnVisualizerWithOverrides()` (or `uploadShotFromHistoryWithOverrides()` if no id yet and auto-upload is also on).
+- On MCP shot metadata write (any field), if the shot has a `visualizer_id` and `visualizerAutoUpdate` is on, fire a PATCH.
+
+**Non-Goals:**
+
+- First-upload of un-uploaded shots from MCP edits (MCP never triggers a first upload; it only PATCHes existing cloud shots).
+- Batch re-sync of all shots with local edits not yet on visualizer.
+- Changes to the reconciliation path defined in `visualizer-upload-persistence`.
+- Any change to the auto-upload-on-shot-end path.
+
+## Decisions
+
+### Decision: Trigger on page Component.onDestruction, not a dedicated "save" signal
+
+**Rationale:** `PostShotReviewPage` already autosaves every field edit to the local DB incrementally. A page-close trigger (via `Component.onDestruction` or the StackView `onStatusChanged: Page.Inactive` exit) is the natural point to fire a single consolidated PATCH rather than patching per-keystroke. This matches how the existing manual re-upload button works (it calls `autosave()` then fires the request).
+
+**Alternative considered:** Patch on every field edit (debounced). Rejected — too chatty, unnecessary API calls, and the existing autosave already deduplicates local writes.
+
+**Chosen approach:** Track a `pendingVisualizerUpdate` boolean on `PostShotReviewPage` that is set when any field edit is saved. On page deactivation/destruction, if `pendingVisualizerUpdate` is true and the shot has a `visualizer_id`, call `updateShotOnVisualizerWithOverrides()` with the current override map; if no `visualizer_id` and auto-upload is also on, call `uploadShotFromHistoryWithOverrides()`.
+
+### Decision: MCP trigger lives in the MCP tool handler, not MainController
+
+**Rationale:** The MCP tool for shot-update already has access to the shot id, the fields being written, and can read the stored `visualizer_id` from the DB result. Adding a post-write call to `VisualizerUploader` directly from the tool handler keeps the trigger co-located with the mutation and avoids a new signal/slot chain.
+
+**Alternative considered:** Emit a signal from `ShotHistoryStorage` on any shot update and connect MainController to it. Rejected — too broad (would fire on every internal shot write, not just user-facing edits) and harder to gate on `visualizerAutoUpdate`.
+
+### Decision: Auto-Update is subordinate to Auto-Upload in the UI
+
+The toggle for Auto-Update is disabled (grayed out) when Auto-Upload is off. This communicates the dependency clearly: if you're not uploading, updating makes no sense. The underlying setting remains independently settable via MCP for programmatic control.
+
+### Decision: No separate "auto-update on first edit" first-upload from MCP
+
+If the MCP edits a shot that has never been uploaded (`visualizer_id` is empty), the auto-update path skips it silently. A first upload from MCP would require credentials and a full shot payload fetch; that scope belongs to a future change. The QML page path (PostShotReviewPage close) does trigger a first upload if `visualizerAutoUpload` is also on.
+
+## Risks / Trade-offs
+
+- **Double-PATCH on page close + manual button**: If the user manually taps the upload button and then immediately closes the page, two PATCHes fire in close succession. Visualizer.coffee is idempotent on PATCH so this is safe, just slightly wasteful. Mitigation: clear `pendingVisualizerUpdate` when the manual upload succeeds.
+- **MCP trigger fires without user awareness**: An MCP agent editing a shot will silently PATCH visualizer. This is the intended behavior, but users who share their API credentials broadly should understand this. The toggle lets them opt out.
+- **No retry on failure**: Auto-triggered updates do not queue for retry. A transient network error silently drops the update. The user can still manually re-upload from `PostShotReviewPage`. Full retry/queuing is out of scope.
+
+## Open Questions
+
+- Should the Auto-Update toggle also trigger a PATCH when the MCP `shots_update` tool is called and there is no `visualizer_id` yet but a prior upload exists that wasn't linked? — Deferred; the reconciliation pass in `visualizer-upload-persistence` handles this.
--- a/openspec/changes/auto-update-visualizer-shots/proposal.md
+++ b/openspec/changes/auto-update-visualizer-shots/proposal.md
@@ -1,0 +1,33 @@
+## Why
+
+When a user edits shot metadata (notes, rating, TDS, etc.) in the post-shot review page or via the MCP API, those changes are saved locally but are not automatically reflected on visualizer.coffee. The user must manually tap the upload button to sync edits, which is easy to forget and creates drift between local and cloud state.
+
+## What Changes
+
+- Add a `visualizerAutoUpdate` boolean setting (default `true`) to `SettingsVisualizer`, stored as `"visualizer/autoUpdate"`.
+- Add an **Auto-Update Shots** toggle to `SettingsVisualizerTab.qml`, positioned directly below the existing Auto-Upload toggle, enabled only when Auto-Upload is also on.
+- When the toggle is on and the shot already has a `visualizer_id`:
+  - Automatically PATCH the shot on visualizer.coffee when `PostShotReviewPage` is closed and local changes were made.
+  - Automatically PATCH the shot when an MCP tool writes any metadata field (notes, rating, TDS, beans, etc.) to a shot that has a `visualizer_id`.
+- If a shot has no `visualizer_id` but auto-upload is also on, the existing auto-upload path already handles first-upload at shot end; no additional first-upload is triggered from these two paths.
+
+## Capabilities
+
+### New Capabilities
+
+- `visualizer-auto-update`: Automatic sync of edited shot metadata back to visualizer.coffee when the shot already has a `visualizer_id`, triggered on PostShotReviewPage close-with-changes and on MCP metadata writes.
+
+### Modified Capabilities
+
+- `settings-ui`: New toggle added to the Visualizer settings tab.
+
+## Impact
+
+- **`src/core/settings_visualizer.h/.cpp`** — new `visualizerAutoUpdate` property.
+- **`src/core/settingsserializer.cpp`** — serialize/deserialize the new setting.
+- **`src/mcp/mcptools_settings.cpp`** — expose new setting via MCP.
+- **`qml/pages/settings/SettingsVisualizerTab.qml`** — new toggle UI.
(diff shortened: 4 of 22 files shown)


skialpine:
### Code review

Found 4 issues:

1. `pendingVisualizerUpdate` not cleared before dispatching manual upload — duplicate PATCH/POST possible if user exits before response

`uploadArea.onClicked` calls `autosave()` (which calls `saveEditedShot()`, setting `pendingVisualizerUpdate = true` if edits are pending) then dispatches the network call, but does not clear the flag. `onUploadSuccess`/`onUpdateSuccess` clear it asynchronously. If the user exits the page before the HTTP response arrives, `Component.onDestruction` → `maybeAutoUpdateVisualizer()` fires a second request against the same edits. Fix: add `pendingVisualizerUpdate = false` at the top of `onClicked` before the network dispatch.

https://github.com/Kulitorum/Decenza/blob/c8c3dba8329aace35aae8bdee6125cf9cfe1bcae/qml/pages/PostShotReviewPage.qml#L1649-L1659

2. `onUploadFailed` does not clear `pendingVisualizerUpdate` — silent retry fires on page exit after a known failure

`onUploadSuccess` and `onUpdateSuccess` both clear the flag, but `onUploadFailed` does not. A user who sees an upload error and exits the page will silently trigger another request via `maybeAutoUpdateVisualizer()` in `Component.onDestruction`.

https://github.com/Kulitorum/Decenza/blob/c8c3dba8329aace35aae8bdee6125cf9cfe1bcae/qml/pages/PostShotReviewPage.qml#L585-L589

3. New Auto-Update toggle uses `onCheckedChanged` instead of `onToggled` — codebase has an explicit prior fix for this

Commit `bc9a0d9c` ("Fix profile editor falsely detecting unsaved changes") migrated switches from `onCheckedChanged` to `onToggled` to prevent handlers firing during component initialization. The `SettingsVisualizer` `NOTIFY` signals on write create a cycle risk: binding init → `onCheckedChanged` → `setVisualizerAutoUpdate` → signal → binding re-eval → `onCheckedChanged` again. The new toggle (and the four pre-existing ones in this file) still use the old pattern.

https://github.com/Kulitorum/Decenza/blob/c8c3dba8329aace35aae8bdee6125cf9cfe1bcae/qml/pages/settings/SettingsVisualizerTab.qml#L241-L246

4. `Component.onDestruction` comment overstates the guarantee

The comment says "Fire deferred visualizer PATCH exactly once, when the page is truly being removed" but `maybeAutoUpdateVisualizer()` only fires when three conditions hold: `pendingVisualizerUpdate`, `Settings.visualizer.visualizerAutoUpdate`, and `MainController.visualizer` non-null. A developer reading the comment would expect a PATCH to always fire on destruction.

https://github.com/Kulitorum/Decenza/blob/c8c3dba8329aace35aae8bdee6125cf9cfe1bcae/qml/pages/PostShotReviewPage.qml#L37-L44

---

<details>

- [Worth a look] `visualizerUpdateTriggered` in the MCP `shots_update` response is set when `requestShot` is queued, not when the PATCH is actually dispatched — `visualizerUpdateQueued` would be more accurate for LLM consumers (score: 72)

</details>

skialpine:
**Additional finding** (missed in initial review):

5. Missing `shot.isValid()` guard in MCP `shotReady` lambda — zero-field PATCH could wipe visualizer metadata

The `shotReady` lambda in `mcptools_write.cpp` calls `updateShotOnVisualizerWithOverrides` without first checking `shot.isValid()`. The canonical pattern in `maincontroller.cpp` (line 2699) explicitly guards on this before any PATCH and logs + bails early. If `requestShot` returns a default-constructed `ShotProjection` (shot deleted in the narrow window between `updateShotMetadataStatic` and the callback, or any DB read failure), the PATCH fires with `id == 0` and empty/zero fields — effectively wiping the shot's metadata on visualizer.coffee.

Fix mirrors the maincontroller pattern:
```cpp
QObject::disconnect(*conn);
if (!shot.isValid()) {
    qWarning() << "MCP shots_update: shot" << shotId << "no longer exists; skipping visualizer PATCH";
    return;
}
visualizerUploader->updateShotOnVisualizerWithOverrides(visualizerId, shot, vizOverrides);
```

https://github.com/Kulitorum/Decenza/blob/c8c3dba8329aace35aae8bdee6125cf9cfe1bcae/src/mcp/mcptools_write.cpp#L199-L207

skialpine:
**Two more findings** (from late-completing review agent):

6. MCP `shots_update` `requestShot` broadcasts globally and clobbers in-progress `PostShotReviewPage` edits

The `shotReady` connection in `mcptools_write.cpp` calls `shotHistory->requestShot(shotId)`. `shotReady` is a global signal — `PostShotReviewPage`'s own listener (line 130) receives every emission and unconditionally executes `editShotData = shot` for a matching shotId (line 132), replacing all in-progress edits with the DB values. The page author's existing comment at line 194 documents this exact hazard ("onShotReady, clobber an in-progress edit, and orphan the undo"). A user editing a shot via the UI while an MCP client calls `shots_update` on the same shot loses their in-progress changes.

The fix is to skip the `requestShot` round-trip entirely: the `visualizer_id` SELECT inside the `withTempDb` block already has the DB connection open — load the full `ShotProjection` there and queue the PATCH directly without broadcasting `shotReady`.

https://github.com/Kulitorum/Decenza/blob/c8c3dba8329aace35aae8bdee6125cf9cfe1bcae/src/mcp/mcptools_write.cpp#L199-L211

https://github.com/Kulitorum/Decenza/blob/c8c3dba8329aace35aae8bdee6125cf9cfe1bcae/qml/pages/PostShotReviewPage.qml#L129-L133

---

7. `maybeAutoUpdateVisualizer` passes stripped `editShotData` as base shot, clearing `profile_title` on visualizer.coffee on every auto-update

After the first `saveEditedShot()` call, `editShotData` becomes a plain JS object (via `Object.assign({}, editShotData)`). Q_GADGET properties like `profileName` are non-enumerable in V4 and are silently dropped — this is documented in the existing upload-button comment at line 1684. When `maybeAutoUpdateVisualizer()` calls `updateShotOnVisualizerWithOverrides(visualizerId, editShotData, overrides)`, the C++ method reads `shotData.profileName` at line 319 via `setStr("profile_title", &ShotProjection::profileName)` (no empty-string guard) and includes `"profile_title": ""` in the PATCH body, clearing the shot's profile title on visualizer.coffee.

Since `pendingVisualizerUpdate` is set on every `saveEditedShot()` and the feature defaults to enabled, this fires on every normal edit-and-close cycle.

Fix: capture `profileName` into a `property string _profileName` before the first `Object.assign` strip (in `onShotReady` when `editShotData` is still a full `ShotProjection`), and include it in `buildVisualizerOverrides()`.

https://github.com/Kulitorum/Decenza/blob/c8c3dba8329aace35aae8bdee6125cf9cfe1bcae/qml/pages/PostShotReviewPage.qml#L490-L510

https://github.com/Kulitorum/Decenza/blob/c8c3dba8329aace35aae8bdee6125cf9cfe1bcae/src/network/visualizeruploader.cpp#L316-L321

skialpine:
### Code review

Found 7 issues:

1. `visualizerId` stripped by `Object.assign` in `saveEditedShot()` — after the first edit, `editShotData.visualizerId` is `undefined`, so `maybeAutoUpdateVisualizer()` falls into the `else if (visualizerAutoUpload)` branch and fires a duplicate POST instead of a PATCH (bug: the `_profileName` workaround at line 80 proves the author was aware Q_GADGET fields are non-enumerable, but `visualizerId` was not given the same treatment)

https://github.com/Kulitorum/Decenza/blob/5701e852cbc634d218849fcd44b7efa96ab737ea/qml/pages/PostShotReviewPage.qml#L499-L518

https://github.com/Kulitorum/Decenza/blob/5701e852cbc634d218849fcd44b7efa96ab737ea/qml/pages/PostShotReviewPage.qml#L550-L559

2. `grinderBurrs` is populated in `vizOverrides` (and in `buildVisualizerOverrides()`) but the PATCH JSON builder in `visualizeruploader.cpp` has no `grinder_burrs` API field — the only grinder field is `grinder_model` (brand + model combined). Editing burrs via MCP or the review page saves locally but silently never reaches visualizer.coffee.

https://github.com/Kulitorum/Decenza/blob/5701e852cbc634d218849fcd44b7efa96ab737ea/src/mcp/mcptools_write.cpp#L147-L151

https://github.com/Kulitorum/Decenza/blob/5701e852cbc634d218849fcd44b7efa96ab737ea/src/network/visualizeruploader.cpp#L299-L322

3. `settingsSetVisualizerAutoUpdate` test passes `nullptr` for `visualizerUploader`, so the gate at line 209–221 (`if (ok && visualizerUploader && !visualizerId.isEmpty() && ... visualizerAutoUpdate())`) never executes. The test verifies the setting round-trips through `settings_set` but provides zero coverage of the auto-update dispatch logic; inverting or removing the gate would not be caught.

https://github.com/Kulitorum/Decenza/blob/5701e852cbc634d218849fcd44b7efa96ab737ea/tests/tst_mcptools_write.cpp#L137-L142

4. `visualizerAutoUpdateDefaultIsTrue` test calls `setVisualizerAutoUpdate(true)` then reads back `true` — it only tests the round-trip (already covered by `visualizerAutoUpdateRoundTrip`). The comment says "restore to original after temporarily clearing the stored value" but no clearing happens; if the default in `settings_visualizer.cpp` were changed from `true` to `false`, this test would still pass.

https://github.com/Kulitorum/Decenza/blob/5701e852cbc634d218849fcd44b7efa96ab737ea/tests/tst_settings.cpp#L96-L101

5. `Component.onDestruction` comment says the call fires "only when `pendingVisualizerUpdate && visualizerAutoUpdate` are both true" but omits the third required gate (`MainController.visualizer` non-null, line 549) and the entire first-upload branch (`else if (visualizerAutoUpload)`, line 555) — a future reader maintaining this handler will miss a whole code path.

https://github.com/Kulitorum/Decenza/blob/5701e852cbc634d218849fcd44b7efa96ab737ea/qml/pages/PostShotReviewPage.qml#L38-L47

6. `skipReason` is set to `"shot %1 no longer exists after update"` — but the shot was just successfully updated; the real cause is that `loadShotRecordStatic` returned a default-constructed `ShotProjection` (`id == 0`), likely a query error. This message is surfaced as `visualizerUpdateSkippedReason` to MCP callers (including LLMs), where "no longer exists" implies deletion rather than a load failure (CLAUDE.md: "Use human-readable strings... An AI seeing X doesn't know...").

https://github.com/Kulitorum/Decenza/blob/5701e852cbc634d218849fcd44b7efa96ab737ea/src/mcp/mcptools_write.cpp#L216-L221

7. MCP response field is named `visualizerUpdateQueued` but the PATCH is dispatched immediately and synchronously — nothing is queued. The OpenSpec tasks called for `visualizerUpdateTriggered`, which accurately describes the behavior. An LLM consumer seeing "Queued" may infer the update hasn't happened yet (CLAUDE.md MCP data conventions: field names must be unambiguous to AI consumers).

https://github.com/Kulitorum/Decenza/blob/5701e852cbc634d218849fcd44b7efa96ab737ea/src/mcp/mcptools_write.cpp#L221-L226

---

<details>

- [OK to skip] `maybeAutoUpdateVisualizer()` is only called from `Component.onDestruction`, not from `handleBack()` or `StackView.onDeactivating` — theoretically fragile, but `StackView.pop()` reliably destroys pages in standard Qt usage (score: 50)
- [OK to skip] `updateShotOnVisualizer()` doesn't check `m_uploading` on entry, so concurrent MCP + QML calls produce two in-flight PATCHes and the spinner clears on the first response — cosmetic only, no data corruption since both PATCHes send full metadata (score: 50)
- [Worth a look] `qWarning()` in `onUpdateFinished` failure branch is currently unreachable in tests (visualizerUploader is nullptr), but once tests cover the failure path they will need `QTest::ignoreMessage(QtWarningMsg, ...)` guards per TESTING.md's no-WARN rule (score: 50)

</details>

skialpine:
### Code review

No issues found at the reporting threshold (75+) on the current head ([9205947a](https://github.com/Kulitorum/Decenza/commit/9205947a309d0d827074f22e0876f1743b84e6c8)). All 2205 tests pass with no warnings.

The PR converged across seven review iterations. Substantive bugs caught and fixed along the way:

1. `editShotData.visualizerId` stripped by `Object.assign` in `saveEditedShot()` would have caused every auto-update after the first edit to fire a duplicate POST instead of a PATCH ([commit e822f112](https://github.com/Kulitorum/Decenza/commit/e822f112), [PostShotReviewPage.qml lines 82-87](https://github.com/Kulitorum/Decenza/blob/9205947a309d0d827074f22e0876f1743b84e6c8/qml/pages/PostShotReviewPage.qml#L82-L92)).
2. The first-POST branch of `maybeAutoUpdateVisualizer()` would have silently failed `isValid()` because `Object.assign` also dropped `id` and `durationSec` from `editShotData` ([commit ae152c6f](https://github.com/Kulitorum/Decenza/commit/ae152c6f), [PostShotReviewPage.qml lines 565-583](https://github.com/Kulitorum/Decenza/blob/9205947a309d0d827074f22e0876f1743b84e6c8/qml/pages/PostShotReviewPage.qml#L565-L583)).
3. The shared `VisualizerUploader` success signals had no caller identity, so an MCP-triggered PATCH on another shot would clear this page's `pendingVisualizerUpdate` and overwrite `_visualizerId` — fixed with `_firstUploadInFlight` / `_patchInFlight` filtering ([commit 66511834](https://github.com/Kulitorum/Decenza/commit/66511834), [PostShotReviewPage.qml lines 88-99](https://github.com/Kulitorum/Decenza/blob/9205947a309d0d827074f22e0876f1743b84e6c8/qml/pages/PostShotReviewPage.qml#L88-L99)).
4. The shot-completion auto-upload finishing while the page was open never refreshed `_visualizerId` because `onUploadSuccess`'s `shotId === editShotId` guard compared a Visualizer cloud string to a local int — replaced with `onUploadSucceededForShot` which carries `dbShotId` ([commit c9e52a51](https://github.com/Kulitorum/Decenza/commit/c9e52a51), [PostShotReviewPage.qml lines 598-613](https://github.com/Kulitorum/Decenza/blob/9205947a309d0d827074f22e0876f1743b84e6c8/qml/pages/PostShotReviewPage.qml#L598-L613)).
5. The maintenance-profile and too-short rejection paths in `validateUpload` emitted `uploadFailed`, which `MainController` treats as a potential migration-16 PATCH failure and unconditionally aborts the drain — split into a new `uploadSkipped` signal ([commit ed15cba5](https://github.com/Kulitorum/Decenza/commit/ed15cba5), [visualizeruploader.h lines 117-122](https://github.com/Kulitorum/Decenza/blob/9205947a309d0d827074f22e0876f1743b84e6c8/src/network/visualizeruploader.h#L117-L123)).
6. `grinderBurrs` and `beverageType` were in the visualizer override maps despite the API having no fields for them — removed and documented ([commit ed15cba5](https://github.com/Kulitorum/Decenza/commit/ed15cba5)).
7. The default-value test was a tautology that wouldn't catch a flipped hardcoded default — replaced with a fresh-instance + raw-key-removal pattern that defeats QSettings caching ([commit e822f112](https://github.com/Kulitorum/Decenza/commit/e822f112), [tst_settings.cpp lines 97-110](https://github.com/Kulitorum/Decenza/blob/9205947a309d0d827074f22e0876f1743b84e6c8/tests/tst_settings.cpp#L97-L110)).

Below-threshold items intentionally not addressed: pre-existing `Theme.scaled(N)` numeric literals across `SettingsVisualizerTab.qml` (file-wide pattern, out of scope), and the `shots_update` auto-update gate's lack of unit-test coverage (documented in the test comment; would require a mock `VisualizerUploader`).

PR #1242, opened 2026-05-20
feat(mcp): add shots_upload_to_visualizer for first-time history uploads

skialpine:
## Summary

Adds a dedicated MCP tool for first-uploading a historical shot to visualizer.coffee. This is the missing piece next to PR #1241's `shots_update` auto-PATCH path — `shots_update` only triggers a PATCH when the shot already has a `visualizer_id`, leaving no way through MCP to upload a shot that was never auto-uploaded (e.g. an older shot recorded before credentials were set up, or one where the shot-completion auto-upload failed). The manual button on the post-shot review page was previously the only path.

## What it does

`shots_upload_to_visualizer(shotId)`:
- Loads the shot's `ShotProjection` from the local DB on a background thread via `withTempDb` + `loadShotRecordStatic`, sidestepping the Q_GADGET-strip hazard the QML side has to work around with `_visualizerId` / `_profileName` captures.
- Refuses to re-upload a shot that already has a `visualizer_id` — returns an error pointing the caller at `shots_update` for PATCH instead.
- Dispatches the POST through `VisualizerUploader::uploadShotFromHistoryWithOverrides` with empty overrides so the freshly-loaded projection's current metadata (notes, enjoyment, bean info, etc.) is sent as-is.
- Responds before the network round-trip completes; the new `visualizer_id` is persisted by `MainController` via the existing `uploadSucceededForShot` signal once the POST lands, matching the manual button flow.

## Response shape

```json
{
  "success": true,
  "uploadTriggered": true,
  "message": "Upload dispatched for shot 894; the visualizer id will land in the local DB once the response arrives"
}
```

Error cases:
- Shot not found → `{"error": "Shot N not found"}`
- Already uploaded → `{"error": "Shot N is already uploaded to visualizer (id …); use shots_update to PATCH instead"}`
- DB load failure → `{"error": "Failed to load shot N for upload"}`

## Test plan

- [x] Unit test added: `shotsUploadToVisualizerRejectsMissingShotHistory` — verifies the dependency guard rejects calls when `shotHistory` is unwired. Full dispatch-path coverage still requires a mock `VisualizerUploader`; the integration test in `scripts/test_mcp.sh` covers the live path.
- [x] `2206/2206` tests pass, build clean.
- [ ] After install on the tablet: call `shots_upload_to_visualizer(shotId=894)` (the historical shot identified during the bulk-PATCH session as not yet on visualizer) and confirm a fresh `visualizer_id` lands in the local DB plus a corresponding shot appears on visualizer.coffee.

### DIFF
--- a/docs/CLAUDE_MD/MCP_SERVER.md
+++ b/docs/CLAUDE_MD/MCP_SERVER.md
@@ -87,7 +87,7 @@ Each tool has a `category` that determines the minimum access level required:
 | Category | Min Access Level | Tools |
 |----------|-----------------|-------|
 | `read` | 0 (Monitor) | machine_get_state, app_get_info, machine_get_telemetry, shots_list, shots_get_detail, shots_get_debug_log, shots_compare, profiles_list, profiles_get_active, profiles_get_detail, profiles_get_params, profiles_get_auto_load, settings_get, dialing_get_context, dialing_get_grinder_calibration |
-| `control` | 1 (Control) | machine_wake, machine_sleep, machine_start_espresso, machine_start_steam, machine_start_hot_water, machine_start_flush, machine_stop, machine_skip_frame, shots_update, backup_now, mqtt_connect, mqtt_disconnect, mqtt_publish_discovery, devices_connect_de1, devices_disconnect_scale, devices_reset_scale_priority |
+| `control` | 1 (Control) | machine_wake, machine_sleep, machine_start_espresso, machine_start_steam, machine_start_hot_water, machine_start_flush, machine_stop, machine_skip_frame, shots_update, shots_upload_to_visualizer, backup_now, mqtt_connect, mqtt_disconnect, mqtt_publish_discovery, devices_connect_de1, devices_disconnect_scale, devices_reset_scale_priority |
 | `settings` | 2 (Full) | profiles_set_active, profiles_edit_params, profiles_save, profiles_delete, profiles_create, shots_delete, settings_set, reset_saw_learning, clear_flow_calibration, apply_theme |
 
 ### Tool → Confirmation Level Mapping
@@ -111,6 +111,7 @@ Two confirmation mechanisms are used depending on where the user is:
 | shots_delete | **Confirm** | Confirm | Chat |
 | settings_set | **Confirm** | Confirm | Chat |
 | shots_update | No confirm | No confirm | — |
+| shots_upload_to_visualizer | No confirm | No confirm | — |
 
 When confirmation level is 0 (None), all tools execute immediately regardless of mechanism.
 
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 5 issues at iter-1 head ([4f7b883d](https://github.com/Kulitorum/Decenza/commit/4f7b883d)), all addressed in [3cdb4b44](https://github.com/Kulitorum/Decenza/commit/3cdb4b44):

1. New concurrent caller of `uploadShotFromHistoryWithOverrides` violates the "mutually exclusive callers" assumption explicitly documented in the uploader header. When `PostShotReviewPage` has a manual upload in flight and this MCP tool fires concurrently with a shot that hits `validateUpload`'s reject path (missing credentials, maintenance profile, or too-short), the synchronous `uploadFailed` / `uploadSkipped` signal would be mis-attributed by the page.

https://github.com/Kulitorum/Decenza/blob/4f7b883d56ad57ac6c93a85efa2ad75ec77c2dc6/src/network/visualizeruploader.h#L162-L171

2. Tool responded `{success: true, uploadTriggered: true}` even when `validateUpload` rejected the upload synchronously before any network call (credentials missing, maintenance profile, too short). The LLM caller would confidently report a successful upload that never happened.

https://github.com/Kulitorum/Decenza/blob/4f7b883d56ad57ac6c93a85efa2ad75ec77c2dc6/src/mcp/mcptools_write.cpp#L317-L322

3. Test comment claimed "The dispatch path is covered by the MCP integration test." False — `scripts/test_mcp.sh` has zero coverage for `shots_upload_to_visualizer`. The line was copy-pasted from the sibling test without verifying.

https://github.com/Kulitorum/Decenza/blob/4f7b883d56ad57ac6c93a85efa2ad75ec77c2dc6/tests/tst_mcptools_write.cpp#L253-L257

4. MCP_SERVER.md (which the `get_agent_file` MCP tool serves to external AI clients as their behavioral system prompt) had no entry for the new tool — invisible to those clients.

https://github.com/Kulitorum/Decenza/blob/4f7b883d56ad57ac6c93a85efa2ad75ec77c2dc6/docs/CLAUDE_MD/MCP_SERVER.md#L170-L179

5. `shotId <= 0` validation guard was unreachable from the test fixture because the `!shotHistory` guard fired first (different ordering from `shots_update`, which validates `shotId` before dependencies).

https://github.com/Kulitorum/Decenza/blob/4f7b883d56ad57ac6c93a85efa2ad75ec77c2dc6/src/mcp/mcptools_write.cpp#L261-L274

### Fixes in `3cdb4b44`

- The MCP handler now mirrors `validateUpload`'s three reject conditions (credentials / maintenance / duration) on the main thread before dispatching, so the shared uploader's synchronous-reject paths are never entered from this tool. That fixes both the signal-misattribution hazard (#1) and the false-success response (#2).
- Test comment rewritten honestly; added `shotsUploadToVisualizerRejectsInvalidShotId` for the `shotId <= 0` path now that guard order is fixed.
- MCP_SERVER.md updated in all three places: Shot History tool table, `control` access-level list, and confirmation-level table.
- Guard order: `shotId` validation moved above the dependency guards (matches `shots_update`).

2207/2207 tests pass on `3cdb4b44`.

PR #1243, opened 2026-05-20
ui(post-shot): replace emoji quick-rate row with inline label on slider

skialpine:
## Summary
- Removes the `QuickRatingRow` (three emoji face buttons) from the post-shot review page
- Replaces the small "Rating" label above the slider with a large "How was this shot?" label displayed inline to the left of the slider
- Eliminates the redundant dual-rating UI (faces + slider served the same purpose)

## Test plan
- [ ] Open post-shot review for an unrated shot — confirm "How was this shot?" label appears to the left of the rating slider
- [ ] Confirm no emoji face buttons appear
- [ ] Rate using the slider — confirm value persists correctly
- [ ] Open post-shot review for an already-rated shot — confirm layout looks correct

### DIFF
--- a/qml/pages/PostShotReviewPage.qml
+++ b/qml/pages/PostShotReviewPage.qml
@@ -1011,6 +1011,7 @@ Page {
                     fallback: "How was this shot?"
                     color: Theme.textColor
                     font: Theme.bodyFont
+                    Layout.preferredWidth: Math.min(implicitWidth, parent.width * 0.45)
                     Accessible.ignored: true
                 }
 
@@ -1028,7 +1029,7 @@ Page {
                         anchors.fill: parent
                         anchors.margins: Theme.scaled(4)
                         value: editEnjoyment
-                        accessibleName: TranslationManager.translate("rating.quick.prompt", "How was this shot?") + " " + value + " " + TranslationManager.translate("postshotreview.unit.percent", "percent")
+                        accessibleName: TranslationManager.translate("rating.quick.prompt", "How was this shot?")
                         onValueModified: function(newValue) {
                             editEnjoyment = newValue
                             postShotReviewPage.autosave("rating")
--- a/qml/pages/PostShotReviewPage.qml
+++ b/qml/pages/PostShotReviewPage.qml
@@ -1001,40 +1001,22 @@ Page {
                 visible: postShotReviewPage.advancedMode && (editShotData.phaseSummaries || []).length > 0
             }
 
-            // Rating (moved to top, right after graph)
-            // QuickRatingRow — issue #1055 Layer 2. Three-icon one-tap
-            // rating row, visible when the user hasn't rated this shot
-            // yet. The precision slider below is the fine-tuning surface.
-            QuickRatingRow {
+            RowLayout {
                 Layout.fillWidth: true
-                visible: postShotReviewPage.isEditMode && editEnjoyment === 0
-                onRateClicked: function(score) {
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 4 issues:

1. **`QuickRatingRow.qml` is now dead code but not removed** (CLAUDE.md says "If you are certain that something is unused, you can delete it completely")

The PR removes the only call site but leaves `QuickRatingRow.qml` in the repo and registered in `CMakeLists.txt`. The component file, its `CMakeLists.txt` entry, and the `QuickRatingRow` import in `PostShotReviewPage.qml` should all be deleted.

https://github.com/Kulitorum/Decenza/blob/686d6b570e41de9a500cb9df362415a6eb4e1347/qml/components/QuickRatingRow.qml#L1-L10

2. **Translation key `rating.quick.prompt` is shared between the new slider label and the still-present `QuickRatingRow` component**

Both now resolve to the same translation entry. A translator who has already localized this key for the emoji-face context (e.g. a short exclamation) will see that same localized string appear as the label above the numeric slider. The two surfaces have diverged semantically. Either delete `QuickRatingRow.qml` (fixing issue #1 above also resolves this), or give the slider its own key such as `postshotreview.label.howWasThisShot`.

https://github.com/Kulitorum/Decenza/blob/686d6b570e41de9a500cb9df362415a6eb4e1347/qml/pages/PostShotReviewPage.qml#L1008-L1017

3. **`accessibleName` on `RatingInput` announces the rating value twice** (CLAUDE.md: "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'")

The parent passes `accessibleName` including `+ " " + value + " " + ... "percent"`. Inside `RatingInput.qml`, `Accessible.name` is built as `accessibleName + " " + value + "%"` — so the value is announced twice (e.g. "How was this shot? 85 percent 85%"). Fix: drop the value/percent suffix from the `accessibleName` passed by the parent and let `RatingInput` own the value announcement, which it already does.

https://github.com/Kulitorum/Decenza/blob/686d6b570e41de9a500cb9df362415a6eb4e1347/qml/pages/PostShotReviewPage.qml#L1033-L1038

4. **`Tr` label has no `Layout.preferredWidth` or `wrapMode` in the new `RowLayout`**

"How was this shot?" takes its full implicit width, leaving the remainder for the slider via `Layout.fillWidth`. On narrow phone screens (portrait, ~360 dp) this may produce a very cramped slider. Similar `Tr`+slider `RowLayout` patterns elsewhere in the file constrain the label width. Add a `Layout.preferredWidth` (e.g. a fraction of the row) or set `wrapMode: Text.Wrap` with a capped width.

https://github.com/Kulitorum/Decenza/blob/686d6b570e41de9a500cb9df362415a6eb4e1347/qml/pages/PostShotReviewPage.qml#L1004-L1017

---

<details>

- [OK to skip] `font: Theme.bodyFont` on `Tr` instance flagged as a font-group conflict — false positive; this pattern is used in 10+ other `Tr` instances throughout the codebase without issue (score: 0)
- [OK to skip] `autosave("rating", true)` → `autosave("rating")`: the `true` (finalize) arg was needed for the one-tap QuickRatingRow; the slider correctly finalizes via `onActiveFocusChanged` instead — cosmetic undo-grouping difference only, no data loss (score: 25)
- [Worth a look] `// Notes (moved to top, right after rating)` — archaeology comment, the "moved to top" phrasing is no longer meaningful; simplify to `// Notes` (score: 35)

</details>

PR #1246, opened 2026-05-21
feat(scale): add WiFi support for Half Decent Scale via WebSocket

skialpine:
## Summary

- Adds a `DecentScaleWifi` sibling driver that talks to the Half Decent Scale over the firmware's `/snapshot` WebSocket (`Kulitorum/openscale`).
- User-initiated "Scan for devices" probes `hds.local` via mDNS in parallel with the BLE scan and lists a `Decent Scale (WiFi)` row when reachable. On-demand only — no idle probing.
- The WiFi driver implements **full BLE-parity** from day one: `tare`, `timer start/stop/reset`, `display on/off`, `soft_sleep on/off`, `led <r> <g> <b>`, `events on`, `rate 10k`, plus `status` frame readbacks (battery, charging, firmware version) and `button` / `power` events.
- Adds `Q_PROPERTY(bool charging)` to `ScaleDevice`. The BLE driver previously lost the charging signal by encoding it as `battery=100`; now both transports surface `charging` correctly (incidental BLE improvement, no behavior regression).
- Saved-scale addresses gain a `wifi:` prefix so #440's primary-only auto-reconnect works unchanged across transports. The `(WiFi)` suffix flows through the synthesized `name` field to every UI surface — list, status badge, toasts, diagnostics, MCP output — with no per-surface branching.
- iOS `Info.plist` extended with `NSBonjourServices` (`_http._tcp`) and the local-network usage description amended.

OpenSpec change: [openspec/changes/add-wifi-scale-support/](openspec/changes/add-wifi-scale-support/). Proposal, design (14 decisions including BLE-machinery-we-deliberately-don't-replicate, button-event encoding, sleep semantics), tasks, and a `wifi-scale-discovery` capability spec.

**Progress: 62 / 77 tasks complete.** 15 deferred and clearly marked in tasks.md — Android `NsdManager` spike, iOS plist manual verification, timing-sensitive reconnect/503 tests, BLEManager integration test harness, manual test matrix.

## Test plan

- [x] Build via Qt Creator MCP — 0 errors / 0 warnings (148s clean, 4s incremental).
- [x] `tst_DecentScaleWifi` — 12 / 12 passing. Fake `QWebSocketServer` harness covers: snapshot parsing, malformed-frame drop, connect-time `rate 10k` + `events on` + `status` ordering, all command frames, `sleep()` → `sleepCompleted` emission, LED clamping, status-frame battery/charging/firmware-version, button event encoding (`0x1101` / `0x1202`), power-event reconnect suppression.
- [x] `tst_WifiScaleDiscovery` — 3 / 3 passing. `localhost` resolves, invalid hostname times out cleanly, rapid probes cancel in-flight.
- [x] `tst_ScaleProtocol` — 37 / 37 passing including two new BLE charging tests verifying `chargingChanged` fires on `battByte == 0xFF` while preserving `batteryLevel == 100`.
- [x] `openspec validate add-wifi-scale-support --strict --no-interactive`.
- [ ] **Manual** — scan on macOS / Windows desktop with a real HDS on the LAN: WiFi row appears, weight streams, tare works, timer / LED / display commands take effect on the scale.
- [ ] **Manual** — switch BLE ↔ WiFi for the same physical scale and confirm clean swap (`physicalScale->type()` change path in main.cpp).
- [ ] **Manual iOS** — first scan prompts the local-network permission dialog; deny path is graceful.
- [ ] **Manual Android** — confirm `QHostInfo::lookupHost("hds.local")` actually resolves on the Galaxy Tab S9. If not, the deferred `NsdManager` JNI spike (task 6.x) lands as a follow-up.
- [ ] **Manual** — saved scale rehydration: save WiFi entry, restart, auto-reconnect runs.


skialpine:
### Code review

Found 7 issues:

1. **`WifiScaleDiscovery` lazy-init has two construction sites with different `scaleFound` lambdas; whichever runs first wins.** App starts with saved WiFi scale → `tryDirectConnectToScale` installs the auto-reconnect-only lambda → later user taps Scan → `if (!m_wifiDiscovery)` short-circuits → the list-populating lambda is never connected → WiFi scale never appears in the picker.

https://github.com/Kulitorum/Decenza/blob/61fa1d35d28ce4d27d804f1afab7cdd6290c9764/src/ble/blemanager.cpp#L1081-L1103

https://github.com/Kulitorum/Decenza/blob/61fa1d35d28ce4d27d804f1afab7cdd6290c9764/src/ble/blemanager.cpp#L1011-L1045

2. **`DecentScaleWifi::onRecognitionTimeout` uses `QTimer::singleShot(0, ...)` to defer the hostname re-attempt "so the close settles before we open again."** The comment admits the timer is waiting on a socket-close state transition. ([root CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/61fa1d35d28ce4d27d804f1afab7cdd6290c9764/CLAUDE.md) says *"Never use timers as guards/workarounds. … 'suppress X until Y has happened' should be a boolean cleared by the Y event, not a timer."*) Same construct also creates a race: if `singleShot(0)` runs before `disconnected` fires, the lambda clears `m_userInitiatedShutdown` first, so `onDisconnected` then sees `wasExpected=false` and schedules an unwanted second reconnect that latches `m_reconnectAttempted`, blocking the next real reconnect.

https://github.com/Kulitorum/Decenza/blob/61fa1d35d28ce4d27d804f1afab7cdd6290c9764/src/ble/scales/decentscalewifi.cpp#L260-L275

3. **The 3s reconnect path in `onDisconnected` opens the socket directly, bypassing `attemptTarget` and the recognition timer.** `m_recognized` is never reset and `m_recognitionTimer` is never started — every reconnect after an unexpected drop skips the "first HDS frame within 5s or close" guard that the rest of the design relies on. The IP-cache update never re-runs either.

https://github.com/Kulitorum/Decenza/blob/61fa1d35d28ce4d27d804f1afab7cdd6290c9764/src/ble/scales/decentscalewifi.cpp#L118-L132

4. **WiFi branch in `BLEManager::tryDirectConnectToScale` never starts `m_scaleConnectionTimer`.** BLE paths (iOS and Android/desktop) both start it; the timeout drives `scaleConnectionFailed = true` and emits `flowScaleFallback()`. With a saved WiFi scale that's unreachable, neither the "scale not found" UI nor the FlowScale fallback dialog ever fires.

https://github.com/Kulitorum/Decenza/blob/61fa1d35d28ce4d27d804f1afab7cdd6290c9764/src/ble/blemanager.cpp#L1078-L1104

5. **`DecentScaleWifi::type()` returns `"decent"` but the `scaleDiscovered` signal carries `"decent-wifi"`, so the type-change guard always recreates the driver.** `resolveScaleType("decent") != resolveScaleType("decent-wifi")` — the "same-type reuse" branch (which calls `connectToHost` on the existing driver and re-wires IP-cache callbacks) is dead code for WiFi. Manifests when `scaleReconnectTimer` re-fires after internal driver retries fail.

https://github.com/Kulitorum/Decenza/blob/61fa1d35d28ce4d27d804f1afab7cdd6290c9764/src/main.cpp#L1380-L1421

https://github.com/Kulitorum/Decenza/blob/61fa1d35d28ce4d27d804f1afab7cdd6290c9764/src/ble/scales/decentscalewifi.h#L37-L40

6. **`ScaleDevice::setConnected(false)` calls `setBatteryLevel(-1)` but doesn't reset `m_charging`.** After a reconnect, if the scale is no longer charging, the `charging` property stays stale at `true` until the next status packet — up to 30s on BLE (next LED-response keepalive). One-line symmetric fix next to the existing `setBatteryLevel(-1)` call.

https://github.com/Kulitorum/Decenza/blob/61fa1d35d28ce4d27d804f1afab7cdd6290c9764/src/ble/scaledevice.cpp#L80-L88

7. **`bleTransport()` comment doesn't mention WiFi as a null-transport case.** The PR adds `DecentScaleWifi` (factory bypasses `makeScale<T>`, so `m_bleTransport` stays null), but the comment still lists only "FlowScale / USB / simulated."

https://github.com/Kulitorum/Decenza/blob/61fa1d35d28ce4d27d804f1afab7cdd6290c9764/src/ble/scaledevice.h#L41-L50

<details>

- [OK to skip] `wifiScaleIp` added directly to `Settings` instead of `SettingsNetwork` (root CLAUDE.md says *"Settings go in their domain sub-object"*). Mitigated: SETTINGS.md explicitly carves out scale settings as "candidate for a future SettingsHardware extension"; the PR follows the pre-existing pattern of `scaleAddress`/`scaleType`/`scaleName` on `Settings`. (score: 50)
- [OK to skip] 3s `QTimer::singleShot` reconnect delay is a debounce/backoff (CLAUDE.md says *"Everything else — including debounce — should use event-based flags"*). No useful event to wait on (ESP32 readiness isn't observable); matches established pattern in `bookooscale.cpp` / `acaiascale.cpp`. (score: 50)

</details>

skialpine:
### Code review

Found 7 issues:

1. **Two new test files use `QTEST_MAIN` instead of `QTEST_GUILESS_MAIN`.** [`docs/CLAUDE_MD/TESTING.md`](https://github.com/Kulitorum/Decenza/blob/ebf66f41865ecdfc560e16594f1568261d3f2ce4/docs/CLAUDE_MD/TESTING.md#L205-L207) says *"Create `tests/tst_yourtest.cpp` with `QTEST_GUILESS_MAIN(tst_YourTest)`"* — neither test needs a GUI loop (one drives a fake `QWebSocketServer`, the other only calls `QHostInfo::lookupHost`).

https://github.com/Kulitorum/Decenza/blob/ebf66f41865ecdfc560e16594f1568261d3f2ce4/tests/tst_decentscalewifi.cpp#L423-L427

https://github.com/Kulitorum/Decenza/blob/ebf66f41865ecdfc560e16594f1568261d3f2ce4/tests/tst_wifiscalediscovery.cpp#L57-L60

2. **`beginWifiFallbackToBleScan` arms the flag + connection timer BEFORE the `isBluetoothAvailable()` check.** If BT is off, the function returns early but `m_wifiFallbackToBleActive = true` and the 20s timer stay armed. If the user then triggers a manual scan and BT becomes available, the `isFallbackCandidate` gate in `onDeviceDiscovered` auto-connects to any Decent BLE scale found — bypassing the user-scan-is-list-only guard. Fix: move the BT-availability check above the flag/timer arming.

https://github.com/Kulitorum/Decenza/blob/ebf66f41865ecdfc560e16594f1568261d3f2ce4/src/ble/blemanager.cpp#L833-L862

3. **Stale doc comment in `decentscalewifi.h:46` references `Settings::wifiScaleIp` which this commit moved to `SettingsNetwork`.** Production wiring is `settings.network()->wifiScaleIp(...)`. The comment should match.

https://github.com/Kulitorum/Decenza/blob/ebf66f41865ecdfc560e16594f1568261d3f2ce4/src/ble/scales/decentscalewifi.h#L45-L53

4. **Comment in `handlePowerFrame` still references the deleted 3s reconnect.** *"Tag this disconnect as expected so `onDisconnected` suppresses the 3 s reconnect attempt"* — but this same commit REMOVED the in-driver 3s reconnect. `m_userInitiatedShutdown = true` now serves a different purpose (log the disconnect as expected, skip `setConnected(false)` in the fallback path).

https://github.com/Kulitorum/Decenza/blob/ebf66f41865ecdfc560e16594f1568261d3f2ce4/src/ble/scales/decentscalewifi.cpp#L228-L235

5. **Class docstring claims downstream code can use `ScaleFactory::resolveScaleType` to treat WiFi and BLE drivers as the same product.** It can't — `resolveScaleType("decent")` → `DecentScale`, `resolveScaleType("decent-wifi")` → `DecentScaleWifi` (different enum values). The actual working pattern (`settings_calibration.cpp` lines 421-424) is to branch on the string explicitly. The docstring's own `main.cpp` reference at #L1383 RELIES on them being different. Recommend dropping the `resolveScaleType` suggestion and keeping only the string-branch guidance.

https://github.com/Kulitorum/Decenza/blob/ebf66f41865ecdfc560e16594f1568261d3f2ce4/src/ble/scales/decentscalewifi.h#L17-L26

6. **`kReconnectDelayMs` and `m_reconnectAttempted` are dead after this commit removed the in-driver reconnect.** Neither is read or written anywhere after the removal (only the reset assignment in `connectToHost()` remains, and even that is a no-op now). Leaving them inflates the class with misleading state.

https://github.com/Kulitorum/Decenza/blob/ebf66f41865ecdfc560e16594f1568261d3f2ce4/src/ble/scales/decentscalewifi.h#L91-L114

7. **WiFi→BLE fallback success silently overwrites the saved WiFi address with the BLE MAC.** When the fallback connects via BLE, main.cpp's scaleDiscovered handler unconditionally runs `settings.setPrimaryScale(deviceId)` + `bleManager.setSavedScaleAddress(...)` with the BLE MAC. The user's explicit WiFi preference is lost on next launch — the app will try BLE first. Fallback should be temporary (don't rewrite `primaryScale` when `m_wifiFallbackToBleActive` triggered the connect), or the user's saved WiFi should be retained as the preferred transport.

https://github.com/Kulitorum/Decenza/blob/ebf66f41865ecdfc560e16594f1568261d3f2ce4/src/main.cpp#L1430-L1442

PR #1247, opened 2026-05-21
feat(scale): Android NsdManager-based mDNS for WiFi scale

skialpine:
## Summary

- Adds Android-specific path for resolving the Half Decent Scale's WiFi endpoint, fixing the v1.7.5 gap where mDNS auto-discovery silently failed on Android and the app fell back to BLE.
- New Java helper acquires a `WifiManager.MulticastLock` at runtime (the `CHANGE_WIFI_MULTICAST_STATE` manifest perm alone is not enough — Android filters multicast frames without an active lock) and uses `NsdManager.discoverServices()` for `_http._tcp` to find the scale by service-instance name, then resolves to an IP.
- `WifiScaleDiscovery::probe()` gets an `#ifdef Q_OS_ANDROID` branch that runs the blocking JNI call on a `QThreadPool` worker and posts back to the Qt thread via `QMetaObject::invokeMethod` with a generation counter so late callbacks after cancel/timeout get dropped.
- No public API change. Desktop/iOS paths untouched. `tst_WifiScaleDiscovery` and `tst_DecentScaleWifi` both still pass (22/22).

This is the follow-up flagged in the openspec proposal (`6.3.1: Add a Kotlin helper ... wrapping NsdManager.resolveService`) — written in Java to match the existing `BleHelper.java` / `AndroidUsbScale.java` pattern rather than adding the Kotlin toolchain.

## Test plan

- [x] macOS build via Qt Creator: 0 errors, 0 warnings
- [x] `tst_WifiScaleDiscovery` + `tst_DecentScaleWifi`: 22/22 pass
- [ ] **Android device verification (required before merge):**
  - Install the APK on an Android device on the same WiFi as a powered HDS
  - Open Settings → Connections → Scan for scales
  - Confirm the HDS appears with a "(WiFi)" suffix and connects
  - Confirm tare, weight readings, and disconnect behave like the BLE driver
  - Confirm `adb logcat` shows `DecenzaWifiNsd` log lines: `acquireMulticastLock: held`, `onServiceFound: hds`, `onServiceResolved: hds -> <ip>`
- [ ] Bonus: turn off the scale's WiFi mid-session and confirm WiFi→BLE fallback still works (BLEManager's fallback path is unchanged but worth one cycle).

## Notes

- The Java helper matches services case-insensitively by name prefix (`hds*`) — so collision-renamed instances (`hds (2)`, `hds-2`, etc.) still match.
- `MulticastLock` is reference-counted, so nested probes do the right thing — each `acquire()` is paired with exactly one `release()` in the `finally` block.
- If the user's network has multiple `_http._tcp` advertisers (printer, IoT hub), they'll be enumerated by `onServiceFound` but skipped by the name filter — only an HDS-named instance gets resolved. We could further harden this by also pinging the WebSocket endpoint, but that's premature optimization until we see a false-match report in practice.

### DIFF
--- a/android/src/io/github/kulitorum/decenza_de1/WifiScaleNsdHelper.java
+++ b/android/src/io/github/kulitorum/decenza_de1/WifiScaleNsdHelper.java
@@ -1,0 +1,215 @@
+package io.github.kulitorum.decenza_de1;
+
+import android.content.Context;
+import android.net.nsd.NsdManager;
+import android.net.nsd.NsdServiceInfo;
+import android.net.wifi.WifiManager;
+import android.util.Log;
+
+import org.qtproject.qt.android.QtNative;
+
+import java.net.InetAddress;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Android helper for discovering the Half Decent Scale's WiFi endpoint.
+ *
+ * Why this exists: Android's stock resolver (getaddrinfo via QHostInfo) does
+ * NOT do mDNS reliably for ".local" names — Qt's hostname lookup will return
+ * NXDOMAIN. We use NsdManager (the Android-blessed NSD/mDNS API) to discover
+ * _http._tcp services, match the one whose instance name matches the requested
+ * hostname (e.g. "hds"), resolve it to an InetAddress, and hand the IP string
+ * back to the Qt caller.
+ *
+ * The caller (WifiScaleDiscovery on Android) MUST run discoverHdsBlocking on
+ * a non-UI thread — the method blocks on a CountDownLatch up to `timeoutMs`.
+ *
+ * Multicast lock: Android's WiFi driver filters multicast frames by default,
+ * even with CHANGE_WIFI_MULTICAST_STATE declared in the manifest. We acquire
+ * a WifiManager.MulticastLock for the duration of the discovery and release
+ * it when finished. The lock is reference-counted, so nested acquires are
+ * safe and only the matching release lowers the count.
+ */
+public class WifiScaleNsdHelper {
+    private static final String TAG = "DecenzaWifiNsd";
+
+    // openscale firmware advertises an _http._tcp service. We compare the
+    // service instance name to the bare hostname stem (e.g. "hds" for
+    // "hds.local") case-insensitively.
+    private static final String SERVICE_TYPE = "_http._tcp.";
+
+    // Multicast lock is a single shared instance keyed by class — Android's
+    // setReferenceCounted(true) means nested probes do the right thing.
+    private static WifiManager.MulticastLock sMulticastLock = null;
+
+    /**
+     * Acquire (or re-acquire) the WiFi multicast lock. Idempotent: the underlying
+     * MulticastLock is reference-counted, so each acquire must be paired with
+     * exactly one release. Returns true on success, false if the WifiManager is
+     * unavailable (the caller can still try discovery — sometimes works, but
+     * usually returns nothing).
+     */
+    public static synchronized boolean acquireMulticastLock() {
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Initial scan surfaced 6 issues scoring 50+, all addressed in [5f29dde5](https://github.com/Kulitorum/Decenza/commit/5f29dde5cad544c5950df538ad20986d2d21e76a). Re-review on the fix commit found no new high-confidence issues.

Resolved findings:

1. (75) Concurrent `NsdManager.resolveService` crashes on API 28–33 with `IllegalArgumentException(\"listener already in use\")` when two matching services arrive. Fixed via strict single-flight gate (`AtomicInteger.compareAndSet(0,1)`). [WifiScaleNsdHelper.java#L186-L221](https://github.com/Kulitorum/Decenza/blob/5f29dde5cad544c5950df538ad20986d2d21e76a/android/src/io/github/kulitorum/decenza_de1/WifiScaleNsdHelper.java#L186-L221)

2. (75) `NsdManager.resolveService` two-arg form is deprecated at API 34; project pattern is `@SuppressWarnings(\"deprecation\")` (per `DeviceShutdownService.java`, `ApkInstaller.java`). Added. [WifiScaleNsdHelper.java#L196-L200](https://github.com/Kulitorum/Decenza/blob/5f29dde5cad544c5950df538ad20986d2d21e76a/android/src/io/github/kulitorum/decenza_de1/WifiScaleNsdHelper.java#L196-L200)

3. (68) `cancelInFlight()` only bumped the generation counter; the Java `DiscoveryListener` stayed registered until end-of-timeout, so rapid re-probes hit `FAILURE_ALREADY_ACTIVE`. Now JNI-calls a new `WifiScaleNsdHelper.cancelDiscovery()` that synchronously stops the listener and unblocks the worker. [wifiscalediscovery.cpp#L132-L153](https://github.com/Kulitorum/Decenza/blob/5f29dde5cad544c5950df538ad20986d2d21e76a/src/network/wifiscalediscovery.cpp#L132-L153)

4. (65) Header comment said \"cancels the previous lookup\" — accurate on desktop but overstated for Android. Reworded to split the two transports' semantics explicitly. [wifiscalediscovery.h#L8-L20](https://github.com/Kulitorum/Decenza/blob/5f29dde5cad544c5950df538ad20986d2d21e76a/src/network/wifiscalediscovery.h#L8-L20)

5. (50) Comment misattributed the `_http._tcp` advertisement to \"openscale firmware\" — the HDS runs genuine Decent firmware. Fixed. [WifiScaleNsdHelper.java#L39-L42](https://github.com/Kulitorum/Decenza/blob/5f29dde5cad544c5950df538ad20986d2d21e76a/android/src/io/github/kulitorum/decenza_de1/WifiScaleNsdHelper.java#L39-L42)

6. (50) Comment \"stopServiceDiscovery() runs ... at the end of the timeout\" — also runs on the common early-success path. Comment removed (the new explicit `cancelDiscovery()` path makes this irrelevant). [wifiscalediscovery.cpp#L132-L153](https://github.com/Kulitorum/Decenza/blob/5f29dde5cad544c5950df538ad20986d2d21e76a/src/network/wifiscalediscovery.cpp#L132-L153)

<details>
<summary>4 below-threshold issues (25-49)</summary>

- [OK to skip] Qt-side \`m_timeoutTimer\` races the Java-side \`CountDownLatch.await\` on the same timeout — generation check absorbs the race; legitimate watchdog for worker-never-runs case (score: 25)
- [OK to skip] \`QThreadPool\` saturation could let the Qt timer fire before the worker starts — same generation check handles it; rare in practice (score: 42)
- [OK to skip] \`MulticastLock.acquire()\` throwing after partial refcount increment is heavily guarded by \`isHeld()\` check on release; contradicts documented Android behavior (score: 25)
- [OK to skip] Timer-as-guard concern on the watchdog timer — CLAUDE.md rule targets event-sequencing/debounce abuse, not multi-layer external-async fences with generation-based idempotency (score: 25)

</details>

PR #1250, opened 2026-05-22
Auto switch back to WiFi primary scale when it returns

skialpine:
## Summary
- While connected to the BLE **backup** (the WiFi→BLE fallback), the app now proactively switches back to the saved **WiFi primary** scale once it's reachable again — instead of staying on the backup until the next disconnect/restart.
- The availability check is **non-disruptive**: a TCP dial of the WiFi scale's cached IP:80 that never touches the live BLE link. Only a positive result triggers the switch, so the working backup never blips while WiFi is still down.
- **Never interrupts a shot:** the check + switch only run while the machine is idle (`Disconnected`/`Sleep`/`Idle`/`Ready`); any active phase blocks it, and idle is re-validated after the ~3 s probe.

## Details
- `BLEManager::probeWifiPrimaryReachable(ip)` — per-probe `QTcpSocket` + timeout; emits `wifiPrimaryReachable(bool)` exactly once (first of connect/error/timeout wins).
- `BLEManager::switchToWifiPrimary()` — drops the backup and connects the WiFi primary via `connectToHost()`'s cached-IP fast path; arms the connection timer so a stale cached IP still falls back to BLE.
- `main.cpp` — a 30 s poll, armed only on the WiFi-primary/BLE-backup combination, gated on idle machine phase.

Independent of #1249 (depends only on the WiFi-scale infrastructure already on `main`). Builds clean (macOS, 0 warnings).

## Test plan
- [ ] WiFi scale as primary; force a WiFi→BLE fallback (power-cycle the scale's WiFi). Confirm it lands on the BLE backup.
- [ ] Bring the WiFi scale back; on the idle page, confirm it switches back to WiFi within ~30 s (log: `Probing WiFi primary…` → `switching back from backup`).
- [ ] Start a shot while on the backup with WiFi available — confirm it does NOT switch mid-shot.
- [ ] With WiFi still down, confirm the BLE backup never blips (no disconnect/reconnect churn from the probe).

### DIFF
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -151,6 +151,7 @@ BLEManager::~BLEManager() {
     if (m_scanning) {
         stopScan();
     }
+    cancelWifiProbe();  // tear down any in-flight WiFi-primary reachability probe
     if (s_instance == this) s_instance = nullptr;
 }
 
@@ -1057,8 +1058,10 @@ void BLEManager::switchToWifiPrimary() {
     // scaleDiscovered emission then (re)creates the WiFi driver and connects via
     // connectToHost()'s cached-IP fast path (the IP we just probed reachable).
     // Arm the connection timer so a surprise failure (e.g. the cached IP was
-    // reassigned between probe and dial) still falls back to BLE via
-    // onScaleConnectionTimeout instead of leaving us on FlowScale.
+    // reassigned between probe and dial) is recovered: onScaleConnectionTimeout
+    // routes into beginWifiFallbackToBleScan() (a BLE scan) rather than leaving
+    // the scale stranded on FlowScale; FlowScale is reached only if that BLE
+    // scan also times out.
     m_wifiFallbackToBleActive = false;
     m_scaleConnectionTimer->start();
     emit disconnectScaleRequested();
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -19,6 +19,7 @@
 #include <QDir>
 #include <QTextStream>
 #include <QDateTime>
+#include <QTcpSocket>
 
 #ifdef Q_OS_ANDROID
 #include <QJniEnvironment>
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 2 issues:

1. The `wifiPrimaryReachable` signal comment says it is "emitted exactly once per probe", but a probe superseded by a later `probeWifiPrimaryReachable()` call is torn down by `cancelWifiProbe()` without emitting — so a cancelled probe emits zero times. "at most once per probe" (or a cancel caveat) would be accurate. (Today's only caller is the 30 s poll, which never overlaps a 3 s probe, so it's a contract-comment accuracy nit rather than a live bug.)

https://github.com/Kulitorum/Decenza/blob/31f10a153176cbb247851bc1769838d504f567df/src/ble/blemanager.h#L399-L402

2. The idle-gate comment lists the blocking phases as "(preheat/preinfusion/pour/steam/hot-water/flush/clean/descale)", but the code is an allow-list (`Disconnected`/`Sleep`/`Idle`/`Ready` switch; everything else blocked via `default:`). The list omits `Heating`, `Ending`, and `Refill`, which are also blocked — clearer to phrase as "any phase other than {Disconnected, Sleep, Idle, Ready}".

https://github.com/Kulitorum/Decenza/blob/31f10a153176cbb247851bc1769838d504f567df/src/main.cpp#L1285-L1290

<details>

- [Worth a look] `~BLEManager()` doesn't call `cancelWifiProbe()`, so an in-flight probe socket/timer isn't explicitly torn down at destruction. Harmless today (BLEManager is a process-lifetime singleton and the probe objects are parented to `this`), but a one-line `cancelWifiProbe();` in the destructor would match the file's cleanup discipline. (score: 50)
- [OK to skip] `switchToWifiPrimary()`'s comment ("falls back to BLE via onScaleConnectionTimeout instead of leaving us on FlowScale") is directionally correct but elides that the timeout first routes through `beginWifiFallbackToBleScan()` (a BLE scan), reaching FlowScale only on a second timeout. (score: 55)

</details>

Note: the deeper bug-class candidates raised during review (duplicate connect via timer re-arm during `physicalScale.reset()`; double/missed emit in the probe; single-scale-invariant bypass) were each verified as false positives — `unique_ptr::reset()` nulls the pointer before the destructor runs (so the reconnect re-arm branch is skipped), and the probe's emit-once guard is correct.

PR #1251, opened 2026-05-22
Don't nag on transient scale disconnects; clearer wording

skialpine:
## Summary
- The FlowScale "Scale Disconnected" notice no longer pops on **every** disconnect. A transient drop — e.g. the WiFi WebSocket dropping on screensaver wake, which auto-reconnects in ~5 s — used to immediately open a modal that then sat there, already stale, until manually dismissed.
- The notice is now surfaced only when a (re)connect attempt **actually gives up** (via `flowScaleFallback`), and it **auto-dismisses** the moment the scale reconnects (new `BLEManager::scaleConnected` signal).
- Wording is now transport-neutral — it previously said "**Bluetooth** scale" even for WiFi scales.

## Behavior
- Transient drop that reconnects in seconds → **no dialog at all**.
- Genuine failure → notice appears after the reconnect attempts give up; a mid-session drop shows "Your scale has disconnected", a startup-with-no-scale shows "No scale was detected".
- A reconnect at any point closes the notice (and removes a queued one), so a recovered drop leaves no trace.

This came out of an on-device report: putting the app to sleep and waking it popped "Scale Disconnected" even though the WiFi scale was fine — the log showed the WebSocket dropping on wake and the cached-IP reconnect restoring it ~5 s later, with the modal left stale.

Independent of #1250 (touches the pre-existing FlowScale-fallback dialog on `main`).

## Test plan
- [ ] WiFi scale connected; sleep the app (screensaver) and wake it — confirm NO "Scale Disconnected" modal (the WS blip auto-reconnects).
- [ ] Power the scale off entirely — confirm the notice appears only after the reconnect attempts give up, with transport-neutral wording.
- [ ] Bring the scale back while the notice is showing — confirm it auto-dismisses.
- [ ] Startup with a saved-but-absent scale still shows "No scale was detected".

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -1417,12 +1417,21 @@ ApplicationWindow {
             root.scaleDropPending = true
         }
         function onScaleConnected() {
-            // Scale (re)connected — clear the pending-drop state and dismiss any
-            // FlowScale notice (open or queued) so a recovered drop leaves no trace.
+            // Scale (re)connected — clear the pending-drop state and dismiss the
+            // open scale-disconnect / no-scale notice so a recovered drop leaves no
+            // trace. Queued-but-unshown popups are purged by the ScaleDevice
+            // reconnect handler ("Discard stale scale popups…" below), so we don't
+            // duplicate removeQueuedScalePopups() here.
             root.scaleDropPending = false
             if (scaleDisconnectedDialog.opened) scaleDisconnectedDialog.close()
             if (flowScaleDialog.opened) flowScaleDialog.close()
-            removeQueuedScalePopups()
+        }
+        function onDisconnectScaleRequested() {
+            // A deliberate scale switch/forget ends the mid-session-drop context, so
+            // the next FlowScale fallback isn't mislabeled "Scale Disconnected" for a
+            // freshly-selected (never-connected) scale. disconnectScaleRequested
+            // fires from connectToScale / clearSavedScale / setDisabled.
+            root.scaleDropPending = false
         }
     }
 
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 4 issues:

1. The wording was made transport-neutral for the visible message, but the **accessibility announcement** string was missed — screen-reader users still hear "No Bluetooth scale detected" while the dialog shows "No scale was detected" (CLAUDE.md: "All user-visible text must be internationalized"; accessibility announcements should match the visible label).

https://github.com/Kulitorum/Decenza/blob/21c5854cc95e36cb7504e78cd455de4bf6160c09/qml/main.qml#L1447-L1450

2. `scaleDropPending` can stick `true`: it is cleared **only** in `onScaleConnected` (a physical-scale reconnect), not when the scale is forgotten or when the screensaver-deferral clears. After drop → forget → pair a new asleep scale, `onFlowScaleFallback` reads the stale flag and shows "Your scale has disconnected" instead of the correct "No scale was detected".

https://github.com/Kulitorum/Decenza/blob/21c5854cc95e36cb7504e78cd455de4bf6160c09/qml/main.qml#L1419-L1426

3. The new `scaleConnected` partly **duplicates** the existing `Connections { target: ScaleDevice; onConnectedChanged }` handler — both call `removeQueuedScalePopups()` on reconnect. Worth consolidating. (Minor: a USB scale connecting while the dialog is open won't auto-dismiss, since USB doesn't route through `BLEManager::scaleConnected`.)

https://github.com/Kulitorum/Decenza/blob/21c5854cc95e36cb7504e78cd455de4bf6160c09/qml/main.qml#L3792-L3801

4. Comments label the notice "a FlowScale notice", but the handler dismisses **two distinct** dialogs — `scaleDisconnectedDialog` ("Scale Disconnected") and `flowScaleDialog` ("No scale detected"). "FlowScale" is the fallback measurement mode, not the disconnect dialog — misleading for maintainers (same wording at `blemanager.cpp:904` and the QML `onScaleConnected` comment).

https://github.com/Kulitorum/Decenza/blob/21c5854cc95e36cb7504e78cd455de4bf6160c09/src/ble/blemanager.h#L397-L400

Note: a "FlowScale's own connect wipes the queued notice (missing `!isFlowScale` guard)" bug was raised and checked — it's a false positive: FlowScale is never assigned as `BLEManager::m_scaleDevice` (`setScaleDevice` only ever gets `nullptr` or the physical scale), so `scaleConnected` never fires for FlowScale. A defensive `!isFlowScale` guard on `onScaleConnected` is optional. The "second `flowScaleFallback` suppressed by the once-per-session latch" concern was also checked — every drop episode is preceded by a connect that resets the latch, so the notice fires once per episode (intended).

PR #1258, opened 2026-05-23
Fix WiFi scale status-frame routing + add disconnect diagnostics

skialpine:
## Summary
- **Fix:** WiFi `status` frames were being swallowed as weight snapshots. The driver detected snapshots by the presence of a `grams` field, but per the openscale WS protocol a `status` frame **also** carries `grams`, so `handleStatusFrame` never ran — meaning **firmware_version, battery, and charging were never parsed over WiFi**. Now routes by the protocol's actual contract: a frame with no `type` is a snapshot; typed frames dispatch by `type`.
- **Diagnostics:** log the WebSocket `closeCode`/`closeReason` on disconnect, so a mid-stream drop reveals a clean close (1000/1001) vs an abnormal one (1006, no close frame).
- **Diagnostics:** log one sample of each distinct typed frame per connect, surfacing the scale's actual WS surface.
- **Test:** regression test for a status frame carrying `grams` (the real-firmware shape). The prior status tests omitted `grams`, so they passed against the buggy routing.

## Context
While investigating a stop-at-weight overshoot (a shot ran 36 g target → 51 g) it turned out the WiFi scale's WebSocket had dropped mid-pour. Chasing that surfaced two things:
1. The disconnect cause isn't visible in the logs yet → the close-code logging.
2. The WiFi scale never logged its firmware version like the BLE link does → which turned out to be the status-frame routing bug (status frames carry `grams` and were mis-handled as snapshots).

The periodic ~1–2 min WiFi drops themselves are a separate, still-open issue (not addressed here); the close-code logging is the next diagnostic step for it.

## Test plan
- [x] `tst_decentscalewifi` — 18/18 pass (incl. new `statusFrameWithGramsIsNotMistakenForSnapshot`)
- [x] Builds clean (Qt 6.11.1, 0 warnings)
- [ ] On device: a WiFi connect now logs `Firmware version: FW: x.y.z` and a `First 'status' frame this connect: {…}` line
- [ ] On device: confirm WiFi battery/charging now populate
- [ ] On device: the next mid-stream drop carries a `[closeCode=…]`

### DIFF
--- a/src/ble/scales/decentscalewifi.cpp
+++ b/src/ble/scales/decentscalewifi.cpp
@@ -243,27 +243,28 @@ void DecentScaleWifi::onTextMessageReceived(const QString& message) {
     }
     const QJsonObject obj = doc.object();
 
-    // Diagnostic: log one sample of each distinct non-snapshot frame shape per
-    // connect, so the firmware's actual WS surface is visible — in particular
-    // whether it ever sends a status frame carrying firmware_version (the BLE
-    // link reports the version; over WiFi we have never observed it). Weight
-    // snapshots are excluded — they are the only high-rate frame.
-    if (!obj.contains(QStringLiteral("grams"))) {
-        const QString shape = obj.value(QStringLiteral("type")).toString(QStringLiteral("(untyped)"));
-        if (!m_loggedFrameShapes.contains(shape)) {
-            m_loggedFrameShapes.insert(shape);
-            WIFI_LOG(QString("First '%1' frame this connect: %2")
-                     .arg(shape, QString::fromUtf8(bytes.left(200))));
-        }
-    }
-
-    // Contract per protocol: absence of "type" means weight snapshot.
-    if (obj.contains(QStringLiteral("grams"))) {
+    // Frame routing (per the openscale WS README): a frame with NO "type" is a
+    // weight snapshot; every typed frame (status/button/power/rate/error)
+    // carries "type". A status frame ALSO carries a "grams" field, so snapshots
+    // must be discriminated by the ABSENCE of "type" — keying on the presence
+    // of "grams" (as we used to) swallowed every status frame as a snapshot,
+    // which is why firmware_version, battery, and charging were never seen over
+    // WiFi.
+    const QString type = obj.value(QStringLiteral("type")).toString();
+    if (type.isEmpty()) {
         handleSnapshotFrame(obj);
         return;
     }
 
-    const QString type = obj.value(QStringLiteral("type")).toString();
+    // Diagnostic: log one sample of each distinct typed frame per connect, so
+    // the firmware's actual WS surface is visible (frame types and, for status,
+    // its fields incl. firmware_version).
+    if (!m_loggedFrameShapes.contains(type)) {
+        m_loggedFrameShapes.insert(type);
+        WIFI_LOG(QString("First '%1' frame this connect: %2")
+                 .arg(type, QString::fromUtf8(bytes.left(200))));
+    }
+
     if (type == QStringLiteral("status")) {
         handleStatusFrame(obj);
     } else if (type == QStringLiteral("button")) {
@@ -273,7 +274,8 @@ void DecentScaleWifi::onTextMessageReceived(const QString& message) {
     } else if (type == QStringLiteral("rate")) {
         handleRateFrame(obj);
     }
-    // Unknown types are silently ignored (forward-compat).
+    // "error" frames and any unknown type are captured by the diagnostic log
+    // above; no further action.
 }
 
 void DecentScaleWifi::handleSnapshotFrame(const QJsonObject& obj) {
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 3 issues:

1. The close-code logging won't reliably distinguish abnormal drops from clean closes — the very discriminator this feature is built on. Qt's `QWebSocket::closeCode()` defaults to `1000` and is only updated when a real close frame is received; abnormal TCP/WiFi drops, and the `abort()` paths in `onRecognitionTimeout`/`onError`, never set `1006`. The single `QWebSocket` is created once and reused across reconnects via `open()`, which does not reset the code/reason, so after one clean close the value also goes stale on the next drop. Net: a mid-shot abnormal drop logs `[closeCode=1000]` (the opposite of the comment), and the existing `cachedIpValidationTimeoutFallsBackToHostname` test already shows an `abort()` teardown logging `[closeCode=1000]`. Consider driving clean-vs-abnormal off `error()` / the `m_userInitiatedShutdown` state instead, and only appending `reason` when a close frame was actually received.

https://github.com/Kulitorum/Decenza/blob/7bb73697b130e5cff9b2967faae6987821229375/src/ble/scales/decentscalewifi.cpp#L174-L199

2. `s_lastLoggedProtoVersion` is `static thread_local`, so it never resets across reconnects (the driver is single-threaded, so this is effectively a process-global static). Unlike `m_firmwareVersion` and `m_loggedFrameShapes` (both cleared in `onDisconnected`), "Protocol version: N" logs at most once per process rather than once per connect. Make it a member cleared in `onDisconnected`, or drop it (the first-`status`-frame diagnostic already captures it). This path is exercised for the first time by this PR's routing fix.

https://github.com/Kulitorum/Decenza/blob/7bb73697b130e5cff9b2967faae6987821229375/src/ble/scales/decentscalewifi.cpp#L317-L326

3. The header class-doc still says snapshot frames are `{ "grams", "ms" }` without noting that `status` frames also carry `grams` — the exact protocol property this PR's routing fix depends on. As written it implies "has `grams` => snapshot", which could lead a future maintainer to revert the fix. Update it to state that the snapshot discriminant is the absence of `type`.

https://github.com/Kulitorum/Decenza/blob/7bb73697b130e5cff9b2967faae6987821229375/src/ble/scales/decentscalewifi.h#L13-L17

---

<details>

- [OK to skip] The `statusFrameWithGramsIsNotMistakenForSnapshot` comment says "battery / charging / firmware_version" but the test only asserts battery and firmware_version; charging is covered by the separate `statusFrameUpdatesBatteryAndCharging` test (score: 50)

</details>

skialpine:
### Code review (round 2)

The three findings from the first review are all properly addressed (close-code reliability, the protocol-version member, and the header class-doc). The reworked disconnect classification introduced one new issue.

Found 2 issues:

1. The reworked disconnect log can be self-contradictory. `onDisconnected` derives the prefix from `m_userInitiatedShutdown` (-> "expected") and separately appends "— transport error: ..." whenever `m_socketErrorThisConnect` is set. But `onError` sets BOTH flags for the four mapped socket errors (`HostNotFoundError`/`ConnectionRefusedError`/`NetworkError`/`SocketTimeoutError`), so an abnormal mid-session drop that surfaces as `NetworkError`/`SocketTimeoutError` (common WiFi-loss codes) logs `WebSocket disconnected (expected) — transport error: ...` and is labeled "expected" — the exact abnormal-drop case this rework set out to label correctly. (`RemoteHostClosedError` is not in the switch, so it logs "(unexpected)".) Logging-only, but it defeats the diagnostic. A transport error should force the "(unexpected)" prefix, and the transport flag should not be set for the deliberate `abort()`/503/user-close paths (which also reach `onError`). Root: `onError` setting both flags ([decentscalewifi.cpp#L498-L503](https://github.com/Kulitorum/Decenza/blob/53c4123b063ab1cc5c8351af3a11dd2c276b0054/src/ble/scales/decentscalewifi.cpp#L498-L503)) plus the prefix/suffix split here:

https://github.com/Kulitorum/Decenza/blob/53c4123b063ab1cc5c8351af3a11dd2c276b0054/src/ble/scales/decentscalewifi.cpp#L188-L203

2. Stale doc comment (pre-existing, doc-only): the `m_userInitiatedShutdown` comment says "Five sites set this to true" but there are six — the Android mDNS-resolution-failure branch in `attemptHostname()` also sets it and isn't listed.

https://github.com/Kulitorum/Decenza/blob/53c4123b063ab1cc5c8351af3a11dd2c276b0054/src/ble/scales/decentscalewifi.h#L143-L151

---

<details>

- [Worth a look] `m_socketErrorThisConnect`/`m_lastSocketErrorString` are reset in `attemptTarget()` but not in `connectToHost()`; on the Android mDNS-failure path (which never calls `attemptTarget`) a stale value can survive into a later cycle and mislabel a subsequent clean close as "transport error". Easy fix: reset them in `connectToHost` alongside the other per-cycle resets (score: 65)
- [OK to skip] The "peer close" branch reads `closeCode()`; in the rare case Qt fires `disconnected()` with neither a socket error nor a received close frame, it logs a stale "peer close (code 1000)" — the in-code comment already acknowledges this (score: 50)

</details>

PR #1259, opened 2026-05-23
Gate USB Decent scale on the Serial USB toggle

skialpine:
## Summary
- The **Serial USB toggle** now gates the USB Half Decent Scale, not just the DE1. Previously `usbScaleManager.startPolling()` ran unconditionally, so a USB-tethered scale always took over via USB even with the toggle off — you couldn't test the same scale over Bluetooth/WiFi without unplugging.
- The USB scale now polls **only when the toggle is on** (matching the DE1's existing `usbManager` gating).
- **Disabling the toggle mid-session releases a live USB scale** — new `UsbScaleManager::disconnectScale()` tears it down and emits `scaleLost`, so the existing handler falls back to BLE / WiFi / FlowScale. (`stopPolling()` alone only halts discovery; it doesn't release a connected scale.)
- Updated the toggle's label/description/accessibility text to reflect it covers the DE1 **and** the scale.

## Test plan
- [x] Builds clean (Qt 6.11.1, macOS, 0 warnings)
- [ ] Desktop: with a Half Decent Scale on USB, turn Serial USB **off** → USB scale releases; connect the same scale over BT/WiFi
- [ ] Turn it back **on** → USB scale reconnects
- [ ] DE1-over-USB gating unchanged

Note: no unit test — `UsbScaleManager` is serial/JNI hardware-bound with no existing test harness; this mirrors the established DE1 `usbManager` toggle pattern.

### DIFF
--- a/qml/pages/settings/SettingsConnectionsTab.qml
+++ b/qml/pages/settings/SettingsConnectionsTab.qml
@@ -657,14 +657,14 @@ Item {
                         Layout.fillWidth: true
                         spacing: 0
                         Text {
-                            text: TranslationManager.translate("connections.usb.serialLabel", "Serial USB (DE1 USB-C)")
+                            text: TranslationManager.translate("connections.usb.serialLabel", "Serial USB (DE1 & Scale)")
                             font.pixelSize: Theme.scaled(14)
                             color: Theme.textColor
                             Accessible.ignored: true
                         }
                         Text {
                             Layout.fillWidth: true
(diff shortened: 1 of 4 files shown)


skialpine:
Superseding this approach. Instead of gating the USB scale behind the Serial USB toggle, USB will become a first-class selectable entry in the discovered/known-devices list (like BLE and WiFi). Closing in favor of that.

PR #1261, opened 2026-05-23
Make USB scale a selectable device (not auto-connecting)

skialpine:
## Summary
Reworks the USB Half Decent Scale to behave like the BLE and WiFi scales: when detected it appears as a **selectable entry** in the device list instead of auto-connecting (which pre-empted testing the same scale over Bluetooth/WiFi). This supersedes the closed #1259 toggle-gating approach.

- **`UsbScaleManager`**: probe-confirm now records *availability* (`usbScaleAvailable`) instead of opening the scale. `connectToScale()` opens it on demand — on user selection, or auto-reconnect when it's the saved primary. Unplug clears availability (and emits `scaleLost` if connected). Single-scale invariant preserved (won't re-probe while available/connected).
- **`BLEManager`**: `setUsbScaleAvailable()` adds/removes a synthetic `"usb:decent"` entry in `discoveredScales`, mirroring the WiFi `"wifi:<host>"` entry. `connectToScale()`'s `usb` branch emits `usbConnectRequested()`. `connectToSavedScale()` / `tryDirectConnectToScale()` skip the BLE machinery for `usb:` addresses (it's not a MAC).
- **`main.cpp`**: lists USB on availability, auto-connects **only** when `usb:decent` is the saved primary, routes selection → `connectToScale()`, and registers USB via `addKnownScale`/`setPrimaryScale` on connect — so it joins Known Devices and reconnects on next launch like BLE/WiFi.
- **No QML changes** — the discovered-list + `connectToScale` path is transport-generic.

## Test plan
- [x] Builds clean (Qt 6.11.1, macOS, 0 warnings)
- [x] `tst_decentscalewifi` 20/20 (shared BLEManager paths unaffected)
- [ ] Desktop: plug in USB scale → appears in Scan/device list, does **not** auto-connect; select it → connects + shows in Known Devices
- [ ] With USB as saved primary → auto-reconnects on next launch
- [ ] With a BT/WiFi scale connected, the USB scale stays listed (not grabbed)
- [ ] Unplug → removed from list / falls back to BT/WiFi/FlowScale

### DIFF
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -18,6 +18,7 @@
 #include <QStandardPaths>
 #include <QDir>
 #include <QTextStream>
+#include <algorithm>
 #include <QDateTime>
 #include <QTcpSocket>
 
@@ -552,8 +553,17 @@ void BLEManager::connectToSavedScale() {
 
     // USB saved-scale switch: USB needs neither Bluetooth nor a direct-wake. Hand
     // off to UsbScaleManager (usbConnectRequested) and skip the BLE guards below.
-    // It connects only if the USB scale is currently available (plugged in).
     if (m_savedScaleAddress.startsWith(QStringLiteral("usb:"), Qt::CaseInsensitive)) {
+        // Bail BEFORE dropping the live scale if the USB scale isn't currently
+        // plugged in: UsbScaleManager::connectToScale() no-ops when nothing is
+        // available, which would otherwise strand the user on FlowScale with no
+        // feedback. The USB entry is present in m_scales only while probe-confirmed.
+        const bool usbPresent = std::any_of(m_scales.cbegin(), m_scales.cend(),
+            [](const ScaleEntry& e) { return e.transport == QStringLiteral("usb"); });
+        if (!usbPresent) {
+            appendScaleLog(QStringLiteral("USB scale switch ignored — scale not plugged in"));
+            return;
+        }
         if (m_scaleDevice && m_scaleDevice->isConnected())
             emit disconnectScaleRequested();
         emit usbConnectRequested();
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -450,7 +450,12 @@ void BLEManager::connectToScale(const QString& address) {
             emit disconnectScaleRequested();
         }
 
-        if (entry.transport == QStringLiteral("wifi")) {
+        if (entry.transport == QStringLiteral("usb")) {
+            // USB connect is owned by UsbScaleManager: it creates + opens the
+            // UsbDecentScale and emits its own scaleDiscovered. Don't emit
+            // scaleDiscovered here (no QBluetoothDeviceInfo / factory path).
+            emit usbConnectRequested();
+        } else if (entry.transport == QStringLiteral("wifi")) {
             // Strip the "wifi:" prefix to get the bare hostname.
             m_pendingWifiHostname = address.mid(QStringLiteral("wifi:").size());
             emit scaleDiscovered(QBluetoothDeviceInfo{}, entry.type);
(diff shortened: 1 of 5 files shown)


skialpine:
### Code review

Found 3 issues:

1. Switching from a connected USB scale to BT/WiFi double-feeds weight — the USB scale is never torn down. The USB `scaleDiscovered` handler wires the USB scale's `weightChanged`/`weightSampleReceived` into the weight pipeline but never calls `bleManager.setScaleDevice(usbScale)` (the BLE/WiFi path does). So `BLEManager::m_scaleDevice` stays null while USB is active, and when the user selects a BLE/WiFi entry, `connectToScale()`'s disconnect guard (`m_scaleDevice && isConnected()`) never fires and nothing tears down `usbScaleManager.scale()`. Both scales then feed `processWeight` simultaneously, corrupting weight/flow/stop-at-weight. This is the PR's primary intended flow (switch USB to BT/WiFi). Fix: call `setScaleDevice(usbScale)` on USB connect and/or explicitly disconnect+close the USB scale in the BLE/WiFi `scaleDiscovered` path.

https://github.com/Kulitorum/Decenza/blob/c2cfccd025c75da69c73b7e0719c58e62c02b141/src/main.cpp#L1930-L1949

2. `connectToScale()` emits `scaleDiscovered` even when `open()` failed -> a stuck "connected-but-dead" scale, and on desktop it permanently disables USB re-discovery. The probe closes the desktop serial port, then `connectToScale()` immediately reopens a new `QSerialPort` on the same port (a CH340 close/reopen race); on failure `m_scale` is non-null and `scaleDiscovered` fires anyway, wiring a dead scale active. The next poll deletes it and clears `m_confirmedPortName`, but the port stays in `m_knownPorts`, so it is never re-probed — USB discovery is dead until a physical replug. Fix: only emit when `m_scale->isConnected()`; on failure delete+null and leave availability set so the poll re-probes (or hold one open handle through to connect).

https://github.com/Kulitorum/Decenza/blob/c2cfccd025c75da69c73b7e0719c58e62c02b141/src/usb/usbscalemanager.cpp#L84-L91

3. `connectToSavedScale()`'s USB branch drops the live scale before confirming the USB scale is available -> the user is stranded on FlowScale with no feedback. Selecting USB when `usb:decent` is the saved primary but the scale is unplugged emits `disconnectScaleRequested()` (drops the current BT/WiFi scale) then `usbConnectRequested()` -> `connectToScale()` returns silently (not available). No error reaches the UI and there's no way back without manually re-selecting. Fix: don't drop the current scale unless the USB scale is actually available (track availability in BLEManager, or surface an error on failure).

https://github.com/Kulitorum/Decenza/blob/c2cfccd025c75da69c73b7e0719c58e62c02b141/src/ble/blemanager.cpp#L553-L562

---

<details>

- [Worth a look] Android: a stale-but-present USB connection is never recovered — no `errorOccurred`/disconnect wiring on the USB scale, so `m_scale` stays "connected" with a dead feed (desktop self-heals via its poll's `isConnected()` check; Android has no equivalent) (score: 65)
- [Worth a look] `scaleLost` handler's weight-signal `disconnect()`s are dead: `m_scale` is nulled before `scaleLost` is emitted, so `usbScaleManager.scale()` returns null and the disconnects are skipped (score: 58)
- [Worth a look] USB unplug emits no "Scale disconnected" UI notice and doesn't stop `scaleReconnectTimer`, unlike the BLE/WiFi disconnect path (score: 55)
- [OK to skip] Desktop `scaleLost` (connected-port-lost) branch doesn't reset `m_hasLoggedInitialPorts`, so the port-found diagnostic won't re-log on replug (Android clears it) (score: 50)

</details>

skialpine:
### Code review (round 2)

The 3 prior findings are resolved, but the fix introduced one critical regression.

Found 1 issue:

1. Android: `teardownConnectedScale()` re-enters itself via the new `connectedChanged` watchdog, causing a NULL-pointer crash. The watchdog wired in `connectToScale()` fires synchronously from `m_scale->close()` inside `teardownConnectedScale()`; at that point `m_scale` is not yet nulled and `isConnected()` is false, so the guard passes and it re-enters `teardownConnectedScale()`, which deletes + nulls `m_scale`. The outer frame then calls `m_scale->deleteLater()` on the now-null pointer -> crash. This fires on the exact unplug / stale-link recovery path the watchdog was added for, and `disconnectScale()` has the same flaw (its `close()` also triggers the watchdog). Fix: disconnect the watchdog (`disconnect(m_scale, &ScaleDevice::connectedChanged, this, nullptr)`) before calling `close()` in both `teardownConnectedScale()` and `disconnectScale()`.

https://github.com/Kulitorum/Decenza/blob/90f5af411e7c8d188756161d7ba703064c437ab2/src/usb/usbscalemanager.cpp#L131-L142

---

<details>

- [Worth a look] `scaleReconnectTimer` loops while the USB scale is the active/primary scale: the USB `scaleDiscovered` handler never stops it (the BLE/WiFi path does), and the app-resume handler treats a USB connection as "disconnected" (`physicalScale` is null for USB) and re-arms it; each tick hits the `usb:` early-return in `tryDirectConnectToScale()` and self-re-arms. Side effects are minor but it runs indefinitely (score: 65)
- [OK to skip] `teardownConnectedScale()` comment "UsbDecentScale::close() delegates this to us" is backwards — `close()` deliberately does NOT close the Android JNI handle; UsbScaleManager must. Risk: a maintainer removes the `AndroidUsbScaleHelper::close()` call thinking it's a double-close (score: 50)

</details>

skialpine:
### Code review

No issues found.

The latest commit ([`9562490`](https://github.com/Kulitorum/Decenza/commit/9562490f5d48c747542cb8406336d179fa24aadb)) adds `usb:`-prefix guards to the four BLE/WiFi `scaleReconnectTimer` arm-sites (`flowScaleFallback`, physical-scale-disconnect, app-resume, DE1-wake) so a USB primary never arms that timer. Previously it would fire once and self-terminate at the existing timeout-lambda guard — harmless but wasteful. Reviewed the four guards for correctness (operator precedence, short-circuit safety, no broken BLE/WiFi reconnect path) and CLAUDE.md compliance: all clean.

<details>
<summary>Investigated, below threshold</summary>

- [OK to skip] `~UsbScaleManager()` releasing the Android JNI USB handle — verified a non-issue. The destructor calls `stopPolling()` → `cleanupAndroidProbe(true)` → `AndroidUsbScaleHelper::close()`, so the JNI connection is released at shutdown (no leak):

  https://github.com/Kulitorum/Decenza/blob/9562490f5d48c747542cb8406336d179fa24aadb/src/usb/usbscalemanager.cpp#L17-L25

</details>

PR #1263, opened 2026-05-24
fix(ble): pause Decent scale heartbeat during DE1 BLE discovery (#1176)

skialpine:
## Summary

Decent Scale was disconnecting mid-startup on Samsung Tab A8 (#1176) even after the persistent dual-HIGH-incapable latch dropped both links to BALANCED. Root cause: when DE1 reconnects and runs service+characteristic discovery (~3 s burst), the local BLE controller is saturated, and a scale heartbeat write landing in that window fails with `CharacteristicWriteError` and the scale drops.

In the user's log (session 3 in `debug-2.log`):
- +14.1 s scale CONNECTED at BALANCED
- +19.0 s DE1 reconnect attempt 1 → DE1 char discovery +20.8 → +23.7 s
- **+22.8 s scale `CharacteristicWriteError` storm → DISCONNECTED** (a heartbeat collided with DE1 char discovery)
- Repeated at +38 s and +143 s on subsequent DE1 traffic windows

## Fix

`BleTransport` (DE1) now brackets its `discoverDetails()` window with a `serviceDiscoveryActiveChanged(bool)` signal. `DE1Device` forwards it; `BLEManager` tracks the state and forwards to a co-resident `DecentScale` via `setHeartbeatsPaused(bool)`. The periodic 1 s heartbeat skips for the duration of DE1 discovery — the next tick resumes once chars are ready. The Decent Scale tolerates skipped heartbeats fine; what it doesn't tolerate is a write storm under controller contention.

Wake-sequence writes (LCD, CCCD enables) are deliberately unaffected — they happen before any DE1 reconnect on this code path (gated by the existing `m_de1DirectConnectInFlight` serializer), and gating them would risk losing the timing-sensitive de1app pattern.

Reset paths (controller-error, `onControllerDisconnected`, `disconnect()`) also clear the flag so the scale doesn't stay paused forever after a connect-time DE1 failure.

## What I considered and rejected

- **Move the existing gate's release from "Connected" to "chars ready"** — turned out to be a no-op. `BleTransport::isConnected()` already requires `m_characteristicsReady`, so the gate already releases at chars-ready (`src/ble/bletransport.cpp:293-299`).
- **Cap-cancel on reconnect-exhausted** — practically a no-op in this log. The cap is a 15 s backstop; the natural `wasActive && !isActive` release fires within ~10–13 s on every failed DE1 attempt.
- **Slow DE1 char discovery / per-write retries on the scale** — the writes aren't lost, the link is. Retrying just extends the storm.

## Test plan

- [x] Build clean (Qt Creator, macOS Debug): 0 errors, 0 warnings, all unit tests link.
- [ ] Field verification on @GCDE-VER1's Tab A8 (cold-DE1 startup on next beta after CI runs).

Signatures to watch for in the new debug log:
- `[BLE DecentScale] Pausing heartbeats — DE1 BLE discovery in progress` appears once per DE1 reconnect.
- `[BLE DecentScale] Resuming heartbeats — DE1 BLE discovery complete` follows within ~3 s.
- Zero `CharacteristicWriteError` on the scale in the first 60 s.

Closes/addresses comments in #1176.

### DIFF
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -2,6 +2,7 @@
 #include "blecapability.h"
 #include "scaledevice.h"
 #include "protocol/de1characteristics.h"
+#include "scales/decentscale.h"
 #include "scales/scalefactory.h"
 #include "refractometers/difluidr2.h"
 #include "../core/settings_hardware.h"
@@ -1026,6 +1027,20 @@ void BLEManager::setScaleDevice(ScaleDevice* scale) {
         // Connect scale's debug log to our logging system
         connect(m_scaleDevice, &ScaleDevice::logMessage,
                 this, &BLEManager::appendScaleLog);
+        // Push current DE1-discovery state immediately so a scale that connects
+        // mid-discovery (e.g. after the gate timed out) starts in the right
+        // pause state instead of waiting for the next edge.
+        if (auto* ds = qobject_cast<DecentScale*>(m_scaleDevice)) {
+            ds->setHeartbeatsPaused(m_de1ServiceDiscoveryActive);
+        }
+    }
+}
+
+void BLEManager::setDe1ServiceDiscoveryActive(bool active) {
+    if (m_de1ServiceDiscoveryActive == active) return;
+    m_de1ServiceDiscoveryActive = active;
+    if (auto* ds = qobject_cast<DecentScale*>(m_scaleDevice)) {
+        ds->setHeartbeatsPaused(active);
     }
 }
 
--- a/src/ble/blemanager.h
+++ b/src/ble/blemanager.h
@@ -404,6 +404,12 @@ public slots:
     // Called by main.cpp when the DE1's direct-wake connection resolves
     // (connected, or the attempt ended).
     void onDe1ConnectionSettled();
+    // Track whether the DE1 transport is in BLE service+characteristic discovery
+    // and forward the state to a co-resident DecentScale so its heartbeat pauses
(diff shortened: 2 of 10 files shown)


skialpine:
### Code review

<details>

- [Worth a look] `DE1Device::disconnect()` severs all transport→this signal connections *before* invoking `m_transport->disconnect()`, so the new `setServiceDiscoveryActive(false)` emission inside `BleTransport::disconnect()` is dropped. `BLEManager::m_de1ServiceDiscoveryActive` can remain `true` (with `DecentScale::m_heartbeatsPaused` stuck on) until a subsequent BLE DE1 reconnect runs another discovery cycle that ends — narrow window, but real on the USB-takes-over path where no further BLE discovery comes. Easiest fix: reorder so `m_transport->disconnect()` runs *before* `QObject::disconnect(m_transport, ...)`, or push `false` to BLEManager from `DE1Device::disconnect()` defensively (score: 65). https://github.com/Kulitorum/Decenza/blob/411175cf9d92738d09f41c82e3bc650e36505263/src/ble/de1device.cpp#L361-L374

- [OK to skip] The doc comment on `setServiceDiscoveryActive` says "three reset call sites (chars-ready, disconnect, controller-error)" but there are four sites — `disconnect()`, `onControllerDisconnected()`, `onControllerError()`, and chars-ready. Loose grouping under "disconnect", not a behaviour bug (score: 50). https://github.com/Kulitorum/Decenza/blob/411175cf9d92738d09f41c82e3bc650e36505263/src/ble/bletransport.h#L68-L74

</details>

PR #1264, opened 2026-05-24
Normalize scaleType to canonical type-ids (rename-safe SAW/sensorLag keys)

skialpine:
## Summary
- `scaleType` doubled as a display label **and** the persistence/lookup key for per-(profile, scale) SAW learning, `sensorLag()`, the global bootstrap lag, and the known-scales registry. For BLE scales it stored the display name (e.g. `"Decent Scale"`), so renaming a display name silently orphaned SAW learning and broke `sensorLag()` lookups — the #1260 WiFi rename only escaped this because WiFi was already id-keyed.
- Normalize `scaleType` to stable **type-ids** everywhere so display names can change freely. No user-facing change (display names + UI unchanged).
- New dependency-free `ScaleTypeIds` unit (`src/ble/scales/scaletypeids.{h,cpp}`) owns the `ScaleType` enum + `scaleTypeId()` / `scaleTypeName()` / `normalizeScaleTypeId()`. The enum now models the Half Decent Scale's **three transports**: `DecentScale` (BLE), `DecentScaleWifi`, `DecentScaleUsb`.
- `BLEManager::getScaleType(device)` returns the id — this also **repairs the previously-dead WiFi→BLE fallback match** (`isFallbackCandidate` had been comparing a display name against `"decent"`).
- `Settings` normalizes on write (`setScaleType`, `addKnownScale`) + a one-time, flag-guarded migration of `scale/type`, known-scale entries, and all SAW storage (preserves/merges existing learning, no loss).
- `SettingsCalibration`: `sensorLag()` rekeyed to ids (incl. `decent-usb`); SAW reads/writes normalize via `sawPairKey()` and by-value params.
- Implements the `scale-type-identity` OpenSpec change (proposal/design/specs/tasks included).

## Test plan
- [x] `tst_saw_settings`: 24/24 pass (6 new — id round-trip, sensorLag by id + legacy name, migration rekey / idempotent / collision-merge / id-preserve, addKnownScale stores id). Existing tests use the display name `"Decent Scale"` and still pass, proving read/write normalization is symmetric.
- [x] Cluster (settings / SAW / scale / dialing / MCP): 19/19 pass — no regressions.
- [x] Build: 0 errors / 0 warnings (app + all test targets).
- [ ] Real-device: WiFi→BLE fallback — set a WiFi scale primary, make WiFi unreachable (BLE still advertising), confirm the scale log shows `WiFi fallback: connecting to BLE Decent scale <name>` and it reconnects over Bluetooth.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -244,6 +244,7 @@ set(SOURCES
     src/ble/bletransport.cpp
     src/ble/scaledevice.cpp
     src/ble/scales/scalefactory.cpp
+    src/ble/scales/scaletypeids.cpp
     src/ble/scales/decentscale.cpp
     src/ble/scales/decentscalewifi.cpp
     src/ble/scales/acaiascale.cpp
@@ -425,6 +426,7 @@ set(HEADERS
     src/ble/bletransport.h
     src/ble/scaledevice.h
     src/ble/scales/scalefactory.h
+    src/ble/scales/scaletypeids.h
     src/ble/scales/decentscale.h
     src/ble/scales/acaiascale.h
     src/ble/scales/felicitascale.h
@@ -1161,6 +1163,7 @@ if(NOT ANDROID AND NOT IOS)
         src/core/settings_network.cpp
         src/core/settings_app.cpp
         src/core/settings_calibration.cpp
+        src/ble/scales/scaletypeids.cpp
     )
     target_link_libraries(saw_parity PRIVATE Qt6::Core Qt6::Sql Qt6::Network Qt6::Gui Qt6::Bluetooth)
     target_include_directories(saw_parity PRIVATE ${CMAKE_SOURCE_DIR}/src)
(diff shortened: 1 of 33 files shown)


skialpine:
### Code review

Reviewed both commits (`9cb3d228`, `f064ff24`) across five aspects (CLAUDE.md, deep bug scan, git history, recent merged PRs, comment accuracy). The migration is idempotent, flag-guarded, correctly sequenced after the #1256 orphan-heal, and read/write SAW keys stay consistent. Three candidate issues were investigated and scored below threshold:

- [OK to skip] Migration collision-merge could trim newer entries on an unfavorable key sort — but the collision needs both a display-name and an id bucket for the same pair pre-migration, which is impossible since only display names were ever written before this change (score: 25)
- [OK to skip] `scale/typeIdsMigrated` flag set even if `m_calibration` were null — but it is constructed unconditionally, so never null (score: 25)
- [OK to skip] Old `"Decent Scale (WiFi)"` display name not mapped by `normalizeScaleTypeId` — false positive: WiFi has always persisted the id `decent-wifi`, never the display name (score: 25)

PR #1265, opened 2026-05-24
fix(scale): WiFi sleep matches BT semantics, and close WS during DE1 sleep

skialpine:
Two related changes to align WiFi scale lifecycle with BT semantics:

## 1. `DecentScaleWifi::sleep()` does firmware-level power-off (matches BT)

WiFi was sending \`soft_sleep on\` — the lighter reversible park state. BT \`DecentScale::sleep()\` sends \`0A 02 00\`, the firmware-level power-off that requires a physical button press to wake. Same logical "sleep" gesture should produce the same outcome on the scale regardless of transport.

Replaced with \`{\"command\":\"power\",\"action\":\"off\"}\` (HDS firmware's equivalent of \`0A 02 00\`).

**Practical impact**: on a battery-only HDS, \`soft_sleep on\` left the ESP32 radio active to maintain the WS, draining the battery while the DE1 was asleep. The new command fully powers the scale down.

The three existing call sites that invoke \`physicalScale->sleep()\` polymorphically are unchanged and now get matched behavior on both transports:
- App exit (\`main.cpp:2823\`)
- \`keepScaleOn=false\` on DE1 sleep (\`main.cpp:2727\`)
- MCP / MQTT \`machine_sleep\` (\`main.cpp:2689\`)

## 2. Graceful WS close on DE1 sleep when \`keepScaleOn=true\` (WiFi only)

No reason to hold an active TCP session open to the HDS while the app is idle. Android's WiFi power-save reliably parks the radio anyway, and the HDS's AsyncTCP 30 s ACK timeout then reaps us — leaving a stale dirty disconnect that the app only notices on DE1 wake when it tries to send the LCD-on command.

Under \`keepScaleOn=true\` (the default), the DE1-sleep path on WiFi now:
1. Calls \`disableLcd()\` (sends \`display off\`, unchanged).
2. Sets \`scaleAutoReconnectSuppressed = true\`.
3. Calls \`disconnectFromScale()\` (graceful WS close).

**BT is unchanged**: the BLE link stays connected through DE1 sleep, matching long-established behavior. BLE doesn't have the same idle-link pathology.

DE1 wake routes through the existing \`scaleAutoReconnectSuppressed\` branch (now shared by both \`keepScaleOn=false\` and the new \`keepScaleOn=true\`+WiFi path) to re-arm \`scaleReconnectTimer\`. \`DecentScaleWifi::onConnected()\` now sends \`display on\` post-handshake — idempotent on a scale whose LCD is already on, required to restore LCD after the DE1-sleep close.

## Test plan

- [x] Build clean (Qt 6.11.1, macOS Debug): 0 errors, 0 warnings, all unit tests link.
- [ ] WiFi scale + MCP \`machine_sleep\` → scale powers off, WS drops, requires button press to wake.
- [ ] WiFi scale + \`keepScaleOn=false\` + DE1 sleep → scale powers off, WS drops, button press to wake.
- [ ] WiFi scale + \`keepScaleOn=true\` (default) + DE1 sleep → \`display off\` lands, WS closes gracefully, scale stays powered. DE1 wake → WS reconnects automatically, LCD restored.
- [ ] BT scale (any \`keepScaleOn\`) unchanged.

### DIFF
--- a/openspec/changes/add-wifi-scale-support/design.md
+++ b/openspec/changes/add-wifi-scale-support/design.md
@@ -187,13 +187,11 @@ The BLE driver carries machinery that exists to work around GATT-layer quirks an
 
 Decision:
 
-- WiFi `sleep()` → `m_socket.sendTextMessage("{\"command\":\"power\",\"action\":\"off\"}")`, the HDS firmware's equivalent of BT `0A 02 00`. Emits `sleepCompleted` immediately after the send call returns.
-- The BLE driver waits for `characteristicWritten` before emitting `sleepCompleted` ([decentscale.cpp:362-364](src/ble/scales/decentscale.cpp:362)). The semantic intent of that wait is "we confirmed the bytes left our radio." On WiFi, `sendTextMessage` returning is the precise analog (the TCP write went out). We do not wait for the firmware's status-ack JSON form — that would change the latency profile vs BLE.
-- If the socket is not open, emit `sleepCompleted` anyway (matches the BLE driver's "transport unavailable → emit immediately" fallback at [decentscale.cpp:358-361](src/ble/scales/decentscale.cpp:358)).
+- WiFi `sleep()` → `send("{\"command\":\"power\",\"action\":\"off\"}")`, the HDS firmware's equivalent of BT `0A 02 00`. Emits `sleepCompleted` unconditionally after the send attempt returns, so callers (app-exit waitLoop, DE1-sleep handler) don't hang waiting for an ack that won't come.
+- The BLE driver waits for `characteristicWritten` before emitting `sleepCompleted` ([decentscale.cpp:362-364](src/ble/scales/decentscale.cpp:362)). The semantic intent of that wait is "we confirmed the bytes left our radio." On WiFi, `send()` returning success (socket was in `ConnectedState`) is the precise analog (the TCP write entered the kernel buffer). We do not wait for the firmware's status-ack JSON form — that would change the latency profile vs BLE.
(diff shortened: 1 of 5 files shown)


skialpine:
### Code review

Found 2 issues:

1. PR contradicts an active OpenSpec design decision without updating it. `openspec/changes/add-wifi-scale-support/design.md` decision 12 (still in `changes/`, not archived) explicitly states "The matching WiFi command is `\"soft_sleep on\"`, NOT `\"power off\"` (which would be irreversible and drop the WS link)" and "`power off` (irreversible full power-off) is **not** wired to any `ScaleDevice` virtual today. It would be a new capability — not in scope here." The PR ships exactly that command. Either the design decision was wrong and should be updated/superseded, or the PR oversteps. Recommendation: update design.md decision 12 to reflect the new BT-parity rationale, or land an OpenSpec change/supersede entry alongside this PR. https://github.com/Kulitorum/Decenza/blob/8d35a8d33f263cca70c0296fbb276521766053dd/openspec/changes/add-wifi-scale-support/design.md#L184-L196

2. Stale diagnostic log line. `scaleAutoReconnectSuppressed` is now set in two paths (pre-existing `keepScaleOn=false`, new `keepScaleOn=true`+WiFi), but the `connectedChanged` handler still hardcodes `(keepScaleOn=false)` in its log — so a deliberate disconnect from the new WiFi path will be misreported. The post-sleep-handler block at `main.cpp:2757-2764` already generalized its log; this one was missed. Same class of stale-log issue called out in the round-2 review of #1258. https://github.com/Kulitorum/Decenza/blob/8d35a8d33f263cca70c0296fbb276521766053dd/src/main.cpp#L1705-L1712

<details>

- [Worth a look] Startup-guard log at `main.cpp:1282` has the same `(keepScaleOn=false)` mislabel — same class as #2, lower-impact (printed once at app startup) (score: 25)
- [OK to skip] `display on` in `onConnected()` runs on every WS connect, not only DE1-wake recovery — mirrors BT's `wake()` on every char-discovery, intentional transport-parity (score: 5)

</details>

PR #1272, opened 2026-05-24
wifi-scale: suppress "Scale shut down" dialog when app initiated the power-off

skialpine:
## Summary

When the DE1 goes to sleep with `keepScaleOn=false`, Decenza calls `DecentScaleWifi::sleep()` which sends `{"command":"power","action":"off"}` to the HDS. The firmware then echoes back a `power_off` frame with `reason: "disabled"` (code 0), and `handlePowerFrame` was treating it as an unexpected event — popping a "Scale shut down" dialog at the user. From the user's perspective this was noise: we knew the scale was about to power off because we just asked it to.

Confirmed live in Jeff's log:
```
[2835.088] [BLE DecentScaleWifi] First 'power' frame this connect: {"type":"power","event":"power_off","reason":"disabled","reason_code":0,"ms":12339844}
[2835.092] WARN  [BLE DecentScaleWifi] Scale shut down: disabled (code 0)
```

## Approach

Track an `m_powerOffInitiatedByApp` one-shot latch:
- **Armed** by `sleep()` right before the send (so a synchronous reply still sees it).
- **Consumed-and-cleared** by `handlePowerFrame` on the next `power_off` frame: log at INFO with a "— app-initiated, dialog suppressed" suffix (diagnostic trail preserved), skip the `errorOccurred` emit.
- **Also cleared** in `connectToHost` so a `sleep()` whose echo got lost (e.g. socket dropped before the firmware could reply) can't silently swallow a real firmware-initiated `power_off` on the next connection.

Genuinely-firmware-initiated `power_off` frames (low battery, button long-press, fault) still surface the dialog as before — the latch is one-shot and only set on app-initiated paths.

The disconnect-classification bookkeeping (`m_userInitiatedShutdown`) is unchanged, so `onDisconnected` still logs `"(expected) — scale power-off: <reason>"`.

## Test plan

Two new tests added; all 22 in `tst_decentscalewifi` pass locally:

- `appInitiatedPowerOffSuppressesDialog` — after `sleep()`, the echoed `{type:power, event:power_off, reason:disabled}` does NOT emit `errorOccurred`.
- `firmwareInitiatedPowerOffStillDialogsAfterAppInitiated` — after an app-initiated power_off is consumed, a subsequent firmware-initiated `low_battery` power_off still surfaces the dialog. Verifies the latch is one-shot, not a permanent mute.

Manual check (next build):
- [ ] DE1-sleep with `keepScaleOn=false` no longer pops a dialog.
- [ ] Manually pulling the scale's battery (or simulating low_battery via firmware) still pops the dialog.

### DIFF
--- a/src/ble/scales/decentscalewifi.cpp
+++ b/src/ble/scales/decentscalewifi.cpp
@@ -595,8 +595,9 @@ void DecentScaleWifi::sleep() {
     //
     // Mark the imminent power_off echo as app-initiated so handlePowerFrame
     // doesn't pop a "Scale shut down: disabled" dialog at the user. Set
-    // BEFORE the send so a synchronous reply from the firmware (rare but
-    // possible) still sees the flag.
+    // BEFORE the send to keep the assignment+send pair locally correct —
+    // any future caller or mock that delivers the echo synchronously inside
+    // send() still sees the flag set.
     m_powerOffInitiatedByApp = true;
     const bool sent = send(QStringLiteral("{\"command\":\"power\",\"action\":\"off\"}"));
     if (!sent) {
@@ -605,6 +606,14 @@ void DecentScaleWifi::sleep() {
         // dropped power-off command isn't masked by a misleading
         // "drained successfully" log downstream. Common at app exit after a
         // DE1-sleep keepScaleOn=true+WiFi close (socket already in ClosingState).
+        //
+        // Clear the latch locally: no echo will arrive (we didn't send the
+        // command), so if a firmware-initiated power_off (low battery, button)
+        // happens before the next connectToHost() cycle resets it, it must
+        // surface, not be silently suppressed. The connectToHost() reset is
+        // still the backstop for the case where send() succeeded but the
+        // socket dropped before the echo arrived.
+        m_powerOffInitiatedByApp = false;
         WIFI_WARN("sleep(): power-off command not delivered (socket not connected)");
     }
     // BT waits for `characteristicWritten` as a "command left the radio" ack.
--- a/src/ble/scales/decentscalewifi.cpp
+++ b/src/ble/scales/decentscalewifi.cpp
@@ -88,6 +88,11 @@ void DecentScaleWifi::connectToHost(const QString& hostname) {
     m_pendingHostnameFallback = false;
     m_socketErrorThisConnect = false;
     m_lastSocketErrorString.clear();
+    // Reset the app-initiated-power-off latch — if a prior cycle armed it via
+    // sleep() but the firmware echo never arrived (e.g. socket dropped first),
+    // the next connection's first real power_off frame (low battery, button)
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

<details>

- [Worth a look] `m_powerOffInitiatedByApp` left armed when `send()` fails in `sleep()` — the existing `connectToHost()` reset covers the gap before any next firmware-initiated `power_off` arrives, but a one-line `m_powerOffInitiatedByApp = false;` inside the `if (!sent)` block at https://github.com/Kulitorum/Decenza/blob/909a23161811d76aa37b2fc7c84dbcabaad3c6e3/src/ble/scales/decentscalewifi.cpp#L605-L616 would make the invariant local rather than depending on a distant reset (score: 72)
- [Worth a look] No test for the `send() == false` path in `sleep()` — both new tests exercise the happy path (socket connected, echo received). The "common at app exit" branch in https://github.com/Kulitorum/Decenza/blob/909a23161811d76aa37b2fc7c84dbcabaad3c6e3/src/ble/scales/decentscalewifi.cpp#L605-L616 is untested (score: 50)
- [OK to skip] Stale doc in the `m_userInitiatedShutdown` comment at https://github.com/Kulitorum/Decenza/blob/909a23161811d76aa37b2fc7c84dbcabaad3c6e3/src/ble/scales/decentscalewifi.h#L157-L168 — says `handlePowerFrame (scale told us it's powering down)` but it's now set for any `power_off`, app- or firmware-initiated (score: 55)
- [OK to skip] The `sleep()` rationale comment at https://github.com/Kulitorum/Decenza/blob/909a23161811d76aa37b2fc7c84dbcabaad3c6e3/src/ble/scales/decentscalewifi.cpp#L595-L600 says "Set BEFORE the send so a synchronous reply from the firmware (rare but possible) still sees the flag" — but Qt's `QWebSocket` delivers `textMessageReceived` exclusively via the event loop, so a synchronous reply is architecturally impossible. The BEFORE-send ordering is still defensible (defensive habit, test-mock robustness) but the stated reason is wrong (score: 58)
- [OK to skip] Test comment at https://github.com/Kulitorum/Decenza/blob/909a23161811d76aa37b2fc7c84dbcabaad3c6e3/tests/tst_decentscalewifi.cpp#L513-L516 says "a low-level INFO log records the event" but `WIFI_LOG` expands to `qDebug()` (debug, not info); the same comment claims `m_userInitiatedShutdown` "still fires" but no assertion verifies it (score: 50)
- [OK to skip] Test verifies absence of dialog but not presence of the diagnostic log line at https://github.com/Kulitorum/Decenza/blob/909a23161811d76aa37b2fc7c84dbcabaad3c6e3/tests/tst_decentscalewifi.cpp#L532-L536 — TESTING.md doesn't mandate positive-assertion testing for logging side effects (score: 50)

</details>

PR #1274, opened 2026-05-24
wifi-scale: fast-fail to hostname fallback on cached-IP socket error

skialpine:
## Summary

When the cached-IP dial fires a fatal transient socket error (`HostNotFoundError`, `ConnectionRefusedError`, `NetworkError`, `SocketTimeoutError`), don't sit on the 5 s recognition timer waiting for it to time out — the kernel just told us the cached IP is dead, so the hostname fallback should run immediately.

## Motivation

Reproduced live in Jeff's macOS log from this evening:

```
[BLE DecentScaleWifi] Trying cached IP 192.168.10.242 for hds.local
[BLE DecentScaleWifi] Connecting to ws://192.168.10.242/snapshot (cached IP)
[BLE DecentScaleWifi] WebSocket disconnected (unexpected) — peer close (code 1000)
[BLE DecentScaleWifi] WebSocket error: Host unreachable (code 7)
...
( 5 s of nothing useful happens )
...
[BLE DecentScaleWifi] No recognizable HDS frame within 5000 ms from 192.168.10.242
[BLE DecentScaleWifi] Cached IP 192.168.10.242 didn't validate as HDS — falling back to hostname hds.local
```

During that 5 s wait the BLEManager connection-timer also fired the WiFi→BLE fallback path. CoreBluetooth wasn't powered on yet, so BLE also failed. The cascade tripped the FlowScale "no scale" dialog at the user before either transport had a fair chance, even though the WiFi scale itself was up and happy.

With this PR the same cascade collapses to ~milliseconds — the hostname fallback dials immediately, and connects on the first try.

## Approach

In `DecentScaleWifi::onError()` (`src/ble/scales/decentscalewifi.cpp`):
- The fast-fail switch already classified four error codes (`HostNotFoundError`, `ConnectionRefusedError`, `NetworkError`, `SocketTimeoutError`) — we now also set a local `fastFailError` flag.
- If we're on the cached-IP path (`!m_currentTargetIsHostname && !m_triedHostnameFallback && !m_pendingHostnameFallback`), trigger the hostname fallback now:
  - If the socket is still in a non-final state → mirror `onRecognitionTimeout`'s pattern: set `m_pendingHostnameFallback = true`, abort the socket, and `onDisconnected` runs `attemptHostname()` inline when the close lands.
  - If the socket has already disconnected (signal-order race: `disconnected` fired before `errorOccurred`) → dial `attemptHostname()` directly via `QMetaObject::invokeMethod(..., Qt::QueuedConnection)` for re-entrancy safety.

The 5 s recognition timer stays in place as a backstop for the case where the WS upgrade hangs without any clean kernel error.

## Deferred follow-up

Jeff also asked about WiFi/USB scale connects being gated on QML loading at ~4.5 s into startup. This PR doesn't address that — moving the direct-wake earlier requires careful auditing of the `scaleDiscovered` re-entry path (the WiFi→BLE fallback machinery has subtle ordering). Worth a separate PR.

For this specific symptom, the fast-fail change alone is sufficient — even if the cached-IP attempt still happens at 4.5 s post-startup, it now recovers in ~50 ms instead of 5+ s.

## Test plan

New test `cachedIpFastFailsToHostnameFallback`:
- Configure the driver with an `IpResolver` returning `127.0.0.1:1` (port 1, no listener → `ConnectionRefusedError` fires near-instantly).
- The hostname target points at a real `FakeHdsServer`.
- Dial via `connectToHost(server.host())`.
- Assert the full WS handshake completes within 2 s wall-clock (well under the 5 s recognition timeout budget). Pre-fix this would have taken ~5 s.

All 24/24 tests pass with zero warnings.

### DIFF
--- a/src/ble/scales/decentscalewifi.cpp
+++ b/src/ble/scales/decentscalewifi.cpp
@@ -671,17 +671,54 @@ void DecentScaleWifi::onError() {
     // user. The 503 "another client connected" case handled above is the one
     // socket error we DO surface here, because the retry loop can never resolve it.
     // #1253
+    bool fastFailError = false;
     switch (err) {
     case QAbstractSocket::HostNotFoundError:
     case QAbstractSocket::ConnectionRefusedError:
     case QAbstractSocket::NetworkError:
     case QAbstractSocket::SocketTimeoutError:
         m_userInitiatedShutdown = true;
+        fastFailError = true;
         break;
     default:
         // Other errors fall through — the recognition timeout will catch the
         // case where the WS upgrade hangs without a clean socket error.
         break;
+    }
+
+    // Fast-fail: if a cached-IP attempt fires a fatal transient socket error,
+    // don't sit on the 5 s recognition timer waiting to time out — start the
+    // hostname fallback immediately. The kernel just told us the cached IP
+    // is dead; there's nothing to gain from waiting longer. Real symptom this
+    // addresses: on macOS cold start the network stack may still be warming
+    // up, the cached scale IP isn't routable yet, and the 5 s wait was long
+    // enough for the BLE-scale fallback machinery to also trigger and fail
+    // (CoreBluetooth not powered on), pushing the FlowScale-fallback dialog
+    // at the user before the WiFi scale could simply be retried via hostname.
+    if (fastFailError &&
+        !m_currentTargetIsHostname &&
+        !m_triedHostnameFallback &&
+        !m_pendingHostnameFallback) {
+        WIFI_LOG(QString("Cached IP %1 unreachable (%2) — fast-failing to hostname %3 fallback")
+                 .arg(m_currentTarget, errStr, m_hostname));
+        m_recognitionTimer->stop();
+
+        if (m_socket && m_socket->state() != QAbstractSocket::UnconnectedState) {
+            // Socket is mid-teardown. Use the existing flag mechanism that
+            // onRecognitionTimeout uses: abort to force a clean disconnect,
+            // and onDisconnected sees m_pendingHostnameFallback and runs
+            // attemptHostname() inline once the close completes.
+            m_pendingHostnameFallback = true;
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

<details>

- [OK to skip] Else-branch in `onError()` fast-fail is likely dead code for the four guarded error types (Qt fires `errorOccurred` before `disconnected` for connection-phase errors). Safe if ever reached, but propagates a transient `setConnected(false)` to consumers before the fallback runs, unlike the if-branch which suppresses that via the pending-flag → onDisconnected short-circuit. Defensive code that's worth keeping in case Qt signal ordering changes. https://github.com/Kulitorum/Decenza/blob/fee863665c1a83c8db1c46f6dfa9c4ed25c25f81/src/ble/scales/decentscalewifi.cpp#L744-L760 (score: 60)
- [OK to skip] Comment near https://github.com/Kulitorum/Decenza/blob/fee863665c1a83c8db1c46f6dfa9c4ed25c25f81/src/ble/scales/decentscalewifi.cpp#L727-L731 says "The kernel just told us the cached IP is dead" — slightly overstates for `SocketTimeoutError` (slow network, not necessarily dead host) and `NetworkError` (catch-all). The inline error message at line 740 (`"Cached IP %1 unreachable (...)"`) is already more precise; the prose comment could match (score: 65)
- [OK to skip] The "raced ahead" comment at https://github.com/Kulitorum/Decenza/blob/fee863665c1a83c8db1c46f6dfa9c4ed25c25f81/src/ble/scales/decentscalewifi.cpp#L751-L755 describes a real platform-specific scenario (e.g. macOS NetworkError where the OS closes the TCP socket synchronously before posting the error) but doesn't specify WHICH error types produce that ordering, leaving readers unsure whether the branch applies to all four guarded errors (it doesn't — ConnectionRefusedError fires error-first, not disconnected-first) (score: 50)

</details>

PR #1275, opened 2026-05-24
wifi-scale: fresh QWebSocket per connect attempt

skialpine:
## Summary

`DecentScaleWifi` held a single `QWebSocket` for its lifetime and re-used it across every connect attempt (initial cached-IP try, hostname fallback, reconnect cycles, `abort()`-then-retry). Reusing a `QWebSocket` across multiple `open()`/`abort()` cycles surfaces a Qt-internal state-staleness bug: subsequent attempts' WS handshakes complete against the server, then immediately close with peer-close `code 1000` + `UnknownSocketError(-1)` — likely because Qt's internal close-state and error bookkeeping carry over from the prior `abort()`.

Stacks on top of PR #1274 (fast-fail to hostname fallback). #1274 fires the hostname attempt sooner; this PR ensures that hostname attempt uses a fresh socket.

## How we caught it

Reproduced live on a macOS instance this evening: the WiFi scale auto-reconnect loop failed in a tight cycle for several minutes — every attempt got `peer close (code 1000)` + `Unknown error (-1)` within milliseconds of WS handshake completion. Manually invoking `devices_connect_scale` via the Decenza MCP triggered the `Scale type changed from \"decent\" to \"decent-wifi\" - creating new scale` code path, which happens to construct a brand-new `DecentScaleWifi` and thus a fresh `QWebSocket`. The very next dial connected in 93 ms — same IP, same scale, same code path past the constructor, but a fresh socket.

```
[291.789] [BLE DecentScaleWifi] Trying cached IP 192.168.10.242 for hds.local
[291.790] [BLE DecentScaleWifi] Connecting to ws://192.168.10.242/snapshot (cached IP)
[291.883] [BLE DecentScaleWifi] WebSocket connected — peer=192.168.10.242:80 localPort=53860
[291.884] [Scale] \"Half Decent Scale (WiFi)\" CONNECTED
[291.895] [BLE DecentScaleWifi] First 'rate' frame this connect: {...}
```

Conclusive client-side state bug.

## Fix

Add `recreateSocket()` private helper. Call it at the start of `attemptTarget()` so every connect attempt starts from a clean slate. The helper:
1. Stops the recognition timer (it would otherwise fire `m_socket->abort()` against the new socket).
2. Disconnects all signals from the old socket to `this` so any late-emitted/queued signals can't run our handlers and mistakenly reference the new `m_socket`.
3. Aborts the old socket if it's not already `UnconnectedState`, then `deleteLater()`s it.
4. Constructs a fresh `QWebSocket` and re-wires `connected`/`disconnected`/`textMessageReceived`/`errorOccurred`.

Construction overhead is microseconds. Architecturally this matches Qt's general one-shot pattern for transient network resources (`QNetworkReply`, `QHttpMultiPart`, etc.) and eliminates an entire class of state-staleness bugs without diving into Qt's private internals to debug which specific `QWebSocket` member is sticky.

## Test plan

- All 24/24 existing tests pass with zero warnings.
- `cachedIpFastFailsToHostnameFallback` (added in PR #1274) exercises a full cached-IP-attempt → recreate-via-attemptTarget → hostname-attempt cycle. Would have failed if the recreation broke signal wiring.
- Manual verification needed after merge: confirm the auto-reconnect loop on macOS now actually reconnects without requiring the scale-type-change workaround.

### DIFF
--- a/src/ble/scales/decentscalewifi.cpp
+++ b/src/ble/scales/decentscalewifi.cpp
@@ -50,21 +50,18 @@ template struct AccessBypass<WsPrivateSocketTag, &QWebSocketPrivate::m_pSocket>;
 
 DecentScaleWifi::DecentScaleWifi(QObject* parent)
     : ScaleDevice(parent)
-    , m_socket(new QWebSocket(QString(), QWebSocketProtocol::VersionLatest, this))
+    , m_socket(nullptr)
     , m_recognitionTimer(new QTimer(this))
 {
     m_recognitionTimer->setSingleShot(true);
     connect(m_recognitionTimer, &QTimer::timeout,
             this, &DecentScaleWifi::onRecognitionTimeout);
 
-    connect(m_socket, &QWebSocket::connected,
-            this, &DecentScaleWifi::onConnected);
-    connect(m_socket, &QWebSocket::disconnected,
-            this, &DecentScaleWifi::onDisconnected);
-    connect(m_socket, &QWebSocket::textMessageReceived,
-            this, &DecentScaleWifi::onTextMessageReceived);
-    connect(m_socket, QOverload<QAbstractSocket::SocketError>::of(&QWebSocket::errorOccurred),
-            this, &DecentScaleWifi::onError);
+    // Initial socket — recreateSocket() will swap it on every connect attempt,
+    // but we need something non-null in m_socket so the various state guards
+    // (m_socket && m_socket->state() != UnconnectedState in the destructor etc.)
+    // don't crash if the driver is destroyed before its first connect attempt.
+    recreateSocket();
 }
 
 DecentScaleWifi::~DecentScaleWifi() {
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

No issues remain. Checked for bugs and CLAUDE.md compliance. One comment-accuracy issue at score 75 was found and fixed in [2a7d67e1](https://github.com/Kulitorum/Decenza/commit/2a7d67e1); a second wording issue at score 50 was fixed in the same commit while in the file.

<details>
<summary>Resolved findings</summary>

- **[Fixed in 2a7d67e1]** Constructor inline comment cited the destructor as the crash site recreateSocket() guards against — but the destructor already null-checks. The real reason is the "m_socket is always live" invariant relied on by signal handlers (e.g. `onRecognitionTimeout` calls `m_socket->abort()` without a null check). Rewrote comment to state the actual rationale (score: 75)
- **[Fixed in 2a7d67e1]** `disconnect(this)` comment claimed it guards against signals "queued by Qt's event loop" — but these connections are AutoConnection→DirectConnection (same thread), so they fire synchronously. The actual risk is QSocketNotifier / QAbstractSocket state-change events posted to the event loop before our `abort()` causing the dying socket to re-emit. Rewrote to describe the mechanism accurately (score: 50)

</details>

PR #1277, opened 2026-05-25
wifi-scale: drop scale-error modal paths; existing dialogs are enough

skialpine:
## Summary

The 1.7.5 WiFi-scale work introduced two new user-facing modals via `bleErrorDialog`:

- **"Scale shut down: \<reason\>"** — fired on a firmware-initiated `power_off` frame (low battery, physical button)
- **"Another client is connected to the scale"** — fired on a WebSocket 503 from HDS firmware

Neither is actionable:

- A firmware power_off already drives the existing **Scale Disconnected** notice when auto-reconnect gives up — the new modal is a duplicate.
- HDS firmware now allows multiple concurrent clients, so a 503 just means the cap is briefly saturated. The original single-connection design was short-lived.

Both got removed. `bleErrorDialog` itself stays — it's still the right surface for genuine BLE permission/adapter problems (Bluetooth off, Location off, permission denied).

The app-initiated power-off latch (`m_powerOffInitiatedByApp`) is retained but now selects `WIFI_LOG` vs `WIFI_WARN` log level for the power_off echo (instead of selecting "dialog or no dialog").

### Scale dialog inventory after this change

| Dialog | Trigger | Gated by `Settings.showScaleDialogs` |
|---|---|---|
| **No Scale Found** | Saved scale never connected at startup | ✅ |
| **Scale Disconnected** | Previously-connected scale dropped + reconnect gave up | ✅ |
| **Shot Stopped — scale not connected** | User pressed Start with scale offline (action-feedback, not nag) | n/a |
| **Add WiFi Scale** | User-invoked in Settings → Connections | n/a |
| **Known Scales** picker | User-invoked | n/a |

## Test plan

- [ ] Power-cycle the WiFi scale (or pull its battery) mid-session → no "Scale shut down" modal; "Scale Disconnected" appears only once reconnect actually gives up
- [ ] Connect a second client to the HDS while the app is already connected → no "Another client is connected" modal
- [ ] Confirm a real BLE permission denial (deny Bluetooth on macOS / disable Location on Android) still pops `bleErrorDialog` with the correct title
- [ ] Run `tst_decentscalewifi` — all power-frame / 503 tests still pass with `errorSpy.count() == 0`

### DIFF
--- a/qml/main.qml
+++ b/qml/main.qml
@@ -1301,10 +1301,8 @@ ApplicationWindow {
         // wrong instruction.
         Tr { id: trEnableLocation;     key: "main.dialog.enableLocation.title";  fallback: "Enable Location";    visible: false }
         Tr { id: trEnableBluetooth;    key: "main.dialog.enableBluetooth.title"; fallback: "Enable Bluetooth";   visible: false }
-        // Generic title for non-permission errors. Transport-neutral because this
-        // dialog also surfaces WiFi-scale errors (e.g. "WiFi scale not found"),
-        // which are NOT Bluetooth errors. The Location / Bluetooth-permission cases
-        // above keep their specific titles.
+        // Generic title for non-permission errors. The Location /
+        // Bluetooth-permission cases above keep their specific titles.
         Tr { id: trBleErrorGeneric;    key: "main.dialog.bleError.title";        fallback: "Connection Error";   visible: false }
         Tr { id: trErrorPrefix;        key: "common.accessibility.errorPrefix";  fallback: "Error:";             visible: false }
 
--- a/src/ble/scales/decentscalewifi.cpp
+++ b/src/ble/scales/decentscalewifi.cpp
@@ -445,10 +445,10 @@ void DecentScaleWifi::handlePowerFrame(const QJsonObject& obj) {
     if (event != QStringLiteral("power_off")) return;
 
     // Reject malformed power_off frames that omit BOTH the reason string and
-    // the numeric reason code — without either, we can't tell the user why
-    // the scale is going down, and marking the disconnect as "expected" would
-    // hide a potentially-real failure. Let the following unexpected disconnect
-    // log surface so it's diagnosable.
+    // the numeric reason code — without either, we have no diagnostic detail
+    // to log, and marking the disconnect as "expected" would hide a
+    // potentially-real failure. Let the following unexpected disconnect log
+    // surface so it's diagnosable.
     if (reason.isEmpty() && reasonCode == -1) {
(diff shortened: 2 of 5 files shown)


skialpine:
### Code review

Found 4 issues:

1. Dead `UiTranslator` infrastructure — `translateUiString()` had only two call sites (the two `emit errorOccurred(...)` lines this PR removes), so the typedef, `setUiTranslator()`, `m_uiTranslator` member, and `translateUiString()` helper are now unreachable. The wiring calls in `src/main.cpp` are also no-ops. CLAUDE.md says: "If you are certain that something is unused, you can delete it completely."

https://github.com/Kulitorum/Decenza/blob/5e5b77edeca8eb94df17f8de1a67f4de1ce00c42/src/ble/scales/decentscalewifi.h#L67-L72

https://github.com/Kulitorum/Decenza/blob/5e5b77edeca8eb94df17f8de1a67f4de1ce00c42/src/ble/scales/decentscalewifi.h#L200-L205

2. Stale 503 comment contradicts new behavior — the comment block right after the 503 early-return still says "The 503 'another client connected' case handled above is the one socket error we DO surface here, because the retry loop can never resolve it." This PR just removed that surface path; the sentence is now factually false and will mislead the next reader.

https://github.com/Kulitorum/Decenza/blob/5e5b77edeca8eb94df17f8de1a67f4de1ce00c42/src/ble/scales/decentscalewifi.cpp#L708-L718

3. Test function name is now misleading — `firmwareInitiatedPowerOffStillDialogsAfterAppInitiated` claims the test verifies the dialog still fires, but the body now asserts `errorSpy.count() == 0` and is documented as "dialog path removed in both cases." The test is correct (it now enforces the WARN-vs-LOG log-level invariant via `ignoreMessage`), but the name should be renamed to match — e.g. `firmwareInitiatedPowerOffStillWarnsAfterAppInitiated`.

https://github.com/Kulitorum/Decenza/blob/5e5b77edeca8eb94df17f8de1a67f4de1ce00c42/tests/tst_decentscalewifi.cpp#L540-L545

4. Stale inline comment `// dialog suppressed` — the comment dates from when the latch suppressed a dialog. Now that the emit is gone in both branches, there's nothing to suppress. Rewording to `// errorOccurred not emitted (dialog path removed)` matches the docstring above the test.

https://github.com/Kulitorum/Decenza/blob/5e5b77edeca8eb94df17f8de1a67f4de1ce00c42/tests/tst_decentscalewifi.cpp#L532-L536

<details>

- [Worth a look] `handlePowerFrame` malformed-frame comment still says "we can't tell the user why the scale is going down" — the user-facing message no longer exists in any branch; minor wording rot (score: 50)

</details>

PR #1278, opened 2026-05-25
fix(scale): keep USB scale switchable via the unified scale view

skialpine:
## Summary

A connected USB scale "took over" Settings → Connections: the whole scale panel was replaced by a USB-only view (status badge, weight, tare, log) with **no scan button, no Known Devices picker, and no Add WiFi Scale link**. Once on USB, the user had no way to switch back to a Bluetooth or WiFi scale.

This removes the separate USB-only view and always shows the unified scale view, so the USB scale behaves like the BT and WiFi scales — it shows as the connected/selected scale and stays switchable to the other transports.

## Why no C++ change was needed

The unified view already supports USB everywhere it reads state:
- When the USB scale connects, `main.cpp` points the QML `ScaleDevice` context property at it, so the status row, "Connected:" name, weight, and tare all reflect USB.
- The synthetic `usb:decent` entry is already in both `BLEManager.discoveredScales` (unified discovered list) and `Settings.knownScales` (Known Devices picker).
- Selecting a BT/WiFi scale already tears the USB scale down first: the `scaleDiscovered` handler calls `usbScaleManager.disconnectScale()` before connecting the new transport, and `usbScaleAvailable` only auto-reconnects USB when it is still the saved primary.

So the QML view swap was the only thing blocking the switch.

## Changes

- `qml/pages/settings/SettingsConnectionsTab.qml` — delete the `usbAvailable && UsbScaleManager.scaleConnected` USB-only `ColumnLayout`; drop the `visible:` gate on the unified scale view so it always shows (3 insertions, 135 deletions).

## Test plan

- [ ] Plug in a USB Decent scale → it connects and shows as "Connected: Half Decent Scale (USB)" in the unified view, with live weight and a working Tare
- [ ] With USB connected, open Known Devices → pick a Bluetooth or WiFi scale → connection switches; USB is torn down (no double weight feed)
- [ ] Switch back to USB from the picker → reconnects via UsbScaleManager
- [ ] Unplug USB while it's the active scale → falls back to FlowScale / BLE with the standard "Scale Disconnected" notice
- [ ] iOS unaffected (USB path is compiled out; `usbAvailable` is false there anyway)


skialpine:
### Code review

Found 1 issue (now fixed in a818c68d):

1. USB scale diagnostics were orphaned from the scale log. Deleting the USB-only view removed the sole UI consumer of `UsbScaleManager::logMessage`, which `main.cpp` forwards only to `BLEManager::de1LogMessage` (the app/DE1 log) — not to the scale log. After the deletion, USB probe/connect/error diagnostics no longer appeared in the unified scale panel and were excluded from the scale "Share Log" export (which serializes `m_scaleLogMessages`). Coarse lifecycle events ("Found …", "unplugged") still reached the scale log via `appendScaleLog`, so it wasn't fully empty.

https://github.com/Kulitorum/Decenza/blob/e1443049c1b359c493496a2e6904a57359cae302/src/main.cpp#L2109-L2112

Fix (a818c68d): also forward `UsbScaleManager::logMessage` to `BLEManager::appendScaleLog`, so USB diagnostics land in both the scale log (and the scale Share Log) and the app/DE1 log.

https://github.com/Kulitorum/Decenza/blob/a818c68d044cbfc5e27cc699560d3367631203db/src/main.cpp#L2109-L2117

<details>

- [OK to skip] `▼` Unicode glyph used as the dropdown chevron at SettingsConnectionsTab.qml is on CLAUDE.md's "no Unicode glyphs as icons" list, but it's an app-wide convention used in 7 places including the shared `StyledComboBox.qml`; fixing one site in isolation would be inconsistent churn unrelated to this PR — better as a dedicated codebase-wide change (score: 60)
- [OK to skip] Header comment "switchable to other transports via Known Devices" — accurate (USB is added to `knownScales` via `addKnownScale("usb:decent")`), minor wording only (score: 50)

</details>

PR #1279, opened 2026-05-26
fix(scale): accurate battery + 'Charging' indicator on the Decent Scale

skialpine:
## Summary

A user reported the BT Decent Scale showing **100%** battery while the same scale over WiFi reported a real percent (40 %, then 86 %, etc.). Root cause turned out to be a combination of three things, all addressed here:

1. **We weren't polling.** The BT Decent Scale only reports battery in the LED-response packet it sends in reply to the display-on command (`0A 01 01 00 01`). Decenza only sent that command during the wake sequence on connect, so the battery reading went stale and was never refreshed for the rest of the session.
2. **We had no log for the parsed byte**, so the "100%" couldn't be read out of the debug log without inferring from the UI.
3. **The UI was showing "100%" while the scale was charging.** When the scale is charging, the BT firmware emits `d[4]=0xFF` as a charging sentinel (per `openscale include/ble.h:710-711, 742-762`) which our parser correctly maps to `batteryLevel=100` + `charging=true`. WiFi reports `charging:true` alongside the (still slightly inflated) percent. Either way, the UI should surface "charging" instead of a percent that isn't meaningful while plugged in.

Reference for the polling cadence: reaprime's `_sendOledOn` is called on connect and again every 75 4-second ticks (~5 min). Decenza's heartbeat fires every 1 s, so we use 240 ticks (~4 min).

## Changes (4 commits)

1. **`71dab0e6` — diag log of the LED-response packet.** Log the parsed battery byte `d[4]` plus the raw 7-byte packet (`Battery byte d[4]=0x… (…) — LED response raw: 03 0a … … … … …`) mirroring the existing firmware-version log's "log first, warn on change" shape. Cleared on disconnect alongside `m_firmwareVersion`.
2. **`f11edb16` — periodic battery polling.** Piggyback on the existing 1 s heartbeat timer — every 240 ticks (~4 min) re-send the display-on command. The scale replies with an LED-response packet whose `d[4]` is parsed as battery. Heartbeat already stops on `sleep()` and gates on `m_characteristicsReady` / `m_heartbeatsPaused`, so polling auto-pauses when the scale sleeps or when DE1 BLE discovery is in flight. No new QTimer; just a tick counter + threshold.
3. **`5a02f434` — scale battery widget: charging icon + label.** When `ScaleDevice.charging` is true, swap to a new `battery-charging.svg` icon (same outline + cap as the existing `battery-N.svg` set, lightning bolt in place of level bars) and replace the percent with a translated "Charging" label. Tint with `successColor` so the color isn't tied to the bogus 100 sentinel. Both transports already drive `ScaleDevice.charging` — the widget just had nothing wired to it.
4. **`b614d831` — Settings → Connections battery row.** Mirror the same charging icon + "Charging" label in the Settings → Connections "Connected: <scale>" row, which had its own inline copy of the battery-icon/percent logic.

## Tests

- `tst_scaleprotocol` (37 tests) — updated `QTest::ignoreMessage` entries for the new battery-byte debug log in the six LED-response cases. **37/37 passing, no warnings** (verified locally via Qt Creator MCP).
- No new unit tests for the polling cadence — heartbeat-driven, hard to test without a transport mock. The same trade-off the existing heartbeat code makes.
- UI changes (icon swap, label) are visual and can only be verified by running the app with a connected scale on charge / off charge.

## Test plan

- [ ] Build and reconnect a Decent / Half Decent Scale via BT. Expect a single `Battery byte d[4]=0x… (…) — LED response raw: 03 0a …` debug line at connect; firmware-version log fires beside it.
- [ ] Leave the scale connected ≥4 min. Expect `Polling battery (display-on refresh)` log line on the heartbeat tick that crosses 240 (~4 min after connect/wake). With the dedup latch, no new `Battery byte` line unless the byte actually changed.
- [ ] Plug USB into the scale (or take it off). The `charging` signal flips. In Settings → Connections the row should swap from "X %" to a charging-bolt icon + "Charging" label; the scale-battery widget should do the same on the Idle layout.
- [ ] Sleep the scale (DE1 sleep with `keepScaleOn=false`, or explicit sleep). Expect heartbeat timer to stop (`Stopping heartbeat timer`), so no further polls fire. On wake: heartbeat restarts, ~4 min window starts fresh.


skialpine:
### Code review

Found 2 issues:

1. Missing `QTest::ignoreMessage` for the new battery-byte debug log in `decentLedResponseSkipsChecksum`. The PR adds `Battery byte d[4]=…` debug-logging on every first-LED-response parse and updates the other six LED-response tests to ignore it, but this one was missed (it only has the firmware-version ignore at line 449). The test will emit a stray `qDebug` line; project rule per [docs/CLAUDE_MD/TESTING.md](https://github.com/Kulitorum/Decenza/blob/main/docs/CLAUDE_MD/TESTING.md) is to mark every unexpected diagnostic.

https://github.com/Kulitorum/Decenza/blob/b614d831dcb7e5877e0435038ef644a29f924569/tests/tst_scaleprotocol.cpp#L446-L452

2. USB Decent Scale path doesn't drive `ScaleDevice.charging`, so the new charging UI won't activate on USB. `UsbDecentScale::processPacket` parses the same `0x0A` LED-response packet but on `d[4] == 0xFF` only calls `setBatteryLevel(100)` — no `setCharging(true)`. The BT path (decentscale.cpp:215-220) and the WiFi path (decentscalewifi.cpp:402-405) both call `setCharging`, so the new "Charging" icon + label in ScaleBatteryItem.qml and SettingsConnectionsTab.qml fires for BT/WiFi users on charge but not for USB users — they continue to see "100%" with a full battery icon. Pre-existing, but the PR makes the gap user-visible.

https://github.com/Kulitorum/Decenza/blob/b614d831dcb7e5877e0435038ef644a29f924569/src/usb/usbdecentscale.cpp#L282-L292

<details>
<summary>Below-threshold notes</summary>

- [Worth a look] `SettingsConnectionsTab.qml:806` battery row visibility gate is `batteryLevel >= 0 && <= 100` — doesn't include `charging` like `ScaleBatteryItem.qml` does. Both transports currently set `batteryLevel` and `charging` together so no user-visible bug, but the two display surfaces are now inconsistent in intent (score: 35).
- [OK to skip] Several minor comment-wording polishes (e.g. "WiFi percent unreliable" phrasing in SettingsConnectionsTab.qml could match ScaleBatteryItem.qml's more accurate "WiFi reports real percent, suppressed for transport-neutral UI"; terminology consistency between "display-on" and "LCD enable" in decentscale.cpp). No correctness impact.

</details>

PR #1282, opened 2026-05-26
saw: preserve settled weight when cup is lifted mid-settle (#1280)

### LINKED ISSUE #1280: [Bug] > [FR] - AI advice is distorted, possibly due to different behaviour of stepdown basket
### What happened?

See video. Here is a copy of the conversation, my comments in **BOLD**

I did a turbo turbo shot using advice from AI and it offered the same grind setting that I calculated using the UGS chart... good! I then asked:

What did you think of this turbo turbo shot

[Assistant]
This shot (May 26, 09:56) was a TurboTurbo profile aiming for a high-extraction turbo shot.

You stopped the shot manually at 38.5g yield when the target was 42g. 

**Nope, I don't think I stopped it manually, so not sure why it had stopped a bit short, nevertheless, since there is a SAW, the comment about yield is observant, maybe interesting, but not relevant for dialling in.**

The total duration was 17 seconds, which is a bit fast even for a turbo shot (the profile targets 20-30s). The flow rate during the "decline" phase was also lower than the target (around 2.5-3.3 ml/s vs. target 5.06 ml/s). This indicates that the grind is likely too fine for this profile, causing more resistance than intended.

**So here is where it is possible the different characteristics of a 32mm step down basket  are confusing analysis... it thinks I need a coarser grind despite the fact it also mentioned the duration was 17 secs, faster than the target... which would indicate needing to grind finer!**

Since you didn't provide tasting notes, I'll focus on the shot mechanics.  To get closer to the profile's intended behavior (faster flow and the full yield), I recommend making your grind coarser. This should increase the flow rate and allow you to reach your target yield within the expected duration.

**Again, yield nonsense.... and confusion about the duration VS flow rate at odds with each other.**

I do feel Basket should be included in the Brew Settings. Ratio could be half a line, as could Roasted date... Getting empirical data WRT the impact of changing from one basket to another will not be easy, but I did notice my Dflow curve shape/behaviour changing when moving to the 32mm stepdown.

### Steps to reproduce

ask AI

### Decenza version

1.7.5 3420

### Platform

Android

### Device / OS version

12

### Scale (if applicable)

_No response_

_No response_

_No response_

### Screenshots or video

_No response_

### Anything else?

_No response_

skialpine:
## Summary

Fixes [#1280](https://github.com/Kulitorum/Decenza/issues/1280). Mark's TurboTurbo 42 shot 5470 was SAW-stopped on target (cup settled at 42.3 g for ~700 ms) but persisted `finalWeightG=38.5g` because cup-lift spike artifacts polluted `m_weight` before the cup-removed detector fired. The AI advisor then reasoned from `{yield: 38.5, target: 42}` and invented a "you stopped manually" framing, self-contradicting on grind direction (fast time → finer, low flow → coarser).

### Root cause (from the debug log)

```
SAW Stop triggered: weight=41.2, target=42
Settling: 41.5, 41.7, 42.0, 42.2, 42.3×7, 42.4×2, 42.5  ← stable plateau ~700 ms
                                                          (just shy of SETTLING_STABLE_MS=1000 ms)
Then the cup is lifted:
Settling: 44, 48.4, 51   ← ShotDataModel rejects these spikes;
                            ShotTimingController accepts all three → m_weight=51
Settling: 38.5           ← under cup-removal threshold relative to 51; accepted → m_weight=38.5
Settling: -28            ← cup-removed branch finally fires, but m_weight already poisoned
[SAW] Cup removed during settling — skipping learning
Saved shot 5470: finalWeightG=38.5
```

### Fix

1. **`ShotTimingController` tracks the last clean rolling-window avg.** Only updated when the existing stability gate holds (`avgDrift < 0.3`, `weight ≤ avg + 0.2`, `avg ≥ m_weightAtStop − 0.5`). On cup-removed, restore `m_weight` via the fallback chain: clean avg → `m_weightAtStop` floor (drip only adds) → unchanged.
2. **`ShotSummarizer::buildShotBlock` emits `stoppedBy`** on the standalone shot JSON block when the value is in the `{manual, weight, volume}` allowlist — same gating already used by `dialing_blocks.cpp` on dial-in surfaces. Gives the LLM a stop-reason anchor on sub-target shots instead of leaving it free to hallucinate.
3. **`shot_eval --settling`** — new offline mode that parses the embedded `app.data.debug_log` from saved shots and reports pre/post-fix `finalWeight` for each. Useful for scanning a personal shot corpus.

### Validation against shot 5470 via the new tool

```
$ shot_eval --settling shot5470.json
file                               stopW  recorded  preFix  postFix       Δ  cupLift  note
shot5470.json                       41.2      38.5    38.5     42.4     3.9      yes  cup-lifted — recovered last clean avg ***
```

Recovers the actual settled weight (42.4 g) instead of the spike-artifact value (38.5 g).

## Test plan

- [x] New `tst_Settling::cupLiftMidSettlePreservesLastStableAvg_1280` — replays shot 5470's exact sample stream and asserts `currentWeight()` ≈ 42.3 g. Was red before the fix (`got 38.50 g`).
- [x] New `tst_Settling::cupLiftBeforeAnyCleanAvgFloorsAtStopWeight` — exercises the `m_weightAtStop` fallback path when no clean avg is observable.
- [x] Four new `tst_ShotSummarizer` tests covering the `stoppedBy` allowlist (`weight` and `manual` emit; `profileEnd` and empty omit).
- [x] Full `ctest` suite — 51/51 pass, 2250 individual test cases.
- [x] `shot_eval --validate tests/data/shots/manifest.json` — 21/21 detector regressions unchanged (fix is orthogonal to detector logic).
- [ ] On-device manual confirmation: pull a sub-target shot, lift the cup as soon as the stop overlay appears, confirm Shot Detail page yield matches what the scale read at lift.

## OpenSpec

`changes/preserve-settled-weight-on-cup-lift/` — amends `stop-at-weight-learning` (settle → finalWeight contract) and `advisor-user-prompt` (`stoppedBy` on standalone shot block). `openspec validate --strict` passes.

### DIFF
--- a/docs/CLAUDE_MD/SAW_LEARNING.md
+++ b/docs/CLAUDE_MD/SAW_LEARNING.md
@@ -156,9 +156,10 @@ Before issue [#1280](https://github.com/Kulitorum/Decenza/issues/1280), `m_weigh
 
 The cup-removed branch now applies this fallback chain to restore `m_weight`:
 
-1. **Last clean settling avg** — the value captured by the stability gate above. This is the truth when at least one stable sample was observed (the common case for any settle that lasted long enough for the user to lift the cup).
-2. **`m_weightAtStop` floor** — if no clean avg was ever observed (cup lifted before the window filled), raise `m_weight` to at least the SAW trigger weight. Post-stop drip can only add weight; persisting a value below the trigger weight is physically impossible.
-3. **Unchanged** — if the fallback can't help (no SAW trigger captured, or `m_weight` already above stop weight), leave it as today's value.
+1. **Last clean settling avg, if plausible** — the value captured by the stability gate above, **provided its overshoot above `m_weightAtStop` does not exceed `MAX_PLAUSIBLE_POST_STOP_DRIP_G = 5 g`**. Real drip is 0.5–3 g; a "stable" reading 10+ g above the trigger weight is almost certainly a scale fault (freeze, drift, sensor glitch), not a settled cup weight.
+2. **Scale-fault snap-to-trigger** — if the clean avg was captured but failed the plausibility check, treat the entire post-stop weight stream as corrupted and set `m_weight = m_weightAtStop`. The corpus scan found one such shot where the scale froze at ~75 g during settling on a ~40 g target; without this guard the fix would amplify the glitch from "recorded yield wrong" to "recorded yield very wrong".
+3. **`m_weightAtStop` floor** — if no clean avg was ever observed (cup lifted before the window filled), raise `m_weight` to at least the SAW trigger weight. Post-stop drip can only add weight; persisting a value below the trigger weight is physically impossible.
+4. **Unchanged** — if the fallback can't help (no SAW trigger captured, or `m_weight` already above stop weight with no implausible clean-avg signal), leave it as today's value.
 
 Learning is still skipped on the cup-removed path — corruption of the post-stop drip measurement is the only reason this branch exists. The fallback only affects the persisted `finalWeightG` (and therefore the AI advisor's `yieldG`, the visualizer export, and `DyeSettings::dyeDrinkWeight`).
 
--- a/docs/CLAUDE_MD/SAW_LEARNING.md
+++ b/docs/CLAUDE_MD/SAW_LEARNING.md
@@ -132,6 +132,40 @@ Median rejects single outliers but does not detect the case where the user chang
 
 A new profile starting from `sensorLag(scaleType) + 0.1 s` will be off until 3 shots populate its history (one batch). With `globalSawBootstrapLag`, it starts from "what we have learned about this user's machine on this scale across their other profiles" — a much closer prior. Flow cal uses the same idea (median of espresso per-profile multipliers) and it noticeably reduces the cold-start error.
 
+## Settling and final-weight capture
+
+After SAW fires the stop command, [`ShotTimingController`](../../src/controllers/shottimingcontroller.cpp) runs a settling state machine that watches the scale stabilize. The cup gains a bit more weight from post-stop drip (typically 0.5–1.5 g) before reaching its true final weight, and learning needs both `m_weightAtStop` (the trigger weight) and the settled value (`m_weight`) to compute drip and update the model.
+
+### Stability gate and clean-avg capture
+
+Each settling sample is added to a 6-sample rolling window. The "stable enough to commit" gate fires when all three hold:
+
+- `avgDrift < SETTLING_AVG_THRESHOLD` (0.3 g) — window mean has stopped shifting
+- `weight ≤ avg + SETTLING_ABOVE_AVG_MARGIN` (0.2 g) — current reading has caught up to the mean (drip has effectively stopped)
+- `avg ≥ m_weightAtStop − 0.5` — the average isn't still recovering from pump-vibration artifacts
+
+Once the gate has held for `SETTLING_STABLE_MS = 1000 ms`, `onSettlingComplete` writes `m_weight = avg` and the learning point is committed.
+
+Every time the gate condition is satisfied (even before the 1000 ms accumulates), the avg is also captured into `m_lastCleanSettlingAvg`. This is the last-known-good settled value — held in reserve for the cup-lift case below.
+
+### Final-weight on cup removal
+
+If the user lifts the cup before settling completes, the cup-removal detector fires on a >20 g drop (single-step OR cumulative-from-peak). The handler **skips learning** (a corrupted weight stream can't teach the predictor) but the shot still needs a `finalWeightG` to persist.
+
+Before issue [#1280](https://github.com/Kulitorum/Decenza/issues/1280), `m_weight` was left at whatever value the last accepted sample had. In practice that was often a *cup-lift spike artifact* that squeaked past the 20 g cup-removal threshold — e.g. shot 5470 persisted `38.5 g` even though the cup had actually settled at `42.3 g` for ~700 ms and SAW had correctly stopped at `41.2 g`. The AI advisor then reasoned from `yield=38.5, target=42` and invented a "you stopped manually" narrative.
+
+The cup-removed branch now applies this fallback chain to restore `m_weight`:
+
+1. **Last clean settling avg** — the value captured by the stability gate above. This is the truth when at least one stable sample was observed (the common case for any settle that lasted long enough for the user to lift the cup).
+2. **`m_weightAtStop` floor** — if no clean avg was ever observed (cup lifted before the window filled), raise `m_weight` to at least the SAW trigger weight. Post-stop drip can only add weight; persisting a value below the trigger weight is physically impossible.
+3. **Unchanged** — if the fallback can't help (no SAW trigger captured, or `m_weight` already above stop weight), leave it as today's value.
+
+Learning is still skipped on the cup-removed path — corruption of the post-stop drip measurement is the only reason this branch exists. The fallback only affects the persisted `finalWeightG` (and therefore the AI advisor's `yieldG`, the visualizer export, and `DyeSettings::dyeDrinkWeight`).
+
+### Offline replay
+
+`./shot_eval --settling ~/shot_corpus/*.json` parses the `app.data.debug_log` embedded in each saved shot and replays the settling sample stream offline, printing `recorded` (today's `finalWeightG`) vs `postFix` (what the new fallback chain would persist). Useful for scanning a personal shot corpus for affected shots — most rows will print equal values; cup-lift-mid-settle shots surface with a `***` flag.
+
 ## User Experience
 
 - **Default ON, automatic** — there is no setting; SAW learning is always on.
@@ -173,9 +207,12 @@ A user who hits "Reset all" gets a clean slate (legacy + new keys both wiped) an
 
 - [src/core/settings.h](../../src/core/settings.h) / [src/core/settings.cpp](../../src/core/settings.cpp) — schema, batch accumulator, read-path fallback chain, bootstrap recompute. The new public API mirrors flow cal: `sawLearnedLagFor`, `getExpectedDripFor`, `sawLearningEntriesFor`, `sawModelSource`, `resetSawLearningForProfile`, `globalSawBootstrapLag`, `addSawLearningPoint(…, profileFilename)`.
 - [src/main.cpp](../../src/main.cpp) — wires `ProfileManager::baseProfileName()` into the WeightProcessor snapshot path and the `sawLearningComplete` handler, and emits the per-shot `model:` / `accuracy:` log lines.
+- [src/controllers/shottimingcontroller.cpp](../../src/controllers/shottimingcontroller.cpp) — settling state machine, stability gate, `m_lastCleanSettlingAvg` capture, cup-removed fallback chain (#1280).
 - [qml/pages/settings/SettingsCalibrationTab.qml](../../qml/pages/settings/SettingsCalibrationTab.qml) — Calibration tab UI changes (source suffix, per-profile reset).
 - [src/mcp/mcptools_control.cpp](../../src/mcp/mcptools_control.cpp) — `reset_saw_learning_for_profile` tool.
 - [tests/tst_saw_settings.cpp](../../tests/tst_saw_settings.cpp) — per-pair isolation, batch commit at N=3, dispersion-rejection, bootstrap recompute, fallback chain, reset behaviour, legacy-path preservation.
+- [tests/tst_settling.cpp](../../tests/tst_settling.cpp) — settling-state-machine tests including the #1280 cup-lift-mid-settle regression replays.
+- [tools/shot_eval/main.cpp](../../tools/shot_eval/main.cpp) — `--settling` mode for offline replay against a saved shot corpus.
 
 ## Related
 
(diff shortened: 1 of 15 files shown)


skialpine:
**Update — tightened capture criterion after corpus scan**

Ran the new `shot_eval --settling` mode against the maintainer's 953-shot local history. The original capture rule (fire on every gate match) flagged 7 affected shots, but 4 of them were false positives — noisy/oscillating settles where the rolling-window avg only transiently satisfied the gate.

Added `SETTLING_CLEAN_CAPTURE_MS = 250 ms` guard: \`m_lastCleanSettlingAvg\` is only written when the stability gate has held *continuously* for that long (≈ 3 consecutive samples at typical scale rates). Filters transients while preserving Mark's ~700 ms plateau on shot 5470.

Re-scan results:

| Shot | Before tightening | After tightening | Verdict |
|---|---|---|---|
| 729 | floor 33.7 g | (mirror artifact, unchanged in production) | — |
| 825 | +18.5 g (7 consec stable) | +18.5 g | legitimate — 700 ms genuine plateau |
| 831 | +3.5 g (4 consec stable) | +3.5 g | legitimate — drip recovery |
| **908** | +16.4 g (1 transient fire) | **eliminated** | false positive fixed ✓ |
| 909 | +18.7 g (1 transient) | +6.9 g (floor at SAW trigger) | tightened — recorded 28.3g was below SAW trigger 35.2g, physically impossible |
| 963 | (mirror artifact) | (mirror artifact, unchanged in production) | — |
| **988** | +5.0 g (non-consec fires) | **eliminated (Δ=0)** | false positive fixed ✓ |

Net: 3 production-meaningful corrections remain, all defensible. 0 regressions.

Added new test \`cupLiftAfterNoisyPlateauDoesNotCaptureTransients_1280\` that asserts \`m_lastCleanSettlingAvg\` stays zero on a noisy/oscillating settle. Existing 5470 test bumped to \`QTest::qWait(50)\` so the controller's stability clock accumulates past the new threshold.

Full ctest: 51/51 pass. Detector corpus: 21/21 pass.

skialpine:
**Update 2 — implausibility cap added after maintainer review**

Maintainer confirmed shot 825's "7 consecutive stable samples at avg=74.8 g" was a scale freeze, not a real settled cup weight (they don't pull 70g shots). The continuous-stability gate alone can't distinguish a true plateau from a frozen reading — both produce arbitrarily many consecutive stable samples.

Added \`MAX_PLAUSIBLE_POST_STOP_DRIP_G = 5 g\` cap: a captured clean avg whose overshoot above \`m_weightAtStop\` exceeds this is treated as a scale fault, and the post-stop stream is considered corrupt (so \`m_weight\` at cup-removal is also unreliable). \`finalWeightG\` snaps to \`m_weightAtStop\` — the only physically defensible minimum-truth value.

Re-scan after this guard:

| Shot | Pre-cap postFix | Post-cap postFix | Verdict |
|---|---|---|---|
| **825** | 74.8 g (+18.5 over recorded) | **39.5 g (-16.8)** | scale fault correctly rejected; snapped to SAW trigger |
| 831 | 42.9 g (+3.5) | 42.9 g (unchanged) | within 5 g cap, drip recovery intact |
| 909 | 35.2 g (+6.9 floor) | 35.2 g (unchanged) | floor unchanged |
| Mark's 5470 | 42.4 g (+3.9) | 42.4 g (unchanged) | within 5 g cap, recovery intact |

5 g cap is conservative — real drip is 0.5–3 g, even slow-flow profiles stay under 5 g. The existing SAW learning model already implicitly bounds drip via its \`drip/flow > 4 s\` rejection.

New unit test \`implausibleCleanAvgIsRejectedAsScaleFault_1280\` exercises a synthetic frozen-scale scenario (74.8 g for 15 samples) and asserts \`finalWeight\` snaps to \`m_weightAtStop\` (39.5 g), not the glitch value.

All tests: 4/4 settling regression, 51/51 ctest, 21/21 detector corpus.

Confidence is now meaningfully higher — the corpus scan revealed two distinct edge-case classes (single-sample transients in noisy settling, and scale-fault freezes) and both have dedicated regression tests.

skialpine:
**Update 3 — review feedback applied**

All issues from the PR review (`/pr-review-toolkit:review-pr`) addressed in commit \`9d860d5\`:

**Important fixes:**
- **E1**: Each branch of the cup-removed fallback chain now logs which path fired. The scale-fault rejection emits a \`qWarning\` with the overshoot magnitude; recovery/floor/no-op branches emit \`qDebug\`. Per-shot debug logs can now distinguish the four branches at post-mortem.
- **E2**: One \`qDebug\` per settling cycle on the first clean-avg capture (uses the existing per-cycle reset, no new state).
- **E3**: Cross-reference comment at the cup-removal detection site warning maintainers that \`tools/shot_eval/main.cpp\` mirrors the logic and must be updated in lockstep.
- **T1**: \`cleanAvgClearedBetweenShots_1280\` — asserts the cross-shot reset invariant on both \`startShot()\` and \`startSettlingTimer()\`.
- **T3**: \`cleanAvgSurvivesPostCaptureGateFailure_1280\` — captured value persists when gate fails post-capture (protects against a "tighten the gate" refactor that would break Mark's actual reproducer in a way the existing tests don't catch).
- **C1**: Replaced two stale \`shotsummarizer.cpp:1405\` anchors with stable string anchors — the rubric has drifted to line 1417.
- **C2**: Updated \`shot_eval\`'s "Mirrors" doc block to list \`MIN_CONSECUTIVE_STABLE_FIRES\` and \`MAX_PLAUSIBLE_POST_STOP_DRIP_G\` (added in later commits).

**Coverage additions:**
- **T2**: \`carriesStoppedByVolume\` — the third allowlist case for SAV shots.
- **T4**: \`summarize_livePath_carriesStoppedByThroughBuildShotBlock_1280\` — exercises the new defaulted parameter on the live \`summarize()\` overload.

**Polish:**
- **M3**: Clarified the "15 × 50 ms = 9 × 50 ms = 450 ms" math in the frozen-scale test comment.

**Verification:**
- 51/51 ctest pass
- 21/21 detector corpus pass
- 6/6 settling regression tests pass (4 original + 2 new)
- 6/6 stoppedBy tests pass (4 original + 2 new)
- \`shot_eval --settling shot5470.json\` still reports the 38.5 → 42.4 g recovery for Mark's shot

The PR now has dedicated regression coverage for every invariant the fix depends on, and the per-shot debug log carries enough trail to diagnose any future report.

skialpine:
### Code review

Found 6 issues:

1. Missing `m_lastCleanSettlingAvg = avg` capture in the `onDisplayTimerTick` stable-completion path. The PR's own [design.md](https://github.com/Kulitorum/Decenza/blob/9d860d5997ac2ba98bdce279b76da5950b20aa1e/openspec/changes/preserve-settled-weight-on-cup-lift/design.md#L46-L52) commits to "Capture in BOTH `onWeightSample` and `onDisplayTimerTick` stable branches" and [tasks.md step 2.3](https://github.com/Kulitorum/Decenza/blob/9d860d5997ac2ba98bdce279b76da5950b20aa1e/openspec/changes/preserve-settled-weight-on-cup-lift/tasks.md) lists the same. The code only captures in `onWeightSample` — the BLE-silence path at the timer-tick stable branch writes `m_weight = avg` but not `m_lastCleanSettlingAvg`. Not a current bug (cup-removed only fires from `onWeightSample`) but a spec-conformance and future-safety gap exactly as the design contemplates.

https://github.com/Kulitorum/Decenza/blob/9d860d5997ac2ba98bdce279b76da5950b20aa1e/src/controllers/shottimingcontroller.cpp#L529-L538

2. Three live-path `summarize()` callers in `aimanager.cpp` silently default the new `stoppedBy` parameter to empty. [tasks.md step 3.5](https://github.com/Kulitorum/Decenza/blob/9d860d5997ac2ba98bdce279b76da5950b20aa1e/openspec/changes/preserve-settled-weight-on-cup-lift/tasks.md) says "Update all callers... to pass the already-resolved `stoppedBy`" and [advisor-user-prompt spec Scenario 1](https://github.com/Kulitorum/Decenza/blob/9d860d5997ac2ba98bdce279b76da5950b20aa1e/openspec/changes/preserve-settled-weight-on-cup-lift/specs/advisor-user-prompt/spec.md) says the live path SHALL carry it. The QML-facing Q_INVOKABLE `analyzeShot` plus the email/share export paths all emit shot prompts without the stop-reason anchor. Production AI advisor today routes through `summarizeFromHistory` instead, so not user-visible — but the spec is unmet.

https://github.com/Kulitorum/Decenza/blob/9d860d5997ac2ba98bdce279b76da5950b20aa1e/src/ai/aimanager.cpp#L714-L718

3. `m_lastCleanSettlingAvg` member doc lists three gate conditions (drift, weight-above-avg, avg-above-stop) but omits the fourth: `avgStableMs >= SETTLING_CLEAN_CAPTURE_MS` (250 ms continuous stability). A reader of the header alone would assume single-sample capture; the field actually needs 250 ms of held stability.

https://github.com/Kulitorum/Decenza/blob/9d860d5997ac2ba98bdce279b76da5950b20aa1e/src/controllers/shottimingcontroller.h#L187-L196

4. Stale line-number reference in `shot_eval`. Comment says "matches src/controllers/shottimingcontroller.cpp line 313" but line 313 is now a section-header comment (`// Rolling average stability detection`); the actual three-condition stability gate is at line 367. PR review feedback already replaced other stale `:1405` line refs with stable string anchors in this PR — same fix applies here.

https://github.com/Kulitorum/Decenza/blob/9d860d5997ac2ba98bdce279b76da5950b20aa1e/tools/shot_eval/main.cpp#L890-L895

5. Inconsistent scale-rate assumptions in adjacent header comments. `SETTLING_WINDOW_SIZE` claims "~1.5s at ~4Hz" (250 ms/sample); `SETTLING_CLEAN_CAPTURE_MS` two lines later claims "~100 ms scale cadence" (10Hz). 6 samples × 100 ms = 600 ms, not 1.5 s. CLAUDE.md says "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'."

https://github.com/Kulitorum/Decenza/blob/9d860d5997ac2ba98bdce279b76da5950b20aa1e/src/controllers/shottimingcontroller.h#L161-L175

6. Stale PR attribution. Comment reads `// #1280: forwarded into buildShotBlock` but `ShotProjection::stoppedBy` was added by #1161 (verifiable via `shothistorystorage.h` and `shotprojection.h` line 127). The forwarding is correctly #1280's contribution; the comment as written may mislead readers tracing the field's origin.

https://github.com/Kulitorum/Decenza/blob/9d860d5997ac2ba98bdce279b76da5950b20aa1e/src/ai/shotsummarizer.cpp#L391-L395

<details>

- [Worth a look] Fallback-chain comment lists 3 branches but code implements 4 (scale-fault snap is omitted from the inline list, though SAW_LEARNING.md has the full 4-item version). One-line fix in code already changed. (score: 72)
- [Worth a look] No test asserts `wasSawTriggered() == true` after `shotProcessingReady` fires on the cup-removed path; a refactor that mistakenly cleared `m_stopAtWeightTriggered` would misclassify all cup-lift SAW shots as "profileEnd" with no failing test. One-line addition to `cupLiftMidSettlePreservesLastStableAvg_1280`. (score: 72)
- [Worth a look] Corpus stats inconsistency: code comment says "4 false-positive captures" while SAW_LEARNING.md says "2/953 false-positive transients". Likely reconcilable (2 shots × 2 events = 4) but readers can't tell. (score: 72)
- [OK to skip] `shot_eval` mirror doc block omits `SETTLING_STABLE_MS = 1000` (intentionally not mirrored — the offline tool only replays the cup-removal fallback chain). Adding one explanatory line would help maintainers but no current bug. (score: 72)

</details>

skialpine:
### Code review

Re-review of latest commit \`819019c6\` (which addressed all 10 findings from the prior review). Verified 9 of 10 fixes landed cleanly. Found 1 partial fix:

1. F3 was only fixed in one of the two places it appears. \`src/ai/shotsummarizer.cpp\` now correctly reads "ShotProjection.stoppedBy was introduced by #1161 (see shotprojection.h:127); #1280 added the forwarding into buildShotBlock." But the new \`analyzeShotWithMetadata\` doc-comment in \`aimanager.h\` still attributes \`stoppedBy\` to "(#1280)" without that qualification. A reader of the header tracing the field's lineage would be misled the same way the original F3 finding described. Same trivial fix — add the two-part attribution.

https://github.com/Kulitorum/Decenza/blob/819019c6959a4445e3f2574018f26a015eb12448/src/ai/aimanager.h#L91-L100

<details>
<summary>Re-verification of prior findings</summary>

All 9 other findings verified clean in commit \`819019c6\`:

- I1: \`m_lastCleanSettlingAvg = avg\` capture added in \`onDisplayTimerTick\` stable branch
- I2: \`stoppedBy\` threaded through all three live-path \`summarize()\` callers
- C1: Fallback-chain comment now lists 4 branches (matches code structure)
- C2: Stale "matches line 313" replaced with stable string anchor
- C3: Member doc-comment now documents the \`SETTLING_CLEAN_CAPTURE_MS\` time gate
- C4: Conflicting 4Hz/10Hz claims consolidated to single "4–10 Hz across supported hardware" block
- F2: \`wasSawTriggered()\` assertion added to cup-removed regression test
- F3 (shotsummarizer.cpp site): #1161 vs #1280 attribution clarified
- A1: Corpus stats now read "2 false-positive transients vs 2 legitimate recoveries" in both code comment and SAW_LEARNING.md
- C6: \`SETTLING_STABLE_MS\` omission from shot_eval mirror now documented as intentional

No new bugs introduced in the latest commit (verified by independent deep-bug-scan, git-history, and CLAUDE.md-compliance reviewers).

</details>

PR #1283, opened 2026-05-27
wifi-scale: validate IP cache + manual entries; surface mDNS in log; fix primary invariant

### LINKED ISSUE #1281: [Bug] scale wifi constant disconnects
### What happened?

With HDS Firmware 3.0.9 eventually managed to  get stable connection to home network, but on build 3420 kept getting constant disconnects in Decenza. No automatic failover to Bluetooth. 

[scale_debug_log.txt](https://github.com/user-attachments/files/28286434/scale_debug_log.txt)

### Steps to reproduce

Turn on scale. Wait for steady network icon. Start Decenza and give it time to connect (takes a while on network compared to BT). After connection, dropouts every few seconds. Had to drop back to BT

### Decenza version

3420

### Platform

Android

### Device / OS version

Samsung A8

### Scale (if applicable)

HDS

_No response_

_No response_

### Screenshots or video

_No response_

### Anything else?

_No response_

skialpine:
## Summary

Fixes #1281 (WiFi scale stuck in a 30 s WiFi↔BLE failover loop, BLE failover repeatedly clobbered) and a coupled primary-scale state-management bug surfaced during the investigation.

- **Cached-IP validation**: any WS error on a cached-IP attempt (incl. 403 from a poisoned IP — the router) now evicts the cache and falls back to mDNS; recognition timeout also evicts. Today only HostNotFound/Refused/Network/Timeout triggered the fallback, so a 403 dead-ended and re-triggered every cycle.
- **HDS-aware proactive switch-back probe**: was a bare TCP connect on port 80 (router/printer/NAS passed); now opens `ws://<ip>/snapshot` and requires a valid HDS frame within 3.5 s. Stops the 30 s "switch back to WiFi" timer from chasing phantoms.
- **Manual "Add WiFi Scale" entry validates before saving**: persistence is deferred until `DecentScaleWifi::recognizedAsHds` confirms the typed endpoint really is a scale. A typo or wrong IP no longer poisons the saved primary; instead a "Couldn't verify a scale at *<address>*" dialog appears. Opening the dialog also fires an mDNS probe and offers a one-tap "Use" banner if the HDS is on the LAN.
- **mDNS diagnosability**: `WifiScaleDiscovery::logMessage` routes through `appendScaleLog`, so future user-shared debug logs show whether mDNS replied / timed out / hit a multicast issue without adb logcat.
- **Primary-scale invariant**: a non-empty Known Devices list now always has exactly one primary. Fixes the picker-with-nothing-selected state where the user has to manually tap a scale to set primary even though one was previously connected. Three coupled edits: addKnownScale auto-promotes; orphan-heal repairs `scale/address` ← `primaryAddress` drift in both directions; main.cpp startup load prefers `primaryScaleAddress()` over the legacy single-scale key.

## Test plan

- [ ] Power on HDS, open Settings → Connections → Add WiFi Scale. mDNS banner should appear within ~2 s offering hds.local. Tap **Use** → scale connects, banner closes, scale shows as primary.
- [ ] Same dialog, type a wrong IP (e.g. your router's IP). Tap Connect → after ~3-5 s, a "Couldn't verify a scale at <ip>" dialog appears. Saved primary is unchanged. Check `scale_debug_log.txt`: should NOT contain the typed IP under known scales.
- [ ] Simulate the original #1281 bug: in QSettings, set `scale/wifiIp/hds.local` to your router's IP. Launch app. Within one connect cycle the cache should self-evict (log line: `Cached IP X unreachable (…) — evicting cache and falling back to hostname hds.local`) and recover via mDNS.
- [ ] Power off the HDS, open Add WiFi Scale. After ~2-5 s the dialog should NOT show a banner, and `scale_debug_log.txt` should contain `Manual-entry mDNS: no HDS scale responded`.
- [ ] Pair both a WiFi and a BLE Decent scale across two sessions. Confirm primary persists across restarts and the Known Devices picker always shows one selected.
- [ ] Connect a WiFi scale, then plug in USB scale + Decent BLE scale. Switch primary between all three from the picker; each switch should disconnect the previous and connect the new without leaving the picker in a "nothing selected" state.
- [ ] During brewing, do NOT trigger a WiFi-primary switch-back (the proactive probe is idle-only). Verify the new probe doesn't accidentally fire mid-shot.

### DIFF
--- a/qml/pages/settings/SettingsConnectionsTab.qml
+++ b/qml/pages/settings/SettingsConnectionsTab.qml
@@ -204,9 +204,24 @@ Item {
         width: Math.min((parent ? parent.width : Theme.scaled(420)) * 0.85, Theme.scaled(420))
         padding: 0
 
+        // mDNS lookup state while the dialog is open. Set by the BLEManager
+        // signals below. The hostname is shown as a one-tap shortcut so the
+        // user doesn't have to type it (or guess between "hds.local" and an IP).
+        property bool mdnsSearching: false
+        property string mdnsHostname: ""
+        property string mdnsIp: ""
+        readonly property bool mdnsFound: mdnsHostname.length > 0
+
         onOpened: {
             wifiScaleHostField.text = ""
             wifiScaleHostField.forceActiveFocus()
+            // Reset prior probe state and kick off a fresh mDNS lookup. If the
+            // scale is on the LAN we offer it as a one-tap option above the
+            // text field, sidestepping the typo-and-bad-IP failure mode entirely.
+            mdnsHostname = ""
+            mdnsIp = ""
+            mdnsSearching = true
+            BLEManager.probeMdnsForManualEntry()
         }
 
         function submitWifiScale() {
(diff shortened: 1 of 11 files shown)


skialpine:
### Code review

Found 7 issues:

1. `connectToWifiScale` docstring contradicts the new deferred-persistence behavior — still claims "main.cpp records it in Known Devices + as primary when the connect is initiated, not on success", but the PR's central change defers persistence for manual entries until `DecentScaleWifi::recognizedAsHds` (comment-rot).

https://github.com/Kulitorum/Decenza/blob/fe15c51b9066ffe6adbf4ac64814071628658daa/src/ble/blemanager.h#L333-L347

2. `main.cpp` proactive switch-back comment still describes the old probe as "a TCP dial of the cached IP" — the PR rewrote it to a WebSocket HDS identity probe (comment-rot).

https://github.com/Kulitorum/Decenza/blob/fe15c51b9066ffe6adbf4ac64814071628658daa/src/main.cpp#L1320-L1330

3. `m_userInitiatedShutdown` docstring says "Six sites set this to true" and enumerates them, but the PR added a 7th site in `onError`'s cached-IP eviction path (`decentscalewifi.cpp:737`). Count and list are stale (comment-rot).

https://github.com/Kulitorum/Decenza/blob/fe15c51b9066ffe6adbf4ac64814071628658daa/src/ble/scales/decentscalewifi.h#L179-L191

4. `recognitionFailed` lambda in `main.cpp` emits `manualWifiValidationFailed` and notes "the driver's onRecognitionTimeout aborts the socket itself, so the half-open scale tears down on its own" — but `physicalScale` (the `unique_ptr<ScaleDevice>`) is never reset. A zombie disconnected `DecentScaleWifi` lingers; if the reconnect-timer or any other trigger reaches the `scaleDiscovered` handler's type-unchanged branch, it re-wires to the same dead driver and re-dials the unvalidated hostname (bug).

https://github.com/Kulitorum/Decenza/blob/fe15c51b9066ffe6adbf4ac64814071628658daa/src/main.cpp#L1816-L1830

5. `probeWifiPrimaryReachable` declaration comment in `blemanager.h` still says "NON-disruptive TCP liveness check of the WiFi scale's cached IP:80" — the PR replaced the TCP probe with a WS handshake that requires a valid HDS frame. The member-variable comment for `m_wifiProbeWebSocket` was updated; this public-API comment was not (comment-rot).

https://github.com/Kulitorum/Decenza/blob/fe15c51b9066ffe6adbf4ac64814071628658daa/src/ble/blemanager.h#L104-L116

6. `onManualWifiMdnsDiscovered` comment says "only surface the suggestion if the user still has the dialog open and hasn't already started typing", but the code only checks `!addWifiScaleDialog.opened` — a `wifiScaleHostField.text.length` guard is missing, so the mDNS result can populate state under partially-typed input (comment claims a guard that isn't implemented).

https://github.com/Kulitorum/Decenza/blob/fe15c51b9066ffe6adbf4ac64814071628658daa/qml/pages/settings/SettingsConnectionsTab.qml#L468-L480

7. `manualWifiMdnsProbeFinished` signal comment says "QML uses this to drop a 'Searching...' indicator", but no UI element in `SettingsConnectionsTab.qml` binds to `mdnsSearching` — the property is set/cleared but never read by any `visible:`/`opacity:`/`text:` binding (doc/code mismatch).

https://github.com/Kulitorum/Decenza/blob/fe15c51b9066ffe6adbf4ac64814071628658daa/src/ble/blemanager.h#L471-L478

PR #1287, opened 2026-05-27
feat(scale): parse session_info + log debug frames for Decent WiFi scale

skialpine:
## Summary

Two related fixes for the Decent WiFi scale driver, grouped so they land together.

### 1. Parse `session_info` frames (commit 1)

Newer HDS firmware moved `firmware_version` and `protocol_version` out of the live `status` frame and into a dedicated `session_info` frame sent unsolicited once per connect; it also added a new `reset_reason` field there. Confirmed in production logs from the test tablet today.

- New `handleSessionInfoFrame()` captures `firmware_version` and `protocol_version` with the same change-detection logic used for status, and logs `reset_reason` (`poweron` / `brownout` / `watchdog` / …) so a watchdog reset mid-session is distinguishable from a clean power-cycle at triage time.
- Old-firmware status-frame parsing for `firmware_version` / `protocol_version` left in place. Both paths assign to the same members, so a transitional firmware that emits both shapes only logs once.
- `m_lastResetReason` cleared on disconnect alongside the other per-connect state.

### 2. Log debug frames + request snapshot on app suspend (commit 2)

The HDS firmware sends debug-event frames unprompted once `events on` is opted into (`stall_start` / `stall_end` / `adc_recovery` / `temp_peak`), and responds to a `debug` text command with a full-state snapshot (`soc_temp_c`, `soc_temp_max_c`, `weight_stalled`, `stall_count`, `last_stall_ms`, `last_stall_temp_c`, `adc_recovery_count`, `ms`).

- `handleDebugFrame()` logs each debug frame verbatim as compact JSON. Both shapes land here and get one log line carrying every field — no filtering, since we don't yet know which fields matter most.
- New `DecentScaleWifi::requestDebugSnapshot()` sends the `debug` text command.
- `main.cpp`'s existing `applicationStateChanged → Qt::ApplicationSuspended` hook calls it (qobject_cast guard so other scale types are a no-op) so the scale log captures firmware state immediately before the app backgrounds.

## Test plan

- [x] `tst_DecentScaleWifi` — all 34 tests pass with no warnings.
- New tests in this PR:
  - `sessionInfoFrameLogsFirmwareProtocolAndResetReason` — first-log, re-send-silent, mid-connect-change.
  - `debugFramesAreLoggedVerbatim` — covers both the event-driven and request-response debug-frame shapes.
  - `requestDebugSnapshotSendsDebugCommand` — verifies the suspend-time request sends the `debug` text command.
- [ ] On a real tablet: connect a WiFi HDS, confirm scale log shows `Firmware version: …`, `Protocol version: …`, `Scale reset reason: …` once per connect (from session_info on new firmware, from status on old firmware).
- [ ] Background the app and confirm the scale log shows `Debug frame: {"type":"debug","status":"ok",…}` capturing the full firmware state.

### DIFF
--- a/src/ble/scales/decentscalewifi.cpp
+++ b/src/ble/scales/decentscalewifi.cpp
@@ -446,10 +446,11 @@ void DecentScaleWifi::handleStatusFrame(const QJsonObject& obj) {
 // protocol_version) and the boot-cause (reset_reason) into its own session_info
 // frame, sent unsolicited right after the WS handshake; the live status frame
 // no longer carries firmware_version / protocol_version. Older firmware put
-// them in status — handleStatusFrame still reads them there. Both paths assign
-// to the same m_firmwareVersion / m_loggedProtoVersion members with
-// change-detection, so a transitional firmware that sends both shapes only
-// logs once.
+// firmware_version and protocol_version in status — handleStatusFrame still
+// reads them there. For those two fields both paths assign to the same
+// m_firmwareVersion / m_loggedProtoVersion members with change-detection, so
+// a transitional firmware that sends both shapes only logs each once.
+// reset_reason has no legacy counterpart in status and is session_info-only.
 void DecentScaleWifi::handleSessionInfoFrame(const QJsonObject& obj) {
     onRecognizedAsHds();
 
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Found 1 issue:

1. The new suspend-time comment claims iOS `bluetooth-central` background mode helps keep the WiFi WebSocket connected through suspension. That mode only preserves the CoreBluetooth/BLE stack — it does **not** extend TCP socket lifetime or keep the event loop running. A plain `QWebSocket` over TCP gets no grace from `bluetooth-central` on iOS; the socket and the event loop will be suspended together. The honest description is the second half of the same paragraph ("the response usually lands … If it doesn't, the request is still on the wire and costs nothing — and the answer arrives on the next resume"). The `bluetooth-central` parenthetical should be dropped from the iOS clause, or replaced with an accurate note that iOS will suspend the WS and the response lands on resume.

https://github.com/Kulitorum/Decenza/blob/ce58b9bb2cc4d4c0a4b87eb8223efd0f379b7f22/src/main.cpp#L2773-L2784

<details>

- [Worth a look] A mid-connect change in `reset_reason` logs at debug level (`WIFI_LOG`), but the comparable mid-connect `firmware_version` change a few lines above logs at WARN. The PR's own comment argues `reset_reason` is operationally significant (disambiguates watchdog from clean reboot), so an asymmetric severity is hard to justify — one-line fix in code already being changed (score: 50)
- [OK to skip] The `handleSessionInfoFrame` doc-comment says "Both paths assign to the same `m_firmwareVersion` / `m_loggedProtoVersion` members" — accurate for those two fields, but `reset_reason` is session_info-only. A reader auditing what "both paths" covers could be momentarily misled. Pure clarity nitpick, no functional impact (score: 50)

</details>

PR #1288, opened 2026-05-27
connections: replace Known Scales picker with a proper ComboBox

skialpine:
## Summary
- Replaces the custom rectangle + hidden `SelectionDialog` picker on the Connections settings page with a real Qt `ComboBox`. The previous control looked like a status indicator (red-bordered card, star icon) but was actually a dropdown trigger — and only when `knownScales.length > 1` (the chevron was conditional). The refractometer row right beneath it used a nearly identical visual but wasn't interactive at all, which made the picker even more confusing.
- The closed control now shows: live connection dot · primary scale name · transport badge (`BLE` / `WiFi` / `USB`) · always-present chevron.
- The popup is capped at **3 visible rows**, with a `ScrollIndicator` for longer lists — works cleanly when the user has 5+ saved scales without opening a modal dialog.
- Per-row delegate: filled star on the primary, dimmed star on others, name, transport badge.
- Same selection wiring as before: `Settings.setPrimaryScale` + `BLEManager.setSavedScaleAddress` + `connectToSavedScale`. The standalone `knownScaleDialog` `SelectionDialog` is deleted.

## Test plan
- [ ] With 0 saved scales: Known Devices section stays hidden (unchanged behavior).
- [ ] With 1 saved scale: ComboBox shows the scale, chevron present; tapping opens a popup with one entry that re-selecting is a no-op.
- [ ] With 2–4 saved scales: popup shows all rows, no scrolling.
- [ ] With 5+ saved scales: popup shows 3 rows and scrolls; star indicator clearly marks the primary.
- [ ] Selecting a non-primary row promotes it: closed control updates, `Connected:` status switches over to the new scale.
- [ ] `Forget` removes the primary and the ComboBox re-syncs to the next primary (existing behavior preserved).
- [ ] Mixed transports: badge reads `WiFi` for `decent-wifi`, `USB` for `decent-usb`, `BLE` for everything else.
- [ ] Refractometer row below still renders and Forget still works.

### DIFF
--- a/qml/pages/settings/SettingsConnectionsTab.qml
+++ b/qml/pages/settings/SettingsConnectionsTab.qml
@@ -1157,7 +1157,18 @@ Item {
                                     border.width: 1
                                 }
 
-                                // Closed-state content: live status dot + name + transport badge + chevron.
+                                // Custom indicator (chevron) replacing the default
+                                // so we don't end up with two stacked chevrons.
+                                indicator: Text {
+                                    x: scalePicker.width - width - Theme.scaled(10)
+                                    y: (scalePicker.height - height) / 2
+                                    text: "\u25BC"
+                                    font.pixelSize: Theme.scaled(10)
+                                    color: Theme.textSecondaryColor
+                                }
+
+                                // Closed-state content: live status dot + name + transport badge.
+                                // The chevron is drawn by `indicator` above.
                                 contentItem: RowLayout {
                                     spacing: Theme.scaled(6)
 
@@ -1197,6 +1208,7 @@ Item {
                                         height: Theme.scaled(18)
                                         radius: Theme.scaled(9)
                                         color: Qt.rgba(Theme.accentColor.r, Theme.accentColor.g, Theme.accentColor.b, 0.22)
+                                        Layout.rightMargin: Theme.scaled(24)
                                         Text {
                                             id: badgeText
                                             anchors.centerIn: parent
@@ -1205,14 +1217,6 @@ Item {
                                             font.pixelSize: Theme.scaled(10)
                                             font.bold: true
                                         }
-                                    }
-
-                                    Text {
-                                        Layout.rightMargin: Theme.scaled(10)
-                                        text: "\u25BC"
-                                        font.pixelSize: Theme.scaled(10)
-                                        color: Theme.textSecondaryColor
-                                        Accessible.ignored: true
                                     }
                                 }
 
@@ -1278,7 +1282,9 @@ Item {
                                     topPadding: Theme.scaled(4)
                                     bottomPadding: Theme.scaled(4)
                                     // Cap visible height at `visibleRows`; longer lists scroll.
-                                    height: Math.min(scalePicker.count, scalePicker.visibleRows) * scalePicker.rowHeight
+                                    // Drive sizing off the model length directly — ComboBox.count
+                                    // can lag behind a QVariantList model after refresh.
+                                    height: Math.min(Settings.knownScales.length, scalePicker.visibleRows) * scalePicker.rowHeight
                                             + topPadding + bottomPadding
 
                                     contentItem: ListView {
--- a/qml/pages/settings/SettingsConnectionsTab.qml
+++ b/qml/pages/settings/SettingsConnectionsTab.qml
@@ -1147,11 +1147,18 @@ Item {
 
                                 onActivated: function(index) {
                                     var scales = Settings.knownScales
-                                    if (index < 0 || index >= scales.length) return
+                                    if (index < 0 || index >= scales.length) {
+                                        // TOCTOU: scan/forget mutated the model between
+                                        // the user's tap and this handler. Log so we can
+                                        // tell this apart from a no-op re-select.
+                                        console.warn("scalePicker: stale activated index", index,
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 1 issue:

1. `ComboBox` uses the native `Popup` for the selection list (ACCESSIBILITY.md says: *"Never use `Popup` for lists users must navigate. TalkBack can't trap focus inside Qt `Popup` elements. Use `Dialog { modal: true }` with `AccessibleButton` delegates instead"*). The project already has [`qml/components/StyledComboBox.qml`](https://github.com/Kulitorum/Decenza/blob/695ecd8d55b4be059101c958435637970af3c363/qml/components/StyledComboBox.qml#L50-L66) which suppresses the native popup and routes selection through a `SelectionDialog` — this PR was an explicit UX choice to avoid the SelectionDialog modal in favor of the inline 3-row popup, so this is a knowing tradeoff rather than an oversight. Flagging because it's a documented project rule, but no auto-fix applied.

https://github.com/Kulitorum/Decenza/blob/695ecd8d55b4be059101c958435637970af3c363/qml/pages/settings/SettingsConnectionsTab.qml#L1288-L1306

Fixed in [695ecd8d](https://github.com/Kulitorum/Decenza/commit/695ecd8d55b4be059101c958435637970af3c363) (pushed):
- Added `Accessible.focusable: true` on the ComboBox (was missing per ACCESSIBILITY.md rule 1 table).
- Added `Accessible.role`/`name`/`focusable`/`onPressAction` to the per-row `ItemDelegate`, plus `Accessible.ignored: true` on its child Text/badge — required for "List delegate" per ACCESSIBILITY.md rule 1.
- Dropped the stale "Connected BLE scale name" comment at `SettingsConnectionsTab.qml:972` — that row has been transport-agnostic since [#1246](https://github.com/Kulitorum/Decenza/pull/1246) (pre-existing violation in a file the PR touched, per CLAUDE.md *"Fix pre-existing violations in any file you touch"*).

<details>
<summary>5 below-threshold issues (50–74)</summary>

- [OK to skip] `modelData.name` flagged as the reserved-word footgun from `QML_GOTCHAS.md`. Empirical screenshots show names rendering correctly — QML's `QVariantMap`-to-JS conversion exposes `name` fine, unlike plain JS-array models. (score: 25)
- [OK to skip] `▼` (U+25BC) as the chevron flagged as a CLAUDE.md "no Unicode glyphs as icons" violation. The project's own [`StyledComboBox` indicator](https://github.com/Kulitorum/Decenza/blob/695ecd8d55b4be059101c958435637970af3c363/qml/components/StyledComboBox.qml#L33-L40) uses the same glyph — project precedent treats it as accepted. (score: 30)
- [OK to skip] `currentIndex: indexOfPrimary()` is a one-time binding; flagged as a potential desync risk. The `Connections { onKnownScalesChanged }` handler reassigns it imperatively whenever the model mutates — covered. (score: 30)
- [Worth a look] `Forget` button `accessibleName` re-evaluates after the primary is removed, momentarily announcing just "Forget" before the next primary's name latches. Minor a11y regression that would need a captured-local-name workaround. (score: 40)
- [Worth a look] Comment at `SettingsConnectionsTab.qml:1160-1163` says "draw a single chevron…so there's only one arrow," which describes the de-duplication intent but doesn't note the behavioral change that the chevron is now unconditionally visible (the old code hid it for single-item lists). Cosmetic. (score: 50)

</details>

skialpine:
### Re-review against current HEAD

Re-ran review across three specialized agents (code-reviewer, comment-analyzer, silent-failure-hunter). All previously flagged issues are resolved; this pass surfaced two small refinements, both now fixed in [8be42363](https://github.com/Kulitorum/Decenza/commit/8be4236378a2c3f5e6dcc4b8d83648b89cd0820a):

- Added `Accessible.onPressAction` on the ComboBox so TalkBack double-tap opens the dropdown (matches the canonical [`StyledComboBox` pattern](https://github.com/Kulitorum/Decenza/blob/8be4236378a2c3f5e6dcc4b8d83648b89cd0820a/qml/components/StyledComboBox.qml#L66)).
- Wrapped the dropdown's `onOpened: forceActiveFocus()` in `Qt.callLater` so delegate items are realized before focus is requested.

https://github.com/Kulitorum/Decenza/blob/8be4236378a2c3f5e6dcc4b8d83648b89cd0820a/qml/pages/settings/SettingsConnectionsTab.qml#L1143-L1149

<details>
<summary>5 below-threshold findings (skipped or rationale)</summary>

- [OK to skip] `modelData.name` reserved-property concern — flagged at conf 95 by the agent but empirically working in screenshots; the QML_GOTCHAS rule applies to JS-array models, not QVariantList-of-QVariantMap. Same conclusion as last review.
- [OK to skip] `onActivated` silently no-ops on out-of-range index — TOCTOU race window is closed when the Dialog dismisses on selection.
- [OK to skip] `currentIndex = -1` desync when no primary exists — `removeKnownScale` auto-promotes a new primary; the parent ColumnLayout's `visible:` hides the picker when knownScales is empty.
- [OK to skip] Delegate accessible name could be " " if both name and type are falsy — C++ side guarantees populated values.
- [OK to skip] Comment polish around "first row (or current primary)" — accurate enough for the common case.

</details>

PR #1290, opened 2026-05-28
fix(scale): recover lost WiFi scale connection without user intervention

skialpine:
## Summary

Two interacting bugs that left the WiFi scale stuck "disconnected" for ~9 minutes in a real-world log, recoverable only by manually switching scales (BT → WiFi). The scale was reachable on the network the entire time.

### (1) User-initiated scan refused to auto-connect to the saved primary

\`BLEManager::onDeviceDiscovered\` (BLE path) and the WiFi \`scaleFound\` handler both gated auto-connect on \`!m_userInitiatedScaleScan\`. The original anti-hijack concern (don't grab the first non-primary scale found) only applies to non-primary scales — matching the saved primary is itself the anti-hijack guard, and a user's "Scan" tap when their scale is offline is a "please reconnect" intent.

Restructured both paths so the user-scan guard runs *after* the saved-primary match check:
- Saved primary discovered → auto-connect (any scan source).
- WiFi→BLE fallback candidate → auto-connect (existing #440-violating opt-in).
- Anything else during a user scan → list-only.
- Anything else in a background scan → ignore.

### (2) Reconnect ladder died after WiFi→BLE fallback failed

\`flowScaleFallback\` is intentionally gated to one shot per saved-scale cycle (otherwise the "No Scale Found" dialog re-pops on every retry). The reconnect timer in \`main.cpp\` was only armed via that signal, so when the WiFi→BLE fallback also failed:

1. \`scaleReconnectTimer\` had been stopped by the scale-type change (\`"Scale reconnect: timer stopped due to scale type change"\`).
2. The second \`onScaleConnectionTimeout\` hit the fall-through, but \`m_flowScaleFallbackEmitted\` was already \`true\` → no \`flowScaleFallback\` emission → timer stayed stopped.
3. Auto-recovery dead-ended.

Added a separate \`scaleRetryNeeded\` signal that fires unconditionally from the same fall-through, paired with a small re-arm handler in main.cpp. The timer's own slot self-perpetuates once started, so this only kicks it off once per failure cycle — at the long-tail 60 s delay (immediate hammering is wasteful since the failure just happened). Saved primary is then retried until it comes back online.

## Test plan

- [ ] Repro the log scenario: start the app while the WiFi scale is in deep sleep (radio off), let the WiFi→BLE fallback fail, then leave the app idle. Within ~60 s of the scale's WiFi coming back, the app should auto-connect — no manual switch required.
- [ ] User scan with no scale connected: confirm the WiFi scale auto-connects as soon as it's discovered (instead of just appearing in the list).
- [ ] User scan with a different scale picker intent: confirm non-primary scales still appear in the list without auto-connecting.
- [ ] Forget scale → background scan should NOT silently re-save (same as before — guarded by the empty-saved-address check).

### DIFF
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -893,14 +893,25 @@ void BLEManager::onDeviceDiscovered(const QBluetoothDeviceInfo& device) {
             && m_savedScaleType == QStringLiteral("decent-wifi")
             && scaleType == QStringLiteral("decent");
 
-        // Auto-connect rules:
-        //   • Saved primary → always auto-connect (even on user-initiated scans).
-        //     The user's "Scan" tap when their scale is offline is itself a
-        //     "please reconnect" — the original "don't hijack the user's pick"
-        //     concern doesn't apply when the found scale IS what they previously
-        //     chose as primary. This restores the natural "scan got my scale
-        //     back" flow that was previously only reachable via an awkward
-        //     BT-then-WiFi switch dance.
+        // Auto-connect rules (BLE path — WiFi has its own handler in
+        // the WifiScaleDiscovery::scaleFound lambda further down):
+        //   • Saved BLE primary → always auto-connect (even on user-initiated
+        //     scans). The user's "Scan" tap when their scale is offline is
+        //     itself a "please reconnect" — the original "don't hijack the
+        //     user's pick" concern doesn't apply when the found scale IS what
+        //     they previously chose as primary. The companion "don't silently
+        //     re-save a forgotten scale via addKnownScale()" concern is also
+        //     inert here: the scale is already saved, so the re-save is a
+        //     no-op against the same record. This restores the natural "scan
+        //     got my scale back" flow that was previously only reachable via
+        //     an awkward BT-then-WiFi switch dance.
+        //
+        //     Note: this rule only fires for BLE primaries. When the saved
+        //     primary is `wifi:hostname`, deviceIdentifiersMatch can never
+        //     return true here (BLE devices don't carry hostnames), so a
+        //     WiFi-primary user scanning while looking at BLE devices falls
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

No issues at the 75+ threshold remain on the latest commit. The review found one critical issue and two below-threshold ones; all three have been addressed in [b5a28a9](https://github.com/Kulitorum/Decenza/commit/b5a28a9259d2475b6bf8502b9161aef2dfe474df).

Fixed:

1. `beginWifiFallbackToBleScan` Bluetooth-unavailable early-return emitted `flowScaleFallback` but not `scaleRetryNeeded`, recreating the same dead-end the PR was meant to fix (WiFi scale offline + BLE disabled = stuck on FlowScale). The header's "EVERY connection-failure path" contract was inaccurate.

https://github.com/Kulitorum/Decenza/blob/b5a28a9259d2475b6bf8502b9161aef2dfe474df/src/ble/blemanager.cpp#L1189-L1207

<details>

- [Worth a look] The new auto-connect-rules comment claimed "saved primary always auto-connects" without qualifying that it's BLE-only by structural necessity (`deviceIdentifiersMatch` can't match `wifi:hostname` against a BLE device). Tightened in the same commit (score: 50)
- [Worth a look] The new comment dropped explicit acknowledgement that the old `addKnownScale()` re-save concern is inert on the saved-primary path (the scale is already saved, so the re-save is a no-op). Restored in the same commit (score: 50)

</details>

Skipped (score: 0):
- A `static_cast<int>(reconnectDelays.size())` line that looked like a `qsizetype` violation — `reconnectDelays` is a `std::vector<int>`, not a Qt container, so the CLAUDE.md `qsizetype` rule (which explicitly targets `QVector`/`QList`/`QString`) doesn't apply.

PR #1292, opened 2026-05-28
fix(scale): silence connection-error dialogs for remaining scales + DiFluid R2

skialpine:
## Summary

Follow-up to [#1285](https://github.com/Kulitorum/Decenza/pull/1285), which silenced the same modal "Connection Error" dialogs but only for the Decent Scale.

The identical `errorOccurred` emits were added across **all 11 scale drivers** by the Nordic BLE refactor in 8ca9da33, plus the DiFluid R2 refractometer driver. They are ungated by the scale-alerts toggle and surface on every transient BLE hiccup — annoying and repetitive. This PR applies the same fix to the remaining 10 scale drivers and the DiFluid R2.

For each driver, the connect-phase `errorOccurred` emits are dropped (keeping the existing `<BRAND>_WARN` logging):

- `errorOccurred("No transport available")` in `connectToDevice()` → replaced with `<BRAND>_WARN("connectToDevice called with no transport")`
- `errorOccurred("<Brand>... connection error")` in `onTransportError()` → dropped (transport-error WARN already logs)
- `errorOccurred("<Brand>... service not found")` in `onServicesDiscoveryFinished()` → dropped (service-not-found WARN already logs)

Drivers touched: Acaia, Atomheart Eclair, Bookoo, Difluid (Microbalance), Eureka Precisa, Felicita, Hiroia Jimmy, Skale, SmartChef, Timemore, Varia Aku, DiFluid R2 refractometer.

(Bookoo's `onTransportError` already had no emit — it intentionally keeps the connection alive on transport errors per its CCCD-tolerance comment — so only the other two paths apply there.)

Also dropped two post-handshake emits that were initially kept but are equally noisy:

- Acaia `"Scale not responding to ident"` (fires after 10 ident retries fail) — `setConnected(false)` + WARN log is enough
- Varia Aku `"scale not sending weight updates"` — `setConnected(false)` on the line above already triggers a `scaleDisconnected` dialog, so the emit was producing a **second** modal for the same event

After these changes, no BLE scale driver emits `errorOccurred` on transient/recoverable connect-phase noise. The forwarder comment at `main.cpp:1617-1624` has been updated to reflect the new contract.

If the scale stays gone, the existing (gated) "No Scale Found" path still surfaces.

## Left in place (different conditions)

- DiFluid R2 measurement-side emits — `"No liquid detected"`, `"Beyond range"`, `"R2 reported an out-of-range value"`, etc. These are meaningful device-reported conditions about the actual measurement, not BLE connect noise.
- DecentScaleWifi 503 / unrecoverable errors — already log-only for transient WiFi failures per #1253; the surviving emits are actionable (e.g. "Another client is connected to the scale").

## Test plan

- [x] macOS debug build clean (no errors, no warnings)
- [ ] Verify on Android tablet: trigger a transient BLE failure on a non-Decent scale (e.g. Skale) and confirm no modal pops up
- [ ] Confirm the gated "No Scale Found" path still appears when a scale is unreachable

### DIFF
--- a/src/ble/refractometers/difluidr2.cpp
+++ b/src/ble/refractometers/difluidr2.cpp
@@ -114,7 +114,7 @@ bool DiFluidR2::isR2Device(const QString& name) {
 
 void DiFluidR2::connectToDevice(const QBluetoothDeviceInfo& device) {
     if (!m_transport) {
-        emit errorOccurred("No transport available");
+        R2_WARN("connectToDevice called with no transport");
         return;
     }
 
@@ -204,7 +204,6 @@ void DiFluidR2::onTransportError(const QString& message) {
                              : QStringLiteral("connect attempt failed before ready (was not connected)"),
                  QString::number(reinterpret_cast<quintptr>(this), 16)));
     R2_WARN(QString("Transport error: %1").arg(message));
-    emit errorOccurred("DiFluid R2 connection error");
     m_measurementTimer.stop();
     m_initTimer.stop();
     m_connected = false;
@@ -228,7 +227,6 @@ void DiFluidR2::onServicesDiscoveryFinished() {
     if (!m_serviceFound) {
         R2_WARN(QString("DiFluid R2 service %1 not found!")
                     .arg(Refractometer::DiFluidR2::SERVICE.toString()));
-        emit errorOccurred("DiFluid R2 service not found");
         return;
     }
     m_transport->discoverCharacteristics(Refractometer::DiFluidR2::SERVICE);
--- a/src/ble/scales/acaiascale.cpp
+++ b/src/ble/scales/acaiascale.cpp
@@ -53,7 +53,7 @@ void AcaiaScale::stopAllTimers() {
 
 void AcaiaScale::connectToDevice(const QBluetoothDeviceInfo& device) {
     if (!m_transport) {
-        emit errorOccurred("No transport available");
+        ACAIA_WARN("connectToDevice called with no transport");
         return;
     }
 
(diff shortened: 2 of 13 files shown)


skialpine:
### Code review

Reviewed by 5 parallel agents covering CLAUDE.md compliance, deep bug scan, recent git history, related PR context, and code-comment adherence. The only consumer of `ScaleDevice::errorOccurred` is the modal-dialog forwarder at [main.cpp:1623-1624](https://github.com/Kulitorum/Decenza/blob/7661210f92332c3216abdb4bc504ca3b1020b7f5/src/main.cpp#L1622-L1626), so removing the emits cannot regress any non-dialog behavior. All transport-error handlers still call `setConnected(false)` and the WARN logging is preserved.

Note: the DiFluid R2 follow-up commit was added after the review agents ran, but the change is mechanically identical to the scale-driver pattern they all cleared.

<details>
<summary>3 below-threshold issues (all < 50, discarded)</summary>

- [OK to skip] Acaia `"Scale not responding to ident"` left in place — PR explicitly scopes out post-connect handshake timeouts; fires only after 10 retries / 5s, an unrecoverable terminal condition, not transient connect noise (score: 18)
- [OK to skip] Varia Aku watchdog `errorOccurred` left in place — PR explicitly scopes out the weight-watchdog timeout; only fires after MAX_WATCHDOG_RETRIES (score: 15)
- [OK to skip] Removing Acaia "No compatible service found" emit makes that path silent until the 20s `m_scaleConnectionTimer` falls back to "No Scale Found" — requires an Acaia-named device that advertises neither Pyxis nor IPS service UUIDs (vanishingly rare; WARN log still fires) (score: 42)

</details>

PR #1305, opened 2026-06-01
fix(ble): stop scale reconnect from wedging the DE1 link (#1303)

### LINKED ISSUE #1303: [Bug] BLE connection to DE1 gets dropped after some time on P80X
### What happened?

Decenza is dropping the BLE connection to the DE1 after some time (on my P80X). I tested and found this behaviors in the last 2 builds. 
Restarting the DE1 doesn't work for restoring the connection, and neither does turning bluetooth off and on in the tablet. Exiting the Decenza app and re-opening it solves it temporarily, though.

### Steps to reproduce

Open Decenza app and connect to DE1
Leave it open for some time (a few hours)
It will lose the connection to the DE1

### Decenza version

1.7.6 (build 3430)

### Platform

Android

### Device / OS version

Android 9 - P80X tablet

### Scale (if applicable)

Eureka PRecisa

[debug.zip](https://github.com/user-attachments/files/28468184/debug.zip)

_No response_

### Screenshots or video

_No response_

### Anything else?

_No response_

skialpine:
Fixes #1303 — on some Android tablets (confirmed P80X) the DE1 BLE link drops after hours and never recovers until the app is restarted.

## Root cause (from log analysis)
The saved-scale reconnect fires a direct `connectToDevice()` to the absent scale's MAC **every 60s**, and `onScaleConnectionTimeout()` only cleared a flag — it **never closed the `QLowEnergyController`**. Each attempt parked the Android BLE stack in `Connecting` for the full ~30s supervision timeout, starving the DE1 link until it dropped with an `AuthorizationError` the reconnect loop couldn't recover from (it bailed on `isConnecting()` and never re-armed).

**Controlled evidence:** a known-good setup (scale connected; refractometer hunted via passive `startScan()` ~1825×) showed **zero** DE1 BT errors over 3 days, vs the reporter's P80X (scale absent, hunted via direct connect ~744×) which wedged. Same firmware (v1352), same 60s USB-charger keepalive — the keepalive succeeds silently when the radio is uncontended, so it's a **victim, not the cause**. de1app avoids this by closing its direct connect after ~4s and relying on a passive scanner.

## Changes
- **Background scale reconnect is scan-only** — `tryDirectConnectToScale` gains `allowDirectConnect`; the 60s ladder passes `false`. A saved scale still auto-connects via `onDeviceDiscovered` when seen advertising (the refractometer path that already coexists with the DE1).
- **Foreground triggers** (device-picker switch / startup / DE1-wake) keep a direct-connect but **bound it to ~4s** (`kScaleDirectConnectAbortMs`); on the deadline `abortScaleDirectConnectIfPending()` tears down the parked controller and the parallel scan continues. `onScaleConnectionTimeout()` also aborts a lingering parked controller.
- **DE1 connect-hang watchdog** in `BleTransport` (`CONNECT_WATCHDOG_MS=35s`, exceeds the slowest real connect ~26s and Android's ~30s timeout): armed on `Connecting`, stopped on any resolution; if it fires while still wedged it aborts the hung controller and synthesizes `disconnected()` so the reconnect loop recreates it — **DE1 self-recovers without an app restart**.
- **Logging** — scale-reconnect lines routed through `appendScaleLog()` so they land in both the scale log and the system debug log.

DE1 keeps its sustained direct-connect (it sleeps but stays connectable); no supported scale needs connect-to-wake.

## Design / spec
OpenSpec change `fix-scale-reconnect-de1-wedge` (proposal/design/specs/tasks) included.

## Build
Clean via Qt Creator — app + all test targets, 0 errors / 0 warnings.

## Testing (pending, on-device)
- Scale **off** → DE1 stays connected, no errors (scale log shows `Auto-reconnect: scanning for saved scale`).
- Scale **on**, switch/startup → still connects fast.
- If a wedge occurs → `Connect watchdog: stuck in Connecting for 35s` + auto-recovery.
- Refractometer reconnect unchanged.

### DIFF
--- a/openspec/changes/fix-scale-reconnect-de1-wedge/.openspec.yaml
+++ b/openspec/changes/fix-scale-reconnect-de1-wedge/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-01
--- a/openspec/changes/fix-scale-reconnect-de1-wedge/design.md
+++ b/openspec/changes/fix-scale-reconnect-de1-wedge/design.md
@@ -1,0 +1,79 @@
+## Context
+
+Issue #1303: the DE1 BLE link drops after hours on a Decent P80X (Android 9) and never recovers until the app is restarted. Restarting the DE1 or toggling tablet Bluetooth does not help — only a process restart does, which points to wedged per-process Android BLE state rather than an adapter or machine fault.
+
+Log analysis (the reporter's #1303 log vs. a known-good 3-day log on identical firmware v1352) established the mechanism:
+
+- When a saved BT scale is **absent**, `BLEManager::tryDirectConnectToScale()` issues a direct `connectToDevice()` to the scale's MAC. With the device gone, the Android stack parks the `QLowEnergyController` in `Connecting` for the full ~30 s supervision timeout. This repeats every 60 s.
+- `onScaleConnectionTimeout()` (the ~20 s app-level timer) clears `m_directConnectInProgress` but **never closes the controller**, so the connect keeps running to the ~30 s Android timeout.
+- During those windows the DE1's writes (e.g. the 60 s USB-charger keepalive on MMR `0x803854`) fail; occasionally — coincident with the scale connect's teardown — the DE1 controller fires `AuthorizationError` and the link dies.
+- The DE1 reconnect timer in `main.cpp` then bails on `de1Device.isConnecting()` ("already connected/connecting, stopping retries") and never rearms, because the wedged connect stays in `Connecting` forever and never emits the error that would re-trigger retries. Permanent wedge.
+
+Controlled evidence: the known-good setup pursues its absent refractometer with a passive `startScan()` (~1825×) and shows **zero** DE1 errors over 3 days; the failing setup pursues its absent scale with direct connect (~744×) and wedges. The USB-charger keepalive is identical on both and succeeds silently when the radio is uncontended — it is a victim, not a cause.
+
+Reference design: de1app tries a direct connect only briefly (~4 s), then `ble close`s it and relies on a shared passive scanner, and defers scale connects under DE1 command backpressure (`bluetooth.tcl`).
+
+Current Decenza BLE structure: a single shared `QBluetoothDeviceDiscoveryAgent` (`m_discoveryAgent`), but fragmented orchestration — DE1, scale, and refractometer each have their own reconnect timer that independently calls `startScan()` (10+ call sites) and/or a parallel direct-connect.
+
+## Goals / Non-Goals
+
+**Goals:**
+- Stop background scale reconnect from parking a direct-connect against an absent address; make it scan-then-connect like the refractometer path.
+- Keep scale connects fast at intentional moments via a single bounded (~4 s) direct-connect that actually aborts on timeout.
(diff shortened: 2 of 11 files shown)


skialpine:
### Code review

Found 1 issue:

1. **The scale-side teardown is a no-op for BLE scales — the foreground 4s abort and the timeout cleanup never actually close the parked controller.** `abortScaleDirectConnectIfPending()` and `onScaleConnectionTimeout()` call `m_scaleDevice->disconnectFromScale()`, but transport-based BLE scales (`DecentScale`, `EurekaPrecisaScale`, etc.) park their controller in `QtScaleBleTransport::m_controller`, not the base `ScaleDevice::m_controller`. No BLE scale assigns the base `m_controller` (it's always null), and only `DecentScaleWifi` overrides `disconnectFromScale()` — so the base implementation runs and tears down nothing but `setConnected(false)`. The radio stays held in `Connecting` for the full ~30s, exactly as before. The adjacent comment ("`disconnectFromDevice()` disconnects the controller's signals first…") describes a path that never executes. The correct teardown is the transport's own `m_transport->disconnectFromDevice()` (what `DecentScale`'s destructor/error paths use).

   Scope: the **primary** #1303 fix (background scan-only) does not depend on this teardown and is intact; the DE1 watchdog uses the transport's own controller correctly. Only the **foreground** direct-connect abort is defeated.

https://github.com/Kulitorum/Decenza/blob/92c063c7e7522d94f1bd800f1852e40b58f78ccd/src/ble/blemanager.cpp#L1134-L1141

(same no-op call in `onScaleConnectionTimeout` at `src/ble/blemanager.cpp#L1159`)

<details>

- [Worth a look] The Android BLE-dead recovery path calls `tryDirectConnectToScale()` at the default `allowDirectConnect=true`, issuing a parked direct-connect to a possibly-absent scale right after a `DeadObjectException` — the highest-risk moment. Likely should pass `false`. (score: 60) https://github.com/Kulitorum/Decenza/blob/92c063c7e7522d94f1bd800f1852e40b58f78ccd/src/main.cpp#L902-L907
- [OK to skip] The connect-watchdog lambda doesn't `m_connectWatchdogTimer.stop()` before `disconnectFromDevice()`; harmless because it's single-shot, but a synchronous re-entrant `stateChanged(Connecting)` could re-arm it. One-line defensive fix. (score: 50) https://github.com/Kulitorum/Decenza/blob/92c063c7e7522d94f1bd800f1852e40b58f78ccd/src/ble/bletransport.cpp#L160-L174

</details>

skialpine:
Addressed the review in 86f2f103:

- **Critical (teardown no-op):** the abort and timeout paths now call `m_scaleDevice->bleTransport()->disconnectFromDevice()` instead of the base `disconnectFromScale()`, so the parked transport controller is actually closed. The misleading comment is corrected.
- **BLE-recovery path** (`main.cpp`): scale reconnect now uses `allowDirectConnect=false` (scan-only).
- **4s abort** is now a cancellable `QTimer` member (`m_scaleDirectAbortTimer`), stopped on connect/reset/abort — no leaked stale shots.
- **Timeout ordering:** `m_directConnectInProgress` is cleared before teardown to prevent re-entrancy.
- **Watchdog lambda** stops its own timer first.

Note: the two CLAUDE.md header-placement findings were false positives — `abortScaleDirectConnectIfPending` and `kScaleDirectConnectAbortMs` are in the `private:` section, not `private slots:`.

Build clean (app + all tests, 0/0).

PR #1316, opened 2026-06-07
fix(refractometer): show "R1 Off" when paired model is a DiFluid R1

skialpine:
## Summary

- The disconnected-refractometer button on the post-shot review page was hardcoded to "R2 Off" from before R1 support landed in [#1308](https://github.com/Kulitorum/Decenza/pull/1308). Users with a DiFluid R1 paired saw the wrong model name whenever the device was asleep.
- Detect R1 vs R2 from the saved BLE advertised name using the same `"DFT_TDJ"` prefix check `DiFluidR1::isR1Device` uses in [src/ble/refractometers/difluidr1.cpp:128](https://github.com/Kulitorum/Decenza/blob/main/src/ble/refractometers/difluidr1.cpp#L128), and split the translation key into `r1off` / `r2off`.

Reported by a tester with an R1.

## Test plan

- [ ] With an R1 paired and disconnected, the post-shot review button reads "R1 Off"
- [ ] With an R2 paired and disconnected, the button still reads "R2 Off"
- [ ] When connected, the button still reads "Read TDS" / "..." while measuring

### DIFF
--- a/qml/pages/PostShotReviewPage.qml
+++ b/qml/pages/PostShotReviewPage.qml
@@ -852,11 +852,13 @@ Page {
                     }
                 }
 
-                // Read TDS button (R2 refractometer)
+                // Read TDS button (DiFluid R1 / R2 refractometer)
                 Rectangle {
                     id: readTdsButton
                     property bool refConnected: BLEManager.refractometerConnected
                     property bool refMeasuring: refConnected && typeof Refractometer !== "undefined" && Refractometer && Refractometer.measuring
+                    // R1 advertises with names starting "DFT_TDJ_*" (see DiFluidR1::isR1Device).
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Verified:
- `Settings.savedRefractometerName` has `NOTIFY savedRefractometerChanged` so the `isR1` binding re-evaluates reactively.
- `(name || "").toLowerCase().indexOf("dft_tdj") === 0` is semantically equivalent to the C++ `DiFluidR1::isR1Device` (`startsWith("DFT_TDJ", Qt::CaseInsensitive)`).
- The dropped `postshotreview.refractometer.off` translation key has no other consumers in the repo — clean rename.
- Empty/null `savedRefractometerName` falls through to "R2 Off" (same as the pre-PR default — no regression for unpaired or legacy state).

PR #1334, opened 2026-06-12
fix: stabilize Visualizer shot + bag sync (server read-back, descriptive fields, transient retries)

skialpine:
Makes both halves of the Visualizer upload — the **shot** record and the **coffee-bag** record — reliable. Grounded in the live visualizer.coffee API and its open source (verified, not inferred).

## The bug you'd see

A coffee bag with full origin metadata (country, region, producer, tasting notes) uploads a shot, but the **bag on visualizer.coffee shows none of it** — only roaster/name/roast date/roast level.

## Root cause (from the Visualizer source)

The upload parser auto-creates and links the bag on **every** upload (`parsers/base.rb` `set_coffee_bag`), keyed on roaster+name+roast_date, but sets only name/roaster/roast_date/roast_level — **never** the descriptive fields. Decenza's old client chain raced that auto-create with its own roast_date-matching find-or-create, and its find-and-link branch never pushed the descriptive fields. It also couldn't self-heal: a bag deleted on visualizer.coffee left a stale local `visualizer_bag_id` that the link path kept reusing forever.

### Bag sync — rebuilt around an authoritative read-back
- **`reconcileShotBag`**: after upload, `GET /api/shots/:id` to read the `coffee_bag_id` the **server** assigned. Empty → Coffee Management is off (cache the negative, clear any stale id). Present → capture the authoritative bag/roaster ids (self-healing a stale id for free) and enrich.
- **`enrichRemoteBag`**: `GET` the bag and `PATCH` only the descriptive fields the server left **blank** (fill-blanks — never clobber a value you set on Visualizer; never touch server-managed name/roast_date/roast_level). 404 → bag deleted mid-flight, clear local id.
- **`enrichRemoteRoaster`**: best-effort verified-roaster badge (link canonical when blank).
- Removes the now-dead `probeCmState` / `resolveRoaster` / `findRemoteBag` / `createRemoteBag` / `linkShotToBag` — the server already does the create+link, so the client only enriches. Also avoids ever PATCHing a shot with an unverified bag id (`shots.coffee_bag_id` has a FK → a dead id would 500).

### Shot side — resilience to transient failures
- **Shot save**: `saveShotStatic` now retries the transaction (bounded, short backoff) on a transient SQLITE_BUSY/locked, instead of dropping the shot. (A real save failure dropped a test shot entirely — no row, no upload.)
- **Upload POST**: bounded auto-retry of the same payload on transport/5xx blips only (401/422/429 stay permanent).

## Verified live against the API
- `GET /api/shots/:id` returns `coffee_bag_id` even for bags the client never stored → read-back works.
- `GET` deleted bag → clean `404` → self-heal signal.
- Descriptive-field `PATCH` → `200` (ParamsWrapper handles wrapped or top-level).
- `dependent: :nullify` confirms deleting a bag nulls the shot's link (bean tags remain).

## Follow-up (separate, upstream)
The server *could* fill descriptive fields itself for canonical/Bean-Base-sourced bags (the `CanonicalCoffeeBag` already has them) and currently even nulls the `canonical_coffee_bag_id` the client sends. Proposing that as a Visualizer issue/PR (link the auto-created bag to its canonical + surface its attributes) — benefits all clients. The client enrich here is still needed for manual/free-text bags the canonical never had.

## Testing
- Builds clean (0 errors, 0 warnings); all uploader/storage test targets relink.
- No unit test added for the CM network sync: it is entirely `QNetworkAccessManager` callback-driven with no existing mock harness, and the behavior was instead verified directly against the live API.

### DIFF
--- a/src/controllers/maincontroller.cpp
+++ b/src/controllers/maincontroller.cpp
@@ -1957,6 +1957,14 @@ void MainController::onShotEnded() {
     if (finalWeight <= 0 && m_profileManager->currentProfile().targetWeight() > 0)
         finalWeight = m_profileManager->currentProfile().targetWeight();
 
+    // Record at the scale's real resolution (~0.1 g). The raw cumulative-weight
+    // reading carries float noise (e.g. 35.36518272805495 g); rounding once here,
+    // at the source, keeps every consumer rational — the shot record, the DYE
+    // drink-weight metadata, the Visualizer upload, MCP, and exports — without
+    // sprinkling rounding across each of them.
+    if (finalWeight > 0)
+        finalWeight = std::round(finalWeight * 10.0) / 10.0;
+
     // Trim trailing zero-pressure samples from SAW settling period before saving.
     // During settling the DE1 reports 0 pressure/flow while the scale settles — these
     // cause a vertical drop to 0 at the end of the graph. Weight data is preserved.
--- a/src/history/shothistorystorage.cpp
+++ b/src/history/shothistorystorage.cpp
@@ -1411,6 +1411,20 @@ qint64 ShotHistoryStorage::saveShotStatic(const QString& dbPath, const ShotSaveD
 
     qint64 shotId = -1;
     withTempDb(dbPath, "shs_save", [&](QSqlDatabase& db) {
+        auto isLockError = [](const QSqlError& e) {
+            const QString code = e.nativeErrorCode();
+            return code == QLatin1String("5") || code == QLatin1String("6")
+                || e.text().contains(QLatin1String("locked"), Qt::CaseInsensitive)
+                || e.text().contains(QLatin1String("busy"), Qt::CaseInsensitive);
+        };
+        // A transient SQLITE_BUSY/locked from a concurrent writer (the
+        // reconciliation drain, daily backup, WAL checkpoint) must not drop a
+        // real shot. busy_timeout covers most contention, but a WAL write-write
+        // race can still surface as "database is locked" — retry the whole
+        // transaction a few times before giving up.
+        for (int attempt = 1; attempt <= 4; ++attempt) {
+        shotId = -1;
+        bool locked = false;
         // Use do-while(false) so error paths 'break' out while db/query are still
         // in scope.
         do {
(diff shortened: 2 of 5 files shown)


skialpine:
### Verified live (June 12, 1:05 PM shot, build `12:41:14`)

Both halves confirmed in the debug log on a real shot:

**Shot-save lock retry** — the same lock that previously dropped a shot was caught and recovered:
```
Failed to insert shot: "database is locked"
shot save hit a transient lock, retrying ( 1 of 4)
Saved shot 1039
```

**Bag read-back + enrich + self-heal** — after deleting the bag on visualizer.coffee, the server auto-created a fresh one (new id), and the read-back enriched it:
```
Upload successful, ID: 21cc9a01… for local shot 1039
Visualizer CM: state Unknown -> Active
Visualizer CM: enriched bag "07deaff8-…" with descriptive fields
```
The new bag id (vs the deleted one) shows the stale-id self-heal working as designed.

skialpine:
### Multi-agent review pass (code / errors / comments / tests)

Ran a comprehensive review. Findings and resolutions (commit `3fc09929`):

**Correctness (fixed):**
- **Unchecked `db.commit()`** in `saveShotStatic` — a WAL race at COMMIT could report a positive `shotId` for a transaction that never landed (silent shot loss). Now checked + retried.
- **`db.transaction()` (BEGIN) lock not classified** — the most likely lock point wasn't retried. Now classified. Both fixes came with restructuring the retry into an `attemptSave` lambda, which also resolves the flat loop-body indentation.
- **Ambiguous empty read-back cached as permanent `PremiumNoCm`** — an empty `coffee_bag_id` (e.g. server returns the shot before its bag link is visible) would suppress sync all session. No longer caches a negative off it; only the definitive 403 caches off.
- **JSON parse failures treated as data** — all three read-back GETs now check `QJsonParseError`. Critical for `enrichRemoteBag`, where an unparseable 200 read as "all blank" would have built a full-overwrite PATCH clobbering server values.

**Tests:** extracted the fill-blanks diff into a pure `buildBagEnrichBody()` and unit-tested it in `tst_coffeebags` (fill-blanks contract + blob→API field mapping). Full `tst_coffeebags` (30) and `tst_dbmigration` (28) green.

**Comments:** fixed a stale call-site comment and a member doc; comment-accuracy review otherwise clean.

The full CM network chain remains covered by production logging rather than a mock-network harness (disproportionate to build for this PR — confirmed against the repo's testing norms).

PR #1335, opened 2026-06-12
fix: eliminate the shot-save database lock (BEGIN IMMEDIATE)

skialpine:
## Symptom
Every shot save logged `ShotHistoryStorage: Failed to insert shot: "database is locked"` and recovered on retry 1 — a lock on **100% of saves**, masked only by the retry added in #1334.

## Root cause (confirmed with temporary instrumentation, then removed)
At shot end, `MainController::onShotEnded` dispatches the background `shs_save` INSERT and then, on the very next line, dispatches the dose/yield/`lastUsed` bag stamp via `CoffeeBagStorage::requestUpdateBag` → a second background `withTempDb` UPDATE (`bags_update`). Two background writers hit the same WAL `shots.db` at once.

The save used `db.transaction()` = a plain `BEGIN` (**DEFERRED**): it takes a read lock first, so the INSERT then has to *upgrade* to the write lock while `bags_update` holds it, and SQLite returns `SQLITE_BUSY` (code 5) **immediately** — deadlock avoidance skips the busy handler, so the connection's `busy_timeout=5000` never engages.

A diagnostic build confirmed it verbatim:
```
Failed to insert shot: "database is locked …" (sqlite code "5")
shot save hit a transient lock, retrying (1 of 4) - other DB connections open now: "bags_update_… (open 4ms)"
```

## Fix
Switch the save transaction to **`BEGIN IMMEDIATE`**, which acquires the write lock up front while holding no other lock — so the busy handler engages and `busy_timeout` rides out the brief (~4 ms) concurrent writer. The lock no longer surfaces; verified with a clean save (`Saving shot` → `Saved shot N`, no warning).

The retry loop + commit/transaction lock-classification from #1334 stay as a **quiet backstop** (covers the pathological case of a writer held longer than `busy_timeout`, so a shot is never dropped). The temporary per-`withTempDb` connection registry used to find the culprit was removed; the lightweight sqlite-native-code line in the failure log is kept.

## Testing
- Storage tests pass: `tst_coffeebags` 30, `tst_dbmigration` 28, `tst_shotrecord_cache` 7.
- Verified live: shot saves now log no lock/retry warning (previously every shot did).

### DIFF
--- a/src/core/dbutils.h
+++ b/src/core/dbutils.h
@@ -5,45 +5,7 @@
 #include <QSqlError>
 #include <QThread>
 #include <QString>
-#include <QStringList>
-#include <QHash>
-#include <QMutex>
-#include <QDateTime>
 #include <QDebug>
-
-// --- DB-lock diagnostics (temporary, investigate/shot-save-db-lock) ---------
-// Every background DB op goes through withTempDb, so a registry of currently
-// open connections lets us name whoever holds the write lock when another
-// connection (e.g. shs_save) gets SQLITE_BUSY. Logs nothing on its own — a
-// caller dumps dbDiagActiveConnections() only when it actually hits a lock.
-namespace dbdiag {
-inline QMutex& mutex() { static QMutex m; return m; }
-inline QHash<QString, qint64>& active() { static QHash<QString, qint64> a; return a; }
-inline void noteOpen(const QString& name) {
-    QMutexLocker lock(&mutex());
-    active().insert(name, QDateTime::currentMSecsSinceEpoch());
-}
-inline void noteClose(const QString& name) {
-    QMutexLocker lock(&mutex());
-    active().remove(name);
-}
-}  // namespace dbdiag
-
-// Snapshot of withTempDb connections open right now, excluding any whose name
-// starts with `excludePrefix` (the caller's own). Each entry shows how long it
-// has been open, so a long-lived holder stands out. Call this from a lock-error
-// path to identify the concurrent writer.
-inline QString dbDiagActiveConnections(const QString& excludePrefix = QString()) {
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance. The change is a faithful mechanical substitution of `db.transaction()`/`commit()`/`rollback()` for raw `BEGIN IMMEDIATE`/`COMMIT`/`ROLLBACK`; the retry/lock-classification invariants from #1334 are preserved, and the connection is short-lived and isolated so the raw-SQL transaction is safe.

---

<details>

- [Worth a look] The pre-existing outer retry-loop comment ("busy_timeout covers most contention, but a WAL write-write race can still surface") now contradicts the new `BEGIN IMMEDIATE` rationale right below it — the fix is specifically what lets `busy_timeout` absorb that race. One-line wording tweak in code already being modified (score: 73).

  https://github.com/Kulitorum/Decenza/blob/487cd3469a29d1c0da1f6e4087068aa486aedf72/src/history/shothistorystorage.cpp#L1419-L1426

- [OK to skip] On a busy `COMMIT`, the code `ROLLBACK`s and the loop re-does the full insert (row + sample blob + phase markers) instead of retrying just the `COMMIT` (SQLite keeps the txn alive). Pre-existing from #1334, and after a successful `BEGIN IMMEDIATE` in WAL a COMMIT-time `SQLITE_BUSY` is near-impossible, so the fix is disproportionate (score: 50).

  https://github.com/Kulitorum/Decenza/blob/487cd3469a29d1c0da1f6e4087068aa486aedf72/src/history/shothistorystorage.cpp#L1554-L1562

</details>

Issue #1309, opened 2026-06-07
[Bug] BLE Connectivity issue with DE1


Churator:
### What happened?

Similiar to 1303 I belive, My machine is not connected to Decenza, I can only recover the connection after reboot to the tablet.
I've updated earlier and still won't connect (1.7.6 3433)
I use Bookoo scale if that matters

### Steps to reproduce

1. Not sure :) i just start the tablet from clicking on the screen, and the machine won't start (Show disconnected

### Decenza version

1.7.6 3433

### Platform

Android

### Device / OS version

P80X with LingenOS

### Scale (if applicable)

Bookoo Ulta

### Debug log / shot debug log

[debug.zip](https://github.com/user-attachments/files/28680122/debug.zip)

skialpine:
Can you flash the nighly de1 firmware on the about page and see if it solves your problem? most users are already running on that firmware. If still having the issue please add another log here so I can look at if there are any differences.

skialpine:
also load the latest beta. I added a little debugging if the firmware does not solve your issue.

Churator:
Done both just now
Let me check and report back

Thank you

mcastaldelli:
I noticed the problem is back for me too, in a different way. It may be related to the P80X bluetooth stack, as It simply freezes after some time measured in days (probably caused by Decenza). It happened twice for me when I left my machine unused for a day or two due to travel. Turning bluetooth off and on (i.e., tablet bluetooth) restores connectivity, even without exiting the Decenza app.

I'm attaching a log from my Decenza instance here as well. It should contain the times when the connection got dropped. 

[debug-2.zip](https://github.com/user-attachments/files/28699764/debug-2.zip)

skialpine:
@Churator and @mcastaldelli is it better for you the last couple of days?

Churator:
After reboot it usually stay connected for few days, so I'm not sure if its fixed or not, but so far so good.

mcastaldelli:
I just installed the most recent beta build. I'll report if it fixed the problem. I was having the same issue as Churator above. It lost connectivity after a few days, both bluetooth to the machine, and IP connectivity to the internet (MQTT stopped working too). Restarting bluetooth fixed both connections.

mcastaldelli:
Hi @skialpine. I can confirm the problem is still happening in build 3443. I checked this morning and there was no connection to the machine.

 When I turned Bluetooth off and on without even leaving Decenza, it came back right away after a scan.

[...]

PR #1337, opened 2026-06-14
Auto-recover wedged Android BLE stack by power-cycling the adapter (#1309)

skialpine:
## Problem

Issue [#1309](https://github.com/Kulitorum/Decenza/issues/1309): on Teclast P80X tablets the app loses BLE connection to the DE1 (and scale) after some uptime, and the **only** thing that recovers it is toggling the tablet's Bluetooth off/on — both reporters do this by hand.

Analysis of the latest debug log (build 3443, `debug.zip`):
- The Android BLE stack **wedges**: every connect fails for *both* the DE1 and the scale. The DE1 was dead for ~22h while the scale retried **1,381 times** in vain.
- It **survives an app restart** — session 3 in the log only recovers after a `HostPoweredOff → HostConnectable` adapter cycle. So the fault lives in the OS BLE stack, below the app.
- The DE1 reconnect ladder also **gave up permanently** after 12 attempts and a screensaver/screen-tap wake couldn't restart it.

Both reporters run **API ≤ 32** (mcastaldelli: Android 9 / API 28; Churator: Android 12L / API 32 GSI), where a non-privileged app can still cycle the adapter silently. `BLUETOOTH_ADMIN` + `BLUETOOTH_CONNECT` are already in the manifest.

## Fix

**Detector** (`BLEManager`) — treat the stack as wedged only when, sustained for ≥45s:
- a saved DE1 won't connect **and** is throwing controller-error faults, **and**
- the physical scale also can't connect.

`de1LinkFault` fires *only* on real controller errors (`ConnectionError`/`AuthorizationError`/`RemoteHostClosed`/`write-failed`); an absent or sleeping device produces a watchdog timeout with **no** fault. That three-way fingerprint keeps a powered-off machine from ever triggering a needless cycle.

**Recovery** (Android only) — `powerOff()` → on observed `HostPoweredOff` `powerOn()` → on `HostConnectable` re-arm DE1 + scale reconnect. Driven off real adapter transitions, with:
- a 5-min per-cycle backoff,
- a 10s fail-safe that re-enables the radio if the power-off event is missed (never leave BT off),
- respect for a user-disabled adapter (skip if `HostPoweredOff`).

On non-Android platforms the recovery action is a no-op (the existing "Try toggling Bluetooth off/on" guidance stands); iOS can't observe/toggle the adapter at all.

**Also**: the DE1 reconnect ladder no longer gives up permanently after 12 attempts — it falls to a slow 5-min background retry indefinitely. This reconnects a DE1 that was simply off for >10 min *and* keeps the detector fed so recovery stays viable. `bleStackRecovered()` resets the budget and retries immediately once the adapter is back.

## Testing

- Builds clean (Qt Creator, 0 errors / 0 warnings).
- Logic-reviewed against the build-3443 log timeline. Needs on-device confirmation on a P80X — once merged into a beta, ask @mcastaldelli / @Churator to confirm the connection self-heals without manually toggling Bluetooth.

### DIFF
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -100,6 +100,28 @@ BLEManager::BLEManager(QObject* parent)
         }
     });
 
+    // Fail-safe for the BLE-stack-wedge adapter power-cycle (#1309): if
+    // powerOff() never reports HostPoweredOff (so the host-mode handler never
+    // calls powerOn()), force the adapter back on here so a wedged radio is
+    // never left switched off. Normally stopped by the HostConnectable event
+    // before it fires. Single-shot — not a logic guard, a last-resort re-enable.
+    m_adapterRecoverySafetyTimer = new QTimer(this);
+    m_adapterRecoverySafetyTimer->setSingleShot(true);
+    m_adapterRecoverySafetyTimer->setInterval(kAdapterRecoverySafetyMs);
+    connect(m_adapterRecoverySafetyTimer, &QTimer::timeout, this, [this]() {
+#ifndef Q_OS_IOS
+        if (!m_adapterRecoveryInFlight || !m_localDevice) return;
+        qWarning() << "BLEManager: adapter did not report powered-off within"
+                   << (kAdapterRecoverySafetyMs / 1000) << "s — forcing power-on (#1309)";
+        m_localDevice->powerOn();
+        // Clear in-flight so a future wedge can still trigger recovery even if
+        // the HostConnectable event for this cycle is also missed.
+        m_adapterRecoveryInFlight = false;
+        m_wedgeSince = QDateTime();
+        emit bleStackRecovered();
+#endif
+    });
+
     // Eagerly run the Linux capability check so any qWarning lands early in
     // startup logs; subsequent calls hit the cached result.
     (void) BleCapability::linuxMissing();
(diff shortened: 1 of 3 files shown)


skialpine:
## Detector verification against the real #1309 logs

Traced the detector's trigger conditions against the actual debug logs (rather than just reasoning about them).

**The existing "Try toggling Bluetooth off/on" hint never reached these users.** `grep` across all three attached logs (mcastaldelli build 3443, mcastaldelli earlier, Churator) finds **0** occurrences of that message and **0** of the "DE1 service not found" path that emits it. That hint is gated behind a *service-discovery* failure (`bletransport.cpp:619`), but this wedge fails earlier — at the **connect** step, with controller errors — so it never reaches service discovery. Real gap; that's what the new detector covers.

**The new detector fires correctly and early.** mcastaldelli build-3443 timeline with the three conditions overlaid:

```
11311  DE1 AuthorizationError   → fault fresh; DE1 down; scale down  → wedge window STARTS
11350  Scale timeout            → 39s elapsed (< 45s)                → not yet
11381  DE1 ConnectionError      → 70s elapsed (≥ 45s)                → ADAPTER CYCLE FIRES
...    DE1 faults every ~60s, scale timeouts every ~60s, "Scale connected" = 0 in window
12071  old code GIVES UP        ← detector would have recovered ~11 min earlier
92700  user finally wakes it    ← ~22h after onset
```

Confirmed in the active window (t=11311–12071):
- **14 DE1 controller faults**, ~every 60s, all `ConnectionError`/`AuthorizationError` (which emit `de1LinkFault`).
- **Scale connected: 0 times** — scale down the whole window.
- All three conditions hold continuously; the 45s sustain threshold is crossed at **t≈11381, ~70s after onset**.

Fault counts on the other logs: Churator **652** DE1 `ConnectionError`, early mcastaldelli **1,453** — abundant signal on both reporters (both API ≤ 32, so the silent adapter cycle works).

**Caveat the trace exposed:** during the ~22h *dead* window after the old give-up, there are **zero** DE1 faults (the old code stopped trying), so the detector alone would go stale there. It instead fires in the **active** window (early ~12 min), which is the ideal time — we recover before reaching the dead state. The slow-retry change keeps faults flowing if we ever do land in a long-dead state, so the two halves are deliberately interdependent.

skialpine:
## Detector verification against the real #1309 logs

Traced the detector's trigger conditions against the actual debug logs rather than just reasoning about them.

**The existing "Try toggling Bluetooth off/on" hint never reached these users.** `grep` across all three attached logs (mcastaldelli build 3443, mcastaldelli earlier, Churator): **0** occurrences of that message and **0** of the "DE1 service not found" path that emits it. Two reasons, both now addressed:
1. That hint is emitted only after a *successful connect + service discovery that finds no service* (`bletransport.cpp:619`). This wedge fails one step earlier, at **connect** (`ConnectionError`/`AuthorizationError`), so the branch is never reached.
2. Even if it fired, `DE1Device::errorOccurred` had **no consumer** — only the scale's errors were forwarded to the UI dialog. Fixed in this PR (debounced forward to `BLEManager::errorOccurred`).

**The new detector fires correctly and early.** mcastaldelli build-3443 timeline with the three conditions overlaid:

```
11311  DE1 AuthorizationError   → fault fresh; DE1 down; scale down  → wedge window STARTS
11350  Scale timeout            → 39s elapsed (< 45s)                → not yet
11381  DE1 ConnectionError      → 70s elapsed (>= 45s)               → ADAPTER CYCLE FIRES
...    DE1 faults every ~60s, scale timeouts every ~60s, "Scale connected" = 0 in window
12071  old code GIVES UP        ← detector would have recovered ~11 min earlier
92700  user finally wakes it    ← ~22h after onset
```

Confirmed in the active window (t=11311–12071):
- **14 DE1 controller faults**, ~every 60s, all `ConnectionError`/`AuthorizationError` (which emit `de1LinkFault`).
- **Scale connected: 0 times** — scale down the whole window.
- All three conditions hold continuously; the 45s sustain threshold is crossed at **t≈11381, ~70s after onset**.

Fault counts on the other logs: Churator **652** DE1 `ConnectionError`, early mcastaldelli **1,453** — abundant signal on both reporters (both API ≤ 32, so the silent adapter cycle works).

**Caveat the trace exposed:** during the ~22h *dead* window after the old give-up, there are **0** DE1 faults (the old code stopped trying), so the detector alone would go stale there. It instead fires in the **active** window (early ~12 min), which is the ideal time — we recover before reaching the dead state. The slow-retry change keeps faults flowing if we ever do land in a long-dead state, so the two halves are deliberately interdependent.

skialpine:
## PR review summary (multi-agent: code / silent-failure / comments)

**code-reviewer** — no high-confidence blockers. Verified `#ifdef` correctness (Android/iOS, `m_localDevice` only `!Q_OS_IOS`), the wedge discriminator, signal/slot wiring, and that the slow-retry counter clamp prevents unbounded growth. Judged the safety timer a *justified* fail-safe, not a "timers as guards" violation.

**comment-analyzer** — zero issues. Cross-checked every constant literal vs its comment (45s / 7min / 5min / 10s), the discriminator claims (`de1LinkFault` only on controller errors; `m_scaleDevice` always physical), and platform gating. All accurate.

**silent-failure-hunter** — found one genuine merge-blocker, now fixed:

> The power-cycle fail-safe could leave Bluetooth **off** and permanently block future recovery, while emitting `bleStackRecovered()` (a lie) and telling the user nothing — reachable if `powerOff()` takes effect but `powerOn()` doesn't bring the radio back (or Android 13+ consent declined).

Fixed in `e132dfa0`:
- The watchdog now **verifies the actual adapter host mode** instead of assuming success — still off → one more `powerOn()` + finalise as failure; on → finalise as success.
- The watchdog is **re-armed for the power-ON leg** too (previously only the power-OFF leg was covered).
- On failure we surface an **actionable `errorOccurred`** ("Decenza tried to restart Bluetooth but it's still off — toggle it in settings") and do **not** claim recovery.
- `m_recoveryLeftAdapterOff` distinguishes *our* failed cycle from a *user-disabled* adapter, so a stranded radio is retried (with backoff) rather than ignored forever.

Two sub-threshold notes accepted as deliberate: (1) DE1 errors now surface a dialog on a routine DE1 power-off (debounced once per disconnect) — intended, since DE1 errors were previously invisible; (2) `bleStackRecoveryStarted()` has no QML consumer yet — a success-path "restarting Bluetooth…" toast is a reasonable optional follow-up.

Rebuilt clean (0 errors / 0 warnings) after each change.

PR #1341, opened 2026-06-19
Equipment packages: first-class switchable grinder

skialpine:
## Equipment packages

Extracts the grinder out of the bean bag into a first-class, switchable **Equipment package** (parallel to Beans), per `openspec/changes/add-equipment-packages`. Grinder identity is now a pointer + JOIN, not per-shot/bag snapshot columns.

### What ships
- **Equipment window** parallel to Beans — package cards (add/edit/remove), empty state, and an idle **Equipment button** added to the default layout and injected into every existing user's saved layout via an idempotent migration.
- **Brew Settings**: grinder make/model/burrs replaced by a read-only **Equipment package name** + **Switch** + **info** button; **grind setting + RPM** kept as dial-in (RPM shown only when the grinder is adjustable, including custom grinders).
- **Equipment surfaced everywhere the grinder used to be**: Shot Review (read-only + Change Equipment), Edit Bag (summary + Switch/Add), Shot Detail (read-only, incl. RPM). The package **name** is shown instead of the raw grinder identity, with a per-row info dialog listing the package contents; the separate burrs line was dropped.
- **Copy-on-write immutability**: editing a used package's identity forks a new package, retires the old (`superseded_by` lineage), repoints referencing bags/shots; identity merges on collision; `last_grind`/`last_rpm` stay mutable so dialing never churns packages.
- **Dial memory**: grinder-scoped (package) + bean-scoped (bag); switching either applies its last dial.
- **Pure-pointer storage**: `coffee_bags.equipment_id` and `shots.equipment_id` reference a package; grinder identity resolves via JOIN to `equipment_items`. Migration 22 (additive: tables, `equipment_id`+`rpm`, package creation from history, `"24 1400rpm"` split); migration 23 (drop legacy grinder columns from shots + bags, rebuild `shots_fts` without them).
- **MCP**: `equipment_list` / `equipment_select` / `equipment_update`; grinder identity + RPM resolved via the pointer; RPM now surfaced to `dialing_get_context` and the in-app AI advisor (`currentBean.rpm`).
- **Visualizer**: appends rpm to the grind setting on upload.

### Tests
Full suite green — **2397 tests, 0 warnings**. Covers `tst_equipment` (CRUD / copy-on-write + merge / setting+rpm split / rpmCapable), migration 22/23, bag/shot `equipment_id` + rpm, and `currentBean.rpm` equivalence across the MCP and in-app surfaces.

### Review
Reviewed via multi-agent `/review`; all findings addressed — new-bag equipment persistence, removal of stale grinder params from `shots_update`, replace-import clear ordering, accessible-name i18n, and a stale comment.

### Follow-ups (not blocking)
- Wiki Manual page for the Equipment feature (release-time).
- Advisor input-path consolidation (the unused `QVariantMap analyzeShot` / 14-arg `buildMetadata` fan-out) — a separate cleanup PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -416,6 +416,7 @@ set(HEADERS
     src/core/settings_app.h
     src/core/settings_calibration.h
     src/core/grinderaliases.h
+    src/core/basketaliases.h
     src/core/widgetlibrary.h
     src/core/batterymanager.h
     src/core/memorymonitor.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -761,6 +761,7 @@ set(QML_FILES
     qml/components/BagCard.qml
     qml/components/SwitchEquipmentDialog.qml
     qml/components/EquipmentCard.qml
+    qml/components/EquipmentInfoDialog.qml
     qml/components/KeyboardAwareContainer.qml
     qml/components/AIConversationPanel.qml
     qml/components/ConversationOverlay.qml
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -690,6 +690,7 @@ set(QML_FILES
     qml/pages/ProfileImportPage.qml
     qml/pages/VisualizerMultiImportPage.qml
     qml/pages/BeanInfoPage.qml
+    qml/pages/EquipmentPage.qml
     qml/pages/PostShotReviewPage.qml
     qml/pages/AISettingsPage.qml
     qml/pages/DialingAssistantPage.qml
@@ -758,6 +759,8 @@ set(QML_FILES
     qml/components/BeanSummary.qml
     qml/components/ChangeBeansDialog.qml
     qml/components/BagCard.qml
+    qml/components/SwitchEquipmentDialog.qml
+    qml/components/EquipmentCard.qml
     qml/components/KeyboardAwareContainer.qml
     qml/components/AIConversationPanel.qml
     qml/components/ConversationOverlay.qml
@@ -810,6 +813,7 @@ set(QML_FILES
     qml/components/layout/items/HotWaterItem.qml
     qml/components/layout/items/FlushItem.qml
     qml/components/layout/items/BeansItem.qml
+    qml/components/layout/items/EquipmentItem.qml
     qml/components/layout/items/HistoryItem.qml
     qml/components/layout/items/DiscussItem.qml
     qml/components/layout/items/AutoFavoritesItem.qml
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,6 +307,7 @@ set(SOURCES
     src/history/shothistorystorage_serialize.cpp
     src/history/shothistorystorage_queries.cpp
     src/history/coffeebagstorage.cpp
+    src/history/equipmentstorage.cpp
     src/history/unifiedbeansearchmodel.cpp
     src/history/shotprojection.cpp
     src/history/shotdebuglogger.cpp
@@ -489,6 +490,7 @@ set(HEADERS
     src/history/shothistorystorage.h
     src/history/shothistorystorage_internal.h
     src/history/coffeebagstorage.h
+    src/history/equipmentstorage.h
     src/history/unifiedbeansearchmodel.h
     src/history/shotprojection.h
     src/history/shotdebuglogger.h
@@ -1180,6 +1182,7 @@ if(NOT ANDROID AND NOT IOS)
         src/core/settings_brew.cpp
         src/core/settings_dye.cpp
         src/history/coffeebagstorage.cpp
+        src/history/equipmentstorage.cpp
         src/core/settings_network.cpp
         src/core/settings_app.cpp
         src/core/settings_calibration.cpp
--- a/docs/CLAUDE_MD/MCP_SERVER.md
+++ b/docs/CLAUDE_MD/MCP_SERVER.md
@@ -86,9 +86,9 @@ Each tool has a `category` that determines the minimum access level required:
 
 | Category | Min Access Level | Tools |
 |----------|-----------------|-------|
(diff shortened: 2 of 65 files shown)


skialpine:
### Code review

Found 1 issue:

1. `.shot` file import (`importShotRecord`) no longer captures grinder identity — a regression. Migration 23 dropped `grinder_brand/model/burrs`, and this INSERT was updated to drop them too, but unlike the live shot-save path it never sets `equipment_id` (and creates no equipment package for the parsed grinder). Shots imported from `.shot` archives via `ShotImporter` now resolve to a blank grinder, losing identity that `main` preserved (parsed through `GrinderAliases` in `shotfileparser.cpp`). Consider find-or-creating an `equipment_packages` row for the parsed identity and setting `equipment_id`, mirroring the `insertShot` test helper.

https://github.com/Kulitorum/Decenza/blob/192d20a2eed89d34fc9dd62aa26649effb415fd0/src/history/shothistorystorage.cpp#L3323-L3345

<details>

- [Worth a look] `int` loop counters over `.size()` in new code (`equipmentstorage.cpp:455`, `:555`, `settings_network.cpp:374`) — CLAUDE.md says use `qsizetype` for container sizes. No `-Wshorten-64-to-32` actually fires here (comparison, not assignment; build is clean), so it's a trivial consistency fix only. (score: 60)
- [Worth a look] `createTables()` still runs `CREATE INDEX IF NOT EXISTS idx_shots_grinder ON shots(grinder_brand, grinder_model)`, which errors ("no such column") on every launch of a post-migration-23 DB and is silently swallowed. Harmless but relies on silent failure; the line can be dropped. (score: 50)
- [Worth a look] Stale comment in the migration-20 `DECENZA_TESTING` abort block: "(incl. migration 21)" predates migrations 22/23 — the deferred chain now runs through 23. (score: 50)
- [OK to skip] `requestDeletePackage`'s reference check ignores other packages' `superseded_by`, so hard-deleting a fork target could leave a dangling lineage pointer. Rare in practice (fork targets normally accrue bag/shot references). (score: 50)

</details>

Note: an earlier automated pass flagged `getDistinctGrinderBurrsForModel` referencing the `AS burrs` alias in its `WHERE` as a critical bug — this is a false positive. SQLite permits result-column aliases in `WHERE` (verified empirically); the burrs filter works.

skialpine:
Addressed in 96db0ad6:

- **`.shot` import grinder identity (the reported issue)** — `importShotRecord` now find-or-creates an equipment package for the parsed identity and links `equipment_id` (+ `rpm`); added regression test `v23_importShotRecordLinksEquipment`.
- qsizetype loop counters → `qsizetype` (equipmentstorage, settings_network).
- Dropped the stale `CREATE INDEX idx_shots_grinder` from `createTables` (it errored every post-migration-23 launch).
- `requestDeletePackage` now also blocks hard-delete when another package's `superseded_by` points at the target.
- `activeEquipmentId` widened `int` → `qint64` (consistent with the ids used everywhere else).
- Fixed the stale migration-20 test-path comment.

Full suite green (2395 passed, 0 failed).

skialpine:
### Code review

Found 5 issues:

1. **Adding a new bag silently drops the chosen equipment package.** The equipment Switch/Add picker row is shown in both create and edit modes, but `confirmForm()` only adds `equipmentId` to the saved fields in the `formMode === "edit"` branch — the create branch omits it, so a package picked while adding a bag is lost (bag saved with `equipment_id` NULL).

https://github.com/Kulitorum/Decenza/blob/eb8a0578a06d6e4c7404086ed6130dae84f7d5da/qml/components/ChangeBeansDialog.qml#L353-L371

2. **`shots_update` MCP schema advertises grinder params the handler discards.** `grinderBrand`/`grinderModel`/`grinderBurrs` are declared in the tool schema and copied into `metadata`, but `updateShotMetadataStatic`'s field map intentionally omits them (grinder identity now resolves through `equipment_id`). A caller passing only those keys gets an opaque "Failed to update shot N"; mixed with valid keys they're silently dropped. Remove the params or document that grinder identity is set via the equipment package.

https://github.com/Kulitorum/Decenza/blob/eb8a0578a06d6e4c7404086ed6130dae84f7d5da/src/mcp/mcptools_write.cpp#L67-L71

3. **Replace-mode import from a pre-equipment backup leaves orphaned packages.** The "source has no equipment tables" probe returns early (line 922) *before* the replace-mode clear of the destination's `equipment_packages`/`equipment_items` (lines 925-934). Importing (replace) an older backup wipes dest shots/bags but leaves the dest's equipment inventory in place as orphans. Edge case, but inconsistent with replace semantics — move the clear ahead of the early return.

https://github.com/Kulitorum/Decenza/blob/eb8a0578a06d6e4c7404086ed6130dae84f7d5da/src/history/equipmentstorage.cpp#L918-L934

4. **Hardcoded English in an `Accessible.name`** (CLAUDE.md says "All user-visible text must be internationalized" — accessible names are read aloud by TalkBack/VoiceOver). The RPM value's accessible name concatenates a literal `"RPM: "` while the adjacent visible label uses `Tr { key: "shotdetail.rpm" }`.

https://github.com/Kulitorum/Decenza/blob/eb8a0578a06d6e4c7404086ed6130dae84f7d5da/qml/pages/ShotDetailPage.qml#L868-L870

5. **Stale comment contradicts the code.** The header says the grinder cache fields are a "transient display cache … storage never populates them," but `loadBagStatic` now populates all three via a JOIN on `equipment_items`. Update to "populated at load from a JOIN; not persisted as columns."

https://github.com/Kulitorum/Decenza/blob/eb8a0578a06d6e4c7404086ed6130dae84f7d5da/src/history/coffeebagstorage.h#L42-L45

Verified clean: migration 22/23 guards + transactional DDL, FTS rebuild/triggers, `loadShotRecordStatic` column indices (0-43), `equipment_id <= 0 -> NULL` mapping, copy-on-write/`superseded_by` lineage, the `_pendingEquipmentId`-keyed `onPackageReady` autosave guard, and CLAUDE.md basics (QML registration, settings sub-object, withTempDb threading, qsizetype, accessible components, no Unicode-glyph icons). Prior review feedback from the earlier pass is all addressed.

PR #1348, opened 2026-06-20
Weight-based dialing: bean/milk auto-capture, weight-timed steaming, capture ding + steam-page polish

loganross2001:
Hi @Kulitorum — first, thank you for Decenza; this builds entirely on your work. 🙏

**This is an offer, not a request.** Please adopt, adapt, or ignore any of it freely — I'm just sharing it transparently in case some pieces are useful upstream. No pressure at all.

## What this adds

A set of "dial-by-weight" features, built on the **`v1.8.0`** tag:

- **`StableWeightCapture`** — a small, reusable, *event-driven* scale-settle detector (no guard timers per `CLAUDE.md`).
- **Bean auto-capture + dose-cup tare** — rest the dose cup on the scale and the net dose is set automatically, with a pleasant "ding". No button press.
- **Weight-timed steaming** — each steam-pitcher preset gets a *reference milk weight*; steam time then scales proportionally to the measured milk (`duration × measured ÷ reference`) and the DE1 auto-stops there. Works from both the home-screen steam flow and the steam page.
- **Capture "ding"** — synthesized bell, pre-loaded at startup, independent of the accessibility toggle.
- **Steam-page polish** — scrollable editor, Weigh buttons, 0.1 g inputs, and a **Purge** button (editor + live steaming view).
- **Home-screen rework** — three setup shortcuts + a high-contrast bottom stat bar. ⚠️ This part is opinionated and **repurposes the `shotPlan` widget**, so it's the least "drop-in" — easy to leave out.

## Notes for review

- **Based on `v1.8.0`**, so it will likely need a rebase onto current `main` (a couple of files — `IdlePage.qml`, `SteamPage.qml` — have since diverged, so GitHub may flag conflicts). Happy to rebase if any of it interests you.
- **Cherry-pick friendly:** features 1–6 are self-contained and do **not** depend on the home-screen rework. The shared substrate is `StableWeightCapture` + two `SettingsBrew` additions (`doseCupTareWeight`, `setSteamPitcherCalibration`).
- Full rationale, design decisions, and gotchas (including a steam-timeout re-sync bug that silently wiped the scaled value) are written up in [`fork-notes/DETAILS.md`](https://github.com/loganross2001/Decenza/blob/weight-dialing/fork-notes/DETAILS.md) and [`fork-notes/SUMMARY.md`](https://github.com/loganross2001/Decenza/blob/weight-dialing/fork-notes/SUMMARY.md).
[...]

PR #1349, opened 2026-06-20
[1/5] Foundation: StableWeightCapture component and capture ding

loganross2001:
Shared building blocks the other PRs in this stack depend on. **No behavior change on its own — nothing calls these yet.**

- **`qml/components/StableWeightCapture.qml`** (new): an event-driven scale-settle detector. Emits `stableCaptured(grams)` once a bound net weight holds within `tolerance` for `stableMs` (~2.5 s), then latches and re-arms when the load is removed/changed. A 150 ms poll `Timer` runs only while armed (periodic polling, not a guard timer per `CLAUDE.md`).
- **`AccessibilityManager::playCaptureDing()`**: plays `qrc:/sounds/ding.wav` via a dedicated `QSoundEffect`, **pre-loaded at startup** (so the first ding isn't silent) and intentionally **ungated** by the accessibility flag (general UI cue).
- **`resources/sounds/ding.wav`** (new) + registrations in `resources.qrc` and the `qt_add_qml_module` list in `CMakeLists.txt`.

**Files:** `StableWeightCapture.qml`, `accessibilitymanager.{h,cpp}`, `resources.qrc`, `ding.wav`, `CMakeLists.txt`.

---
**📦 Part 1 of 5 — a dependency-ordered stack** splitting #1348 into per-feature PRs (as you asked). Review/merge in order:
1. #1349 — Foundation: StableWeightCapture + capture ding
2. #1350 — Bean auto-capture + dose-cup tare
3. #1351 — Weight-timed steaming
4. #1352 — Steam-page UX polish
5. #1353 — Home-screen rework

Each builds on the previous; its diff narrows to just this feature once the earlier ones merge. Full per-feature technical reference: https://github.com/loganross2001/Decenza-community-fork/blob/community-fork/fork-notes/CHANGES_DETAILED.md


src/core/accessibilitymanager.cpp
```diff
@@ -240,6 +244,25 @@ void AccessibilityManager::initTickSound()
...
+    m_dingSound->setSource(QUrl("qrc:/sounds/ding.wav"));
+    m_dingSound->setVolume(0.9);
+}
+
+void AccessibilityManager::playCaptureDing()
+{
+    if (m_shuttingDown) return;
+    initDingSound();  // no-op after the first call / startup pre-load
+    if (m_dingSound)
+        m_dingSound->play();
```
skialpine, line 263: **Consistency:** `playTick()` gates playback on `status() == QSoundEffect::Ready`, but this calls `play()` after only a null-check. Playing during the async load window (or on a failed load) is a silent no-op. The startup pre-load makes Ready likely by first capture, but doesn't guarantee it. Suggest matching the tick path:

```cpp
if (m_dingSound && m_dingSound->status() == QSoundEffect::Ready)
    m_dingSound->play();
```

src/core/accessibilitymanager.cpp
```diff
@@ -240,6 +244,25 @@ void AccessibilityManager::initTickSound()
     }
 }
 
+void AccessibilityManager::initDingSound()
+{
+    if (m_dingSound)
+        return;  // already loaded
+    // Weight-capture confirmation ding — full volume, independent of the tick /
```
skialpine, line 251: Comment says **"full volume"** but the code sets `setVolume(0.9)` (line 255). Minor, but they disagree — suggest "near-full volume (0.9)" or just drop the adjective.

qml/components/StableWeightCapture.qml
```diff
@@ -0,0 +1,84 @@
...
+//
+// The parent binds `weight` to a net value (e.g. scaleWeight - tare). When that
+// value holds within `tolerance` of a candidate for `stableMs`, this emits
+// stableCaptured(grams) exactly once and latches. It re-arms automatically when
+// the load is removed (drops below `removeThreshold`) or changes materially from
+// the captured value (so topping up milk / re-dosing beans re-captures).
+//
+// Detection is driven by `weight` changes; a slow poll Timer additionally
+// re-checks elapsed stability time so a scale that streams a perfectly constant
+// value still captures. The poll is genuinely periodic, not a guard timer.
```
skialpine, line 13: The mechanism here is legitimate (a perfectly constant stream emits no `onWeightChanged`, so the poll re-drives `_evaluate()`), but **"genuinely periodic, not a guard timer" overstates it** — `running` is gated `active && !_captured`, so it runs only during the arming window, not as an unconditional heartbeat. Suggest just describing it and letting the mechanism justify itself, e.g. *"A 150 ms poll re-runs `_evaluate()` while arming so a non-jittering stream still reaches `stableMs`; it stops once captured."* Reads less like pre-empting a reviewer.

qml/components/StableWeightCapture.qml
```diff
@@ -0,0 +1,84 @@
...
+
+    // --- Inputs (parent may override) ---
+    property real weight: 0           // net weight to watch
+    property bool active: true        // only detect while true
+    property real minWeight: 20       // ignore readings below this (noise / empty)
+    property real maxWeight: 100000   // ignore readings above this (e.g. wrong vessel)
+    property real tolerance: 1.0      // ± grams treated as "the same reading"
+    property int  stableMs: 2500      // must hold this long to count as stable
+    property real removeThreshold: 5  // below this = load removed -> re-arm
+    property real rearmDelta: 6        // change this much from captured value -> re-arm
```
skialpine, line 25: **Defaults look milk-tuned, but both callers in #1350 are bean-dosing** (`minWeight: 5`, `maxWeight: 45`) and leave `removeThreshold`/`rearmDelta` at these defaults. Two things worth a sanity check:

- `rearmDelta: 6` means a top-up smaller than 6 g never re-captures — in a 5–45 g bean window that's a big chunk of an ~18 g dose, yet the header comment promises "re-dosing beans re-captures." Is 6 g intentional for beans?
- `removeThreshold` isn't coupled to `minWeight`. It's fine while `removeThreshold (5) <= minWeight`, but if a future caller raises `minWeight` above 5, a load in the gap never captures *and* never re-arms cleanly. A `removeThreshold: Math.min(5, minWeight)` default (or a one-line contract comment) would prevent that.

Not blocking — just flag whether the defaults should target the bean reality the callers actually use, or be documented as milk-oriented.

skialpine (commented):
Reviewed via a multi-agent pass (correctness, comments, silent-failure, type-design, tests). Nice clean foundation PR — the C++ faithfully reuses the existing tick-sound idiom, the `readonly` outputs defend the public surface, and the hysteresis design (separate `removeThreshold` vs `rearmDelta`) is a thoughtful model of "removed vs changed." Genuinely inert, so zero merge risk on its own.

A few small things below. The two trivial ones (Ready-guard + the "full volume" comment) are worth fixing before this calcifies under the stack; the defaults question is a product-tuning call for you since every caller in #1350 is bean-dosing while the defaults look milk-tuned. Tests can reasonably defer to the first feature PR that wires a caller.

Nothing here blocks merge.

PR #1350, opened 2026-06-20
[2/5] Bean auto-capture and dose-cup tare

loganross2001:
Auto-capture the espresso dose by weight, net of a stored dose-cup tare.

- **`SettingsBrew::doseCupTareWeight`** (new persistent property, QSettings key `espresso/doseCupTareWeight`, clamped ≥ 0): the empty dosing-cup weight, subtracted from the scale so the recorded dose is *net beans*.
- **Bean auto-capture:** `BrewDialog.qml` and `IdlePage.qml` each host a `StableWeightCapture` bound to `scaleWeight − doseCupTareWeight`. On a stable reading the dose is set (IdlePage also sets stop-at-weight = dose × ratio), the "ding" plays, and a brief confirmation shows. Bounded to ≤ 45 g; suppressed in steam/hot-water/flush modes so milk/water never trip it.
- The old "Weigh beans" button becomes a small live net-weight readout with a gentle two-line "Place Beans on Scale / (wait for the beep before removing)" prompt, plus a flashing **"wait for the bell"** hint while the scale settles.

**Files:** `settings_brew.{h,cpp}` (dose-cup tare only), `BrewDialog.qml`, `IdlePage.qml`.

---
**📦 Part 2 of 5 — a dependency-ordered stack** splitting #1348 into per-feature PRs (as you asked). Review/merge in order:
1. #1349 — Foundation: StableWeightCapture + capture ding
2. #1350 — Bean auto-capture + dose-cup tare
3. #1351 — Weight-timed steaming
4. #1352 — Steam-page UX polish
5. #1353 — Home-screen rework

Each builds on the previous; its diff narrows to just this feature once the earlier ones merge. Full per-feature technical reference: https://github.com/loganross2001/Decenza-community-fork/blob/community-fork/fork-notes/CHANGES_DETAILED.md

### DIFF
--- a/qml/components/BrewDialog.qml
+++ b/qml/components/BrewDialog.qml
@@ -90,6 +90,35 @@ Dialog {
     property bool showScaleWarning: false
     property bool lowDoseWarning: doseValue < 3 || showScaleWarning
 
+    // Bean auto-capture: when the dose cup (with beans) rests stable on the scale,
+    // lock the net dose, ding, and show a confirmation. Same net-dose math as the
+    // "Get from scale" button; re-arms when the cup is removed / the load changes.
+    property bool beanCaptureVisible: false
+    property string beanCaptureText: ""
+    Timer { id: beanCaptureTimer; interval: 3500; onTriggered: root.beanCaptureVisible = false }
+    StableWeightCapture {
+        id: beanCapture
+        weight: ScaleDevice.connected ? Math.max(0, MachineState.scaleWeight - Settings.brew.doseCupTareWeight) : 0
+        active: root.visible && ScaleDevice.connected && !ScaleDevice.isFlowScale
+        minWeight: 5
+        maxWeight: 45
+        tolerance: 0.5
+        stableMs: 2500
+        onStableCaptured: function(net) {
+            root.showScaleWarning = false
+            root.targetManuallySet = false
+            root.doseValue = net
+            root.beanCaptureText = TranslationManager.translate("brewDialog.doseCaptured", "Dose set: %1g").arg(net.toFixed(1))
+            root.beanCaptureVisible = true
+            beanCaptureTimer.restart()
+            if (typeof AccessibilityManager !== "undefined") {
+                AccessibilityManager.playCaptureDing()
+                if (AccessibilityManager.enabled)
+                    AccessibilityManager.announce(root.beanCaptureText)
+            }
+        }
+    }
+
     // Recalculate target when dose or ratio changes (unless manually overridden)
     onDoseValueChanged: {
         if (!targetManuallySet) {
@@ -334,6 +363,29 @@ Dialog {
                 }
             }
 
+            // Dose-captured confirmation (auto-dismiss after a few seconds)
+            Rectangle {
+                Layout.fillWidth: true
+                visible: root.beanCaptureVisible
+                color: Theme.surfaceColor
+                border.width: 1
+                border.color: Theme.primaryColor
+                radius: Theme.scaled(8)
+                implicitHeight: beanCaptureLabel.implicitHeight + Theme.scaled(24)
+                Text {
+                    id: beanCaptureLabel
+                    anchors.left: parent.left
+                    anchors.right: parent.right
+                    anchors.verticalCenter: parent.verticalCenter
+                    anchors.margins: Theme.scaled(12)
+                    text: root.beanCaptureText
+                    font: Theme.bodyFont
+                    color: Theme.primaryColor
+                    horizontalAlignment: Text.AlignHCenter
+                    Accessible.ignored: true
+                }
+            }
+
             // Temperature input
             ColumnLayout {
                 Layout.fillWidth: true
@@ -450,11 +502,12 @@ Dialog {
                     accessibleName: TranslationManager.translate("brewDialog.getDoseFromScale", "Get dose from scale")
                     primary: true
                     onClicked: {
-                        var scaleWeight = MachineState.scaleWeight
-                        if (scaleWeight >= 3) {
+                        // Net beans = scale reading minus the stored dosing-cup tare (0 if unset)
+                        var net = MachineState.scaleWeight - Settings.brew.doseCupTareWeight
+                        if (net >= 3) {
                             root.showScaleWarning = false
                             root.targetManuallySet = false  // Reset manual flag
-                            root.doseValue = scaleWeight
+                            root.doseValue = net
                         } else {
                             // Show warning but don't change dose
                             root.showScaleWarning = true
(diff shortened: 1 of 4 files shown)


skialpine (commented):
Multi-agent review (code / silent-failure / comments / tests) of the bean auto-capture + dose-cup tare work.

First, the good: the `doseCupTareWeight` C++ property follows the SETTINGS.md pattern exactly (getter/setter/NOTIFY, `espresso/` key, default 0, negative-clamp, change-guarded emit), i18n uses `TranslationManager.translate` throughout, and the new interactive elements carry accessibility names. The net-dose math is correctly shared with the existing "Get from scale" path rather than forked.

The substance is in the auto-capture wiring. Headline: the feature is **on by default with no coherent default state** — with the tare at its default 0, `weight = gross (cup+beans)`, so out of the box it either captures a wrong dose (light cup) or never fires (heavy cup > 45 g) while telling the user to wait for a bell that never rings. We've agreed on the fix and it's already being applied: **gate auto-capture on `doseCupTareWeight > 0`** (no saved cup ⇒ feature dormant; setting the tare back to 0 turns it off — no new setting). That resolves the default-state, wait-for-bell, and no-opt-out problems together. Remaining items are called out inline + below.

## Critical (separate from the tare-gate fix)
- **Double-fire when the brew dialog is open.** `BrewDialog` is a child of `IdlePage`, and the idle capture's `active` doesn't exclude the dialog being open, so a single stable cup fires *both* handlers — the idle one writes `dyeBeanWeight`/`brewYieldOverride` underneath the dialog's in-progress edits, plus a double ding/announce. The idle capture needs `&& !idleBrewDialog.visible`. (inline on IdlePage `active`)

## Important
- **Serializer parity gap (likely bug).** `doseCupTareWeight` persists via QSettings but was not added to `settingsserializer.cpp` — export block (`~L59-60`) or import (`~L399-400`), which hand-list `targetWeight`/`lastUsedRatio`. So the calibrated tare is **silently lost on settings backup/restore and device-to-device migration**, reverting to wrong net-dose math. (File isn't in this diff so I can't inline it.) Confirm intentional (device-specific?) or add to both sides + a parity test.
- **"Weigh" button silent no-op** on a 0/negative scale reading — no feedback, unlike the sibling "Get dose from scale" which sets `showScaleWarning`. (inline)

## Suggestions
- Inconsistent/dead dose thresholds: idle `if (net < 3) return` is unreachable (`minWeight: 5`), BrewDialog capture has no guard, the button uses `>= 3`. Pick one floor. (inline)
- `brewYieldOverride = net * lastUsedRatio` can be stale vs the active profile after a profile switch. (inline)
- Comment rot: references a "Weigh beans button" that became a readout. (inline)
- Drop the `typeof AccessibilityManager !== "undefined"` guards — it's registered unconditionally before QML loads.
- Add one `doseCupTareWeightRoundTrip` test in `tst_settings.cpp` (round-trip + negative-clamp + no-op idempotency), matching the `targetWeight` pattern.

Heads-up: I rebased this branch onto current `main` locally (dropping the now-squash-merged #1349 foundation commit) — the foundation is in `main` as of #1349.

qml/pages/IdlePage.qml
```diff
@@ -124,6 +124,65 @@ Page {
...
+    // in espresso/beans mode (NOT while showing steam/hot-water/flush presets,
+    // where the scale is for milk/water), and bounded to a dose-plausible weight
+    // (<= 45 g net) so a milk pitcher or water vessel never trips it.
+    property bool beanCaptureShown: false
+    property string beanCaptureText: ""
+    Timer { id: idleBeanCaptureTimer; interval: 3500; onTriggered: idlePage.beanCaptureShown = false }
+    StableWeightCapture {
+        id: beanCapture
+        weight: ScaleDevice.connected ? Math.max(0, MachineState.scaleWeight - Settings.brew.doseCupTareWeight) : 0
+        active: ScaleDevice.connected && !ScaleDevice.isFlowScale
```
skialpine, line 163: **Two issues on this `active` gate.**

1. **No coherent default state (critical).** With `doseCupTareWeight` at its default 0, `weight` = gross cup+beans. Out of the box this either captures a wrong dose (light cup) or, for a cup whose gross > `maxWeight: 45`, never fires. **Agreed fix (being applied):** add `&& Settings.brew.doseCupTareWeight > 0` so the feature is dormant until a cup tare is saved.

2. **Double-fire when the brew dialog is open (critical, separate).** `BrewDialog` is a child of this page and its capture also goes active, so a stable cup fires both handlers — this one writes `dyeBeanWeight`/`brewYieldOverride` under the dialog's edits, plus a double ding. Add `&& !idleBrewDialog.visible` (and ideally exclude any open modal).

qml/pages/IdlePage.qml
```diff
@@ -124,6 +124,65 @@ Page {
...
+    // settling on the scale (something is on the scale but the capture hasn't
+    // fired yet). Disappears the instant it captures (the bell rings).
+    Text {
+        id: waitForBellHint
+        anchors.horizontalCenter: parent.horizontalCenter
+        anchors.top: parent.top
+        anchors.topMargin: Theme.scaled(70)
+        z: 1500
+        horizontalAlignment: Text.AlignHCenter
+        readonly property bool beansSettling: beanCapture.active && !beanCapture.isCaptured
```
skialpine, line 196: **"Wait for the bell" can flash forever.** This visibility is `weight >= minWeight` with no upper bound, but capture gates on `weight <= maxWeight (45)`. For any net weight in `[45, ∞)` the hint blinks indefinitely for a bell the code has guaranteed will never ring — and with default tare 0, a heavy cup's gross reading sits there permanently. Gating capture on `doseCupTareWeight > 0` (the agreed fix) hides this when disabled since `beansSettling` keys off `beanCapture.active`; but even when enabled, consider bounding the hint at `<= maxWeight` or showing a "too heavy for a dose" message instead of an endless wait.

qml/pages/IdlePage.qml
```diff
@@ -124,6 +124,65 @@ Page {
...
+        active: ScaleDevice.connected && !ScaleDevice.isFlowScale
+                && idlePage.activePresetFunction !== "steam"
+                && idlePage.activePresetFunction !== "hotwater"
+                && idlePage.activePresetFunction !== "flush"
+        minWeight: 5
+        maxWeight: 45
+        tolerance: 0.5
+        stableMs: 2500
+        onStableCaptured: function(net) {
+            if (net < 3) return
```
skialpine, line 172: `if (net < 3) return` is dead code — `minWeight: 5` already prevents any capture below 5 g, so this never triggers. Meanwhile the BrewDialog `onStableCaptured` has no such guard and the "Get from scale" button uses `>= 3` — three different floors implying a distinction that doesn't exist. Pick one source of truth (lean on `minWeight`) and drop the dead branch.

qml/pages/IdlePage.qml
```diff
@@ -124,6 +124,65 @@ Page {
...
+            Settings.dye.dyeBeanWeight = net
+            Settings.brew.brewYieldOverride = net * Settings.brew.lastUsedRatio
```
skialpine, line 174: `lastUsedRatio` only updates on brew-dialog OK and defaults to 2.0, so after a profile switch this auto-set `brewYieldOverride` can silently disagree with the active profile's intended yield. The dialog path recomputes from the live `ratio`; consider deriving from the active profile here too, or confirm `lastUsedRatio` is the intended source.

qml/pages/IdlePage.qml
```diff
@@ -124,6 +124,65 @@ Page {
     // Track which function's presets are showing (used by center-zone action items)
     property string activePresetFunction: ""  // "", "steam", "espresso", "hotwater", "flush", "beans"
 
+    // Idle bean auto-capture: when the dose cup (with beans) rests stable on the
+    // scale, set the dose + stop-at-weight (same as the "Weigh beans" button),
```
skialpine, line 152: Comment rot: there is no "Weigh beans button" anywhere in the tree — it became the readout `Text` (as the comment ~30 lines down explicitly says: "this is a readout (not a button)"). Replace the analogy with the concrete behavior, e.g. "sets the dose (dyeBeanWeight) and stop-at-weight (brewYieldOverride = dose × lastUsedRatio)."

qml/components/BrewDialog.qml
```diff
@@ -499,6 +552,52 @@ Dialog {
...
+                }
+
+                AccessibleButton {
+                    Layout.preferredHeight: Theme.scaled(44)
+                    text: TranslationManager.translate("brewDialog.weighCup", "Weigh")
+                    accessibleName: TranslationManager.translate("brewDialog.weighEmptyCup", "Weigh empty cup from scale")
+                    primary: true
+                    onClicked: {
+                        var w = MachineState.scaleWeight
+                        if (w > 0)
```
skialpine, line 573: `if (w > 0) doseCupTareWeight = w` silently no-ops when the scale reads 0 or negative (forgot to tare, empty cup drifting slightly negative): the user taps "Weigh", the field doesn't change, and nothing explains why. The sibling "Get dose from scale" button handles its failure path with `showScaleWarning = true` — mirror that (or a short toast) here.

skialpine:
Closing — this was reworked and merged as #1358 (commit 7488a339), which builds on your bean-dosing work here. Thanks @loganross2001 for the original implementation; the core feature (dose auto-capture + dose-cup tare) shipped with a delta/virtual-zero rewrite for un-zeroed scales, a single shared Brew Settings dialog, an opt-in capture ding, and the review fixes.

The rest of your stack (#1351–#1353) is unaffected by this.

PR #1358, opened 2026-06-20
Bean auto-capture: virtual-zero dose weighing (reworks #1350)

skialpine:
Reworked, hardware-validated bean auto-capture (supersedes #1350; builds on @loganross2001's original work and its foundation #1349). Rebased onto current `main` (includes #1357).

## What it does
Set your dose cup of beans on the scale; when it settles, the app sets the dose (and stop-at-weight = dose × ratio) and shows it — no buttons. Set the empty-cup weight once via **Weigh**; default 0 means the feature is off.

## Core: virtual-zero delta model
Tracks an empty-scale **virtual zero** (any value, incl. negative) and measures `net = (load − virtualZero) − cupWeight`, so an un-zeroed or drifting scale still gives the correct dose. **Validated on-device** (macOS, Half Decent Scale) down to a **−356 g** baseline — capture still produced the correct net dose. Includes stable seed, re-zero on removal/drift, and re-zero on `tareCompleted`.

## Fixes from the #1350 review
- No coherent default / wrong-dose on tare 0 → gated on a saved cup.
- Heavy cup never captures → plausibility gate on **net** beans, not gross (the 122 g Niche steel cup works).
- "Wait for the bell" forever → hint bounded to `[minNet, maxNet]`.
- Double-ding / re-capture on opening brew settings → one persistent capture, latch survives open/close.
- Dose not updating in an open brew-settings window → canonical `dyeBeanWeight` write + a watcher in BrewDialog.
- Manual button redundancy → "Get from scale" shows only when no cup saved.
- "Weigh" silent no-op / offset baked in → stores the cup as an offset-free delta, disabled until the cup is on the scale.
- Tare lost on backup/migration → `doseCupTareWeight` added to the serializer.

## Follow-on changes (from review + review testing)
- **One shared Brew Settings dialog.** The Shot Plan and Scale Weight home-screen tiles now open the single `BrewDialog` owned by IdlePage (via an `openBrewSettings()` parent-walk), instead of each hosting its own copy. Fixes the virtual-zero Weigh/Get-from-scale offset only reaching the idle instance, and deletes two duplicate dialogs.
- **Capture ding is opt-in.** New `SettingsBrew.doseCaptureSoundEnabled` (default **off**, serialized), toggled by a **bell icon** on the Dose cup row (`bell.svg`/`bell-off.svg`). The spoken accessibility announcement is unaffected.
- Readout uses `Theme.labelFont` (no hardcoded size).

## Verification
Builds clean (0 warnings). Settings tests pass (`doseCupTareWeight`, `doseCaptureSoundEnabled` round-trip + clamp). Extensively tested on the macOS build against a real scale: normal capture, negative/non-zeroed baseline, tap/jostle immunity, one-ding-per-placement across dialog open/close, live dose update in the dialog, and the shared dialog from all three entry points.

## Deferred to a follow-up
Separating the **working dose** from the **bag's remembered dose** (so an un-shot override doesn't stick to the bag — bag dose updates only on a completed shot or an explicit most-recent-shot edit). Tracked separately; not in this PR.

### DIFF
--- a/qml/components/BrewDialog.qml
+++ b/qml/components/BrewDialog.qml
@@ -1,6 +1,7 @@
 import QtQuick
 import QtQuick.Controls
 import QtQuick.Layouts
+import QtQuick.Effects
 import Decenza
 
 Dialog {
@@ -99,10 +100,9 @@ Dialog {
     // stays armed across this dialog opening/closing, so a cup already weighed on
     // the home screen is NOT re-captured (no second ding) when you open settings.
     // The capture writes the canonical dyeBeanWeight; this watcher reflects it into
-    // the editable dose whenever a capture lands while THIS dialog is open. (There
-    // are several BrewDialog instances — idle, ShotPlan tile, ScaleWeight tile — so
-    // each one self-updates rather than the capture targeting a specific instance.)
-    // The manual "Get from scale" button below covers the no-dose-cup case.
+    // the editable dose whenever a capture lands while the dialog is open (rather
+    // than the capture poking doseValue directly). The manual "Get from scale"
+    // button below covers the no-dose-cup case.
     Connections {
         target: Settings.dye
         enabled: root.visible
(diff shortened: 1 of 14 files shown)


skialpine:
### Code review

Focus was the latest change — hosting the single Brew Settings dialog at the app root so it's reachable from the global status bar. Confirmed: `Window.window.openBrewSettings()` resolves from every host (status bar, idle content, layout previews) since all live in the one `ApplicationWindow`; the `scaleVirtualZero` bridge via `pageStack.currentItem` is reactive and NaN-safe (falls back to a literal `0` off-idle); no stale `idleBrewDialog`/per-tile-dialog references remain; and the `pendingBrewDialog`/auto-favorite flow is intact. CLAUDE.md: clean (timers, settings domain, Theme, i18n, accessibility, QML gotchas). The prior status-bar-tile regression is resolved.

---

<details>
<summary>2 below-threshold items (50-74)</summary>

- [OK to skip] The dialog's `scaleVirtualZero` is a live binding; if the app navigates away from Idle (programmatically) while the dialog is open, it snaps to `0`, which could affect a `Weigh` on a non-zeroed scale. The dialog is modal so normal user navigation is blocked, making this rare; snapshotting the value on open would fully close it. (score: 50)
https://github.com/Kulitorum/Decenza/blob/20e82cc99dcc4d78ad7209c2936ac3a5ea862ad5/qml/main.qml#L50-L53

- [Worth a look] The `StableWeightCapture` seed comment says an already-loaded scale at startup "can't become the zero," but only a *transient* can't — a sustained load (cup left on the scale across startup) can be adopted as the baseline. Behavior is mitigated by the tare→`reset()` path; the comment slightly overstates. One-line comment fix. (score: 50)
https://github.com/Kulitorum/Decenza/blob/20e82cc99dcc4d78ad7209c2936ac3a5ea862ad5/qml/components/StableWeightCapture.qml#L75-L79

</details>

PR #1353, opened 2026-06-20
[5/5] Home-screen rework (opinionated)

loganross2001:
⚠️ **The most opinionated change — easy to decline on its own.** Tuned to a 1920×1200 tablet and it **repurposes the `shotPlan` widget's full mode**, so please review with that in mind (see the candid integration-risk notes in the linked doc, Appendix B).

- **`ShotPlanItem.qml`** full mode → three centered setup buttons (Espresso-Shot Setup → BrewDialog; Choose Espresso Profile → ProfileSelectorPage; Milk-Steaming Setup → SteamPage). **Compact/bar mode is unchanged.** This silently redefines what `shotPlan` renders in full mode — a cleaner upstream form would be a *new* widget type registered in the usual four places.
- **`IdlePage.qml`:** a high-contrast bottom stat bar (Profile · Scale · Ratio · Beans · Milk); the duplicated center status readouts hidden (`centerStatus visible: false`); a thin status-bar bottom-border line; and the center block re-anchored to the top so it grows downward (buttons never slide behind the header). These bypass/override the configurable zone system, so you may prefer to gate the whole thing behind an opt-in setting.
- **`LayoutCenterZone.qml`:** an equal-cell `distribute` mode for the center status/middle rows.

**Files:** `IdlePage.qml` (layout only), `ShotPlanItem.qml`, `LayoutCenterZone.qml`.

---
**📦 Part 5 of 5 — a dependency-ordered stack** splitting #1348 into per-feature PRs (as you asked). Review/merge in order:
1. #1349 — Foundation: StableWeightCapture + capture ding
2. #1350 — Bean auto-capture + dose-cup tare
3. #1351 — Weight-timed steaming
4. #1352 — Steam-page UX polish
5. #1353 — Home-screen rework

Each builds on the previous; its diff narrows to just this feature once the earlier ones merge. Full per-feature technical reference: https://github.com/loganross2001/Decenza-community-fork/blob/community-fork/fork-notes/CHANGES_DETAILED.md

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1352, opened 2026-06-20
[4/5] Steam-page UX polish

loganross2001:
Quality-of-life improvements to the steam page (no new domain logic).

- **Scrollable editor:** the per-pitcher settings panel (which grew taller with the new rows) is wrapped in a `Flickable` with a `ScrollBar`. The "Off"-preset placeholder switched from `Layout.fillHeight` to a fixed height (fillHeight is meaningless inside a content-sized Flickable).
- **Weigh buttons:** inline buttons set the empty-pitcher weight / reference milk straight from the scale.
- **0.1 g inputs:** the pitcher-weight and reference-milk `ValueInput`s use `stepSize 0.1`, `decimals 1`.
- **Purge button:** in the editor and on the live steaming view, calling `DE1Device.requestIdle()` (the documented steam-purge trigger).

**Files:** `SteamPage.qml` only.

---
**📦 Part 4 of 5 — a dependency-ordered stack** splitting #1348 into per-feature PRs (as you asked). Review/merge in order:
1. #1349 — Foundation: StableWeightCapture + capture ding
2. #1350 — Bean auto-capture + dose-cup tare
3. #1351 — Weight-timed steaming
4. #1352 — Steam-page UX polish
5. #1353 — Home-screen rework

Each builds on the previous; its diff narrows to just this feature once the earlier ones merge. Full per-feature technical reference: https://github.com/loganross2001/Decenza-community-fork/blob/community-fork/fork-notes/CHANGES_DETAILED.md

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1351, opened 2026-06-20
[3/5] Weight-timed steaming

loganross2001:
Scale steam time to the measured milk weight so a half-full and a full pitcher both auto-stop at the right foam point.

- **`SettingsBrew::setSteamPitcherCalibration(index, calibMilkG)`** (new): stores a *reference milk weight* on a steam-pitcher preset (or removes it when ≤ 0). Pairs with the preset's existing `duration` and optional `pitcherWeightG` tare.
- **Formula:** `steamTime = clamp(round(duration × measuredMilk / referenceMilk), 5, 120) s`, where `measuredMilk = scaleWeight − pitcherWeightG`. Uncalibrated presets fall back to the fixed duration. The scaled value is written to `steamTimeout` → the DE1's `TargetSteamLength` (auto-stop denominator).
- **`SteamPage.qml`:** a "Reference milk" input + Weigh button, the scaling helpers, a milk `StableWeightCapture` that locks the scaled time (and dings), an "Expected steam time" readout, and a milk "wait for the bell" hint.
- **Includes a bug fix:** three code paths previously re-synced `steamTimeout` to the *unscaled* duration (on page activate, session end, and pill taps), silently wiping the scaled value once the pitcher was lifted off the scale. A `syncSteamTimeout()` helper now preserves the already-scaled value when the preset is calibrated but no milk is on the scale. (Detail in the linked doc, §3.)
- **`IdlePage.qml`:** the home-screen steam flow gains the same milk capture (`measuredMilkG`) and a fallback so steaming uses the last measured milk when the pitcher is already at the wand.

**Files:** `settings_brew.{h,cpp}` (steam calibration only), `SteamPage.qml` (steaming logic), `IdlePage.qml` (milk capture).

---
**📦 Part 3 of 5 — a dependency-ordered stack** splitting #1348 into per-feature PRs (as you asked). Review/merge in order:
1. #1349 — Foundation: StableWeightCapture + capture ding
2. #1350 — Bean auto-capture + dose-cup tare
3. #1351 — Weight-timed steaming
4. #1352 — Steam-page UX polish
5. #1353 — Home-screen rework

Each builds on the previous; its diff narrows to just this feature once the earlier ones merge. Full per-feature technical reference: https://github.com/loganross2001/Decenza-community-fork/blob/community-fork/fork-notes/CHANGES_DETAILED.md

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1359, opened 2026-06-20
Ratio quick-select on the home screen + editable presets

loganross2001:
Quick-select coffee-to-water ratio from the home screen, with user-editable presets.

- The bottom stat bar's **Ratio** value becomes a tappable pill → opens a **Brew Ratio** chooser with three one-tap styles: **Ristretto / Normale / Lungo**. Picking one sets the ratio *and* recomputes the yield target from the current dose (same math as the bean auto-capture).
- The three preset **values** are editable at the bottom of **Espresso Setup** (e.g. set them to 1:1.5 / 1:2.6 / 1:3.7) — three new persistent settings (`ratioPreset1/2/3`, clamped 0.5–6.0). The home-screen chooser follows them live.
- An expandable **"About brew ratios"** help panel.
- **Attribution:** the ratio-style descriptions are adapted from "[Brew Ratios Around the World](https://home.lamarzoccousa.com/brew-ratios-around-world/)" by Ben Blake & Scott Callender — La Marzocco Home (2014), credited in-dialog with a link. The text is our own wording; only the styles/learning are theirs.

**Files:** new `qml/components/RatioPresetDialog.qml`, `qml/pages/IdlePage.qml`, `qml/components/BrewDialog.qml`, `src/core/settings_brew.{h,cpp}`, `CMakeLists.txt`.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -760,6 +760,7 @@ set(QML_FILES
     qml/components/BeanBaseDetailsRow.qml
     qml/components/BeanSummary.qml
     qml/components/ChangeBeansDialog.qml
+    qml/components/RatioPresetDialog.qml
     qml/components/BagCard.qml
     qml/components/SwitchEquipmentDialog.qml
     qml/components/EquipmentCard.qml
--- a/qml/components/BrewDialog.qml
+++ b/qml/components/BrewDialog.qml
@@ -90,6 +90,35 @@ Dialog {
     property bool showScaleWarning: false
     property bool lowDoseWarning: doseValue < 3 || showScaleWarning
 
+    // Bean auto-capture: when the dose cup (with beans) rests stable on the scale,
+    // lock the net dose, ding, and show a confirmation. Same net-dose math as the
+    // "Get from scale" button; re-arms when the cup is removed / the load changes.
+    property bool beanCaptureVisible: false
+    property string beanCaptureText: ""
+    Timer { id: beanCaptureTimer; interval: 3500; onTriggered: root.beanCaptureVisible = false }
+    StableWeightCapture {
+        id: beanCapture
+        weight: ScaleDevice.connected ? Math.max(0, MachineState.scaleWeight - Settings.brew.doseCupTareWeight) : 0
+        active: root.visible && ScaleDevice.connected && !ScaleDevice.isFlowScale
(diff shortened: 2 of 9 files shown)


skialpine (commented):
The **one-tap named-ratio idea is good** and genuinely isn't reproducible through existing settings — no preset concept exists today, so this is a fair feature. Two things on the implementation, though, both about not duplicating what's already there.

### This adds a second, parallel ratio dialog

Setting a ratio from the home screen and recomputing the yield from the dose already ships. Tapping the shot-plan pill opens `BrewDialog`, which on `main` already has:

- a **Ratio input** (`BrewDialog.qml:480–506`, `value: root.ratio`),
- `targetValue = doseValue * ratio` recompute on every change,
- and it persists `Settings.brew.lastUsedRatio` on confirm.

`RatioPresetDialog.applyRatio()` does the same — `lastUsedRatio = r` then `brewYieldOverride = dyeBeanWeight * r` (the "same math as the bean auto-capture" from your description). So this introduces a *second* home-screen dialog for ratio next to the existing one, re-deriving the same set-ratio-and-recompute logic.

**Suggestion:** add the three preset chips into `BrewDialog`'s existing Ratio row instead of a new dialog. One dialog, one code path, and the chips reuse the recompute BrewDialog already has — the help/attribution panel can live there too (or behind a small info affordance). That folds ~267 new lines back into the existing flow and avoids two entry points that can drift apart.

### Three new settings is a lot for this

`ratioPreset1/2/3` are three persisted doubles plus an editor UI in Espresso Setup. The maintainer leans toward fewer user-facing settings, and 1:1 / 1:2 / 1:3 are about as universal as espresso defaults get. Could the chips ship with fixed sensible defaults (no editor), or key off the one ratio the user already has, rather than three new tunables? If per-user preset *values* turn out to be genuinely wanted, that can be added later — but it's worth starting without them.

### Net

Keep the named-preset concept; drop the parallel dialog and (ideally) the three settings. Reshaped as "preset chips in the existing BrewDialog ratio row with fixed defaults," this is a much smaller change that adds no duplicate code path. Happy to re-review once it's in that shape.

loganross2001:
Superseded by #1364 — ratio quick-select is folded into the independent simplified-home PR (#1364). Closing in favour of #1364.

PR #1360, opened 2026-06-20
Save last steam session; "use as baseline" in steam setup

loganross2001:
Save the most recent steam session (milk weight + actual steam time) so steam setup can adopt it as a new reference baseline.

- Two new persistent settings: `lastSteamMilkG` (saved when a milk capture locks, from either the home flow or the steam page) and `lastSteamTimeS` (the **actual elapsed** steam duration, saved when a steam session ends via a small watcher in `main.qml` — works regardless of which screen you steamed from).
- In **Steam Setup**, a new **"Use as baseline"** button (under the reference-milk row) shows *"Last: X g → Y s"* and, in one tap, sets the current pitcher's reference milk + duration to those values.

**Files:** `src/core/settings_brew.{h,cpp}`, `qml/main.qml`, `qml/pages/SteamPage.qml`, `qml/pages/IdlePage.qml`. Builds on the weight-timed-steaming change.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -760,6 +760,7 @@ set(QML_FILES
     qml/components/BeanBaseDetailsRow.qml
     qml/components/BeanSummary.qml
     qml/components/ChangeBeansDialog.qml
+    qml/components/RatioPresetDialog.qml
     qml/components/BagCard.qml
     qml/components/SwitchEquipmentDialog.qml
     qml/components/EquipmentCard.qml
--- a/qml/components/BrewDialog.qml
+++ b/qml/components/BrewDialog.qml
@@ -90,6 +90,35 @@ Dialog {
     property bool showScaleWarning: false
     property bool lowDoseWarning: doseValue < 3 || showScaleWarning
 
+    // Bean auto-capture: when the dose cup (with beans) rests stable on the scale,
+    // lock the net dose, ding, and show a confirmation. Same net-dose math as the
+    // "Get from scale" button; re-arms when the cup is removed / the load changes.
+    property bool beanCaptureVisible: false
+    property string beanCaptureText: ""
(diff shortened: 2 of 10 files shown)


skialpine (commented):
This is the cleanest of the three morning PRs — there's little to push back on, and unlike #1359/#1361 it isn't reinventing something the existing system already does. On `main` there's no memory of a past steam session at all, so the auto-record + one-tap **"Use as baseline"** is genuinely new convenience that can't be reproduced through existing settings. A few notes, mostly affirming:

- **The steam-end watcher is event-based** (`Connections` on `MachineState` phase/`shotTime` in `main.qml`), not a timer-as-guard — good, that's the project rule and it's easy to get wrong.
- **`lastSteamMilkG` / `lastSteamTimeS` follow the existing `lastUsedRatio` precedent** — persisted last-values, not user-facing tunables — so they don't carry the "new setting" weight that, say, the three `ratioPreset` settings in #1359 do. No concern here.
- **"Use as baseline" is a thin wrapper** over the per-pitcher calibration the stack already adds, rather than a parallel code path — nice.

The one thing worth calling out isn't the code, it's the **coupling**: this rests entirely on the unmerged weight-timed-steaming base (per-pitcher milk→duration calibration). It can't really be evaluated on its own — if that base model lands, this is a clean, well-scoped addition on top; if it changes shape, this moves with it. So my only ask is that this be reviewed/merged after the base is settled, in stack order as you note.

Minor: confirm `lastSteamMilkG` is in fact written from *both* entry points (home flow and steam page) as the description says — worth a quick check that the home-flow capture path saves it, since the watcher in `main.qml` only handles the time.

loganross2001:
Superseded by #1363 — the last-steam baseline is folded into the independent weight-based-steaming PR (#1363). Closing in favour of #1363.

PR #1361, opened 2026-06-20
Icon-led top status bar; relocate Sleep to top-centre

loganross2001:
Redesign the top status bar as a clean, icon-led summary, and move Sleep out of the bottom-left corner.

- **Icon-led datapoints** (tinted via `ThemedIcon`): DE1 status · group-head temp · steam temp on the left; scale status · battery on the right. Ratio and brew weight are removed from the top bar (they live in the home-screen bottom stat bar). Each item dims and shows "—" when its device is disconnected. (Battery uses the device/tablet battery via `BatteryManager`, since many scales don't report battery over BLE.)
- **Sleep relocated** to a clear, labelled button dead-centre in the top bar. It previously shared the bottom-left corner with the Back button, which caused **accidental sleeps during screen-transition lag**. The bottom `SleepItem` now renders nothing (no duplicate). Long-press still quits.

⚠️ Opinionated: this hardcodes the top bar's contents rather than rendering the configurable `statusBar` zone. A cleaner upstream form could keep the zone system and add icon-capable widgets. **Files:** `qml/components/StatusBar.qml`, `qml/components/layout/items/SleepItem.qml`.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -760,6 +760,7 @@ set(QML_FILES
     qml/components/BeanBaseDetailsRow.qml
     qml/components/BeanSummary.qml
     qml/components/ChangeBeansDialog.qml
+    qml/components/RatioPresetDialog.qml
     qml/components/BagCard.qml
     qml/components/SwitchEquipmentDialog.qml
     qml/components/EquipmentCard.qml
--- a/qml/components/BrewDialog.qml
+++ b/qml/components/BrewDialog.qml
@@ -90,6 +90,35 @@ Dialog {
     property bool showScaleWarning: false
     property bool lowDoseWarning: doseValue < 3 || showScaleWarning
 
+    // Bean auto-capture: when the dose cup (with beans) rests stable on the scale,
+    // lock the net dose, ding, and show a confirmation. Same net-dose math as the
+    // "Get from scale" button; re-arms when the cup is removed / the load changes.
+    property bool beanCaptureVisible: false
+    property string beanCaptureText: ""
+    Timer { id: beanCaptureTimer; interval: 3500; onTriggered: root.beanCaptureVisible = false }
+    StableWeightCapture {
+        id: beanCapture
+        weight: ScaleDevice.connected ? Math.max(0, MachineState.scaleWeight - Settings.brew.doseCupTareWeight) : 0
+        active: root.visible && ScaleDevice.connected && !ScaleDevice.isFlowScale
+        minWeight: 5
+        maxWeight: 45
+        tolerance: 0.5
+        stableMs: 2500
+        onStableCaptured: function(net) {
+            root.showScaleWarning = false
+            root.targetManuallySet = false
+            root.doseValue = net
+            root.beanCaptureText = TranslationManager.translate("brewDialog.doseCaptured", "Dose set: %1g").arg(net.toFixed(1))
+            root.beanCaptureVisible = true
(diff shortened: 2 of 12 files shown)


skialpine:
this is an example of what you should not do. 

you can already use the layout page to move the sleep button where ever you want it, dont force that change on everyone.  add new layout widgets or options to current ones to allow changing the look, do not change layouts for everyone to reflect how you like it. 

once you have a great layout you can share it  with everyone using the layout library . 

A little dated but look at Michael’s video
https://youtu.be/DHtUZHtsPgs
to understand the philosophy.

skialpine (commented):
Thanks for this — the **icon-led look is a real improvement** and the bar reads much better at a glance. The concern is purely *how* it's implemented: as written, this replaces the configurable `statusBar` zone with a hardcoded `RowLayout`, which removes top-bar customization for everyone and reimplements widgets that already exist. I think you can get the exact same look while keeping the zone system intact, and with a lot less code.

### The core issue

The old `StatusBar.qml` rendered the user-configurable `statusBar` zone (`Repeater` over `LayoutItemDelegate` reading `Settings.network.layoutConfiguration`). Hardcoding the contents means anyone who customized their top bar loses it and can no longer change it. The datapoints you're hardcoding already have widgets in `qml/components/layout/items/`:

| Hardcoded item | Existing widget |
|---|---|
| DE1 status | `MachineStatusItem` / `ConnectionStatusItem` |
| Group temp | `TemperatureItem` |
| Steam temp | `SteamTemperatureItem` |
| Scale status | `ConnectionStatusItem` / `ScaleWeightItem` |
| Device battery | `BatteryLevelItem` |

Worth noting: your `batteryIcon()` is a verbatim copy of the one already in `BatteryLevelItem.qml` (same `≤10/37/62/87` thresholds) — that one can just be reused.

### Suggested approach (same look, keeps the zone system)

1. **Add an optional `iconSource` + `active`-dimming to the text-only widgets** — `MachineStatusItem`, `TemperatureItem`, `SteamTemperatureItem`, `ConnectionStatusItem` currently render `Text` only. Your `StatusItem` component (icon + value + `active` + `Accessible.name`) is a nice shape — factor it into a shared helper those widgets reuse. Then the icon-led look shows up *anywhere* the widget is placed, not just the top bar, and `StatusBar.qml` stays as the thin zone renderer it was.

2. **Bake the disconnect "dim to 40% + —" behavior into the widgets themselves** (tied to `DE1Device.connected` / `ScaleDevice.connected`). Good touch — just make it general rather than specific to the hardcoded bar.

### On relocating Sleep

I'd drop the Sleep relocation from this PR. The Sleep button is **already movable** through the layout editor — anyone who doesn't like it in the bottom-left corner can move it themselves, so we don't need to move it for everyone (and `SleepItem.qml` then doesn't need to be blanked out to avoid a duplicate). If accidental sleeps during nav lag are the real concern, that's better addressed on its own as an event-based guard than by hardcoding position.

### Net

The valuable deltas here are small: optional icons on ~4 existing widgets (~30–40 lines) plus consistent disconnect-dimming. That gives the same scannable, icon-led bar while keeping the top bar configurable for everyone and dropping the `SleepItem`/`batteryIcon` duplication. Happy to look again once it's reshaped that way.

loganross2001:
Superseded by #1362 — re-submitted as an independent, opt-in PR (rebased onto current main), addressing the request to un-stack these. Closing in favour of #1362.

PR #1362, opened 2026-06-21
Top status bar + Sleep relocation (opt-in)

loganross2001:
## Summary
Optional compact, icon-led top status bar with a centred Sleep button. **Default OFF** — the existing zone-driven status bar is completely unchanged unless enabled in **Settings → Layout → "Use compact status bar"**.

## What's in it
- New `compactStatusBar` theme setting (default `false`) + a toggle in Settings → Layout.
- When on: a fixed bar showing DE1 state, group/steam temps, scale weight, and battery (mobile only), with a centred **single-tap** Sleep button.
- No hidden long-press-to-quit on the Sleep button — deliberate quit stays in the existing `QuitItem` layout widget.
- `SleepItem` collapses to nothing when the compact bar owns Sleep, so there's no duplicate control.

## Independent & tested
- Diffs **only this feature** vs `main` — not stacked on anything else.
- Sleep button uses `AccessibleMouseArea` (screen-reader announce-then-activate, so exploring the bar can't sleep the machine on first touch).
- Builds on macOS; reviewed with the pr-review-toolkit; tested on-device.

Replaces the stacked #1361.

### DIFF
--- a/qml/components/StatusBar.qml
+++ b/qml/components/StatusBar.qml
@@ -32,6 +32,8 @@ Rectangle {
         var win = Window.window
         if (win && typeof win.goToScreensaver === "function")
             win.goToScreensaver()
+        else
+            console.warn("StatusBar.doSleep: no goToScreensaver on the window")
     }
 
     // ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


loganross2001:
**Review completed** (per the request to review each change and iterate until clean).

Ran the `pr-review-toolkit` reviewers (code-reviewer, silent-failure-hunter, comment-analyzer) against this PR's diff, fixed every real finding, and re-reviewed until a pass came back with nothing.

- Caught + fixed: the Sleep button used a raw `MouseArea`, so a screen-reader user *exploring* the bar would sleep the machine on first touch — now an `AccessibleMouseArea` (announce-then-activate). Plus comment-durability nits.
- Builds on macOS; tested on-device.

Independent (diffs only the top-bar files), opt-in, default **off**.

skialpine:
Superseded by #1369 (now merged to `main`).

Rather than a hardcoded compact bar behind a `compactStatusBar` flag, #1369 reworks the same idea into the existing layout engine — composable, no flag, opt-in for everyone — the same approach that replaced the #1364 home-screen work in #1368:

- the top `statusBar` zone now honors the per-zone options (distribution / alignment / style) like every other bar (it was the last special-cased renderer);
- the existing readout widgets (machine status, group/steam temp, scale weight) gain a per-instance **icon / text display mode** that works in any zone, not just a compact bar — no new widgets, no new icons;
- the Sleep widget gets per-instance **`allowQuit`** (remove long-press-to-quit) and **`showIcon`** options;
- a one-tap **"Compact status bar"** populate preset reproduces this PR's look on the status bar;
- `connectionStatus` is merged into `machineStatus` (which already shows "Disconnected" when offline), with a migration so existing layouts update automatically.

So you get this PR's compact, icon-led bar and the centred single-tap Sleep, but composable and customizable, without removing the configurable status bar for other users. Thanks for the original — it's the basis for the reworked version. Closing in favor of #1369.

PR #1364, opened 2026-06-21
Simplified home + ratio quick-select (opt-in)

loganross2001:
## Summary
An optional simplified home composition (fixed brew stat bar + setup buttons) plus a home-screen **ratio quick-select** (Ristretto / Normale / Lungo, editable). **Default OFF** — the zone-driven layout is byte-for-byte unchanged unless enabled in **Settings → Layout → "Use simplified home screen"**.

## What's in it
- New `simplifiedHome` theme setting (default `false`) + a toggle in Settings → Layout.
- When on: a fixed brew stat bar (profile / scale / ratio / beans / milk) and three setup buttons.
- **Ratio quick-select** dialog with editable presets; a tap sets **only** `lastUsedRatio` (it deliberately does not poke the persistent profile target / `brewYieldOverride`).
- Degrades gracefully without the weight-based-steaming PR (the milk cell simply shows "—").

## Independent & tested
- Diffs **only this feature** vs `main`.
- Builds on macOS; reviewed with the pr-review-toolkit; tested on-device.
- Ratio guidance/labels attributed to **La Marzocco**.

Replaces the stacked #1353, #1359.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -760,6 +760,7 @@ set(QML_FILES
     qml/components/BeanBaseDetailsRow.qml
     qml/components/BeanSummary.qml
     qml/components/ChangeBeansDialog.qml
+    qml/components/RatioPresetDialog.qml
     qml/components/BagCard.qml
     qml/components/SwitchEquipmentDialog.qml
     qml/components/EquipmentCard.qml
--- a/qml/components/BrewDialog.qml
+++ b/qml/components/BrewDialog.qml
@@ -791,6 +791,72 @@ Dialog {
                     onTextEdited: root.grindRpm = parseInt(text) || 0
                 }
             }
+
+            // --- Home-screen ratio quick-select presets. Rarely changed, so
+            // they sit at the bottom. These three values are what the
+            // home-screen ratio button offers as Ristretto / Normale / Lungo. ---
+            Rectangle {
+                Layout.fillWidth: true
+                Layout.topMargin: Theme.scaled(12)
+                Layout.preferredHeight: 1
+                color: Theme.borderColor
+            }
+            Text {
+                Layout.fillWidth: true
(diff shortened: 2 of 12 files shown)


loganross2001:
**Review completed** (per the request to review each change and iterate until clean), plus on-device testing.

Ran the `pr-review-toolkit` reviewers against this PR; fixed all real findings and re-reviewed clean:

- Corrected comment-accuracy issues (the `applyRatio`/Espresso-Setup behaviour and a dose-ding comment).
- **Pinned the simplified-home setup buttons** above the brew status bar — they were the last child of a clipped, fixed-height centre column, so an open preset row pushed them out of view. (Found and fixed via on-device testing.)

Builds on macOS; tested on-device. Independent, opt-in, default **off** (the zone-driven layout is byte-unchanged when off).

skialpine:
Superseded by #1368 (now merged to `main`).

Rather than a hardcoded `brewStatusBar` behind a `simplifiedHome` flag, #1368 reworks the same idea into the existing layout engine, so the bar is fully composable from the in-app and web layout editors with **zero impact** on anyone who doesn't opt in:

- a new optional, location-named `lowerMidBar` zone (empty by default, collapses when empty, height-gated);
- the missing readout widgets — `profileName`, `doseWeight`, `milkWeight`, and an interactive `ratioQuickSelect` (ports your `RatioPresetDialog`);
- per-zone options for **all** zones (distribution incl. equal-width, alignment, theme-driven style presets including an `accentBar` that matches this PR's look), plus one-tap **Fill with Brew bar** / **Clear zone**;
- the existing Scale widget gains a per-instance data mode (gross / net-beans / net-milk / context-aware) instead of a parallel widget.

So you get this PR's exact look (accent fill + equal cells, one tap), but customizable and without removing the configurable layout for other users. Thanks for the original — it's the basis for the reworked version. Closing in favor of #1368.

PR #1363, opened 2026-06-21
Weight-based steaming: milk auto-capture + steam UX + last-steam baseline

loganross2001:
## Summary
Auto-captures the milk weight from a BLE scale and scales steam time to it, built on the virtual-zero `StableWeightCapture` from #1358. **Opt-out** in Settings; manual entry always remains.

## What's in it
- **Milk auto-capture**, symmetric with bean capture: fires only when an empty-pitcher weight is saved (dormant otherwise). Records the milk weight when the pitcher rests on the scale.
- **Weight-scaled steam time**: scaled proportionally from a per-pitcher reference (milk weight ↔ duration), clamped to [5, 120] s.
- **Steam-page UX**: pitcher-weigh + reference-milk editor, with a confirmation guard against accidentally weighing a non-empty pitcher.
- **Last-steam baseline**: the (milk, time) pair from the last session is saved atomically at session end; "use as baseline" adopts it in steam setup.
- Pitcher weight + calibration round-trip through settings backup/restore.

## Safe by default
- Auto-capture is disableable (`milkAutoCaptureEnabled`); when off, steam time is never changed automatically.

## Independent & tested
- Diffs **only this feature** vs `main`.
- A pr-review-toolkit pass caught and fixed 3 real serializer-import bugs (calibration written onto disabled presets; unverified append index; selection applied before the preset list was rebuilt).
- Builds on macOS; reviewed; tested on-device.
- Weight-timed steaming concept credited to **DSx2 / Damian-AU** (https://github.com/Damian-AU/DSx2); implementation original.

Replaces the stacked #1351, #1352, #1360.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1365, opened 2026-06-21
Weight-timed steaming (polished): milk auto-capture, scale steam time to milk weight

skialpine:
Scales the DE1 steam auto-stop time to the milk weight on the scale, so a half-full and a full pitcher both stop at the right temperature. **Converged version** — combines @loganross2001's feature + fixes (#1363) with the review polish from this branch. Concept credited to **DSx2 / Damian-AU**. Supersedes #1363.

## Feature (from #1363, @loganross2001)
- Virtual-zero milk auto-capture; steam time = `clamp(round(duration × milk / referenceMilk), 5, 120) s`.
- Per-pitcher reference (milk weight ↔ duration); empty-pitcher tare via `pitcherWeightG`.
- Calibrate from a real pour ("Use as baseline", `Last: Xg → Ys`); opt-out toggle; backup/restore round-trip.
- **Two correctness fixes carried in (cherry-picked, attribution preserved):**
  - Apply the scaled time **at steam-start** via `setSteamTimeoutImmediate` — fixes the case where the loaded pitcher rests on the scale the whole time so the settle-capture never fires.
  - **Home-screen flow** scaling survives the page change (recovered from `sessionMeasuredMilkG`).

## Polish / convergence (this branch)
- **Centralized the scaling math into C++** — `SettingsBrew::scaledSteamTime()` + `netMilkForPitcher()` are the single source of truth; all former QML copies (SteamPage ×2, IdlePage capture + pill-tap) call them. Bounds, clamp, and the toggle gate live in one place.
- **The toggle truly disables the feature** (not just auto-capture): `scaledSteamTime()` returns 0 when off, so every path falls back to fixed duration. Calibration is retained → flip back on resumes instantly. Relabeled **"Weight-timed steaming"** (storage key unchanged, no migration).
- **Merged the two pitcher-weight controls into one "Milk pitcher" row** (editable field + confirm-guarded Weigh + Tare/live/Clear) and **relocated it under Steam Flow**, above the weight-timed rows. One consistent net-milk rule (requires a saved pitcher weight).
- **Manual ±5 survives auto-capture** — `steamTimeoutUserAdjusted` latch (set on ±5; cleared on lift-off-scale, pitcher change, or session end); also honored at steam-start.
- **Fixed double milk-capture** — IdlePage capture gated on `StackView.status === Active` so it can't fire alongside SteamPage's copy.
- **Hardening** — milk floor `minNet: 50` (a bean cup can't trip milk capture); guard a 0/missing duration; Purge buttons get keyboard activation; tab order skips the hidden Tare when no scale; stale-comment + orphaned-doc-comment cleanups.

## Testing
- Builds clean on macOS via Qt Creator incl. the full test suite (0 errors, 0 warnings).
- Editor order: Duration → Steam Flow → Milk pitcher → Reference milk → Baseline → Expected time → Auto-capture → Temperature.


skialpine:
### Code review

Found 4 issues:

1. Settings restore silently re-enables weight-timed steaming for users who turned it off. `importFromJson` applies `milkAutoCaptureEnabled` *before* the preset loop, then `setSteamPitcherCalibration` auto-enables it (`if (calibMilkG > 0) setMilkAutoCaptureEnabled(true)` in settings_brew.cpp), so a backup with the toggle OFF + calibrated presets restores with the feature back ON — against the saved preference. Flagged independently by 3 reviewers. Fix: apply the toggle value after the preset loop so the serialized value wins.

https://github.com/Kulitorum/Decenza/blob/f97e3751443059c97f04851773c2b5a8fd2c0114/src/core/settingsserializer.cpp#L421-L425

(side effect at https://github.com/Kulitorum/Decenza/blob/f97e3751443059c97f04851773c2b5a8fd2c0114/src/core/settings_brew.cpp#L381-L385, calibration import at L448-L451)

2. Stale comment: `netMilkForPitcher` doc says "or the raw reading if no tare is saved", but the implementation now returns 0 when no empty-pitcher weight is saved (one consistent net-milk rule).

https://github.com/Kulitorum/Decenza/blob/f97e3751443059c97f04851773c2b5a8fd2c0114/src/core/settings_brew.h#L136-L140

3. Stale comment: the milk-capture comment says `milkAutoCaptureEnabled` is "default on", but the default is now `false` (off; calibrating opts in). Same wording in IdlePage.qml L227.

https://github.com/Kulitorum/Decenza/blob/f97e3751443059c97f04851773c2b5a8fd2c0114/qml/pages/SteamPage.qml#L1957-L1961

4. Keyboard nav: `refMilkInput.backtab` targets `savePitcherWeightBtn`, which is hidden when no scale is connected (its row is `visible: ScaleDevice.connected && !ScaleDevice.isFlowScale`). The forward path is guarded `(scale) ? tareBtn : refMilkInput` but the backtab isn't, so Shift+Tab lands on an invisible element.

https://github.com/Kulitorum/Decenza/blob/f97e3751443059c97f04851773c2b5a8fd2c0114/qml/pages/SteamPage.qml#L1810-L1813

---

<details>

- [Worth a look] Manual ±5 override can be lost at steam-start if the pitcher is still on the scale — `onLoadPresentChanged` may clear `steamTimeoutUserAdjusted` before `onIsSteamingChanged` reads it (same-transition race). (score 60)
- [Worth a look] CLAUDE.md: serializer stores `.size()` results in `const int` instead of `qsizetype` (the `static_cast<int>` suppresses the warning, but the rule is explicit). (score 55)
- [Worth a look] CLAUDE.md: the new section title/summary use `TranslationManager.translate(...)` inside `Text` instead of the `Tr` component, so they won't re-evaluate on a live locale change. (score 55)
- [Worth a look] No unit tests for the new `SettingsBrew` fields / `scaledSteamTime` branches (the round-trip test pattern PR #1358 established). (score 50)
- [OK to skip] `steamTimeoutUserAdjusted` comment omits the session-end clear point. (score 50)

</details>

PR #1369, opened 2026-06-21
Composable status bar: per-instance display modes; merge status widgets

skialpine:
## Summary

Reworks the "compact status bar" idea (PR #1362) **into the existing layout engine** — composable, no `compactStatusBar` flag, opt-in for everyone — and folds in related ratio/widget cleanups. Continues the #1368 pattern; built on `main` and supersedes #1362.

## What's in it

**Status bar joins the zone system.** `StatusBar.qml` now renders via `LayoutBarZone`, so the top bar honors the same per-zone options (distribution / alignment / style) as the bottom bars — it was the last special-cased bar. `LayoutBarZone` packed mode now fills the row when a spacer is present, preserving the default status-bar layout.

**Per-instance display mode** (`text` / `icon`) on `machineStatus`, `temperature`, `steamTemperature`, `scaleWeight` — icon-led readouts, editable via long-press (`DisplayModeEditorPopup`; `scaleWeight` keeps its `dataMode`). Works in any zone, not just a styled one. No new icons (existing assets) and no new widgets.

**Sleep widget options** (per-instance): `allowQuit` (remove the long-press-to-quit) and `showIcon` (default on), via `SleepEditorPopup` and web inline toggles.

**"Compact status bar" populate preset** (in-app + web) reproduces #1362's look in one tap. "Brew bar" preset is hidden on the `statusBar` zone (irrelevant for an all-pages bar).

**Per-zone "Reset to default"** alongside "Clear zone" (`resetZoneToDefault`).

**Merge `connectionStatus` into `machineStatus`** — they overlapped (machine status already shows "Disconnected" when offline). Alias + migration + default-layout update; `ConnectionStatusItem.qml` deleted. (`vital_stats.h` already treated them identically.)

**Ratio consistency fixes:**
- The `ratioQuickSelect` pill and the ratio chooser now show the **live active ratio** (`target / dose`), matching the scale widget and Brew Settings — fixes a mismatch where the pill showed `lastUsedRatio` (last preset tapped) while a bean's saved yield override made the real ratio different.
- Brew Settings shows the ratio form (`1:X.X`) next to the multiplier, so it's clear `1.8` == `1:1.8`.

## Status

Builds clean (Qt 6.11.1, 0 errors / 0 warnings). All new options default to current behaviour. On-device QA pending. One behavior change to note: the default status bar previously showed Online/Offline (`connectionStatus`); it now shows the machine phase (and "Disconnected" when offline) via the merged `machineStatus`.

OpenSpec change: `openspec/changes/composable-status-bar/` (to be archived before merge).


skialpine:
### Code review

Ran a multi-agent review (CLAUDE.md compliance, deep bug scan, git-history, cross-PR context, comments). Found 5 issues; all addressed in commit 8220c9f471.

1. **connectionStatus→machineStatus migration was applied in-memory but never persisted** (flagged by 3 agents) — `connMigrated` was missing from the save guard, so the rename re-ran on every launch instead of writing once.

https://github.com/Kulitorum/Decenza/blob/8220c9f471934003f34c19c82f0e8faabf0ec22c/src/core/settings_network.cpp#L427-L432

2. **"Compact status bar" populate didn't reset zone options** — it relies on `packed` distribution + spacers to centre Sleep, so applying it after the "Brew bar" preset (which sets `equalWidth`) silently broke the layout. Now resets `distribution`/`alignment`/`style` (in-app + web).

https://github.com/Kulitorum/Decenza/blob/8220c9f471934003f34c19c82f0e8faabf0ec22c/qml/components/layout/ZoneOptionsPopup.qml#L60-L68

3. **"Compact status bar" button was offered on every zone** though it injects status readouts — now restricted to the `statusBar` zone (mirrors the Brew-bar guard).

https://github.com/Kulitorum/Decenza/blob/8220c9f471934003f34c19c82f0e8faabf0ec22c/qml/components/layout/ZoneOptionsPopup.qml#L207-L211

4. **`int` loop counter over `QJsonArray::size()`** in the new migration (CLAUDE.md: *"Use `qsizetype` (not `int`) for container sizes … Assigning to `int` causes `-Wshorten-64-to-32` warnings"*).

https://github.com/Kulitorum/Decenza/blob/8220c9f471934003f34c19c82f0e8faabf0ec22c/src/core/settings_network.cpp#L366-L372

5. **Both `StyledSwitch`es in the Sleep editor lacked `accessibleName`** (announced as "Toggle"). CLAUDE.md / ACCESSIBILITY.md require an accessible name on interactive elements; ACCESSIBILITY.md also says to fix violations in files you touch.

https://github.com/Kulitorum/Decenza/blob/8220c9f471934003f34c19c82f0e8faabf0ec22c/qml/components/layout/SleepEditorPopup.qml#L74-L76

<details>
<summary>3 below-threshold notes (50-74)</summary>

- [OK to skip] Comment wording: "subsumes the old Online/Offline widget" overstates the merge in full mode (scale connection not shown) — compact-bar context is accurate (score: 55)
- [OK to skip] `ScaleWeightItem` comment "status dot + value" — the dot has no phase semantics there (score: 50)
- [OK to skip] Ratio display vs apply use different dose property names, but both resolve to `dyeBeanWeight` (same value) — no real divergence (score: 50)

</details>

PR #1375, opened 2026-06-21
Add Clock widget (time readout) with text/icon modes

skialpine:
Adds a first-class **Clock** widget to the home-screen layout system — a current-time readout usable in the status bar or any zone, with the same text/icon display modes as the other readouts.

## What's included
- **`ClockItem.qml`** — shows the current time, updates each second, and respects the 12/24-hour setting (`Settings.app.use12HourTime`). Uses the **same** format string as the `%TIME%` custom variable (`h:mmap` / `hh:mm`); the Flip Clock screensaver reads the same setting, so all three stay consistent.
- **Per-instance display mode** — `text` (default) or `icon` (a clock glyph ahead of the time, new `clock.svg`), so it looks good on the status line and in center zones.
- **Registered everywhere** per the widget checklist: `CMakeLists.txt`, `LayoutItemDelegate` routing, in-app picker (Readouts) + chip-label map, web `WIDGET_TYPES` + `DISPLAY_NAMES`, `LayoutCenterZone` auto-size list, `resources.qrc`.
- **Configurable in both editors** — added to `typeHasOptions` (C++ + the web JS mirror) and the display-mode routing (in-app gear → Text/Icon; web inline selector).
- **Default layout** — clock (icon mode) added to the far right of the status bar. Existing saved layouts are untouched; users there add it from Readouts → Clock.

## Testing
- Builds clean (Qt 6.11.1 macOS): C++ recompiled, `ClockItem` through qmlcachegen, app + tests link, 0 warnings.
- `tst_settings`: `typeHasOptions` allowlist test updated to include `clock`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -846,6 +846,7 @@ set(QML_FILES
     qml/components/layout/items/BatteryLevelItem.qml
     qml/components/layout/items/ScaleBatteryItem.qml
     qml/components/layout/items/MiniGHCItem.qml
+    qml/components/layout/items/ClockItem.qml
     qml/components/layout/CustomEditorPopup.qml
     qml/components/layout/ZoneOptionsPopup.qml
     qml/components/layout/ScaleWeightEditorPopup.qml
--- a/qml/components/layout/LayoutCenterZone.qml
+++ b/qml/components/layout/LayoutCenterZone.qml
@@ -26,6 +26,7 @@ Item {
             case "scaleWeight":
             case "weather":
             case "ghcSimulator":
+            case "clock":
                 return true
             default:
                 return false
--- a/qml/components/layout/LayoutItemDelegate.qml
+++ b/qml/components/layout/LayoutItemDelegate.qml
@@ -177,6 +177,7 @@ Item {
                 case "settings":         src = "items/SettingsItem.qml"; break
                 case "temperature":      src = "items/TemperatureItem.qml"; break
                 case "waterLevel":       src = "items/WaterLevelItem.qml"; break
+                case "clock":            src = "items/ClockItem.qml"; break
                 // connectionStatus merged into machineStatus (alias): the machine
                 // status shows the phase and "Disconnected" when offline, replacing
                 // the old binary Online/Offline widget.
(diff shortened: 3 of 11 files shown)


skialpine:
### Code review

Ran a multi-agent review (CLAUDE.md compliance, deep bug scan, historical context, prior-PR patterns, comment accuracy) plus confidence scoring. CLAUDE.md compliance is clean and the 4-place widget registration is complete. Found 3 minor real issues, all fixed in a8cb1af33be965883e1c17208fb89a7800a3bfb6:

1. Placed chip for the new clock and the Flip Clock both displayed "Clock" in the editor (indistinguishable). The clock chip now reads "Time" via a descriptive key (also replaces the opaque `chipClock2` key); web placed-chip matched.

https://github.com/Kulitorum/Decenza/blob/a8cb1af33be965883e1c17208fb89a7800a3bfb6/qml/pages/settings/LayoutEditorZone.qml#L693-L695

2. `ClockItem` header comment said the icon is "ahead of the time" — only true in bar mode (it sits above the time in center zones). Reworded.

3. 24-hour format now uses explicit `HH:mm` (equivalent to `hh:mm` without an am/pm specifier, but unambiguous and matches the other dedicated clocks).

https://github.com/Kulitorum/Decenza/blob/a8cb1af33be965883e1c17208fb89a7800a3bfb6/qml/components/layout/items/ClockItem.qml#L23-L25

Two items that scored high but are **not** defects on inspection (no change made):

- "`hh:mm` renders 12-hour" — false positive: Qt's `hh` is 12-hour only when an am/pm specifier is present; without one it's 00–23. The shipped `%TIME%` helper uses the same pattern. (Switched to `HH:mm` anyway for clarity, above.)
- "Removing a clock always prompts confirmation" — intentional and consistent: every configurable readout (temperature, waterLevel, scaleWeight, machineStatus) already prompts on removal by design. `itemIsConfigured` returns true for any type that has options.

https://github.com/Kulitorum/Decenza/blob/a8cb1af33be965883e1c17208fb89a7800a3bfb6/src/core/settings_network.cpp#L615-L627

Below threshold (discarded): existing-user migration (default-layout changes intentionally apply only to fresh installs / Reset — consistent with #1372), timer not synced to the wall-clock second boundary (cosmetic), and `use12HourTime` reactivity (the property is CONSTANT).

PR #1393, opened 2026-06-25
Home action buttons: highlight the selected mode + softer selection outline

loganross2001:
## Summary

On the home screen, the large Espresso / Steam / Flush / Hot-Water action buttons give **no visual cue for which mode is currently selected** (i.e. which preset row is expanded). This PR adds a clear "selected" highlight to the active action button and to the selected preset pill, using Theme tokens only. It also removes a hardcoded grey and an inconsistent "gold" state on the espresso button.

## Why

When you tap Espresso (or Steam/Flush), its preset row expands but the button itself looks identical to the others, so there's no at-a-glance indication of which screen/mode you're in. This is the fix.

## Changes & reasoning

**`CustomItem.qml` — the full/centre-mode action-button renderer (the core fix)**
- In centre zones, action buttons are compiled to `CustomItem` (not `EspressoItem` etc., which only render in *compact/bar* mode). `CustomItem` *wrote* `activePresetFunction` on tap but never *read it back*, so it had no concept of "selected." That's why the button never highlighted.
- Added: a parent-walk to the `IdlePage`, a `_toggleMode` parsed from the button's own `togglePreset:<mode>` action, and an `isActive` flag.
- **Reactivity:** reading `idlePage.activePresetFunction` *through* the `var idlePage` intermediate does **not** register a QML binding dependency (it would evaluate once and never update). So the value is mirrored into a local `string _activeFn` via `Connections { onActivePresetFunctionChanged }`, and `isActive` binds to that.
- When `isActive`, a `Theme.scaled(3)` accent ring is drawn on the button's background Rectangle (both compact and full render paths).
- The action-button default background was a hardcoded `#555555` grey — changed to `Theme.primaryColor` (themed; removes a hardcoded colour per the Theme-token rule).

**`PresetPillRow.qml` — selected preset pill**
- The selected pill already got a `primaryColor` fill, but in a light theme that barely stands out. Added the same accent ring on the selected pill so it reads clearly as selected.

**Selection-outline colour**
- The ring is a **darker shade of the button's own colour** (`Qt.darker(...)`), not a saturated accent — it reads as "selected" without being harsh, and stays coordinated with any theme.

**`ActionButton.qml` + `EspressoItem/SteamItem/FlushItem/HotWaterItem.qml`**
- The same active-mode highlight for the **compact (bar)** rendering path, for consistency wherever the buttons appear.

**`LayoutItemDelegate.qml` — action-button colours**
- The espresso button previously turned `Theme.highlightColor` (a "gold" nudge) when no favourite profile was selected, which looked like an error state against the other buttons. It now stays `Theme.primaryColor` like the rest. Sleep/Quit and other compiled buttons use themed colours.

## Testing

On-device (Android tablet): tapping each action button shows an accent ring while its preset row is open; selecting another mode / clearing moves or removes the ring. Selected preset pill shows the ring. Verified in both light and dark themes.

## Notes for the reviewer

- No hardcoded colours or sizes are introduced; this PR also *removes* the pre-existing `#555555` literal.
- `LayoutItemDelegate.qml` and `PresetPillRow.qml` are also touched — in different sections (widget registration / pitcher labels) — by the "UI presentation + steam plan" PR. The two PRs are independent (no dependency / no stacking); merging the second is an ordinary file-overlap resolution.

### DIFF
--- a/qml/Theme.qml
+++ b/qml/Theme.qml
@@ -275,6 +275,13 @@ QtObject {
         return value
     }
 
+    // Dark-vs-light mode of the active theme. Mirror of Settings.theme.isDarkMode:
+    // the canonical spelling is Settings.theme.isDarkMode (CLAUDE.md settings rule),
+    // but a bare `Theme.isDarkMode` read otherwise resolves to undefined — falsy —
+    // and fails silently, which has already caused real bugs. The mirror makes
+    // either spelling correct.
+    readonly property bool isDarkMode: Settings.theme.isDarkMode
+
     // Dynamic colors - bind to Settings with fallback defaults
     // Wrapped in _c() for flash-to-identify from web theme editor
     property color backgroundColor: _c("backgroundColor", Settings.theme.customThemeColors.backgroundColor || "#1a1a2e")
--- a/qml/components/ActionButton.qml
+++ b/qml/components/ActionButton.qml
@@ -11,12 +11,6 @@ Button {
     property color backgroundColor: Theme.primaryColor
     property int iconSize: Theme.scaled(48)
 
-    // When true, draw a ring to mark this button as the currently-active
-    // mode/screen. A darker shade of the button (not accent) — matches the
-    // CustomItem / PresetPillRow selection outline.
-    property bool active: false
-    property color activeColor: Qt.darker(control._effectiveBackground, 1.5)
-
     // Translation support
     property string translationKey: ""
     property string translationFallback: ""
@@ -48,12 +42,8 @@ Button {
     activeFocusOnTab: true
 
     // Accessibility on the Button itself (not delegated to a child).
-    // Append "selected" when this button is the active mode, so a screen-reader
-    // user re-navigating to it hears the state (the visual ring alone isn't spoken).
-    // Only appended when active, so non-mode ActionButtons are unaffected.
     Accessible.role: Accessible.Button
     Accessible.name: control.text + ". " + control._computedAccessibleDescription
-                     + (control.active ? ", " + TranslationManager.translate("accessibility.selected", "selected") : "")
     Accessible.focusable: true
     Accessible.onPressAction: {
         if (control.enabled) {
@@ -115,16 +105,6 @@ Button {
             if (control._isPressed) return Qt.darker(control._effectiveBackground, 1.2)
             if (control.hovered || control.activeFocus) return Qt.lighter(control._effectiveBackground, 1.1)
             return control._effectiveBackground
-        }
-
-        // Active-mode indicator: a thick darker-shade ring around the button.
-        Rectangle {
-            anchors.fill: parent
-            visible: control.active
-            color: "transparent"
-            border.width: Theme.scaled(3)
-            border.color: control.activeColor
-            radius: parent.radius
         }
 
         // Focus indicator
--- a/qml/components/ActionButton.qml
+++ b/qml/components/ActionButton.qml
@@ -11,6 +11,12 @@ Button {
     property color backgroundColor: Theme.primaryColor
     property int iconSize: Theme.scaled(48)
 
+    // When true, draw a ring to mark this button as the currently-active
+    // mode/screen. A darker shade of the button (not accent) — matches the
+    // CustomItem / PresetPillRow selection outline.
+    property bool active: false
+    property color activeColor: Qt.darker(control._effectiveBackground, 1.5)
+
     // Translation support
     property string translationKey: ""
     property string translationFallback: ""
@@ -41,9 +47,13 @@ Button {
     focusPolicy: Qt.StrongFocus
     activeFocusOnTab: true
 
-    // Accessibility on the Button itself (not delegated to a child)
+    // Accessibility on the Button itself (not delegated to a child).
+    // Append "selected" when this button is the active mode, so a screen-reader
+    // user re-navigating to it hears the state (the visual ring alone isn't spoken).
+    // Only appended when active, so non-mode ActionButtons are unaffected.
     Accessible.role: Accessible.Button
     Accessible.name: control.text + ". " + control._computedAccessibleDescription
+                     + (control.active ? ", " + TranslationManager.translate("accessibility.selected", "selected") : "")
     Accessible.focusable: true
     Accessible.onPressAction: {
         if (control.enabled) {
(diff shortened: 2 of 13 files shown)


loganross2001:

loganross2001:
Ran `pr-review-toolkit` (code review, opus) on this branch, fixed the findings, and re-reviewed to clean. Summary so you don't have to spend tokens re-reviewing:

### Fixed
- **The active ring was invisible** — `ActionButton.active` drew a ring in `Theme.accentColor` *over* an accent-colored fill, so it never showed. Reworked to a **darker-shade ring** (`Qt.darker(_effectiveBackground, 1.5)`) on a `Theme.primaryColor` fill, matching the `CustomItem`/`PresetPillRow` selection outline. The `active` flag now does something visible, and the four items + CustomItem + pills use one consistent selection vocabulary.
- **Screen-reader "selected" state** — `CustomItem` (center-zone renderer) now appends ", selected" to its `Accessible.name` when active (via the existing `accessibility.selected` key), so VoiceOver/TalkBack announce the active mode on re-navigation, matching `ActionButton`.
- **Comment accuracy** — corrected three stale comments (the removed "no-favorite gold" reference; "accent ring" → "darker ring" in two inline spots; "compact-mode EspressoItem" → "dedicated per-mode items").

### Intentional / deferred (not bugs)
- **Espresso "no favorite selected" gold nudge removed** — this is a deliberate de-gold (the button stays `Theme.primaryColor`); the non-favorite pill in the profile popup still cues "pick a favorite". The beans tile keeps its no-inventory gold, so there's an intentional asymmetry — happy to de-gold beans too for consistency if you'd prefer.
- **Beans/equipment-mode highlighting and compact-bar (non-CustomItem) a11y "selected"** are scoped out: the center zone renders all modes via `CustomItem` (covered), and the compact bar signals active via accent icon/text rather than a ring — a second active-state vocabulary that only appears in a different zone.

Verified: re-review reports no must-fix, and the macOS build passes.

loganross2001:
Ran the full `pr-review-toolkit:review-pr` suite (code / silent-failure / comments / types / tests) on this PR and iterated to convergence. It caught a few real things, now fixed:

- **A11y gap:** the compact bar buttons highlighted visually but never announced ", selected" to TalkBack/VoiceOver (the center/CustomItem path did) — added, reusing the existing `accessibility.selected` key.
- **Dead code:** an `ActionButton` active-ring + four `fullContent active:` bindings that never render — the per-mode items only load in *compact* zones; a full-size instance compiles to `CustomItem` (`LayoutItemDelegate` `isCompiled`), so that `fullContent` is unreachable. Removed.
- **`ignoreUnknownSignals: true`** on the 5 highlight `Connections` — bought nothing today (the `activePresetFunction` signal always exists) but would have *silently frozen every highlight* if that property were ever renamed. Dropped, so a rename now warns loudly.
- **Silent no-op → warn:** the `togglePreset` parent-walk for IdlePage now `console.warn`s on a miss instead of doing nothing.
- Comment fix (yield vs "ratio").

Both bug-finders' final pass reports converged/clean.

skialpine:
Maintainer push: rebased onto current `main` and added three commits on top of your work after an on-device review pass (macOS, both themes):

**c52aeb8 — review fixes to the highlight feature**
- **Dark-theme fix (the big one):** the `Qt.darker` ring was invisible on the default dark theme — a darker edge on a bright fill against a near-black page reads as anti-aliasing (verified on-device, and quantifiable: fill `#4e85f4`, ring `#3459a3`, page `#1a1a2e`). The ring now flips by theme: `Settings.theme.isDarkMode ? Qt.lighter(fill, 1.6) : Qt.darker(fill, 1.5)`. Same treatment on the selected preset pill, derived from the pill's own colour. Note: `Theme.isDarkMode` doesn't exist on the QML singleton (silently `undefined`) — the property lives on `Settings.theme`.
- Replaced the five-file `Connections` mirror with plain declarative bindings — reads through a QObject-valued `var` property *are* reactive; verified working on-device. The comments claiming otherwise are corrected (they would have taught future maintainers a false QML rule).
- `togglePreset` handler now reuses the resolved `idlePage` property instead of duplicating the parent walk.
- Compact `CustomItem` now fills with `_effectiveBackground` (removes the last hardcoded `#555555` here; compact brew-settings widgets get the override highlight like full mode).
- `LayoutItemDelegate`: the empty `Loader.Error` branch now warns with widget type/id instead of swallowing load failures.
- Dropped the comment-only `ActionButton.qml` hunk.

**df9b109 / 7e65095 — unrelated fixes bundled by maintainer decision** (same `Theme.isDarkMode` bug class in `ScaleBatteryItem`, and a `ShotPlanText` width binding loop from recent main).

Known-and-accepted from review, not addressed here: a `togglePreset` custom button placed in the status bar (outside IdlePage) is a dead button that only `console.warn`s.

Thanks for the feature — the selected-mode gap was real, and the a11y/i18n work was solid.

skialpine:
One follow-up (2f84875a): restored the gold no-favorite espresso cue — maintainer call. It's the only 'no favorite selected' signal visible without expanding the preset row, and it composes fine with the new active ring (the ring derives from the effective fill, gold included). The PR no longer changes that behavior. Also since the last comment: 0bfbf77e makes a togglePreset custom widget work from the persistent status bar by falling back to pageStack.currentItem when there's no IdlePage ancestor.

skialpine:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance (5 independent review passes: CLAUDE.md/QML-gotchas/accessibility audit, deep bug scan, git-history regression check against #1425/#1426/#1428/#1429, prior-PR feedback check, and comment accuracy). Verified in particular: the six deleted full-mode blocks were unreachable (`isCompiled` routing in [LayoutItemDelegate.qml](https://github.com/Kulitorum/Decenza/blob/e181489750e052c30f4c50697302c21745e39218/qml/components/layout/LayoutItemDelegate.qml#L31-L52)) with no dangling references, all new property accesses resolve, the `ShotPlanText` measurer fix is acyclic and repairs a binding loop introduced by #1426, and no recent-history work (steam pill suffixes, stacked shot plan, GHC start-guards) is regressed.

---

<details>
<summary>2 below-threshold issues (0-74)</summary>

- [Fixed in e1814897] Over-broad comment: CustomItem's active-ring comment implied custom widgets commonly carry togglePreset actions, but neither the in-app nor the web widget editor exposes them — reworded for precision (score: 55) — https://github.com/Kulitorum/Decenza/blob/e181489750e052c30f4c50697302c21745e39218/qml/components/layout/items/CustomItem.qml#L57-L63
- [OK to skip] Full-size compiled preset buttons announce a generic "long-press/double-tap" accessibility hint rather than the per-button wording that existed in the deleted (already-unreachable) full-mode blocks — pre-existing, no user-audible change, and ACCESSIBILITY.md requires a description to exist, not specific wording (score: 25)

</details>

PR #1394, opened 2026-06-25
Weight widgets: live net weight that holds the captured value + two scale data modes

loganross2001:
## Summary

The Dose and Milk weight widgets show a single static number. This PR makes them show the **live net weight while a load is on the scale, then "hold" the value captured when the scale settles** (the beep) — and adds two new data modes to the Scale Weight widget.

## Why

When dosing beans or weighing a milk pitcher, it's useful to watch the live net weight build up, and then have the widget settle on the captured value rather than jumping back to a stored number. This mirrors how the scale-capture flow already behaves elsewhere, but surfaces it on the home-screen widgets.

## Changes & reasoning

**`DoseWeightItem.qml` / `MilkWeightItem.qml` — live → hold**
- While a load is present on the scale, the widget shows the **live net weight** (scale reading minus the saved cup/pitcher tare).
- When the weight stabilises (the capture / "beep"), it **freezes** that captured value and holds it, falling back to the last set/measured value when the scale is empty.
- **Event-based, no timers:** the live/hold transitions are driven by `Connections` on `MachineState.onScaleWeightChanged` / `Settings.dye.onDyeBeanWeightChanged` with boolean/value flags — not a polling or guard timer (per the project's "no timers as guards" rule). A timer would be a fragile heuristic; the capture is a real event.

**`ScaleWeightItem.qml` + `ScaleWeightEditorPopup.qml` — two new data modes**
- `beansHold` — "beans, live then hold the captured dose": same live→hold behaviour as a configurable Scale Weight data mode.
- `expectedYield` — "expected output (dose × ratio)": shows `ProfileManager.targetWeight`, i.e. the target the shot will stop at, rather than a live reading. The `g~` ("approximate live reading") suffix is suppressed for this mode since it's a computed target, not a scale reading.
- Both modes are added to the on-device widget editor (`ScaleWeightEditorPopup`) and the web layout-editor data-mode list (`shotserver_layout.cpp`) so QML and web editors stay in parity.

## Testing

On-device: place a dose cup / milk pitcher → the widget tracks the live net weight → on settle it holds the captured value → lifting the load and re-dosing goes live again. The two new Scale Weight data modes selectable in the layout editor behave as described.

## Notes for the reviewer

- No timers used as guards; no hardcoded colours/sizes; new editor strings are internationalised via `TranslationManager.translate(...)`.
- The re-dose "scale empty" detection uses a small bare weight threshold (`< 1.0 g`) — a physical threshold, not styling.

### DIFF
--- a/qml/components/layout/ScaleWeightEditorPopup.qml
+++ b/qml/components/layout/ScaleWeightEditorPopup.qml
@@ -46,7 +46,9 @@ Dialog {
         { value: "gross",        label: TranslationManager.translate("layoutEditor.scaleGross", "Gross weight") },
         { value: "netBeans",     label: TranslationManager.translate("layoutEditor.scaleNetBeans", "Net beans (minus dose tare)") },
         { value: "netMilk",      label: TranslationManager.translate("layoutEditor.scaleNetMilk", "Net milk (minus pitcher)") },
-        { value: "contextAware", label: TranslationManager.translate("layoutEditor.scaleContext", "Context-aware (milk while steaming, else beans)") }
+        { value: "contextAware", label: TranslationManager.translate("layoutEditor.scaleContext", "Context-aware (milk while steaming, else beans)") },
+        { value: "beansHold",    label: TranslationManager.translate("layoutEditor.scaleBeansHold", "Beans (live, then hold captured dose)") },
+        { value: "expectedYield", label: TranslationManager.translate("layoutEditor.scaleExpectedYield", "Expected output (dose × ratio)") }
     ]
 
     modal: true
--- a/qml/components/layout/ScaleWeightEditorPopup.qml
+++ b/qml/components/layout/ScaleWeightEditorPopup.qml
@@ -48,7 +48,7 @@ Dialog {
         { value: "netMilk",      label: TranslationManager.translate("layoutEditor.scaleNetMilk", "Net milk (minus pitcher)") },
         { value: "contextAware", label: TranslationManager.translate("layoutEditor.scaleContext", "Context-aware (milk while steaming, else beans)") },
         { value: "beansHold",    label: TranslationManager.translate("layoutEditor.scaleBeansHold", "Beans (live, then hold captured dose)") },
-        { value: "expectedYield", label: TranslationManager.translate("layoutEditor.scaleExpectedYield", "Expected output (dose × ratio)") }
+        { value: "expectedYield", label: TranslationManager.translate("layoutEditor.scaleExpectedYield", "Expected output (target)") }
     ]
 
     modal: true
--- a/qml/components/layout/items/DoseWeightItem.qml
+++ b/qml/components/layout/items/DoseWeightItem.qml
@@ -27,8 +27,13 @@ Item {
     // (the _operating phase gate alone doesn't cover the return-to-Idle window).
     property bool _postExtraction: false
     function _liveNet() { return Math.max(0, MachineState.scaleWeight - Settings.brew.doseCupTareWeight) }
+    // Live net-bean display is valid only with a saved dose-cup tare (else scale − 0 = gross, mislabelled
+    // "beans") AND when the net sits in a plausible dose range — a brew cup's net (≫ any real dose) must not
+    // read as beans. When either fails, valueText falls through to the recorded dyeBeanWeight.
     readonly property bool _loaded: root.scaleConnected
+        && Settings.brew.doseCupTareWeight > 0
         && MachineState.scaleWeight > (Settings.brew.doseCupTareWeight + 0.3)
+        && _liveNet() <= 55
 
     // Live net-bean display only makes sense while dosing. During espresso preheat or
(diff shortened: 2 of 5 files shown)


loganross2001:

loganross2001:
Ran `pr-review-toolkit` (code review, opus) — it took a few rounds, and it caught real bugs plus a whole class of phantom readings. All fixed and re-reviewed to clean.

### Real bugs fixed
- **`DoseWeightItem` stale dose** — the captured dose (`_held`) was never cleared when the cup was lifted, so after any weigh it permanently shadowed the idle fallback (a later bean switch / AI dial / manual dose edit showed the old number). Now cleared on lift.
- **Phantom "Beans" readings** — the live "beans = scale − cup tare" path was ungated, so any cup on the scale (empty espresso cup, brew cup) read as beans. Closed the whole class across three windows: **during** an operation (preinfusion/pour/ending/steam/water/flush) via an `_operating` phase gate, during **espresso preheat** (brew cup goes on then), and — the subtle one — the **post-shot Idle window** (brew cup left on after Ending) via a `_postExtraction` latch that clears when the cup is lifted. Applied to both `DoseWeightItem` and `ScaleWeightItem`'s `beansHold` mode; `MilkWeightItem` was already phase-gated.

### Also
- Dropped the inaccurate "scale-independent" claim on the `expectedYield` doc, and refreshed the `DoseWeightItem`/`MilkWeightItem`/`ScaleWeightItem` header comments to match the new live→hold→fallback behavior.

- **Consistency:** the scale-tile `beansHold` mode now falls back to the recorded dose when the cup is off (matching the dedicated Beans tile and its "hold captured dose" label), instead of zeroing out on lift.

### Deferred (acknowledged, not bugs)
- `expectedYield` (a computed dose×ratio target) currently only displays when a scale is configured — showing it without a scale is a small follow-up.
- The `"Expected output (dose × ratio)"` editor label is only literally "dose × ratio" when brew-by-ratio is active (otherwise it's the manual target) — wording nit.

Verified: re-reviewed clean, macOS build passes. No timers used — the live→hold→fallback is entirely event/flag-based.

loganross2001:
Ran the full `pr-review-toolkit:review-pr` suite (code / silent-failure / comments / types / tests) on this PR and iterated to convergence over two rounds. It caught a few real behaviors worth fixing, now resolved:

- **Live "beans" could show the wrong number.** With no dose-cup tare saved, the live net-beans path showed the *gross* cup weight labeled as beans; and even with a tare, a **brew cup** (or any heavy vessel) placed on the scale pre-shot read as e.g. "Beans: 110 g". The live path (`DoseWeightItem._loaded` + `ScaleWeightItem.beansHold`) now requires a saved tare **and** a plausible-dose net (`≤ 55 g` — a real dose is ~5–30 g); otherwise it falls through to the recorded dose. Live capture/hold behavior is unchanged for a genuine dose.
- **`expectedYield` needed no scale but was hidden without one** (it's a computed `ProfileManager.targetWeight`). It now renders regardless of scale connection, while the live-weight modes still hide/warn on disconnect — and I fixed a follow-on where the "--" placeholder overlapped the value in that mode.
- **Wording:** the `expectedYield` "dose × ratio" label/comment over-claimed (`targetWeight` is the brew-yield override when set — dose × ratio only in brew-by-ratio mode — else the profile's fixed target); reworded, editor label → "Expected output (target)".

Both bug-finders' final pass reports converged/clean. No test added — the latch/mode logic is pure QML with no C++ QTest seam, and the sibling scale-modes are likewise untested.

One heads-up: this branch is based on the pre-#1396 `main`; happy to rebase it onto current `main` if that helps the merge.

skialpine:
Thanks for this PR — the ideas in it were genuinely useful, and all three have now landed in some form, so I'm closing this one rather than merging it directly. Disposition:

- **Beans live→hold** → landed reworked as #1429. Rather than computing `scaleWeight − tare` in the widgets, the Beans widget now binds to the existing `StableWeightCapture` engine on IdlePage (the virtual-zero auto-capture from #1358) via two window-root properties. That gives the same live→capture→hold UX with one source of truth: the widget, the panel readout, and the captured value are always the same number, the engine's arming rules (real scale, saved tare, espresso context, 5–45 g plausibility) replace the per-widget phase gates and thresholds, and its re-arm handles "added more beans after capture" without a hold latch.
- **`expectedYield` data mode** → cherry-picked as #1430 (with credit), reworked around a single `isComputedMode` property instead of per-condition mode checks, plus a screen-reader fix for the no-scale case. The `beansHold` scale-widget mode was dropped as redundant with the Beans widget + existing `netBeans` mode.
- **Live net milk in MilkWeightItem** → superseded by #1425, which restored the live milk readout on the idle steam pills. The steaming-phase gate here would have kept the live path from showing at the moment it matters — milk is weighed at idle with the pitcher on the scale, before it goes under the wand.

Appreciate the contribution — the live-dosing UX on the widgets exists because this PR pushed on it.

PR #1395, opened 2026-06-25
Steam: enable milk weight auto-capture by default

loganross2001:
## Summary

Flip the default of `milkAutoCaptureEnabled` (weight-timed steaming / milk weight auto-capture) from **off** to **on**, so milk auto-capture mirrors how bean dose-capture already behaves.

## Why

Bean dose-capture activates as soon as a dose-cup tare is saved — no separate opt-in. Milk auto-capture required an extra manual toggle, so the milk-weighing prompts and weight-scaled steam time never appeared for most users even after they'd set up a pitcher. Making the default consistent with beans removes that redundant step.

## Changes & reasoning

**`src/core/settings_brew.cpp`**
- `milkAutoCaptureEnabled()` default changed `false → true`.
- **Why it's safe:** default-on is inert until both preconditions hold — a saved empty-pitcher weight (`cupWeight > 0`, required by `StableWeightCapture` before it captures) and a calibrated reference milk (`scaledSteamTime()` returns 0 without one). So nothing auto-changes the steam stop time until the user has actually set up a pitcher; the flag can still be turned off to disable weight scaling.
- Updated the explanatory comments (here and at the persist-reference logic) that previously described the old "off by default" behaviour.

**`src/core/settings_brew.h`**
- Updated the `Q_PROPERTY` declaration comment for `milkAutoCaptureEnabled` (it still said "Default OFF", which would now contradict the implementation).

## Testing

With a calibrated pitcher: selecting Steam now surfaces the milk-weighing prompt and weight-scaled steam time without a separate toggle. With no pitcher set up, behaviour is unchanged (inert).

## Notes for the reviewer

- One-line functional change plus comment corrections so the property's own documentation matches its default. Calibrating a pitcher already set this flag, so for set-up users the effective behaviour is unchanged; this only helps users who saved a pitcher weight but hadn't flipped the toggle.

### DIFF
--- a/src/core/settings_brew.cpp
+++ b/src/core/settings_brew.cpp
@@ -160,7 +160,12 @@ void SettingsBrew::setDoseCupTareWeight(double weight) {
 }
 
 bool SettingsBrew::milkAutoCaptureEnabled() const {
-    return m_settings.value("steam/milkAutoCaptureEnabled", false).toBool();  // off by default; calibrating turns it on
+    // On by default so milk auto-capture mirrors bean dose-capture (which activates
+    // as soon as a dose-cup tare is saved). It still does nothing until a pitcher
+    // weight is saved (capture needs cupWeight > 0) and a reference milk is set
+    // (scaledSteamTime() returns 0 otherwise), so default-on only removes a redundant
+    // manual toggle. Users can still turn it off to stop the scale changing steam time.
+    return m_settings.value("steam/milkAutoCaptureEnabled", true).toBool();
 }
 void SettingsBrew::setMilkAutoCaptureEnabled(bool enabled) {
     if (milkAutoCaptureEnabled() != enabled) {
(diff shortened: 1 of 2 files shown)


loganross2001:
Part of a coordinated set of 5 independent PRs (each branched off the latest `main`, **no stacking** — reviewable/mergeable in any order): #1393 action-button highlight, #1394 weight widgets, #1395 milk auto-capture default, #1396 UI presentation + steam/Plan widgets, #1397 °C/°F option.

loganross2001:
Withdrawing this one. On reflection the default change is the least important of the batch, and the actual milk-weighing features are independent of it — the live milk-weight widget (#1394) and the steam plan + "place the milk pitcher on the scale" prompt (#1396) both work regardless of this default (the prompt uses a toggle-independent detector). Without this change, weight-scaled steam auto-capture simply stays opt-in via the existing toggle, as it is today. Happy to revisit if you'd ever want the default flipped. Thanks!

PR #1396, opened 2026-06-25
UI presentation: bigger separators, shot-plan sentence, steam plan + page-aware Plan widget

loganross2001:
> **TL;DR for the reviewer:** this is **primarily UI-presentation polish** — bigger separators, the shot-plan-as-a-sentence, and the steam/Plan layout widgets. The *only* milk-related piece is a small on-screen "place the milk pitcher on the scale" prompt (item 4 below); it's secondary, uses a **toggle-independent** detector, and does **not** depend on any setting default or on the (withdrawn) milk-auto-capture-default change. That prompt can be dropped on its own if you'd prefer, without affecting any of the rest.

## Summary

A group of UI-presentation changes that share a small `Theme` text helper, so they land together fully wired:
1. Info separators render as a **larger, bolder bullet** instead of a tiny dot.
2. The **shot plan** reads as a sentence with a leading cup icon and bold values.
3. A new **steam-plan widget** and a **page-aware "Plan" widget** for the layout editor.
4. The idle **steam preset row** gains a "place the milk pitcher on the scale" prompt and clearer pitcher labels.

## 1. Info separators — larger, bolder bullet

**Why:** the small mid-dot/period separator made info sections (bean, equipment, auto-favourites, settings summaries) blend together.

**Changes:**
- `Theme.qml`: new `escapeHtml(s)` and `joinWithBullet(parts)` helpers. `joinWithBullet` joins parts with a `<font size="+1"><b>•</b></font>` separator — **relative** font sizing (no hardcoded px), and each part is HTML-escaped so a bean/roaster name containing `& < >` renders correctly in StyledText.
- `BeanSummary.qml`, `BagCard.qml`, `BeanBaseDetailsRow.qml`, `ChangeBeansDialog.qml`, `Equipment*`, `WeatherItem.qml`, `AutoFavorite*`, `FlowCalibrationPage.qml`, `ShotHistoryPage.qml`, `SettingsAITab.qml`: switched their info-separator joins to `Theme.joinWithBullet` (display) while keeping a plain string for accessibility labels. The `mL·g/s` units dot and spoken text are left unchanged.

## 2. Shot plan as a sentence

**Why:** the shot plan is the most important thing on the idle screen; a row of dot-separated fragments reads poorly.

**Changes (`ShotPlanText.qml` / `ShotPlanItem.qml`):**
- Reworked into a sentence — "Brew 36.0g of Espresso, using <profile> at 92°C" — with a leading tinted cup icon (`ColoredIcon`, theme-safe) and **bold** live values.
- Shows a **single target output weight** (`ProfileManager.targetWeight`, i.e. dose × ratio) rather than "profile default → target".
- Temperature is rendered via the shared `ProfileManager.temperatureDisplay(...)` so multi-temperature profiles and the override delta tag are shown honestly (consistent with the Brew Settings dialog).
- The "deviated from profile" highlight triggers on a **temperature** override only — yield is excluded because the target is dose × ratio and the measured dose rarely equals the profile's listed dose, which would otherwise keep the line permanently highlighted.
- Preserves the wrapper API, the click-to-open-Brew-Settings, and accessibility.

## 3. Steam plan + page-aware "Plan" widget

**Why:** there's a shot plan for espresso but nothing equivalent for steaming; and a single plan widget that adapts to the page reads better than two.

**Changes:**
- `SteamPlanText.qml` / `SteamPlanItem.qml`: a steam analogue — "Steam 300g of milk, using the Large pitcher for 30s" — reflecting the measured weight after the bell (`Window.window.sessionMeasuredMilkG`, mirrored via its change signal because reading a window sub-property through a `var` intermediate isn't a capturable binding dependency).
- `PlanItem.qml` (new **page-aware** widget): shows the brew plan on the home/espresso side and the steam plan on the steam screen. Page/mode are read from window-root properties (`currentPageObjectName`, `currentOperationMode`) — a status-bar widget loads as a separate component and can't see the page-stack by scope, so the app publishes these on the `ApplicationWindow` root; the widget mirrors them via `Connections` (same reactivity reason as above).
- `main.qml`: publishes `currentPageObjectName` and `currentOperationMode`; `IdlePage.qml` sets the operation mode when a preset row is opened.
- Registered both new widgets in the **four** standard places: `CMakeLists.txt` (qml module file list), `LayoutItemDelegate.qml` (type switch), `LayoutEditorZone.qml` (palette + chip-label map), and `shotserver_layout.cpp` (web editor list + label map).

## 4. Idle steam preset row

**Why:** users steam from the idle screen's steam preset row (the full Steam page is only reached once steaming has started), so the milk-weighing guidance belongs there.

**Changes (`IdlePage.qml`, `PresetPillRow.qml`):**
- A "Place the milk pitcher on the scale" prompt below the pitcher pills (mirroring the existing "Place Beans on Scale" prompt), driven by a toggle-independent load detector so it appears whenever a scale is connected and the scale is empty.
- Pitcher pills now read "Small Pitcher" / "Large Pitcher" (a new optional `PresetPillRow.pillLabelFn`, which skips the disabled "Off" preset and names already containing "pitcher").
- Removed the per-pill milk-weight `(Xg)` readout (it was noise on the buttons).

## Testing

On-device: separators read as distinct; the shot plan reads as a sentence and follows the C/F unit (when the temperature PR is also present); the Plan widget switches between brew and steam info as you move between the home and steam screens; the steam row shows the place-pitcher prompt and "Small/Large Pitcher" labels.

## Notes for the reviewer

- No hardcoded colours/sizes; `<font size="+1">` is relative sizing. User data is HTML-escaped before StyledText.
- Touches `LayoutItemDelegate.qml` and `PresetPillRow.qml`, also touched (different sections) by the action-button-highlight PR — independent (no stacking).
- The shot plan's temperature uses upstream's `ProfileManager.temperatureDisplay`; its °C/°F awareness comes from the separate temperature-unit PR. On this branch alone it shows °C.


loganross2001:

loganross2001:
Ran `pr-review-toolkit` (code review, opus), fixed the findings, and re-reviewed to clean.

### Fixed
- **`SteamPlanText.qml` header comment** claimed temperatures go through `Theme.formatTemperature` — but this widget renders no temperature (milk / pitcher / duration only) and `Theme.formatTemperature` doesn't exist. Corrected the comment.
- Clarified the a11y comment on the `SteamPlanText` root (`Accessible.ignored: true`; `ShotPlanText` reaches the same no-duplicate-node outcome by being a role-less Item).

### Deferred (not bugs — re-review agreed safe)
- The `plan` widget's plan-content show-flags aren't wired into the per-flag editor branches the way `shotPlan` is, so they're not user-editable — but the defaults are correct and the widget renders fine. Wiring `plan` into the editors is a small follow-up, not a defect.
- `LibraryItemCard`'s `getItemDisplayName` label map doesn't include `steamPlan`/`plan`, but that map is already missing several widgets (`milkWeight`, `ratioQuickSelect`, `doseWeight`, `clock`, …) and has a graceful `names[type] || type` fallback — pre-existing incompleteness, not introduced here.

Verified: re-review reports no must-fix/should-fix, and the macOS build passes. The 4-place layout-widget registration for the new `plan`/`steamPlan` widgets is complete.

skialpine:
Thanks — the direction here is good: steam-plan parity and the page-aware Plan widget fill a real gap, and the mechanical work is thorough (widget registration complete in all four places + CMakeLists, accessibility roles/names handled carefully, everything translated, user data HTML-escaped). Review below; **items 1–3 are required changes**, the rest is smaller.

> Note: findings are anchored to file + property/function names rather than line numbers — I verified against your branch rebased onto current `main`. Heads-up: `ShotPlanText.qml` now conflicts with the °C/°F follow-up (#1414) that landed on main; the conflict resolves cleanly in your favor (your `_tempStr` already contains the same `void(Settings.app.temperatureUnit)` fix), so a straightforward rebase is all it needs.

## Required changes

### 1. The Shot Plan display options must keep working

The Shot Plan widget's settings (the in-app popup in `ScreensaverEditorPopup.qml` and the web layout editor) expose five toggles: **Profile & temperature, Roaster, Coffee (grind), Roast date, Dose & yield**. In the rewritten `ShotPlanText.qml`:

- `showRoaster`, `showGrind`, `showRoastDate` are declared (and their backing `Settings.dye` data is still read) but never referenced by either the plain `text` or the `_rich` builder — toggling them does nothing.
- `showDoseYield` now only gates the yield; the **dose is gone entirely** — the `dose` property is declared but unused. The old widget showed "18.0g in".

We want to keep the options, and the sentence needs to incorporate them. Users configured these toggles for a reason; the new look shouldn't change what their widget shows. Suggested shape: keep the sentence as the base rendering for the default toggle set, and when optional fields are enabled either append them as a joined tail or degrade to the fragment-list form — your choice — applied symmetrically to both `text` (accessibility/visibility) and `_rich` (display), with the dose restored under Dose & yield. `ShotPlanItem`/`PlanItem` already pass all five flags through, so this is contained to `ShotPlanText.qml`. Please also fix the comment above the flags ("kept for source compatibility but are not part of the sentence by default") once they work — as written it papers over the fact that there is currently no path where they appear.

### 2. Replace `•` (U+2022) — it's not on the safe-glyph list

The PR swaps the separator `·` (U+00B7) for `•` everywhere: the plain `.join("  •  ")` calls, the `\u2022` escape forms (`ShotHistoryPage`, `WeatherItem`), and inside `Theme.joinWithBullet`'s font markup. Project rule (CLAUDE.md / QML_GOTCHAS): text glyphs are limited to the safe list `° · — → ×` because **glyphs outside it can crash the app on some platforms but not others** — it looks fine on the device you test on and fails elsewhere, so it can't be validated locally.

Two acceptable fixes, both of which keep the visual intent (a more prominent separator):

- Keep the safe `·` and give it `joinWithBullet`'s bigger/bold treatment (`<font size="+1"><b>·</b></font>`), reverting the plain joins and escapes to `·`/`\u00B7`; or
- Use a **graphic** instead of a glyph: an SVG from `qrc:/icons/` — `Text.StyledText` supports inline `` in `joinWithBullet`, and the places using standalone separator `Text` elements (the AutoFavorites pages) can use `Image`.

Affected files: `Theme.qml` (`joinWithBullet`), `ShotPlanText.qml`, `SteamPlanText.qml`, `BagCard.qml`, `BeanBaseDetailsRow.qml`, `BeanSummary.qml`, `ChangeBeansDialog.qml`, `EquipmentInfoDialog.qml`, `EquipmentSummary.qml`, `AutoFavoriteInfoPage.qml`, `AutoFavoritesPage.qml`, `FlowCalibrationPage.qml`, `ShotHistoryPage.qml`, `WeatherItem.qml`, `SettingsAITab.qml`.

### 3. The pitcher prompt must respect "Enable weight-timed steaming"

The new blinking "Place the milk pitcher on the scale (and wait for the beep before removing)" prompt on the idle page is deliberately toggle-independent — the `idlePitcherDetect` comment states it "does NOT require milkAutoCaptureEnabled, so the nudge shows whenever steam is selected and a scale is connected". Verified on-device: with **Enable weight-timed steaming** off (the default — `milkAutoCaptureEnabled` defaults to `false`; calibrating turns it on), the prompt still shows.

That's a broken instruction, not just a bypassed setting: with the toggle off, `idleMilkCapture` is inactive, so placing the pitcher captures nothing and the promised beep never comes — the prompt tells the user to perform an action that does nothing and wait for a signal that can't fire. A feature being off by default doesn't get overridden by a nudge; gate the prompt's detector (or its `visible`) on `Settings.brew.milkAutoCaptureEnabled`, same as `idleMilkCapture`.

## Bug

### 4. `SteamPlanText._preset` goes stale when the selected pitcher's contents change

```qml
readonly property var _preset: Settings.brew.getSteamPitcherPreset(Settings.brew.selectedSteamPitcher)
```

`getSteamPitcherPreset()` is a `Q_INVOKABLE`, so the binding only captures `selectedSteamPitcher`. Renaming the pitcher, toggling its disabled state, or recalibrating `calibMilkG` on SteamPage **without changing the selection** leaves the widget showing the stale name / stale reference milk until the pitcher is reselected. Fix: make the binding also depend on `Settings.brew.steamPitcherPresets` (e.g. a `void(Settings.brew.steamPitcherPresets)` read before the call), or index into the list directly. The "(re-reads when the selection changes)" comment currently reads as a reactivity guarantee — update it either way.

## Worth considering (not blocking)

5. **Displayed steam duration**: the plan shows `Settings.brew.steamTimeout`, which is overwritten on every steam-pill tap with the milk-scaled time — so before the first tap it can show a scaled time computed from a *previous session's* milk. Consider sourcing the displayed duration from the selected preset's base `duration`, or accept it as "last computed time" knowingly.
6. **Removed live milk readout**: the steam pills lost the live "(123g)" suffix. The blinking prompt + plan sentence don't replace watching net milk tick up while pouring; the Milk Weight widget covers it only if the user placed one. Flagging so the regression is a conscious choice.
7. **`.arg()` chaining**: profile/pitcher names containing a literal `%2`/`%3` could be substituted by a later `.arg`. Single-pass multi-arg (`.arg(a, b, c, d)`) avoids it. Cosmetic, low probability.
8. **Comment fixes**:
   - `PlanItem.qml` header: drop the trailing "that scoping is why reading pageStack.currentItem from a status-bar widget failed" — development-history narration; the constraint sentence before it carries the information.
   - The "reading a sub-property through a var-typed intermediate doesn't register a binding dependency" rationale (in both `PlanItem.qml` and `SteamPlanText.qml`) is contradicted by the existing `MilkWeightItem.qml`, which reads `win.sessionMeasuredMilkG` declaratively through the same kind of var reference and relies on it updating live. Please reconcile: either the gotcha is real (then MilkWeightItem has a latent bug worth noting) or it isn't (then the imperative mirroring may be unnecessary). If you hit this empirically, say what specifically failed.
   - `SteamPlanText.qml`'s `Accessible.ignored` parenthetical claims ShotPlanText differs "by being a role-less Item" — both roots are structurally identical role-less Items; correct or drop the parenthetical.
   - `IdlePage.qml`: the standalone "(The 'Place the milk pitcher on the scale' prompt now lives…)" comment narrates within-PR history (the prompt is new in this PR) and duplicates two adjacent comments — remove.
9. **Rename hazard (hardening)**: `PlanItem`, `SteamPlanText`, and `MilkWeightItem` mirror `main.qml`'s `currentPageObjectName` / `currentOperationMode` / `sessionMeasuredMilkG` by name through `ignoreUnknownSignals` Connections — a future rename in `main.qml` fails silently (widgets freeze on their last value). A cross-reference comment on those properties in `main.qml` would prevent that.

## What checked out (verified, no action needed)

- Both new widget types (`steamPlan`, `plan`) registered in all four required places + `CMakeLists.txt`.
- All `ignoreUnknownSignals` handler names match real window-root signals; `openBrewSettings` exists; `MachineStateType.Phase.Steaming`, `"steamPage"`, and `"steam"` all resolve to real registrations/values.
- `PresetPillRow.pillLabelFn` feeds both the width calculation and the rendered label (one code path), with recalc wired.
- `escapeHtml`/`joinWithBullet` call sites all guard their parts (no `undefined` can leak in), and every `joinWithBullet` consumer sets `textFormat: Text.StyledText`.
- Dropping the `typeof ScaleDevice` guard is safe (context property set before engine load, never null); `sessionMeasuredMilkG` reset paths always write `0`.
- The `idlePitcherDetect` StableWeightCapture reuse (`minNet: 999999`, `loadThreshold: 50`) genuinely never captures and only reads `loadPresent`; the "Off"-preset early return still protects the undefined `duration`/`flow` writes; the blinking prompt is an animation, not a timer-as-guard.
- `_publishOperationMode` clears the mode when IdlePage isn't active — no stale steam-mode leak.

---
*Aggregated from a multi-agent review (general code review, comment accuracy, silent-failure passes) run with Claude Code.*

loganross2001:
Thanks for the thorough review — all required items addressed, rebased onto current `main`, and re-reviewed before pushing.

**1. Shot Plan display options + dose (required)** — Fixed. All five toggles (Profile & temperature, Roaster, Coffee/grind, Roast date, Dose & yield) now render, and the **dose ("18.0g in") is restored**. The sentence stays the base rendering for the default toggle set; enabled optional fields trail after a `·`, and if the core can't render it degrades to the fragment-list form — applied **symmetrically to both `text` (accessibility) and `_rich` (display)**. Fixed the misleading "kept for source compatibility" comment.

**2. Replace `•` (required)** — Done. Swapped `•` (U+2022) → `·` (U+00B7) across the 15 files this PR touched; `joinWithBullet` keeps the bigger/bold treatment with the safe glyph. Left the pre-existing-on-`main` `•` (main.qml, ProfileEditorPage, DeviceMigrationDialog, StringBrowserPage) out of scope.

**3. Pitcher prompt gate (required)** — Done. The prompt's detector (`idlePitcherDetect.active`) is now gated on `Settings.brew.milkAutoCaptureEnabled`, so it's hidden when weight-timed steaming is off. Updated the now-accurate comments.

**4. `_preset` stale binding (bug)** — Fixed. The binding reads `Settings.brew.steamPitcherPresets` before the invokable, so it refreshes on rename/disable/recalibrate without a selection change.

**Smaller items:**
- Comment fixes (#8): PlanItem header dev-history line dropped; the var-binding rationale softened; the SteamPlanText `Accessible.ignored` parenthetical corrected; the IdlePage within-PR narration removed.
- `.arg` chaining (#7): QML's `String.arg` is single-argument (the multi-arg `.arg(a,b,c,d)` form is C++ `QString` only), so I kept the chained form — the residual re-substitution risk needs a name to literally contain `%2`–`%4`.
- Left #5 (steam duration source), #6 (removed live milk readout), and #9 (rename cross-refs) as the conscious-choice / hardening items.

**Rebase note:** confirmed — the `ShotPlanText.qml` °C/°F conflict with #1414 resolved cleanly in this branch's favor (the `void(Settings.app.temperatureUnit)` re-render fix was already present).

One thing I explored and dropped: adding a temperature to the *steam* plan for parity with the shot plan — but the app has no milk-target temperature (the only steam temp is the boiler setpoint, ~160 °C), so "steam milk to 160°" would be misleading. Steam sentence stays milk / pitcher / duration.

loganross2001:
Ran the full `pr-review-toolkit:review-pr` suite (code, silent-failure, comments, tests, types) on this PR and iterated it to convergence — **three rounds**. It caught real issues the earlier single pass missed; all are now fixed:

**Functional bugs (round 1):**
- **The page-aware `plan` widget's five display toggles were dead config.** `PlanItem` forwards `shotPlanShow*` into `ShotPlanText`, but `plan` was never added to the options allowlist, so no editor could ever set them. Now wired everywhere `shotPlan` is — the C++ gate (`typeHasOptions`), the native editor (`ScreensaverEditorPopup` **and** the `SettingsLayoutTab` open-router), and the web editor (`shotserver_layout` chipClick/load/save + the JS mirror) — with `PlanItem` consuming the saved props. `steamPlan` stays non-configurable (no toggles).
- **`plan`/`steamPlan` overflowed the center/idle zone** — missing from `LayoutCenterZone.isAutoSized()`. Fixed.

**Round 2 caught that round 1's own `plan`-config fix was incomplete** — the native open-router (`SettingsLayoutTab.qml:125`) still routed `plan` to the *wrong* editor (5 of 6 sites were wired). Two agents independently flagged it; fixed, and round 3 confirmed all sites coherent with no half-wiring.

**Also fixed:** pinned both new widget types in `tst_settings.cpp::typeHasOptionsAllowlist`; consolidated the plain-`text` vs bolded-`_rich` builders in ShotPlanText/SteamPlanText into a single renderer so the accessibility label provably can't drift from the display; `%`-escape user-supplied names before `.arg`; `_beverage` now reacts to a live language switch; plus the comment-accuracy fixes.

**Converged:** the final two rounds report no blockers/should-fixes; macOS build + qmllint + the settings test are clean. Point taken on running it in full and repeatedly — that's how this was done.

skialpine:
### Code review

Found 7 issues:

1. Roaster/Coffee toggle content swapped vs main — the coffee name moved from the "Coffee (grind)" toggle to the "Roaster" toggle. On main, `showRoaster` rendered the roaster brand only and `showGrind` rendered `coffeeName (grindSize)`; at HEAD `_roasterStr` is `roasterBrand + " " + coffeeName` and `_grindStr` is the bare grind number. Saved widget configs keep their old booleans but now show different content (Roaster OFF + Coffee ON drops the coffee name entirely), and the editor labels no longer describe what they control. Restore the original split: Roaster → brand only; Coffee (grind) → coffee name + grind.

https://github.com/Kulitorum/Decenza/blob/241f21a01a8f4c499b0ecdd9166934fdb7ce944f/qml/components/ShotPlanText.qml#L56-L59

2. Steam sentence renders "Large Pitcher pitcher" — the template `"Steam %1 of milk, using the %2 pitcher for %3"` feeds `%2` the raw preset name with no dedupe, while the idle pills both display names as "Small Pitcher"/"Large Pitcher" and guard against exactly this (`pillLabelFn` skips the suffix when the name already contains "pitcher"). A preset named "Large Pitcher" — the naming the pills themselves teach — duplicates the word in the plan sentence and its accessibility label.

https://github.com/Kulitorum/Decenza/blob/241f21a01a8f4c499b0ecdd9166934fdb7ce944f/qml/components/SteamPlanText.qml#L61-L64

3. `ScaleDevice` null guards dropped — `idlePitcherDetect` and the rewritten steam `onPresetSelected` access `ScaleDevice.connected` without the `ScaleDevice && ` guard that #1399 (f72aa24c) added to every reference in this file ("it is a context property that reads null briefly at startup"), reintroducing the startup `TypeError` it fixed. The untouched neighbors (`idleMilkCapture`, the bean readout) still carry the guard. (Also flagged in the earlier automated comment on this PR.)

https://github.com/Kulitorum/Decenza/blob/241f21a01a8f4c499b0ecdd9166934fdb7ce944f/qml/pages/IdlePage.qml#L358-L364

4. The pitcher prompt still promises a beep that won't play by default — the prompt is now correctly gated on `milkAutoCaptureEnabled`, but the capture ding is separately gated on `doseCaptureSoundEnabled`, which defaults to `false` and is not enabled by pitcher calibration. A user who calibrates (auto-enabling weight-timed steaming) is told to "wait for the beep before removing" and no beep ever comes. Either play the ding for milk capture unconditionally (or when the prompt is shown), enable the sound on calibration, or drop the beep wording.

https://github.com/Kulitorum/Decenza/blob/241f21a01a8f4c499b0ecdd9166934fdb7ce944f/qml/pages/IdlePage.qml#L636-L638 / https://github.com/Kulitorum/Decenza/blob/241f21a01a8f4c499b0ecdd9166934fdb7ce944f/qml/pages/IdlePage.qml#L268-L270 / https://github.com/Kulitorum/Decenza/blob/241f21a01a8f4c499b0ecdd9166934fdb7ce944f/src/core/settings_brew.cpp#L194-L196

5. Yield-override arrow silently removed — main showed `"36.0 → 40.0g out"` when the user deliberately dialed a yield override (`hasBrewYieldOverride`); the new `_yieldStr` is just the target weight. The code comment justifies removing the yield *highlight* (measured dose drifts, so a raw diff would always trip), but the arrow on main was gated on the explicit override flag, not a raw diff — that reasoning doesn't cover it. If dropping the arrow is intended, say so; otherwise restore it under `hasBrewYieldOverride`.

https://github.com/Kulitorum/Decenza/blob/241f21a01a8f4c499b0ecdd9166934fdb7ce944f/qml/components/ShotPlanText.qml#L38-L48

6. Page-aware Plan widget's settings don't acknowledge its steam mode — editing a "plan" widget always shows "Shot Plan Settings" with the five shot-plan toggles; while the widget is displaying the steam plan (which has no options) they visibly do nothing. A title like "Plan Settings" plus one line noting the toggles apply to the shot-plan side would remove the confusion.

https://github.com/Kulitorum/Decenza/blob/241f21a01a8f4c499b0ecdd9166934fdb7ce944f/qml/components/layout/ScreensaverEditorPopup.qml#L107-L111

7. Second copy of current-page state — `Theme.currentPageObjectName` already exists (kept updated by main.qml's page-change handler) and, as a singleton, is directly bindable from any widget with no `Window.window` mirroring. The PR adds a parallel `currentPageObjectName` on the window root plus the rename-fragile Connections dance in PlanItem. Prefer binding PlanItem to the Theme property and putting `currentOperationMode` next to it on Theme.

https://github.com/Kulitorum/Decenza/blob/241f21a01a8f4c499b0ecdd9166934fdb7ce944f/qml/main.qml#L664-L669 / https://github.com/Kulitorum/Decenza/blob/241f21a01a8f4c499b0ecdd9166934fdb7ce944f/qml/Theme.qml#L23-L25

---

<details>
<summary>4 below-threshold issues</summary>

- [Worth a look] `ShotPlanItem.qml` has a hardcoded English `Accessible.name` ("Shot plan: … Tap to edit") — real i18n violation and the direct ancestor of the new PlanItem (which fixes it), but the file is untouched by this PR (score: 75, outside strict scope)
- [Worth a look] `BeanSummary.qml` header comment shows the `[· Def Md]`/`[Frozen]` suffix only on the canonical template line, but the code appends it in the history-only branch too (score: 75, comment-only)
- [OK to skip] `ShotPlanText`/`SteamPlanText` re-implement `joinWithBullet`'s separator inline — legitimate, they need per-value bolding the helper can't do; a note on the helper would help (score: 50)
- [OK to skip] `PresetPillRow.pillLayoutName` comment says "layout calculation" but its output (including the new `pillLabelFn` transform) is also the displayed label text (score: 50)

</details>

skialpine:
Note on the review above: we'll take it from here and fix the remaining issues ourselves on this branch — no further action needed on your side. Thanks for the quick turnarounds on the earlier rounds; the feature itself is in good shape.

skialpine:
Pushed `3adbf205` addressing all seven findings from the review above (fixes done maintainer-side as noted):

1. **Roaster/Coffee toggle mapping** — restored and extended: Roaster shows the brand only; a new **Coffee** toggle (`shotPlanShowCoffee`, default ON) carries the bean name; **Grind** now shows the grinder setting plus the recorded RPM ("grind 2.5 · 90 rpm"). Saved `shotPlanShowGrind: false` configs keep grind hidden. Both editors gained the Coffee toggle and the relabeled Grind (new translation keys, so stale translations of the old "Coffee (grind)" label can't mislabel the narrowed toggle).
2. **"Large Pitcher pitcher"** — the sentence uses a second full template ("…using the %2 for %3") when the preset name already contains "pitcher", in both the displayed and accessibility forms.
3. **`ScaleDevice` null guards** — restored at all three sites (`idlePitcherDetect` rawWeight/active, steam `onPresetSelected`), matching the #1399 convention in the rest of the file.
4. **Beep promise** — the prompt's hint line is now conditional: "(and wait for the beep before removing)" only when `doseCaptureSoundEnabled` is on, otherwise "(hold still until the weight registers)". The sound's own gating is unchanged.
5. **Yield-override arrow** — restored: a deliberate override (`hasBrewYieldOverride`, >0.1 g from the profile yield) renders "36.0 → 40.0g". The highlight stays temperature-only per the PR's original reasoning, which covered the highlight but not the arrow.
6. **Plan widget settings** — retitled "Plan Settings" with a note (in-app and web editor) that the display options apply to the shot-plan side; the steam plan has no options.
7. **Duplicate page state** — `PlanItem` now binds `Theme.currentPageObjectName` (existing) and a new `Theme.currentOperationMode` directly; the window-root duplicates and the rename-fragile `Connections` mirroring are gone.

Also in this commit:

- The scaled-or-base steam duration ternary moved into `SettingsBrew::effectiveSteamDurationSec()`, used by the steam plan display and both pill-tap paths (IdlePage + SteamPage) — the plan now shows the *selected* preset's effective time instead of the last tap's `steamTimeout` residue.
- Below-threshold items: `ShotPlanItem`'s hardcoded English `Accessible.name` translated (+ missing `Accessible.onPressAction` added); `BeanSummary` header comment corrected.
- Separator visibility completed with the safe glyph: the remaining data rows (BagCard, BeanBaseDetailsRow, EquipmentSummary, ChangeBeansDialog) now use `Theme.joinWithBullet`'s styled bold `·`, and the AutoFavorites separator dots are bolded. Gated on an empirical check that `elide` + `Text.StyledText` works in Qt 6.11 (it does — verified with an overlong-string harness), and the helper's HTML escaping was verified against `&`/`<`/`>` in user data.

Build is clean (0 errors/warnings) and `tst_settings` passes 40/40. Remaining before merge: on-device pass on the tablet (glyph rendering, prompt wording, narrow-window eliding) and a CI Android test build.

skialpine:
Follow-up `bb46c158` (maintainer decision): the separate **Plan (page-aware)** and **Steam Plan** widget types are gone. The existing **Shot Plan** widget is now page-aware itself — it swaps to the steam sentence in steam context, gated by a new per-instance **Steam plan** toggle (default ON) in its settings dialog and the web editor. Net effect: one widget instead of three for the same concept, −118 lines, and everyone with a Shot Plan widget already placed gets the steam plan for free (turn the toggle off to opt out). `SteamPlanText` remains as the internal renderer. Saved layouts from branch builds that contained the removed types will just render nothing there — re-add the Shot Plan widget. Build clean, `tst_settings` 40/40 (now pinning the removed type names as non-configurable).

skialpine:
Follow-up `4ff65c73`: the sentence's beverage word now follows the profile's `beverage_type` (new `ProfileManager.currentProfileBeverageType`) — **Espresso** for espresso profiles, **coffee** for filter/pourover/other coffee types, **tea** for tea profiles. Cleaning and descale profiles replace the plan entirely with **"Cleaning run with <profile> — no coffee in the portafilter!"** in the warning color, bean/dose segments suppressed. Build clean, tst_settings 40/40.

skialpine:
Tweak `a1651787`: the cleaning/descale notice is now **bold red** (Theme.errorColor), not warning-orange — it's a warning, not a plan.

skialpine:
Follow-up `f23366ee`: the pre-existing **bean** prompt had the same beep lie as the pitcher one — "(wait for the beep)" while the ding is gated on the capture-sound option (default off). Same fix applied: beep wording only when the sound is enabled, otherwise "(hold still until the weight registers)" (hint key now shared by both prompts). Its visibility was already correctly opt-in — it only shows once a dose-cup tare is saved.

loganross2001:
Small sweep of dead leftovers from the plan-widget consolidation (`bb46c158`) — none functional, just orphaned references:

1. **`qml/components/SteamPlanText.qml`** (~L99) — the comment still says *"Always wrapped by SteamPlanItem / PlanItem…"*, but both were deleted; `SteamPlanText` is now the internal renderer for the page-aware `ShotPlanItem`. Comment-only.

2. **Dead `plan` / `steamPlan` type references** now that those widget types are gone (both are from my earlier review-pr round, which your consolidation superseded):
   - `qml/components/library/LibraryItemCard.qml:635` — the `"plan": "Plan", "steamPlan": "Steam Plan"` friendly-name entries. Harmless (the `names[type] || type` fallback covers it), just dead.
   - `qml/pages/settings/SettingsLayoutTab.qml:125` — the `|| type === "plan"` in the editor open-router is now unreachable, since `typeHasOptions("plan")` returns false after you dropped it from `kConfigurable`.

All cosmetic. Happy to push a one-commit sweep if you'd like, or leave them to you.

PR #1399, opened 2026-06-29
fix(hotwater): store water temperature per vessel preset

### LINKED ISSUE #1398: [Bug] changing the dispensing temp in a water profile changes the temp in all of them…
### What happened?

We have several water dispensing profiles with different volumes and (we wish) different temperatures. 

But whenever you set the water temp in one water profile, it (re)sets the temperature for all of the water profiles.

This is on the iOS release and it has been there for a while. I just checked and the bug remains present in 1.8.0.

### Steps to reproduce

1) Create two or more water dispensing profiles with different temps

2) look at first profile and note that it has the same temp as the new profile. 

3) change the temp on first profile. 

4) look at new profile and note that it has changed.

### Decenza version

1.8.0 (build 3447)

### Platform

iOS

### Device / OS version

iPadOS 26.5

### Scale (if applicable)

_No response_

_No response_

_No response_

### Screenshots or video

_No response_

### Anything else?

_No response_https://github.com/user-attachments/assets/8c582009-b528-4c5b-9393-2a55099f6298

skialpine:
Closes #1398

## Problem

Creating a new hot water profile (vessel preset) and setting a temperature changed the temperature for **all** existing hot water profiles. Each preset stored its own `volume`, `mode`, and `flowRate`, but temperature lived in a single global setting (`water/temperature`) shared by every preset. Switching presets also never restored a per-preset temperature.

## Fix

Make hot water temperature per-preset, mirroring how `volume` already works (each preset persists its own value; the active `Settings.brew.waterTemperature` is loaded from / saved to the selected preset and sent to the machine).

**C++**
- `addWaterVesselPreset` / `updateWaterVesselPreset` gain a `temperature` parameter and persist it in the preset JSON (`src/core/settings_brew.{h,cpp}`).
- Settings export/import round-trips the per-preset temperature (`src/core/settingsserializer.cpp`); legacy presets without the field fall back to the global value on export.

**QML (`qml/pages/HotWaterPage.qml`)**
- New `selectVessel()` helper loads volume/mode/flow **and temperature** from a preset; the four duplicated selection blocks (accessibility press, Return, Space, mouse click) now call it.
- Temperature input, `saveCurrentVessel()`, page activation, add-preset, and rename all carry temperature through.

**Migration:** legacy presets have no `temperature` field, so they fall back to the current value on first load — no surprise jumps for existing users.

## Also included

Guard all `ScaleDevice` references in `qml/pages/IdlePage.qml` with the codebase-standard `ScaleDevice && …` null check. `ScaleDevice` is a context property that reads `null` briefly at startup, which was throwing `TypeError: Cannot read property 'connected' of null` on launch. Verified the rest of the QML tree already guards its `ScaleDevice` access.

## Testing

- Build verified in Qt Creator by the author.
- Manual: switching presets changes the shown/active temperature; editing a temperature only affects the selected preset; new presets start at the current temperature; renaming preserves a preset's temperature; IdlePage TypeErrors no longer appear in the log.

### DIFF
--- a/qml/pages/HotWaterPage.qml
+++ b/qml/pages/HotWaterPage.qml
@@ -20,6 +20,7 @@ Page {
             Settings.brew.waterVolume = getCurrentVesselVolume()
             Settings.brew.waterVolumeMode = getCurrentVesselMode()
             Settings.hardware.hotWaterFlowRate = getCurrentVesselFlowRate()
+            Settings.brew.waterTemperature = getCurrentVesselTemperature()
             MainController.applyHotWaterSettings()
             // Tare immediately so display shows 0g instead of current scale weight.
             if (!isVolumeMode) {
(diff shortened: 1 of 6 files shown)


joshhodas:
The identical bug exists in the steam profiles. I forgot that that is actually where we first discovered it a while back.

PR #1129, opened 2026-05-12
Remove temperatureUnstable badge end-to-end (#1128)

skialpine:
## Summary

- Retires the `temperatureUnstable` quality badge across detector, projection, badge UI, MCP, history filter, and DB column. The detector measured average deviation from goal but was labeled "Temp unstable" — the math didn't match the label, and 12+ built-in profiles intentionally produce large gaps the badge mistook for problems.
- Adds `puckfailed:yes` chip to the Search Syntax help (it was in the syntax footer but missing from the per-keyword chip row).

## Why

Issue #1128: waltzingfool was getting persistent false-positive "Temp unstable" badges on Extractamundo Dos! Audit found at least 12 built-in profiles whose actual-vs-goal temperature gap is by design:
- D-Flow / Damian's Q / D-Flow La Pavoni — `espresso_temperature_steps_enabled = 0`; KB explicitly says "DO NOT flag the large gap (94°C target / ~88°C actual) — by design"
- Extractamundo Dos! / TurboBloom — 16°C bloom drops the head physically can't cool to
- 80's Espresso / TurboTurbo / Classic Italian / Blooming Espresso / A-Flow dark / Innovative long preinfusion / Flow Profile straight & milky — 6–10°C designed swings

The `hasIntentionalTempStepping` curve-based suppression was supposed to catch these but is fragile — one Visualizer-shared shot showed the captured `temperatureGoal` series collapsed to a single constant 72.5°C even on a profile that spans 67.5–83.5°C, and suppression failed.

The remaining diagnostic value (cold portafilter crash, heater failure, boiler exhaustion) is rare, already visible in the post-shot temperature chart, and crowded out by the false-positive noise. Tagging profiles individually with a `temp_drift_expected` flag was considered and rejected — every new "intentional drop" profile would need the flag, the false positive would remain permanent until tagged, and the underlying detector still measures the wrong thing.

## What Changes

**Detector (`src/ai/shotanalysis.{h,cpp}`)**
- Delete fields: `tempUnstable`, `tempStabilityChecked`, `tempIntentionalStepping`, `tempAvgDeviationC` from `DetectorResults`
- Delete constants: `TEMP_UNSTABLE_THRESHOLD`, `TEMP_STEPPING_RANGE`, `TEMP_WARMUP_SKIP_C`, `TEMP_MIN_EXTRACTION_SEC`
- Delete static helpers: `hasIntentionalTempStepping` (both overloads), `avgTempDeviation`, `reachedExtractionPhase`
[...]

Issue #1128, opened 2026-05-12
[Bug] Near constant bad puck reports


waltzingfool:
### What happened?

This is one example of a shot this morning that, like almost every shot for the last several weeks has given me either one or both errors for the shots summary, typically both as in this case where the summary indicated temp, unstable and too loose grind. The shot was actually tighter than I wanted, finished about ten seconds later. However, I got the exact same report with a  preceding shot that lasted only 20 seconds, but had the same profile and inputs.

I'm sorry I'm such a doofus I couldn't figure out how to download the debug log, but I copied it and hope you can use it in this form: 

[08:21:39.572] START Shot capture started - 2026-05-11T08:21:39
[08:21:39.572] INFO Profile: Extractamundo Dos! (user)
[08:21:39.580] DEBUG [SAW] model: source= "perProfile" lag= 1.0846 profile= "extractamundo_dos_user" scale= "Decent Scale" historyN= 3
[08:21:39.665] DEBUG ShotDataModel: Registered goal/marker series
[08:21:39.668] DEBUG ShotDataModel: Registered fast renderers (QSGGeometryNode, pre-allocated VBO)
[08:21:39.675] DEBUG [AutoSleep] Reset by phase change: normal=70
[08:21:40.098] DEBUG [Scale] weight= "0.0" phase= "EspressoPreheating" tare= true
[08:21:40.553] DEBUG === SCALE TIMER: Started (espresso extraction began) ===
[08:21:40.575] DEBUG [AutoSleep] Reset by phase change: normal=70
[08:21:40.609] DEBUG FRAME CHANGE: -1 -> 0 name: "preinfusion start" exitWeight: 0
[08:21:40.615] DEBUG EXTRACTION STARTED at frame 0
[08:21:40.636] DEBUG [ExtractionTrack] pressure actual= "0.03" goal= "7.31" delta= "7.29" threshG= "1.83" threshW= "3.66" "RED"
[08:21:42.463] DEBUG FRAME CHANGE: 0 -> 1 name: "preinfusion" exitWeight: 0
[08:21:42.693] DEBUG [ExtractionTrack] pressure actual= "0.68" goal= "3.81" delta= "3.13" threshG= "0.95" threshW= "1.91" "RED"
[08:21:44.746] DEBUG [ExtractionTrack] pressure actual= "2.03" goal= "7.94" delta= "5.91" threshG= "1.98" threshW= "3.97" "RED"
[...]

PR #1411, opened 2026-07-02
Live coaching (during-shot): real-time on-device cues

loganross2001:
## Live coaching (during-shot): real-time on-device cues

Adds a local **"reflex" layer** that comments on an espresso shot **as it runs** — one short, calm line at a time. This is deliberately *not* the AI advisor (which stays post-shot and network-backed): there are **no AI calls, no network, and no DB in the hot path**. Most shots get zero or one cue; it favours silence over false alarms.

### What it does
`LiveShotCoach` (`src/ai/liveshotcoach.{h,cpp}`) subscribes to `DE1Device::shotSampleReceived` (~5 Hz) and `MachineState` phase/scale-weight, and runs cheap checks on the **partial** live curve:

- **Low pressure / no puck** — peak pressure never crosses the floor (2.5 bar).
- **Channeling proxy** — a sharp flow rise while pressure stays flat during a pressure phase (a conservative live proxy; the real dC/dt detector is post-shot only and is never called here).
- **Fast / slow flow vs goal** — rolling average of `flow − goalFlow` during the pour.
- **On pace to target** — from scale flow rate + weight vs target (a positive, encouraging cue).
- **Phase milestone** — the preinfusion→pour transition.

Cues are exposed to QML as individual marshalable `Q_PROPERTY`s (`cueText` / `cueSeverity` / `cueActive` / `cueSpeak`) — never a struct return — so the banner binds directly.

### Presentation & voice
`LiveCoachingBanner.qml` shows one calm line, tinted by severity using Theme tokens, and fades out on its own. Voice reuses the **existing** `AccessibilityManager` extraction-announcement preference, so it never double-announces and respects the user's TTS choice. `Settings.app.liveCoachingEnabled` (default on) gates the banner, with a toggle in Settings → Accessibility.

### Design notes
- The cue **governor** is event/clock-based (dedupe by cue id; spoken-cue spacing measured against `shotTime`), **not** a guard `Timer` — per the project's no-timers-as-guards rule. The only timer is the banner's UI auto-dismiss.
- All strings go through `TranslationManager`; the banner uses only Theme tokens.

### Notes for review
- Independent PR off the latest `main` (no stacking). It shares one **new** file — `qml/components/LiveCoachingBanner.qml` — with the companion steam-coaching PR; the file is **byte-identical** on both branches (the banner takes `coach`/`enabled` as properties), so an add/add merge resolves cleanly.
- Verified: clean macOS build (app + tests).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,6 +301,7 @@ set(SOURCES
     src/ai/aiprovider.cpp
     src/ai/aiconversation.cpp
     src/ai/conductance.cpp
+    src/ai/liveshotcoach.cpp
     src/ai/shotanalysis.cpp
     src/ai/shotsummarizer.cpp
     src/ai/shotsummarizer_kb.cpp
@@ -489,6 +490,7 @@ set(HEADERS
     src/ai/aimanager.h
     src/ai/aiprovider.h
     src/ai/conductance.h
+    src/ai/liveshotcoach.h
     src/ai/shotanalysis.h
     src/ai/shotsummarizer.h
     src/history/shothistory_types.h
@@ -734,6 +736,7 @@ set(QML_FILES
 	qml/components/BrewDialog.qml
     qml/components/CircularGauge.qml
     qml/components/ConnectionIndicator.qml
+    qml/components/LiveCoachingBanner.qml
     qml/components/FavoritesListView.qml
     qml/components/PresetPillRow.qml
     qml/components/ValueInput.qml
--- a/qml/components/LiveCoachingBanner.qml
+++ b/qml/components/LiveCoachingBanner.qml
@@ -1,0 +1,117 @@
+import QtQuick
+import QtQuick.Layouts
+import Decenza
+
+// Live coaching banner. Binds to a live-coach service passed in via `coach`
+// (LiveShotCoach on the espresso page, LiveSteamCoach on the steam page) and
+// shows one short calm cue at a time while an operation runs. Fades in when a
+// cue is active, auto-dismisses after a few seconds (UI auto-dismiss timer is
+// allowed), and tints by severity using Theme tokens. Voice is gated on the
+// user's existing extractionAnnouncements preference so we don't double-announce.
+Item {
+    id: banner
+
+    // The live coach service to bind to (QML-marshalable Q_PROPERTYs: cueText /
+    // cueSeverity / cueActive / cueSpeak). Passed in by each mount so this banner
+    // is reusable across the espresso and steam pages without referencing any
+    // page-specific service directly.
+    property var coach: null
+
+    // Feature gate, passed in by each mount (each page has its own on/off pref).
+    // Named coachEnabled (not `enabled`) to avoid shadowing the built-in Item.enabled.
+    property bool coachEnabled: true
+
+    // Gate: feature pref + an actually-active cue + not yet auto-dismissed.
+    readonly property bool shouldShow: coachEnabled
+                                       && coach
+                                       && coach.cueActive
+                                       && !dismissed
+
+    // Set true by the auto-dismiss timer; reset whenever a new cue arrives.
+    property bool dismissed: false
+
+    // Severity -> Theme color token (no hardcoded colors).
+    function severityColor(severity) {
+        if (severity === "positive") return Theme.successColor
+        if (severity === "caution") return Theme.warningColor
+        return Theme.textColor  // "info"
+    }
+
+    implicitHeight: pill.height
+    height: shouldShow ? implicitHeight : 0
+    visible: opacity > 0.01
+    opacity: shouldShow ? 1.0 : 0.0
+    Behavior on opacity { NumberAnimation { duration: 220 } }
+
+    Accessible.role: Accessible.StaticText
+    Accessible.name: coach ? coach.cueText : ""
+
+    // React to cue changes: re-arm the auto-dismiss timer, and speak if asked.
+    Connections {
+        target: banner.coach
+        function onCueChanged() {
+            if (!banner.coach.cueActive)
+                return
+            banner.dismissed = false
+            if (banner.coachEnabled) {
+                dismissTimer.restart()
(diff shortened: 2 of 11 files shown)


loganross2001:
Companion PR: #1412 (steam coaching). Both are independent off latest `main` (no stacking); they share one new file, `qml/components/LiveCoachingBanner.qml`, which is **byte-identical** on both branches, so the add/add merge resolves cleanly whichever lands first.

loganross2001:
Ran `pr-review-toolkit` (code review, opus), fixed the findings, re-reviewed to clean.

### Fixed
- **Reserved-name collision** — `LiveCoachingBanner`'s feature-gate property was named `enabled`, shadowing the built-in `Item.enabled` (a CLAUDE.md QML gotcha). Renamed to `coachEnabled` (declaration + `shouldShow` + the cue-changed guard) and updated the EspressoPage mount.
- **No-puck cue false-firing on non-espresso profiles** — `LiveShotCoach`'s "Low pressure — did the puck seat?" caution fired on filter/pourover/tea profiles that legitimately run below the pressure floor (its post-shot twin `detectPourTruncated` gates on beverage type; the live one didn't). Rather than plumb beverage type through `MachineState`/`ProfileManager` (risky), I added a **self-contained gate**: only fire once the profile has actually targeted espresso-level pressure (`setPressureGoal` crossed the floor). Non-espresso profiles never target pressure, so the cue is suppressed; the failure mode is silence, which matches the coach's "silence over false alarms" design.
- Removed the unused injected `m_shotDataModel` (ctor param + member + the `new LiveShotCoach(...)` call); corrected the `pushSample` "O(1)" comment (it's an O(n) left-shift over a tiny window).

### Known limitation (conscious trade-off)
- The no-puck gate keys on the profile targeting pressure, so a pure **flow-profiled** espresso (which never sets a pressure goal) also won't get the no-puck cue. That's a blind spot, accepted deliberately: the failure mode is silence (matching the coach's silence-over-false-alarms design), and a universal gate would need the profile's beverage type plumbed into the coach. Noted in the code.

### Deferred (acknowledged)
- Unit tests for the private detectors (needs a `DECENZA_TESTING` friend hook — a follow-up); the rolling-window clock uses `MachineState.shotTime` rather than the sample's own timer (negligible drift).

Verified: re-reviewed clean, macOS build passes. The spoken-cue spacing is clock-based (not a guard Timer); the only Timer is the banner's UI auto-dismiss. Note: `LiveCoachingBanner.qml` is byte-identical to the companion steam-coaching PR #1412's copy (clean add/add).

skialpine:
## Functional weakness review — closing this PR

Thanks for the careful engineering here (event/clock-based governor, no network/DB in the hot path, Theme + i18n throughout). The problem isn't the code quality — it's the premise. **Real-time espresso "coaching" targets a moment when the user can't act, using detectors that are weaker and less profile-aware than the post-shot ones we already ship.** Closing for the reasons below.

### 1. There is nothing to act on in real time

On a DE1 the shot outcome is locked in *before* the pour by grind, dose, distribution, tamp, and puck screen. Once the profile is running the machine executes it automatically (on a GHC machine the shot isn't even app-initiated). Channeling, low pressure, and fast/slow flow cannot be fixed mid-shot — the only available action is stopping, which by the time these detectors fire (>3 s into the pour, after a clear flow jump across a 3 s window) just yields a worse-truncated shot rather than a saved one. `on-pace` and the phase milestone aren't advice at all, they're status.

The moment this information is actually **actionable is when planning the next shot** — adjust grind/dose/prep — and that's exactly where the existing **Shot Summary + quality badges** already deliver it, off the full curve.

### 2. The detectors re-introduce false-positive traps we already fixed

Each live detector is a conservative subset of a post-shot detector, minus the profile-awareness that made the post-shot version safe. It re-enters bugs that specific PRs closed:

- **Flow-vs-goal (`FLOW_DEVIATION_THRESHOLD = 0.4`)** re-enters the "deviation-from-goal = defect" fallacy that got the temperature-unstable badge **retired end-to-end in #1129** (root cause #1128: the gap was *design intent* on D-Flow / Extractamundo / TurboBloom and 9+ built-ins). It also ignores the profile-KB gating that made the grind detector safe — **#1230** (suppress grind badge on the Advanced Spring Lever family) and **#1232** (skip grind Arm 1 when the profile KB is unresolved). The live coach consults no KB, so it will fire "running slow — tight puck" on profiles whose flow deliberately sits under goal.
- **Channeling proxy** ignores the inclusion-window work that keeps the real detector from flagging every lever preinfusion — **#811** (mode-aware analysis to kill lever false positives) and **#1076** (exclude low-pressure soak phases from channeling windows). A "flow jump while pressure flat in a pressure phase" test with no windowing is precisely what those PRs replaced.
- **No-puck / low pressure** mirrors `detectPourTruncated` but ships without the **suppression cascade added in #922 (#903)**: when the puck fails, the other curves stop meaning what they normally mean, so the post-shot pipeline forces channeling/grind to false. The live coach has no cascade and can stack a "possible channeling" cue onto a shot that never built pressure. It also lacks the real-extraction gating from **#898** and **#1078**.

So the net is: **earlier + weaker analysis + no profile-identity awareness** — a downgrade that interrupts the user mid-shot, not a coach.

### 3. The post-shot path is the source of truth — see the docs

`docs/SHOT_REVIEW.md` is the reference for all of this:
- **§1** — badges the user sees + the always-present "Shot Summary" chip.
- **§2** — detector internals (2.1 channeling windowing / severity tiers, 2.2 grind, 2.4 pour-truncated + the suppression cascade).
- **§3** — the Shot Summary dialog.

These recompute on every shot load, are KB/profile-aware, and land the feedback at the point it can be used. Reproducing a thinner version of them on the live curve doesn't add coverage; it adds a false-positive surface with none of the suppression history baked in.

### If we want a genuinely live feature later

The "live" idea only earns its keep where the user has a real-time control loop — **steaming** (the companion PR), or manual/lever paddle profiles where flow is under live control. An automated espresso extraction has no such loop to coach. Closing this one; happy to revisit under that framing.

PR #1412, opened 2026-07-02
Live steam coaching: milk-weight-gated during-steam cues (banner + voice, off by default)

loganross2001:
## Live steam coaching — during-steam cues, gated on the actual milk weight

A local "reflex" layer that coaches milk steaming as it happens — one short line at a time, on an on-screen banner and/or spoken. **No AI calls, no network, no DB in the hot path.** Original concept and implementation by @loganross2001; substantially reworked after a multi-agent review and spec pass (openspec change `refine-live-steam-coaching`, included in this PR).

### What it does

`LiveSteamCoach` (`src/ai/livesteamcoach.{h,cpp}`) follows the elapsed steam time against the session's target duration and emits at most one cue at a time, each once per steam:

- **Stretch** — at the start: keep the tip near the surface to fold in air.
- **Roll** — past ~35%: submerge the tip to texture (skipped on very short steams).
- **Almost** — a heads-up near the end (by fraction or seconds remaining, whichever first).
- **Done** — at the *actual* end of steam flow, driven by a new `MachineState::steamFlowStopped()` event (exactly-once per steam, emitted where the steam shot timer really stops), never by a predicted time window — so clock/firmware drift can't lose it. An early manual stop is a deliberate abort and stays silent.

### Coaching requires the milk weight

A fixed preset duration says nothing about the milk actually in the pitcher — 200 mL against a 60 s preset is ruined long before "Almost there", and a coach narrating that timer would endorse ruining it. **All cues therefore require a milk-derived duration**: weight-timed steaming enabled + calibrated + the pitcher captured on the scale this session (`SteamPage.steamTimeoutScaled` feeds the coach's `durationMilkDerived` property). Without it, the banner shows a single informational pill — *"No coaching — milk weight not captured"* (never spoken) — and the coach otherwise stays silent, including "done".

### Gating & presentation

- **Two independent opt-ins, both OFF by default**: `Settings.app.steamCoachVisualEnabled` (banner) and `steamCoachAudioEnabled` (voice). Neither implies the other. Toggles live in a "Coaching" group at the bottom of the steam page's configuration section (global — not per-pitcher).
- **Banner** (`LiveCoachingBanner.qml`, purely visual): mounts in the steaming view's layout between the preset pills and the countdown; each cue persists until the next one replaces it or the steam ends (no auto-dismiss — easy to miss).
- **Voice**: the coach emits `speakRequested`, wired to a new `AccessibilityManager::announceCoaching()` that bypasses *both* accessibility voice gates (master switch and Voice Announcements) — the coach's audio toggle is its own opt-in, like the capture ding. Screen-reader-preference (no TTS overlap) is preserved. "Done" is assertive; the rest polite.

### Robustness (from the review rounds)

- `steamFlowStopped` is exactly-once (pending-flag armed at observed flow start, consumed by the first stop site; a mid-steam BLE disconnect disarms without emitting; a steam whose flow was never observed never arms).
- The coach never coaches a stopped flow, never trusts a shot clock it hasn't seen tick this session, and does zero per-tick work when fully disabled.
- Pre-existing SteamPage shutdown landmines fixed: 11 direct `ScaleDevice.connected` bindings threw TypeErrors at teardown; all now route through a null-safe `realScaleConnected` property.

### Tests

`tests/tst_livesteamcoach.cpp` (gating, milestone ordering/latches, milk-derived pill, clock-lag completion, exactly-once, early-abort silence, reset between operations, no settings-store pollution), plus producer-side `steamFlowStopped` emission tests in `tst_machinestate.cpp` and `announceCoaching` gate-bypass tests in `tst_accessibility_announcements.cpp`. Full suite green.


loganross2001:
Companion PR: #1411 (during-shot coaching). Both are independent off latest `main` (no stacking); they share one new file, `qml/components/LiveCoachingBanner.qml`, which is **byte-identical** on both branches, so the add/add merge resolves cleanly whichever lands first.

loganross2001:
Ran `pr-review-toolkit` (code review, opus), fixed the findings, re-reviewed to clean.

### Fixed
- **Shared banner reserved-name fix** — carried the `enabled`→`coachEnabled` rename from the companion PR #1411 (the property was shadowing `Item.enabled`). `LiveCoachingBanner.qml` stays **byte-identical** to #1411's copy; the SteamPage mount now binds `coachEnabled`.
- **Safety cue could be silenced** — the spoken "stop" caution (the cue you steam by, eyes off the screen) was subject to the generic 2 s spoken-cue spacing, so on a very short steam it could be silently downgraded to visual-only. `emitCue` now exempts `caution` severity from the spacing gate.
- Corrected the `cueSeverity` doc comment (the steam coach only emits `info`/`caution`; `positive` belongs to the shared banner).

### Deferred (acknowledged)
- The steam-coaching toggle sits under the "Extraction Announcements" (TTS) heading — it works without TTS, grouped with the other coaching accessibility settings; placement nit.
- The banner sets `Accessible.name` and also calls `announce()` — not a double-read in practice (the banner is a non-focusable `StaticText`, which isn't auto-spoken), consistent with #1411.

Verified: re-reviewed clean, macOS build passes. Independence holds — the branch references no #1411-only symbols; milestone logic (stretch→roll→almost→stop) and the clock-based spacing are unchanged.

PR #1414, opened 2026-07-02
History & graphs: honor the °C/°F display unit (follow-up to #1397)

loganross2001:
> **⚠️ Stacks on #1397 — please merge #1397 first.** This branch builds on the `Theme.cToDisplay` / `Theme.tempUnitSuffix` helpers introduced there, so until #1397 lands this PR's diff includes #1397's commits. Once #1397 is merged, this reduces to just the history/graph changes below. Opening now because it completes the °C/°F coverage your round-2 review flagged as a known follow-up.

## °C/°F for the history, graph & comparison views

#1397 converted the live read-outs to the user's temperature unit but deferred the historical/graph/comparison views. This finishes that coverage.

**The rule applied everywhere:** convert the value *and* the unit together — never a °F label on a raw Celsius number. Storage stays Celsius; only display converts, via the existing `Theme` helpers (which read `Settings.app.temperatureUnit`, so the labels refresh live on toggle).

### Graphs — axis is linear, so only labels convert
The temperature curves and axis range stay in Celsius (no re-scaling of the plot); because `cToDisplay` is a linear map, converting only the **tick-label values** and the **axis-title suffix** is exactly correct — a 90 °C point keeps its position, its tick just reads 194 instead of 90. (One cosmetic consequence: the °F ticks are non-round — 104/131/158/185/212 — since the underlying grid stays whole-Celsius. Round °F ticks would require an axis-rescale rework; happy to do that separately if you'd prefer it.)
- `ShotGraph.qml`, `ProfileGraph.qml`, `SteamGraph.qml`, `HistoryShotGraph.qml`: axis title → `tempUnitSuffix()`; temperature tick labels → `cToDisplay(value)`.

### Graph tooltips / inspect bar — value + unit converted together
- `HistoryShotGraph.qml`: the `temperature` / `mixTemp` crosshair + legend series now convert their displayed value via `cToDisplay` and their unit via `tempUnitSuffix()`. Non-temperature series (bar, mL/s, g, …) are untouched.

### Tables & detail views
- `ComparisonDataTable.qml`, `ComparisonShotTable.qml`, `ComparisonInspectBar.qml`, `ShotDetailPage.qml`, `PostShotReviewPage.qml`, `ShotHistoryPage.qml`: each temperature value + unit converted together.

Out of scope (unchanged): the descaling prose and the calibration sliders keep Celsius intentionally (native calibration ranges / documented reference), and `main.qml`'s speech expansion + the `Theme` helper source are the canonical-unit sources.

Build-verified on macOS with the test suite.


loganross2001:
Ran `pr-review-toolkit` (code review, opus), fixed the one coverage gap it found, re-reviewed to clean.

### Fixed
- **`GraphLegend` temperature tooltips still said "°C"** — they're shown in the shot-detail / post-shot views this PR converts, so in Fahrenheit mode the axis read °F while the tooltip said °C. Reworked so the unit follows the setting *and* stays translatable: the translated string now carries a `%1` placeholder filled with `Theme.tempUnitSuffix()` (e.g. "Basket temperature in %1.") — no baked-in unit, no broken string extraction.

### Deferred (acknowledged)
- `ProfileGraph`'s temperature axis converts its tick values but has no unit-suffix title at all (unlike `ShotGraph`/`SteamGraph`). That's pre-existing (it never had a unit title) and not a value/unit mismatch — optional parity add.

Verified: re-reviewed clean, macOS build passes. Reminder: this PR stacks on #1397 (uses its `Theme` helpers) — the diff slims to just these history/graph changes once #1397 merges.

skialpine:
Maintainer update (rebase + review fixes), force-pushed `f543570` → `5f79762`:

**Rebase:** #1397 has landed on main, so the stacked copy of its commit was dropped and the follow-up commit replayed onto `main`. This also restores the two `tst_settings.cpp` tests and the `SettingsMachineTab` `font.family` line from #1397's final revision, which the stale branch would have reverted. The PR now shows only the real follow-up diff.

**Review fixes (second commit):**
- **GraphLegend i18n:** `graph.tip.temp`/`graph.tip.mixTemp` keys bumped to `*2`. The fallback text gained a `%1` unit placeholder, but a previously cached translation of the old key contains a literal "°C" and no `%1` — `.arg()` would log `QString::arg: Argument missing` and the tooltip would never switch to °F. New keys can't resolve to stale cache entries.
- **Single-sourced the C/F conversion math:** the formulas previously lived twice — in `TemperatureDisplay::format()`'s lambdas (tested) and mirrored as JS one-liners in `qml/Theme.qml` (untested). They're now pure primitives in the `TemperatureDisplay` namespace (`cToDisplay`, `displayToC`, delta variants, `unitSuffix`), used by `format()` and exposed to QML via a thin `TemperatureDisplayBridge` context property. `Theme.qml`'s helpers delegate to it, still reading `Settings.app.temperatureUnit` in JS so bindings stay reactive on unit toggle (QML can't track property reads inside C++ methods).
- **Tests:** `tst_temperaturedisplay` extended to cover the primitives and the bridge — 36 pass locally.

skialpine:
### Code review

Found 1 issue:

1. The comparison-graph crosshair accessibility announcement spoke the raw Celsius value ("92.5 degrees") while the visible sibling `ComparisonInspectBar` (fixed by this PR) shows the converted value — in Fahrenheit mode a screen-reader user hears a different number than what is on screen, violating the PR's own rule to "convert the value *and* the unit together". Fixed in cd14f796 by converting via `Theme.cToDisplay` + `Theme.tempUnitSuffix()`, matching the announcer pattern this PR established in `HistoryShotGraph`.

https://github.com/Kulitorum/Decenza/blob/5f79762bad23b3c97cf7a314a29bb8093fc666e5/qml/components/ComparisonGraph.qml#L284-L286

All findings at or above the report threshold were fixed and pushed in cd14f796c048089ac0a8f154337edf7b1d7602e1.

<details>

- [Worth a look — fixed in cd14f796] `TemperatureDisplayBridge` was declared after the `QQmlApplicationEngine`, so it was destroyed before the engine at scope unwind — a latent dangling context-property pointer during teardown, benign today but inverting the file's declare-before-engine convention. Hoisted above the engine. (score: 50) https://github.com/Kulitorum/Decenza/blob/5f79762bad23b3c97cf7a314a29bb8093fc666e5/src/main.cpp#L2449-L2451
- [OK to skip — fixed in cd14f796 anyway] `ProfileGraph.qml` comment "tickCount: 5 on a 80-100 °C range" no longer mentioned that the label text now converts to the display unit (176-212 in °F mode). Comment extended. (score: 50) https://github.com/Kulitorum/Decenza/blob/5f79762bad23b3c97cf7a314a29bb8093fc666e5/qml/components/ProfileGraph.qml#L184-L188

</details>

Checked for bugs, CLAUDE.md compliance, historical context from #1397, prior review guidance, and code-comment accuracy. Conversion math (absolute vs delta), axis gating (temperature vs weight), QML binding reactivity on unit toggle, and the i18n key bump were all verified clean.

PR #1415, opened 2026-07-05
Shot plan: per-instance display format (Sentence / Compact / Stacked / Plain)

loganross2001:
Adds a per-instance **format** option to the Shot Plan widget, chosen in the layout editor (native gear + web editor), so users can pick how the plan reads:

- **Sentence** (default) — unchanged; existing widgets render byte-identically.
- **Compact** — short `·`-separated fragments, no sentence.
- **Stacked** — sentence on line 1, details wrap onto following lines.
- **Plain** — a dot-free recipe line (e.g. *"Pull a 41.8g coffee shot using 19.0 grams of Olympia Blooming Espresso coffee beans"*); it ignores the per-field toggles, which grey out when Plain is selected.

Also included:
- **Overflow fix** — the text is width-capped to the tile, so long plans wrap/elide instead of bleeding past the widget.
- **Implicit-width fix** — the widget reports its *natural* (uncapped) width, so the tile can't ratchet ever-smaller through the `availableWidth: root.width` feedback loop.
- Wired identically in the native `ScreensaverEditorPopup` (segmented control) and the web editor (`<select id="spFormat">`), persisted per-instance as `shotPlanFormat` and validated to the four values on both load paths (defaults to `sentence`).

**Default unchanged:** the `sentence` output is byte-identical to current — the new `blockSep` collapses to the existing separator on the non-stacked path.

**Known limitation (disclosed):** `stacked`/`plain` are the first multi-line formats. The fixed-height status / lower-mid bars don't clip, so a 3-line plan placed there can overflow into neighbours — they're best in the center / growable zones. A follow-up could add `clip: true` to those two fixed bar containers; I left that out to keep this PR to the plan files and avoid touching zone layout. Happy to add it if you'd prefer.

Reviewed with the full `pr-review-toolkit:review-pr` suite to convergence over two rounds — code, silent-failure, comments, type-design, and test-coverage. No test added: the seven existing per-instance `shotPlanShow*` keys are likewise unpinned, and this rides the same generic `setItemProperty` mechanism (view-layer QML with no C++ test seam).

### DIFF
--- a/qml/components/ShotPlanText.qml
+++ b/qml/components/ShotPlanText.qml
@@ -159,13 +159,63 @@ Item {
         return parts.join(sep)
     }
 
+    // "plain" format: a fixed, dot-free two-line recipe template — IGNORES the per-field toggles.
+    //   Line 1: "Brew <yield> of <beverage> from <dose> grams of <roaster> <coffee> coffee beans"
+    //   Line 2: "(Use <profile> at <temp>)"
+    // Single output weight (no "36.0 → 41.8" arrow), no · separators. blockSep joins the two lines
+    // (a line break for display, ". " for the spoken/a11y string so it reads as two sentences).
+    function _buildPlain(fmt, blockSep) {
+        var _ = TranslationManager.translationVersion
+        void(Settings.app.temperatureUnit)
+        // A cleaning/descale run still takes precedence — no recipe, just the warning.
+        if (_isCleaning) {
+            return (profileName.length > 0)
+                ? TranslationManager.translate("shotplan.sentenceCleaning", "Cleaning run with %1 — no coffee in the portafilter!").arg(fmt(profileName, true))
+                : TranslationManager.translate("shotplan.cleaningNoProfile", "Cleaning run — no coffee in the portafilter!")
+        }
+        // Line 1 — brew + (optional) bean clause.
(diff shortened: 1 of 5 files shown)


skialpine:
Thanks for this, Logan — the thinking here was genuinely good, and two of its ideas have now shipped, so I'm closing this as superseded rather than asking for a conflicted rework:

- **#1426** landed a larger rework of the same files prompted by user feedback: the display options became a drag-to-reorder item list, Sentence style became a per-instance toggle (OFF gives your *Compact*), and the overflow/implicit-width problems you identified got fixed with the same width-cap + natural-implicit approach you used — you diagnosed the same feedback loop independently, which was good confirmation.
- **#1428** ports your **Stacked** format onto the new architecture as a "Stacked details" toggle, with credit to you and this PR in the commit. Same rendering trick (line-break block separator, display path only, a11y stays one sentence).

**Plain** is the one piece deliberately not carried over: its ignore-the-toggles design made sense against seven checkboxes, but now that users compose the item list themselves, a format that discards their configuration reads as a contradiction. If the prose-without-dots tone is the part you care about, a plain-style template that *honors* the item list would be very welcome as a fresh PR on top of the new architecture — happy to point at the seams (`ShotPlanText._build` now takes the ordered list; `ShotPlanConfig.js` holds the shared config rules).

Sorry the ground moved under this one mid-review, and thanks again — Stacked was the standout idea and it's in.

PR #1417, opened 2026-07-05
Advisor: order grind/ratio before temperature for sour/bitter shots

loganross2001:
## Order grind/ratio before temperature for sour/bitter shots

The dialing advisor's "Common Espresso Patterns" told it to reach for **temperature first** on a sour or bitter shot — the Sour Shot fix literally led with *"Increase temp 2°C, or pull longer, or grind finer."* Grind and ratio are the primary espresso levers; temperature is a smaller, later adjustment. This reorders the four temp-first fixes to lead with grind/ratio and demotes temperature to a later step.

**What changed (prompt text only, `src/ai/shotsummarizer.cpp`):**
- Reordered the **Sour Shot**, **Bitter Shot**, **Bitter/Harsh**, and **Sour/Sharp Acidity** fixes to lead with grind/ratio; temperature "only after."
- Added a short **"Lever ordering"** note under Common Espresso Patterns.
- Folded a channeling **guard-rail** into the Sour fix: *if the finer grind itself makes the shot channel, finer extracts less — go coarser instead* (the Matter 2020 non-monotonic result).
- Added the missing `(one at a time!)` to the Bitter fix.

**Framed as ordering, not a prohibition.** It explicitly defers to a profile deliberately built around a specific temperature ("Profile Intent is the Reference Frame") — so temperature-designed profiles are unaffected. Backed by both standard dial-in practice and the temperature sensory study (Nature *Sci Rep* 2020: no appreciable sensory effect at fixed TDS/EY).

Ran the review-pr suite (code-reviewer + comment-analyzer) to convergence — it caught a backwards ratio gloss and a stale cross-reference, both fixed. No logic touched.

*Companion note:* I also explored adding a computed days-off-roast number to the `beanFreshness` block, but the block already carries the roast/thaw dates + an aging instruction, and baking a day-count into a cached prompt risks staleness — so I left it out. Happy to revisit if you'd want it.

### DIFF
--- a/src/ai/shotsummarizer.cpp
+++ b/src/ai/shotsummarizer.cpp
@@ -1312,6 +1312,8 @@ The "Common Espresso Patterns" section below tells you the **direction** of grin
 
 ## Common Espresso Patterns
 
+**Lever ordering.** Grind and ratio are the primary espresso levers — settle those first. Temperature is a smaller, later adjustment; don't reach for it to fix sourness or bitterness until grind and ratio are dialed. (Exception: when a profile was deliberately built around a specific temperature, respect the author's intent — see "Profile Intent is the Reference Frame" below.)
+
 ### The Gusher
 - **Symptoms**: Very fast shot (<20s), flow way above target, thin/watery taste
 - **Cause**: Grind too coarse or severe channeling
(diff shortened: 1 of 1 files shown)


loganross2001:
Completed the full `pr-review-toolkit` suite on this (code-reviewer, comment-analyzer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer), to convergence:

- **code-reviewer** — caught a stale "below" cross-reference (the Profile-Intent note splices in *above*); fixed to directionless.
- **comment-analyzer** — caught a backwards ratio gloss ("pull longer (tighten the ratio)" — pulling longer *lengthens* it); fixed to "lengthen", and disambiguated the guard-rail from the Channeler pattern.
- **silent-failure-hunter** — clean; verified the edits sit inside the `R"( )"` literals with no `)"` terminator collision that could truncate the assembled prompt.
- **type-design-analyzer** — N/A (prose only, no type/API surface).
- **pr-test-analyzer** — no test change warranted; confirmed no golden/snapshot test asserts this prose and the two `tests/data/dialing/*_pre_change_baseline.json` fixtures are intentionally frozen "before" snapshots (per their README), so nothing breaks.

No blockers remaining.

skialpine:
Reviewed and pushed a follow-up commit (653492d0) with small wording fixes — the science and the core reordering are solid, these just resolve internal-consistency issues the review turned up:

- **Channeling caveat now scoped** so it doesn't fight The Channeler entry two bullets above: re-check puck prep first, and "go coarser" is now "step back to the previous grind setting" (revert the finer move, not open-ended coarsening).
- **Filter section made self-consistent**: Astringent and Thin/Watery still listed temperature as a coequal fix while your two edited entries demoted it. Added a Lever ordering note atop Common Filter Issues (with the grind-pinned-profile exception, since the filter prompt elsewhere says to try temperature *before* grind on coarse-by-design profiles) and applied the same "only after" phrasing to those two entries.
- **"One change at a time" moved to the front** of the Sour/Bitter fix lines so it governs all options, not just the temp step.
- **Exception clause tightened** to profiles whose *description* calls out temperature as central — every DE1 profile technically specifies temperatures, so the old wording let a literal-minded model route around the rule.
- Added a C++ comment marking the verbatim "Profile Intent is the Reference Frame" cross-reference coupling.

Thanks for the well-sourced PR — the Matter 2020 guard-rail in particular is a great addition.

Issue #1418, opened 2026-07-05
Bag grind memory doesn't record which profile it was dialed for (follow-up to #1413)


skialpine:
Follow-up to #1413 (request 3, raised by @fredphoesh). The two display items from that issue (grind in the shot detail metrics row, equipment card placement on the review page) are handled separately; this issue tracks the remaining design problem.

## Problem

A coffee bag stores the last-used grind setting and RPM, and selecting a bag restores them. But the bag does **not** record which *profile* that dial-in was for. Switch from a Cremina-style profile to Turbo Turbo on the same bag and the restored grind is silently wrong — there is no indication of what the number was dialed against, so users who rotate profiles can't trust it.

As Mark put it: a grind size without mention of what preset it's intended for assumes the user never changes presets.

## Ideas discussed (not yet designed)

- **Stamp the dial-in context**: record the current profile alongside `grinderSetting`/`rpm` when the bag's dial-in is written (the write-through moment already knows it). One column, factual, no inference.
- **Quiet Shot Plan qualifier**: when the active profile differs from the stamped one, render e.g. `Grind 5.5 (dialed for D-Flow)` instead of the bare number. A factual hint, not a red/flashing alert.
- **"Recommend grind for this bean on this profile"**: this is the grinder-calibration advisor's job — covered by the existing cross-profile calibration work for #1223 (within-coffee anchors, extrapolation caps, directional-only guidance when data is thin). This issue should not duplicate that; at most the qualifier can link to it.

Explicitly rejected during discussion: a per-bag "lock grind to profile" checkbox (new setting for something better defaults can solve), and flashing/red alerting (fact over alarm).

## Open questions

- Stamp on the bag vs. derive from the last shot for that bag in history (which already stores grind + profile per shot)?
- What happens on bags dialed before the stamp exists (no context recorded)?
- Should the qualifier also appear in the Change Beans dialog / bag cards, or only the Shot Plan?

fredphoesh:
Just to say,  WRT the relationship with Grind Setting and the way History, Shot Plan and Recipe relate seems logical and intuitive to me... that

1/ if you do a shot and realise you had not entered your Grind Setting properly, so you go to History and change the setting - that will update both your Shot Plan and Recipe. This is making the (usually correct) assumption that your last shot is better dialled in than anything before, so it keeps the Grind Setting for your future shots.
[...]

PR #1419, opened 2026-07-05
Grind visibility: metrics-row grind cell on shot detail, equipment card last on review page

### LINKED ISSUE #1413: [Feature] Grind Setting display in several places

Hi Jeff, long time no see ;)
There are a few parts to this, and they are not really bugs or particularly "new" features

1/ the way Grind Size and RPM have been split now means the full "grind setting" is not visible on the Shot Plan. So the request is to have RPM show up on Shot Plan if you have set it.
2/ looking at your last shot or browsing recent shots in History, grind setting is almost invisible... it is very weirdly situated, in tiny writing, in the Equipment pill. It really does not belong there and deserves TOP LEVEL status. A very common reason to look at past shots is to decide what to do next, to change grind or not. So it's VERY important to have grind setting visible at the very top, above the graph.
3/ I *_quite_* like how changing beans will change your grind setting, I THINK to display the last grind setting you used with that bean. That's pretty helpful, but I was initially shocked to see grind setting attached to/changed by selecting Bean pills. Of course that system is very flawed, as you may have been using a very similar bean, but with Cremina on one and Turbo Turbo on the other, and changing beans gives you no indication at all as to what that grind setting is based on, so you cannot trust it as useful info for your next shot. I'm not sure what an elegant solution might look like, but it's clumsy as it is. imo.

Thanks fella, Mark.

### Proposed solution

1/ show rpm in shot plan
2/ show grind size at the top of last shot/history etc
3/ don't know

### Alternatives considered

_No response_

### Additional context

_No response_

skialpine:
Closes #1413 (requests 1 & 2 — request 1, RPM in the Shot Plan, already shipped in #1396; request 3 is tracked separately in #1418).

## What changed

**Shot detail page** — the top metrics row gains a **Grind** cell between Dose and Output, showing the shot's grind setting with the RPM as a caption suffix (`5.5 · 340 rpm`, matching the Shot Plan's format). The cell hides entirely for shots with no recorded grind (row reflows to five metrics), and long free-text settings elide — the Equipment card below keeps the un-elided grind/RPM line.

**Post-shot review page** — the equipment identity card (grinder + basket + puck prep, with Change Equipment) moves from the top of the field grid to the very bottom, after Preset and Shot date. Editable per-shot fields first, hardware context last. Pure child reorder, no behavior change.

## Verification

- Qt Creator build clean, no new warnings.
- Manually verified: Grind cell renders correctly with and without RPM, hides on grind-less shots, elides long values; review-page grid order is Grind setting → RPM → Barista → Preset → Shot date → equipment card with Change Equipment/autosave/undo intact; no new QML warnings in the app log.
- Wiki Manual.md text updates are staged locally in the wiki clone (uncommitted, alongside the existing buffered manual edits).

OpenSpec change `improve-grind-visibility-shot-review` is included (proposal/design/specs/tasks); archive will land as the last commit before merge.


fredphoesh:
Hi Jeff, I have installed 1.81 on PC and Android, and
1/ Shot Plan "sentence" aspect is taking up space unnecessarily, and the shot plan does not fit, even with a few data options turned off... IMO the words like Brew... of Espresso... are unnecessary.
2/ The the shot detail page still does not show your grind setting without the user having to search for it... a prime reason to look at a recent shot is to recall whether it was dialled in adequately, and then to make a decision about grind setting for the next attempt. So Grind setting is probably THE most important and basic bit of information needed, so should be at the top, above the graph so you don't have to scroll to find it.  Even more importantly, when swiping through history to quickly compare graphs, when you are then required to scroll down to find grind setting it kind of defeats the purpose and flow, as you should be able to see the graph, profile, bean and grind size at the highest level, as those are the things you are most needing to know when dialing in/making tweaks.

There is a lot of empty space for this, in the top where it says Shot detail, then loads of empty blue space around that, also next to and below where it says the Preset name (D-Flow / 42 (88°C) in mine)

Screenshot_20260705_222952
Screenshot_20260705_222820

Screenshot_20260705_223009

PR #1421, opened 2026-07-05
Don't flag 'First step skipped' when frame 0 exited on its own exit condition

loganross2001:
## Don't flag "First step skipped" when frame 0 exited on its own condition

**The bug:** `detectSkipFirstFrame` flags a first frame that runs shorter than half its *configured* max duration — but it never checks *why* the frame ended. DE1 preinfusion/fill frames are designed to exit early on their exit condition (a pressure/flow/weight target), long before their configured max. So a preinfusion frame set to e.g. 8 s that correctly exits at ~1.2 s on its pressure target gets flagged "First step skipped" **every shot**, even though it executed exactly as designed. Fully reproducible for any profile whose first frame has an exit condition.

**The fix:** the first non-zero frame's marker already records `transitionReason` (the reason the preceding frame exited). Gate the detector on it — a `pressure`/`flow`/`weight` exit means frame 0 met its own target and was *not* skipped. Only a `time`/empty (old-data) reason falls through to the existing duration checks. Mirrors the signal `analyzeFlowVsGoal` already uses; recompute-on-load is preserved (`transitionReason` is persisted).

**Coupled change worth your attention — `maincontroller.cpp` transition-reason inference:** the detector can only trust `transitionReason` if `pressure`/`flow`/`weight` always means a *confirmed* sensor exit. Today the ambiguous `else` (exit configured, but neither threshold confirmed nor time expired) **guesses** `pressure`/`flow` from `exitType` (it logs "ambiguous"). That guess, treated as truth, would let the new guard suppress a *genuinely* too-short frame. So this PR changes that ambiguous case to record **`"time"`** instead of a guessed sensor reason.

This is a **shared-semantics change** — `transitionReason` is also read by `detectGrindIssue`/`analyzeFlowVsGoal`, the shot models, the graph marker letters, and the AI summarizer. I traced them all: `"time"` is already a valid, handled value everywhere (no blank letters, no missed grind-trim — the trim only fires on *confirmed* pressure), and the full suite stays green (89/89). The one honest tradeoff: the ambiguous case is often a real sensor exit delayed by BLE sample lag, so it will now show **"Time exit"** in the UI/summarizer where it previously showed a (usually-correct) pressure/flow guess. That's cosmetic and logged — versus the old behavior, which *silently suppressed a real defect*.

If you'd prefer to keep the UI's sensor-exit labeling, a cleaner alternative is a distinct `"unconfirmed"` reason (flags identically in the guard, but the UI/summarizer can render it as an inferred sensor exit rather than "Time exit"). Happy to switch to that — your call on your data model.

**Tests:** added rows to `skipFirstFrameDetection()` — a 0.3 s first frame with `pressure`/`flow`/`weight` → not flagged; with `time` → still flagged (proves the gate is reason-specific, not a blanket off-switch). `docs/SHOT_REVIEW.md` §2.3 updated. Ran the full `pr-review-toolkit` suite (code/silent-failure/comments/types/tests) to convergence.

### DIFF
--- a/docs/SHOT_REVIEW.md
+++ b/docs/SHOT_REVIEW.md
@@ -343,7 +343,16 @@ other reason). Consumers wanting "verified clean" specifically must read
 ### 2.3 Skip first frame
 
 `detectSkipFirstFrame` operates on phase markers only — no curves needed. Two
-distinct branches inside one function, picked from the marker stream:
+distinct branches inside one function, picked from the marker stream.
+
+**Exit-condition guard (checked first).** Before either branch, if the first
+non-zero frame's marker `transitionReason` is `"pressure"`, `"flow"`, or
+`"weight"`, the detector returns `false` — frame 0 exited on its own exit
+condition, so it executed as designed and was not skipped. DE1 preinfusion/fill
+frames routinely exit far earlier than their configured max duration (that is
+their purpose), and without this guard the duration checks below flag them as
+"skipped." Only a `"time"` or empty (old-data) reason falls through to the
+branches. This mirrors how `detectGrind` already uses `transitionReason`.
 
 **FW-bug branch** — frame 0 was never observed before a non-zero frame:
 returns `phase.time < 2.0`. The 2-second window matches the de1app Tcl
--- a/docs/SHOT_REVIEW.md
+++ b/docs/SHOT_REVIEW.md
@@ -347,12 +347,16 @@ distinct branches inside one function, picked from the marker stream.
 
 **Exit-condition guard (checked first).** Before either branch, if the first
 non-zero frame's marker `transitionReason` is `"pressure"`, `"flow"`, or
-`"weight"`, the detector returns `false` — frame 0 exited on its own exit
-condition, so it executed as designed and was not skipped. DE1 preinfusion/fill
-frames routinely exit far earlier than their configured max duration (that is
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Verified in particular: the guard cannot suppress the FW-bug branch (a firmware-skipped frame 0 produces an empty `transitionReason` because `m_lastFrameNumber` is `-1` at shot start, and the ambiguous-exit branch now records `"time"` instead of a guessed sensor reason — without that coupled `maincontroller.cpp` change, a genuinely skipped frame with a configured exit condition could have been recorded as a guessed `"pressure"`/`"flow"` and masked the true positive). Old persisted shots (empty reason) fall through to the existing duration checks unchanged, and the new `docs/SHOT_REVIEW.md` §2.3 section matches the code.

<details>

- [Worth a look] `docs/SHOT_REVIEW.md` §2.2 (grind detector) doesn't mention the second-order effect of the `maincontroller.cpp` change: ambiguous exits previously guessed as `"pressure"` fed the Arm 1 limiter-tail trim (https://github.com/Kulitorum/Decenza/blob/eea3276e183c84f40e46b1a712bf4e30aef0bb71/src/ai/shotanalysis.cpp#L412-L419); those shots now record `"time"` and lose the trim. §2.2 remains factually accurate — a one-line cross-reference would close the loop (score: 55)

</details>

PR #1422, opened 2026-07-06
Unconfirmed transition reasons: keep the grind limiter-tail trim and sensor labels on ambiguous frame exits

skialpine:
## Why

PR #1421 fixed the "First step skipped" false positive by making the skip-first-frame guard trust `transitionReason` as ground truth, and — to keep that trust honest — recorded the ambiguous frame-exit case (exit configured, threshold not confirmed by the transition sample, time not expired) as `"time"` instead of a guessed sensor reason. That collapse traded away two things:

1. **Grind detector**: `analyzeFlowVsGoal`'s limiter-tail trim (drop the last 1.5 s of a flow-mode window when the next marker says `"pressure"`) stopped firing on ambiguous exits, which are usually real pressure-limiter engagements whose threshold crossing fell between ~5 Hz BLE samples. Including that limiter-suppressed tail biases the flow-vs-goal delta toward "too fine".
2. **UI**: the transition pill and graph marker letters show "Time exit"/[T] where a sensor exit almost certainly happened.

The root issue: the two decision-making consumers need **opposite defaults** for the ambiguous case. The skip guard must NOT suppress on it (a firmware-skipped frame lands in the same branch — the very bug the detector catches), while the grind trim SHOULD fire on it (trimming a maybe-clean tail is harmless under the existing ≥1 s post-trim floor; including a contaminated one is not). One value cannot express both. This implements the distinct-unconfirmed-reason alternative offered in #1421's description.

## What

- `MainController` records `pressure_unconfirmed`/`flow_unconfirmed` (hint from `exitType`, as the pre-#1421 guess did) in the ambiguous branch.
- `detectSkipFirstFrame` guard: **no logic change** — exact-match on confirmed `pressure`/`flow`/`weight` only, so unconfirmed still falls through and flags. #1421's safety property is preserved by string comparison, not convention.
- `analyzeFlowVsGoal`: limiter-tail trim fires on `pressure` OR `pressure_unconfirmed`; `flow_unconfirmed` does not trim.
- Displays render unconfirmed as its sensor equivalent (letter, text, color): `ShotGraph.qml`, `HistoryShotGraph.qml`, `EspressoPage.qml` pill (reuses existing translation keys), both embedded-JS switches in `shotserver_shots.cpp`.
- `SHOT_REVIEW.md` §2.2/§2.3 document the vocabulary and the deliberate opposite polarities (includes the cross-reference note deferred from #1421's review); `HistoryPhaseMarker` field comment updated.

## Data / migration

None. The field is a free string end-to-end; old guessed `pressure`/`flow` values and the brief post-#1421 `"time"` window keep their recorded meaning. All display switches degrade gracefully on unknown values, so rollback is fail-safe (unconfirmed values would behave exactly like #1421's `"time"`).

## Tests

- Skip detector: short first frame with `pressure_unconfirmed`/`flow_unconfirmed` still flags (guard is confirmed-only).
- Grind trim: fires on `pressure_unconfirmed` (lever-shape fixture stays clean), does not fire on `flow_unconfirmed` (sustained dip stays in, badge fires).
- Full suite green: 91/91 `tst_shotanalysis`, 645 passed / 0 failed / 0 warnings overall.

Includes the OpenSpec change (`unconfirmed-transition-reason`) with proposal/design/specs; archive will land as the final commit before merge.


skialpine:
### Code review

Verified in particular: every `transitionReason` consumer in the codebase was traced — the skip-first-frame guard remains exact-match on confirmed values so `pressure_unconfirmed` structurally cannot suppress it (https://github.com/Kulitorum/Decenza/blob/48c19a019b7439b8cde6c88ab228900b4b50a9c2/src/ai/shotanalysis.cpp#L700-L704); no `.contains()` check anywhere treats an unconfirmed value as confirmed; `exitType`'s closed value set makes the pressure/flow hint derivation collision-free; DirectController correctly keeps recording confirmed reasons only (it verifies exits itself before recording); and all five display switches plus pass-through persistence handle the new values or degrade gracefully. The SHOT_REVIEW.md §2.2/§2.3 updates match the code, including the deliberate opposite-polarity cross-references, and deliver the doc note deferred from #1421's review.

Two cosmetic review observations (stale one-line test comment, wrong file name in the OpenSpec proposal prose) were fixed on the branch in 09a92cb5 before merge.

PR #1439, opened 2026-07-07
Show a detail line on Bean Base search results to tell duplicates apart

skialpine:
## Problem

Searching for a bag in the Add Bag dialog can show the same coffee several times with nothing to differentiate the rows. Visualizer's canonical DB genuinely holds near-duplicate submissions — e.g. `?q=hometown` returns three "Sweet Bloom Coffee · Hometown Blend" entries with distinct canonical UUIDs, differing only in roast level, elevation, and tasting notes (one comes from a `hometown-blend-copy` product URL). The dialog row only showed roaster + name, so they rendered identically.

## Fix

- `UnifiedBeanSearchModel::mergeLanes()` now composes a `detail` string on canonical rows (tiers 1–2) from the entry's descriptive blob: **roast level · origin · tasting notes** — the fields that actually vary between duplicates. Exposed as a new `DetailRole`.
- `ChangeBeansDialog.qml` shows it in the existing subtitle slot (where inventory rows show their roast date), and appends it to the row's accessible name so TalkBack/VoiceOver users can tell the rows apart too.

Each duplicate stays its own row on purpose: distinct canonical ids are distinct pick targets, and the detail line is what lets the user choose the right one.

## Testing

- New unit test `mergeLanesCanonicalDuplicatesCarryDetail` pins the composition and that same-name entries stay distinct rows (attrs missing → empty detail, subtitle hidden).
- `tst_CoffeeBags`: 36 passed, 0 failed, no warnings.

### DIFF
--- a/docs/CLAUDE_MD/BEAN_BASE.md
+++ b/docs/CLAUDE_MD/BEAN_BASE.md
@@ -46,8 +46,8 @@ Entries are QVariantMaps (QML lingua franca; deliberately not a C++ value type).
 | `visualizerCanonicalId` | = id | — | — |
 | `source` | "visualizer" | "beanbase" | — |
 | `roasterName`, `roastName` | ✓ | ✓ | — |
-| `degree, origin, region, producer, variety, process, harvest, tastingNotes` | — | ✓ | ✓ (remapped from Visualizer columns) |
-| `elevation` (display string) | — | — | ✓ |
+| `degree, origin, region, producer, variety, process, harvest, tastingNotes` | ✓ (remapped from Visualizer columns via `kAttrMap`) | ✓ | ✓ (same values; `canonicalDetails` is a local re-emit) |
+| `elevation` (display string) | ✓ | — | ✓ |
 | `link` (roaster product page) | ✓ (from `url`) | ✓ | — |
 | `minElevationM/maxElevationM` (int), `image, beanType, description, tastingTags, generalTags, soldout, available, roasterRegion, roasterCountry` | — | ✓ | — |
 
--- a/qml/components/ChangeBeansDialog.qml
+++ b/qml/components/ChangeBeansDialog.qml
@@ -634,7 +634,10 @@ Dialog {
                                 Text {
                                     Layout.fillWidth: true
                                     visible: text.length > 0
-                                    text: model.roastDate || ""
+                                    // Bean Base rows have no roast date; their detail line
+                                    // (roast level · origin · notes) is what tells the
+                                    // canonical DB's same-name near-duplicates apart.
+                                    text: model.roastDate || model.detail || ""
                                     font: Theme.captionFont
                                     color: Theme.textSecondaryColor
                                     elide: Text.ElideRight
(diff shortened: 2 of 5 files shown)


skialpine:
### Code review

Found 3 issues:

1. Tier-1 rows re-persist already-corrupted blobs: a non-empty history blob always wins over the fresh canonical entry (`histBlob.isEmpty() ? entryJson : histBlob`), so a history snapshot that already carries the broken `{"link":…}` artifact (the exact corruption this PR prevents) gets copied into a newly created bag instead of being replaced by the full entry. Treating a link-only blob as effectively empty would close the loop.

https://github.com/Kulitorum/Decenza/blob/64e402bd8d58f8f00f922c8de6d71719178f522d/src/history/unifiedbeansearchmodel.cpp#L312-L318

2. `model.detail` (Bean Base free text: roast level, origin, tasting notes) is rendered in a `Text` with no `textFormat`, so Qt's default `Text.AutoText` can interpret markup-looking content; the same file escapes this class of Bean Base data elsewhere via `Theme.joinWithBullet` (per `Theme.qml`'s escaping rationale). The row title (`primaryText`, also Bean Base data) shares the gap. (CLAUDE.md says "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'".) `textFormat: Text.PlainText` on both fixes it.

https://github.com/Kulitorum/Decenza/blob/64e402bd8d58f8f00f922c8de6d71719178f522d/qml/components/ChangeBeansDialog.qml#L636-L644

3. `docs/CLAUDE_MD/BEAN_BASE.md`'s key-mapping table marks `degree/origin/…/tastingNotes` as absent ("—") from canonical `search()` entries, contradicting both the doc's own line 37 ("the search response already includes every descriptive field") and `kAttrMap` in `beanbaseclient.cpp` — and this PR's `canonicalDetail()` is the first code to rely on exactly those fields from that producer, so the stale row is now actively misleading.

https://github.com/Kulitorum/Decenza/blob/64e402bd8d58f8f00f922c8de6d71719178f522d/docs/CLAUDE_MD/BEAN_BASE.md#L48-L50

---

<details>

- [Worth a look] `DetailRole`'s doc comment says "Canonical rows only", but `detail` is set before the tier split, so `Tier::HistoryCanonical` rows carry it too — ambiguous against the `Tier::CanonicalOnly` enum name, and the Tier-1 case is untested (score: 50)
- [OK to skip] Test comment cites "Hometown Blend" while the fixture uses `roastName: "Hometown"` — traceability nit only (score: 50)

</details>

skialpine:
All 3 findings (and both below-threshold notes) addressed in 7937c3f88fb4dabbeeece85e4e1a92827b93546e: link-only history blobs no longer shadow the fresh canonical entry (new test `mergeLanesTier1ReplacesLinkOnlyBlob`), row title/subtitle now render `Text.PlainText`, BEAN_BASE.md key table corrected, `DetailRole` comment made tier-precise with a tier-1 assertion. `tst_CoffeeBags`: 39 passed, 0 failed.

PR #1442, opened 2026-07-08
Shot Plan: 'Final yield only' option (hide the profile-default → arrow)

loganross2001:
## What

Adds a per-Shot-Plan-widget option **"Final yield only"** that shows just the effective target yield (e.g. `40.0g`) instead of the profile-default → target arrow (e.g. `36.0 → 40.0g`). Default **off**, so existing layouts are unchanged.

## Why

#842 added the yield override arrow to match the temperature treatment, and it's genuinely useful for spotting a deliberate override. But when you brew by ratio (or keep a standing yield override), the profile's static `target_weight` (often 36 g) is just noise next to the number you actually care about — the plan permanently reads `36.0 → 40.0g` and the leading `36.0 →` never changes. This gives people who don't want that a one-tap way to collapse it to the final figure, without removing the arrow for those who like it.

It slots naturally into the per-instance Shot Plan config from #1426 — same editor, same storage pattern as `shotPlanSentence` / `shotPlanStacked`.

## How

New per-instance boolean `shotPlanYieldTargetOnly` (absent / `false` = current behavior):

- **`ShotPlanText.qml`** — `_yieldStr` skips the arrow branch when the flag is set, returning only `targetWeight`. Both the sentence and fragment render paths consume `_yieldStr`, so both honor it.
- **`ShotPlanItem.qml`** — reads `modelData.shotPlanYieldTargetOnly` and passes it to the compact and full `ShotPlanText` instances.
- **`ScreensaverEditorPopup.qml`** (the Shot Plan bespoke editor) — adds the toggle with load/save, wires the live preview, and disables it when the *Dose & yield* item isn't in the plan (nothing to affect).
- **`shotserver_layout.cpp`** — mirrors the toggle in the web layout editor (checkbox + load + save), matching the other `shotPlan*` keys.

## Notes

- Non-breaking: default off; no migration; `setItemProperty` already accepts arbitrary boolean keys.
- Scope: 4 files, +32/−2.
- Build-verified standalone on `main` (Android debug, arm64) — the branch is `main` + this one commit, barista/fork code excluded.

### DIFF
--- a/qml/components/ShotPlanText.qml
+++ b/qml/components/ShotPlanText.qml
@@ -37,6 +37,10 @@ Item {
     // Sentence mode only: render the detail tail on its own line(s) below the
     // sentence (display path — the a11y text stays one joined sentence).
     property bool stacked: false
+    // Yield rendering: when true, show only the effective target yield (e.g.
+    // "40.0g") and suppress the "profileDefault → target" arrow. The default
+    // (false) keeps the arrow, mirroring the temperature-override treatment.
+    property bool yieldTargetOnly: false
     // Wrap budget before eliding: 2 in the full-size widget, 1 in compact bars.
     property int maxLines: 2
 
@@ -63,10 +67,11 @@ Item {
     // --- Per-item segments (empty string = hidden). ---
     // Dose & yield: the shot's target output, plus dose-in (e.g. "18.0g in"). A DELIBERATE yield
     // override (the hasBrewYieldOverride flag, not raw drift — measured dose never exactly matches
-    // the profile's) renders as "36.0 → 40.0g", mirroring the temperature-override treatment.
+    // the profile's) renders as "36.0 → 40.0g", mirroring the temperature-override treatment —
+    // UNLESS yieldTargetOnly is set, in which case only the effective target ("40.0g") shows.
     readonly property string _yieldStr: {
         if (!(_has("doseYield") && targetWeight > 0)) return ""
-        if (Settings.brew.hasBrewYieldOverride && profileYield > 0
+        if (!yieldTargetOnly && Settings.brew.hasBrewYieldOverride && profileYield > 0
                 && Math.abs(targetWeight - profileYield) > 0.1)
             return profileYield.toFixed(1) + " → " + targetWeight.toFixed(1) + "g"
         return targetWeight.toFixed(1) + "g"
--- a/qml/components/ShotPlanText.qml
+++ b/qml/components/ShotPlanText.qml
@@ -37,9 +37,8 @@ Item {
     // Sentence mode only: render the detail tail on its own line(s) below the
     // sentence (display path — the a11y text stays one joined sentence).
     property bool stacked: false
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Found 1 issue:

1. The web layout editor's new "Final yield only" checkbox is never disabled/dimmed when the Dose & yield item isn't in the shot plan, so it stays fully interactive while inert. The in-app editor gates its switch with `enabled: popup.shotPlanItems.indexOf("doseYield") !== -1`, and the direct precedent for this pattern is `spSyncStacked()` (added in #1428, commit cbc444c0), which mirrors the in-app enabled-gating for the sibling "Stacked details" checkbox on load and on change. This PR reproduces the QML half but omits the web-side sync — a `spSyncYieldTargetOnly()`-style function called from load and `spAdd()`/`spRemove()` (the label also needs an `id` styling hook, like `spStackedLabel`). Harmless to data (the stored value simply has no effect until Dose & yield is re-added), but it breaks the two editors' UX parity for a directly analogous toggle.

Checkbox without gating: https://github.com/Kulitorum/Decenza/blob/e6278320a30da89d909c7aa7c1e1cdd3578d2bce/src/network/shotserver_layout.cpp#L2031-L2037

In-app gating it should mirror: https://github.com/Kulitorum/Decenza/blob/e6278320a30da89d909c7aa7c1e1cdd3578d2bce/qml/components/layout/ScreensaverEditorPopup.qml#L686-L691

Existing precedent (`spSyncStacked`): https://github.com/Kulitorum/Decenza/blob/e6278320a30da89d909c7aa7c1e1cdd3578d2bce/src/network/shotserver_layout.cpp#L3213-L3221

Everything else checked out: the `shotPlanYieldTargetOnly` key is consistent across the QML editor, `ShotPlanItem`, and the web editor; all reads use `=== true` with a safe false default; both `ShotPlanText` instances receive the property; and Shot Plan is a bespoke-editor type, so no `readoutOptionSchema()` registration is needed.

---

<details>

- [Worth a look] The new property comment in `ShotPlanText.qml` copies the pre-existing "mirroring the temperature-override treatment" phrase, but temperature overrides render as a delta tag (e.g. `92°C +1°` via `TemperatureDisplay::format()`) plus a highlight color, not an arrow — the analogy is about calling out overrides, not the format, and is easy to reword while touching these lines (score: 50)
- [OK to skip] The new `yieldTargetOnly` property doc and the `_yieldStr` comment in `ShotPlanText.qml` now duplicate the same arrow-vs-target-only explanation — purely a documentation-drift risk (score: 50)

</details>

skialpine:
Pushed 5132a7bc addressing the review: added `spSyncYieldTargetOnly()` so the web editor disables/dims the new checkbox when Dose & yield isn't shown (mirroring the in-app switch, same pattern as `spSyncStacked()`), reworded the "mirroring the temperature-override treatment" comment (temperature uses a delta tag, not an arrow), de-duplicated the property doc vs `_yieldStr` comment, and noted the toggle is a visible no-op without an active yield override.

PR #1443, opened 2026-07-09
Shot plan: profile-less "recipe" sentence when the Profile item is hidden

loganross2001:
## What

Adds a **profile-less "recipe" sentence** to the Shot Plan widget. When **Sentence style** is on but the **Profile** item is not in the shown-items list, the plan now reads as the recipe itself:

> **Brew 40.0g of Espresso at 92°C from 18.0g of Onyx Monarch**

instead of collapsing to a dotted fragment list.

## Why

The sentence scaffold added in #1426 anchors on the profile name — `"Brew … using <profile> at …"`. That anchor is a hard requirement: `_build()` gates the whole sentence on `sentence && _profileStr !== ""`, so the moment you **remove the Profile chip** (a perfectly reasonable choice — many people don't want the profile name in their plan), the sentence silently degrades to fragments (`40.0g · 92°C · Onyx · Monarch`). Turning **Sentence style** on then does nothing, which reads as a bug.

This gives the sentence a second anchor — the beverage word, which is always present — so "Sentence on" always yields a sentence, and the recipe reads naturally without the profile name.

## How

New branch in `ShotPlanText._build()` for `sentence && _profileStr === ""`:

- **Anchor** on the beverage word (`_beverage`, always non-empty): `Brew {yield} of {beverage}` (or `Brew {beverage}` with no yield).
- **Consume** dose, roaster and coffee **into** the sentence — `… at {temp} from {dose} of {roaster} {bean}` — so they no longer *also* trail as fragments. Only `grind`/`roastDate` trail after.
- **Each clause stays gated by its item chip** via the existing `_yieldStr`/`_tempStr`/`_doseStr`/`_roasterStr`/`_coffeeStr` getters, so the shown-items list still drives exactly what appears. Degrades cleanly with any piece missing (no dangling "from"/"of"):
  - no temp → *Brew 40.0g of Espresso from 18.0g of Onyx Monarch*
  - no beans → *Brew 40.0g of Espresso at 92°C from 18.0g*
  - no dose → *Brew 40.0g of Espresso at 92°C*
  - no yield → *Brew Espresso at 92°C from 18.0g of Onyx Monarch*
- Built by appending translatable clauses (`shotplan.recipe.*`) onto the head; the a11y (`text`) and rich (bolded) paths share the one `_build` builder, so they can't drift. User values pass through the existing `_argSafe` + `Theme.escapeHtml`, same as the profile-shown branch.

The profile-**shown** branch is unchanged — this only adds behavior for the case that previously fell through to fragments.

## Notes for reviewers

- **Visibility:** the root is `visible: text !== ""`. Previously, sentence-on + profile-hidden + every other item also hidden could produce an empty string and hide the widget; the recipe branch now always renders at least "Brew Espresso", so that (degenerate) configuration becomes visible. Intended — "Sentence on" should never be silently blank.
- **i18n word order:** the recipe is assembled by appending `at %`/`from %` clauses, which bakes in English clause order (a translator can reorder *within* a clause but not move the temp/dose clauses relative to the verb, unlike the single-template profile-shown branch). Deliberate tradeoff to keep the combinatorial template count down; happy to switch to full per-combo templates if you'd prefer.

Only the Shot Plan widget is affected. No new files, imports, or C++.

### DIFF
--- a/openspec/specs/plan-widgets/spec.md
+++ b/openspec/specs/plan-widgets/spec.md
@@ -25,7 +25,13 @@ The seven display items are: Profile (`profile`), Temperature (`temperature`), R
 - profile + temperature, no yield → "Brew {beverage}, using {profile} at {temperature}"
 - yield + profile, Temperature item absent → "Brew {yield} of {beverage}, using {profile}"
 - profile only → "Brew {beverage}, using {profile}"
-- Profile item absent (or no profile name available) → the sentence has no anchor; the text SHALL render in fragment format regardless of the Sentence style option.
+
+Profile item absent (or no profile name available) → the sentence has no profile anchor, so it falls back to the profile-less "recipe" sentence: the scaffold instead consumes the Dose & yield, Temperature, Roaster, and Coffee items (wherever they sit in the list) together with the beverage word, and only the Grind and Roast date items SHALL trail after it as a separator-joined tail in item-list order. The beverage word SHALL always anchor this sentence, so Sentence style SHALL NOT degrade to fragment format while it is ON — regardless of which other items are present:
+
+- yield + temperature + dose + roaster + coffee → "Brew {yield} of {beverage} at {temperature} from {dose} of {roaster} {coffee}"
+- yield only, no temperature/dose/roaster/coffee → "Brew {yield} of {beverage}"
+- no yield, dose present, no roaster/coffee → "Brew {beverage} from {dose}"
+- no yield, no dose, roaster + coffee only → "Brew {beverage} from {roaster} {coffee}"
 
 The beverage word SHALL follow the current profile's `beverage_type`: "Espresso" for espresso (and unset), "tea" for tea types, and "coffee" for any other coffee beverage (filter, pourover, …). A cleaning or descale profile SHALL replace the plan entirely — in both formats — with a cleaning notice carrying a do-not-load-coffee warning, rendered bold in the error (red) color with no item segments.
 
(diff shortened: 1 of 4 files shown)


loganross2001:
Independent of my open #1442 ("Final yield only"). Both touch `ShotPlanText.qml` but different functions — #1442 edits `_yieldStr` (the yield-override arrow), this edits `_build()` (adds the profile-less sentence branch) plus the header doc comment. No stacking; whoever lands second gets at most a trivial context resolution. Reviewable/mergeable in any order.

skialpine:
### Code review

Found 3 issues:

1. New comments say the recipe sentence triggers on "Profile removed" / "drop Profile to get it," but the branch condition is `_profileStr !== ""` being false — it also fires whenever the Profile chip is shown but no profile is currently loaded (`profileName` empty, e.g. fresh install or profile deleted). The pre-existing header comment explicitly called out this second trigger ("or a profile name"); the new wording drops it, and could mislead a future reader about when the fragment-list path used to kick in vs. when the recipe sentence now does.

https://github.com/loganross2001/Decenza-community-fork/blob/b0058576b6cf38c733f7c278efd929ea7a9d5b48/qml/components/ShotPlanText.qml#L11-L17

2. The new comments (file header, inline `_build()` comment, and the `ScreensaverEditorPopup` switch comment) all say the recipe sentence consumes "dose, roaster and coffee" and that "only grind/roastDate trail" — omitting temperature. Temperature is in fact consumed too (via `_tempStr` into the "at %2" clause), and the `rtail` switch has no `"temperature"` case. The comments' own example ("...at 92°C from...") shows this. A future maintainer relying on the prose list could add a `"temperature"` case to `rtail`, believing it's currently dropped, causing it to render twice.

https://github.com/loganross2001/Decenza-community-fork/blob/b0058576b6cf38c733f7c278efd929ea7a9d5b48/qml/components/ShotPlanText.qml#L173-L220

3. `openspec/specs/plan-widgets/spec.md` has a live SHALL requirement and matching scenario stating that with the Profile item absent (or no profile name), the text "SHALL render in fragment format regardless of the Sentence style option" — exactly the behavior this PR replaces with the recipe sentence. No spec update or `openspec/changes/...` proposal accompanies the PR, so after merge the checked-in spec asserts behavior the code no longer implements.

https://github.com/Kulitorum/Decenza/blob/f1338f7d3fb7bcd7251f6999839f0944b104d872/openspec/specs/plan-widgets/spec.md#L27-L29

https://github.com/Kulitorum/Decenza/blob/f1338f7d3fb7bcd7251f6999839f0944b104d872/openspec/specs/plan-widgets/spec.md#L54-L58

Everything else checked out: the `.arg()`/`_argSafe()` chaining in the new sentence builder is safe against `%N`-injection from user-supplied roaster/coffee names, all seven items are consumed or trailed exactly once with no double-render, the shared `_build()` renderer keeps the a11y and rich paths from drifting, the new branch correctly inherits `yieldTargetOnly` gating from the recently-merged #1442, and `RecipesPage.qml`'s `sentence: false` usage is unaffected.

---

<details>

- [Worth a look] `ShotPlanItem.qml`'s maxLines comment ("the profile-anchor fragment fallback still gets the extra line — harmless, just a wider wrap budget") is now stale — that case is a real two-line sentence+tail after this PR, not a fragment list that merely tolerates extra width. The maxLines value itself is still correct. (score: 50)

</details>

skialpine:
Pushed 9037ee97 addressing the review: reworded the recipe-sentence comments in `ShotPlanText.qml` and `ScreensaverEditorPopup.qml` to note the no-profile-loaded trigger and that temperature is consumed (not just dose/roaster/coffee), fixed the now-stale `maxLines` comment in `ShotPlanItem.qml`, and updated `openspec/specs/plan-widgets/spec.md` to describe the recipe-sentence contract instead of the old fragment-fallback requirement it contradicted (plus a scenario for the no-profile-loaded case).

PR #1450, opened 2026-07-09
Web settings: AI model picker parity with the app

skialpine:
## Summary

The web settings page (`http://<tablet>:8888/settings`) only allowed updating AI API keys — the per-provider model selection the app offers (GPT-5.4 mini/GPT-5.4, Sonnet 4.6/Sonnet 5, Gemini 2.5/3.5 Flash) was missing, and the provider buttons hard-coded stale model labels ("GPT-4.1", "Claude"). This brings the web page to parity and centralizes the model metadata so the two UIs can't drift.

### Web page (ShotServer)
- **Generic Model dropdown** appears for any provider whose catalog has more than one model (OpenAI/Anthropic/Gemini today — future catalogs light up with no web-side changes). Unset/stale selection falls back to the catalog's first entry (the recommended default), exactly like the app; the default is never written back.
- Provider-specific guidance line under the dropdown (same copy as the app).
- Provider-button sublabels now show the actual selected model (e.g. "Sonnet 5") instead of hard-coded stale text.

### API
- `GET /api/settings` adds `providerModelCatalogs`, `providerModels` (stored selections), `providerModelNames`, `providerModelHints` — all sourced from `AIManager`.
- `POST /api/settings` (and the shared AI-test path) accepts a `providerModels` object routed through `SettingsAI::setProviderModel()`. Unknown ids are harmless: each provider's validating `setModel()` ignores ids not in its catalog.

### Single source of truth (drift prevention)
- New `AIProvider::modelHint()` sits next to `availableModels()` so the catalog and its guidance copy live in one place per provider.
- `AIManager::modelHint(providerId)` exposes it; `SettingsAITab.qml` now consumes it (keeping its existing `settings.ai.modelHint.<provider>` translation keys — the runtime string-discovery system handles the dynamic key), and the web page serves it from the same source. The duplicated hint strings in QML were removed.

## Test plan
- Full test suite run via Qt Creator: all passed (pre-hint-refactor state; re-run pending).
- Manual: open `/settings` on the tablet, select each of OpenAI/Anthropic/Gemini → Model dropdown + hint appear with correct selection; Save persists; app's AI settings tab reflects the same model. OpenRouter/Ollama keep their free-text model fields.

### DIFF
--- a/qml/pages/settings/SettingsAITab.qml
+++ b/qml/pages/settings/SettingsAITab.qml
@@ -294,24 +294,18 @@ KeyboardAwareContainer {
                         }
                     }
 
-                    // Provider-specific guidance. Gated per provider so a future
-                    // provider that gains multiple models can't inherit wrong copy.
+                    // Provider-specific guidance. The English copy lives in
+                    // AIProvider::modelHint() (next to the model catalog) so the
+                    // app and the ShotServer web settings page share one source;
+                    // the per-provider translation key keeps it translatable.
                     Text {
                         visible: text.length > 0
                         text: {
-                            switch (modelSelect.currentProvider) {
-                            case "gemini":
-                                return TranslationManager.translate("settings.ai.modelHint.gemini",
-                                    "3.5 Flash is the most capable. 2.5 Flash is more available (fewer busy errors).")
-                            case "anthropic":
-                                return TranslationManager.translate("settings.ai.modelHint.anthropic",
-                                    "Sonnet 5 is the most capable. Sonnet 4.6 is the established default.")
-                            case "openai":
-                                return TranslationManager.translate("settings.ai.modelHint.openai",
-                                    "GPT-5.4 is more capable. GPT-5.4 mini is cheaper and faster.")
-                            default:
-                                return ""
-                            }
+                            var hint = MainController.aiManager
+                                ? MainController.aiManager.modelHint(modelSelect.currentProvider) : ""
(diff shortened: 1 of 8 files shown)


skialpine:
Automated review (4 agents: general, comments, silent-failures, tests) ran against the original diff; all real findings fixed in 344e4375:

- **HIGH (silent-failure):** `POST /api/settings` persisted unvalidated `providerModels` entries and always answered `success:true`, even when the provider silently discarded the id (stale web tab after a catalog change, typo'd API client, junk provider keys written to QSettings). "Test Connection" would test the provider *default* and report the discarded selection as working. → Entries are now validated against `AIManager`'s provider list and catalogs; rejections return `success:false` naming the bad entries, and the test endpoint refuses to run.
- **i18n regression (comments):** the QML hint's dynamic translation key was invisible to `scanAllStrings()`, dropping the three hint strings from the batch-translation registry. → main.cpp registers the keys at startup from the same `AIProvider::modelHint()` source.
- **Test gap:** `tst_aiproviders` now pins the anti-drift contract — every multi-model provider has a non-empty `modelHint()` mentioning each catalog entry's display name.
- **LOW:** breadcrumb `qWarning` when `/api/settings` is served without an AIManager; comment-accuracy fixes.

Clean passes: QML binding reactivity, raw-string delimiters, JSON round-trip semantics, dropdown default/fallback logic, JS fetch error paths, CLAUDE.md conventions.

PR #1456, opened 2026-07-10
Grind Quick-Select: brew-bar widget for one-tap grind adjust (current ±5)

loganross2001:
## What

Adds a **Grind Quick-Select** layout widget — the grind counterpart to the existing **Ratio Quick-Select**. Place it on the brew bar; tap it and a popup lists grind values at **current ±5 steps** (11 rows). Tap one and it sets `Settings.dye.dyeGrinderSetting`. The idea: between one shot and the next you almost never move the grinder more than a few steps, so a compact ±5 list is a one-tap dial-in instead of opening a full editor.

## Why

Adjusting grind mid-dial-in currently means going into a settings/dialog flow. The ratio widget already gives one-tap ratio changes on the bar; grind — the thing you actually tune most between shots — has no equivalent. This fills that gap symmetrically.

## How

- New `GrindQuickSelectItem.qml` (the bar pill, shows the current grind) + `GrindPickerDialog.qml` (the ±5 value list, current row highlighted, finer/coarser ends labelled).
- Registered exactly like `ratioQuickSelect`: `widgetCatalogTable()` row, `LayoutItemDelegate` case, `CMakeLists` qml entries.
- **Works whether the grinder uses numbers or letters** — `stepGrind(current, n, step)` is tiered:
  - **Numeric** ("30") → `current ± n·step`, clamp ≥ 0, decimals follow the step (e.g. `25…35` at step 1.0).
  - **Number embedded in text** ("C4") → step the number, keep the letters (`C0…C9`).
  - **Pure letters** ("F") → step by letter position, clamp A–Z (`A…K`).
  - **Unparseable** ("medium-fine") → fall back to the grind settings actually used for this grinder (from shot history), the nearest ~5 below/above.
- **Step is a global preference** — `Settings.brew.grindQuickSelectStep` (default 1.0, clamp 0.1–5.0), edited in **Settings → Machine → Grinder**, serialized for portability. (Kept global rather than per-widget so it's discoverable in Settings, not buried in the layout editor.)
- Selecting a value writes `dyeGrinderSetting`, which already rides the coffee_bags last-used write-through — no extra plumbing.

Theme tokens + i18n + accessibility throughout. Self-contained (its only cross-cutting dependency is the existing `getDistinctGrinderSettingsForGrinder` shot-history query).


skialpine:
Quick note — not a detailed review, just two gaps that stood out on a skim:

1. **RPM-capable grinders aren't addressed.** `dyeGrinderSetting` is only the burr-position/string value. RPM-capable grinders also expose `Settings.dye.dyeGrinderRpm` / `grinderRpmCapable(brand, model)` (see the dedicated RPM field in `BrewDialog.qml`, shown whenever the active grinder is RPM-capable). `GrindQuickSelectItem.qml`'s `stepGrind()` never checks `grinderRpmCapable()` and never touches `dyeGrinderRpm` — for a grinder where RPM is the actual dial-in axis, this widget wouldn't represent the parameter the user needs to adjust.

2. **`grindQuickSelectStep` is global, not per-grinder.** It's a single flat `espresso/grindQuickSelectStep` key, shared across every configured grinder. The app already has a per-equipment grind-memory pattern (`EquipmentPackage.lastGrindSetting`/`lastRpm`), but the step size doesn't follow it — switching between grinders with very different step granularities (e.g. a Niche vs. an EK43) keeps the same step value for both.

loganross2001:
Good catches, thanks. On the important one: the pill now checks `grinderRpmCapable(brand, model)` (reactive on brand/model, same as BrewDialog's RPM field) and, when the grinder is RPM-capable *and* has an RPM set, steps `dyeGrinderRpm` via plain property assignment instead of `dyeGrinderSetting` — relabeled "RPM", Finer/Coarser hint suppressed, and using a dedicated fixed 50-RPM step (not the burr-position `grindQuickSelectStep`, which would only span ~10 RPM; the constant is tunable). I gated on `dyeGrinderRpm > 0` on purpose, since `deriveRpmCapable` treats any unknown/custom grinder as capable — so burr-dialed custom grinders keep the existing stepping until an RPM actually exists. On the per-grinder step: it's a bigger change (new key + serializer + settings UI + migration) so I left the step global to keep this PR tight — happy to do per-equipment step as a follow-up, and the burr-step/RPM-step split makes that a clean separate change.

skialpine:
Followed up on the RPM point by checking coverage against the full grinder catalog (`src/core/grinderaliases.h`), and the gap is bigger than RPM alone — this isn't ready to ship as-is.

Of the 93 grinders in `allGrinders()`:

- **51 (~55%)** — `NumericWithSuffix` notation, non-RPM. These work with the current `stepGrind()` parser, assuming the user types a plain number.
- **17 (~18%)** — `Compound` (rotation) notation: the entire Eureka Mignon/Atom/Helios line (12 models) plus all five 1Zpresso models (JX-Pro, J-Max, K-Max, K-Plus, Q2). These store settings as `"a+b"` (rotations+position). None of `stepGrind()`'s three branches (pure numeric / number-embedded-in-text / pure letters) match that format, so it always falls through to the shot-history fallback — which degenerates to a single-row, non-adjustable picker on a fresh setup or thin history.
- **25 (~27%)** — `variableRpm = true`. Currently gated on `dyeGrinderRpm > 0`, so these don't work until the user has separately entered an RPM once elsewhere first. This gate is also unnecessary: the catalog already knows definitively which grinders are RPM-capable (`GrinderEntry.variableRpm`) — there's no need to wait on a live value for grinders we can already identify from the DB.

So as shipped, this reliably works for about half the catalog and needs either a first-time manual detour or lucky shot history for the rest. A "Grind Quick-Select" that only quick-selects for half the supported grinders isn't a coverage bar we can merge on.

The good news: none of this needs new algorithm work — `grinderaliases.h` already has the full pipeline, proven in production via `dialing_blocks.cpp`:

- `parseGrinderSetting(GrinderEntry, rawString) -> optional<double>` (grinderaliases.h:395-424) parses **both** notations — including Compound `"a+b"` — into one linear value.
- `formatGrinderSetting(GrinderEntry, double) -> QString` (grinderaliases.h:431-443) is the inverse, and already handles rotation/position carry-borrow correctly via `floor(linear/positionsPerRev)`.
- These are already chained as parse → arithmetic → format in `dialing_blocks.cpp:878,1061-1068,1124-1125` for AI dialing.

None of it is exposed to QML yet — `SettingsDye` currently only wraps `grinderRpmCapable`, `suggestedBurrs`, `isBurrSwappable`, `knownGrinderBrands`, `knownGrinderModels` (settings_dye.h:150-155). Ask: add two more `Q_INVOKABLE`s following that same pattern instead of hand-rolling a parser that only covers a slice of the catalog:

1. `QString stepGrinderSetting(brand, model, current, deltaUnits)` — wraps `findEntry` + `parseGrinderSetting` + arithmetic + `formatGrinderSetting` (~10 lines, mirrors the existing math in `dialing_blocks.cpp`). Drop-in replacement for the widget's numeric/embedded-number branches, and picks up Compound notation for free.
2. `bool isKnownRpmGrinder(brand, model)` — same lookup as `grinderRpmCapable()` but returns `false` for unmatched grinders instead of the "unknown → true" fallback, so RPM mode only engages when the catalog actually confirms it. With this, drop the `dyeGrinderRpm > 0` gate — known RPM grinders should work in RPM mode from the first tap, no manual detour required.

Keep the pure-letter branch as a JS-side fallback (that's outside `parseGrinderSetting`'s scope), but numeric and Compound should route through the catalog, not a hand-rolled regex that only covers a slice of it. This isn't a big lift — the risky part (parse/format/carry-borrow correctness) is already written and exercised elsewhere in the app; it's mostly plumbing two methods through to QML and swapping the widget's stepping calls over.

loganross2001:
Thanks — that catalog audit was exactly the right call, and you're right the hand-rolled parser only covered a slice. I've reworked the widget to route through the catalog pipeline as you laid out, rather than the regex.

**Two new `Q_INVOKABLE`s on `SettingsDye`** (mirroring the existing `grinderRpmCapable`/`suggestedBurrs` wrappers):

1. `stepGrinderSetting(brand, model, current, deltaUnits)` — `findEntry` → `parseGrinderSetting` → add → `formatGrinderSetting`. Registry grinders (numeric **and** Compound `"a+b"`) now step through the proven pipeline, so Eureka Mignon/Atom/Helios and the 1Zpresso line round-trip correctly incl. rev/position carry-borrow. Returns `""` for a grinder not in the registry, an unparseable value, or a step below the dial floor (`< 0`) — the widget then keeps its plain-numeric / letter / history fallbacks for those.
2. `isKnownRpmGrinder(brand, model)` — catalog-confirmed `variableRpm` only; an unmatched grinder returns `false` (not the `unknown → true` of `grinderRpmCapable`). The widget's `isRpmMode` now uses this and I dropped the `dyeGrinderRpm > 0` gate.

In the widget, `stepGrind()` tries `stepGrinderSetting` first and only falls through to the JS branches (pure-letter, embedded-number, custom-numeric, history) when the catalog returns `""`.

**Two things I want to flag explicitly, since they go a bit beyond the literal ask and I'd rather you make the call than have it surprise you in re-review:**

- **Dropping the gate makes the pill RPM-only for *every* catalog `variableRpm` grinder**, including someone who dials their DF83V by burr position and has never set an RPM. `variableRpm` in the catalog means "has an adjustable motor," not "dials primarily by RPM," so for burr-first users of these grinders the pill now quick-selects RPM instead of burr position. That's the direct consequence of removing the live-value gate — the catalog can't tell burr-first from RPM-first. If you'd prefer burr-dialers on these grinders to keep burr stepping, the cleanest signal is still "has an RPM been set," i.e. re-introduce a gate but keyed on `isKnownRpmGrinder && dyeGrinderRpm > 0`. Happy to go either way — your repo, your call.
- **When RPM mode is on but no RPM is set yet, the picker seeds a neutral ~1000-RPM anchor** so the first tap is usable (otherwise dropping the gate just gives a dead single-row picker). Nothing is written until the user actually picks a value, and the "current" row isn't highlighted for the seed. The anchor is a tunable constant.

Minor note: catalog numeric grinders now format via `formatGrinderSetting` (1 decimal), so a sub-0.5 `grindQuickSelectStep` (e.g. 0.25) loses the extra place vs. the old JS branch (`20.25 → "20.2"`). Monotonic and harmless, but noting it.

Added 4 tst_settings slots covering numeric stepping, Compound stepping with rev-carry (`1+98`+5 → `2+3`), the custom-grinder `""` fallback, and `isKnownRpmGrinder` catalog-confirmed vs unknown. macOS build + tst_settings green (69/69).

PR #1457, opened 2026-07-10
Steam: global seconds-per-gram weight calibration (pitcher-agnostic)

loganross2001:
## What

Replaces the **per-pitcher** weight-timed steam calibration with a **single global "seconds per gram of milk"** rate. Calibrate once; every pitcher then finishes at the same temperature regardless of milk volume.

## Why

Today each steam-pitcher preset carries its own `calibMilkG` + `duration`, and the stop time is `duration × (milk / calibMilkG)` — **per pitcher**. So calibrating the large pitcher does nothing for the small one; switch pitchers and the calibration is effectively gone. Yet the section's own help text already promises *"calibrate once — a small or large pitcher both finish at the same temperature."* The copy describes the desired behavior; the implementation didn't deliver it.

Physically, steam time to bring milk to temperature is driven by **milk mass** (× ΔT ÷ steam power) — so at a consistent steam flow, **seconds-per-gram is the same across pitchers**. A single global rate is both simpler and what the feature already claims to do.

## How

- New `Settings.brew.steamSecondsPerGram` (default 0 = uncalibrated), serialized for portability.
- `scaledSteamTime()` now returns `qBound(5, qRound(secPerGram × milkG), 120)`, keeping the existing gates (weight-timing enabled, preset not disabled, milk present, rate set). No longer reads per-preset `calibMilkG`/`duration` for scaling.
- `calibrateSteamFromReference(milkG, timeSec)` sets `rate = timeSec / milkG`. The SteamPage **"Use last steam"** button calls it with the last session's measured milk + steam time (calibrate straight from a good pour); a manual ± control is also provided.
- **One-time, sentinel-guarded migration** seeds the global rate from an existing calibrated preset (`duration / calibMilkG`) so upgraders keep their calibration. The sentinel (`steam/steamRateMigrated`), not `rate <= 0`, is the guard — so a user who deliberately sets the rate (including to 0) is never re-seeded.
- The per-pitcher `calibMilkG` / `setSteamPitcherCalibration` are **left in place** — no longer used for scaling, but still read by display code (the steam-plan hint); freshly-calibrated users (`calibMilkG == 0`) degrade gracefully there.
- SteamPage UI: the per-pitcher "Reference milk" stepper is replaced by a global rate readout + manual adjust + "Use last steam". The master weight-timing toggle is unchanged.

Theme tokens + i18n + accessibility throughout; unit tests updated to drive the global rate.

## Note for reviewers

The migration is one-time (sentinel `steam/steamRateMigrated`) and runs at construction. Restoring an **old, pre-feature backup** afterward brings back per-preset `calibMilkG` but won't re-run the seed, so `steamSecondsPerGram` stays 0 (weight-timing off until re-calibrated). It degrades safely — base duration is used, no crash — and the "Weight-timed: Xg→Ys" hint still renders from the restored `calibMilkG`. Flagging as a design edge, not a defect; happy to add an import-time re-seed if you'd prefer.

### DIFF
--- a/src/core/settings_brew.cpp
+++ b/src/core/settings_brew.cpp
@@ -68,6 +68,34 @@ SettingsBrew::SettingsBrew(QObject* parent)
         m_settings.setValue("flush/presets", QJsonDocument(defaults).toJson());
     }
 
+    // One-time migration: weight-timed steaming is now a GLOBAL seconds-per-gram
+    // rate (same steam flow on every pitcher) instead of per-pitcher reference milk.
+    // Existing calibrated users had (calibMilkG, duration) stored per preset; seed
+    // the global rate from the first preset that has both so they aren't reset. Runs
+    // after the preset-seeding block above so the presets are present. Done in the
+    // ctor (not the getter) to avoid a const getter mutating QSettings mid-read.
+    //
+    // Gated by a run-once SENTINEL, not by "rate <= 0" — legacy calibMilkG is left in
+    // place (other readers still use it), so a bare "rate <= 0" guard would re-seed the
+    // old rate every launch and silently undo a user who deliberately uncalibrated
+    // (the new ± control allows 0).
+    if (!m_settings.value("steam/steamRateMigrated", false).toBool()) {
+        if (m_settings.value("steam/steamSecondsPerGram", 0.0).toDouble() <= 0.0) {
+            QByteArray data = m_settings.value("steam/pitcherPresets").toByteArray();
+            QJsonArray arr = QJsonDocument::fromJson(data).array();
+            for (const QJsonValue& v : arr) {
+                QJsonObject preset = v.toObject();
+                double calibMilk = preset.value("calibMilkG").toDouble(0.0);
+                double duration  = preset.value("duration").toDouble(0.0);
+                if (calibMilk > 0.0 && duration > 0.0) {
+                    m_settings.setValue("steam/steamSecondsPerGram", duration / calibMilk);
+                    break;  // first calibrated preset wins
+                }
+            }
+        }
+        m_settings.setValue("steam/steamRateMigrated", true);
+    }
+
     // Load persistent brew overrides into the cache (Settings used to do this).
     m_hasTemperatureOverride = m_settings.value("brew/hasTemperatureOverride", false).toBool();
     if (m_hasTemperatureOverride) {
@@ -202,6 +230,25 @@ void SettingsBrew::setLastSteamTimeS(double s) {
         m_settings.setValue("steam/lastSteamTimeS", s);
         emit lastSteamTimeSChanged();
     }
+}
+
+double SettingsBrew::steamSecondsPerGram() const {
+    return m_settings.value("steam/steamSecondsPerGram", 0.0).toDouble();
+}
+void SettingsBrew::setSteamSecondsPerGram(double secPerGram) {
+    if (secPerGram < 0) secPerGram = 0;  // 0 = uncalibrated; never negative
+    if (!qFuzzyCompare(1.0 + steamSecondsPerGram(), 1.0 + secPerGram)) {
+        m_settings.setValue("steam/steamSecondsPerGram", secPerGram);
+        emit steamSecondsPerGramChanged();
+    }
+}
+
+void SettingsBrew::calibrateSteamFromReference(double milkG, double timeSec) {
+    if (milkG <= 0.0 || timeSec <= 0.0) return;  // guarded — need both to derive a rate
+    setSteamSecondsPerGram(timeSec / milkG);
+    // Weight-timed steaming is off by default; calibrating is the explicit opt-in,
+    // mirroring the old per-pitcher setSteamPitcherCalibration behaviour.
+    setMilkAutoCaptureEnabled(true);
 }
 
 bool SettingsBrew::doseCaptureSoundEnabled() const {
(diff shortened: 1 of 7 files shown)


src/core/settings_brew.cpp
```diff
@@ -436,10 +483,12 @@ int SettingsBrew::scaledSteamTime(int index, double milkG) const {
...
     QVariantMap p = getSteamPitcherPreset(index);
     if (p.isEmpty() || p.value("disabled").toBool()) return 0;
-    double calibMilk = p.value("calibMilkG", 0.0).toDouble();
-    double duration  = p.value("duration", 0.0).toDouble();
-    if (calibMilk <= 0.0 || duration <= 0.0 || milkG <= 0.0) return 0;
-    return qBound(5, qRound(duration * (milkG / calibMilk)), 120);
+    if (milkG <= 0.0) return 0;
+    // Pitcher-agnostic: one global seconds-per-gram rate for every pitcher (same
+    // steam flow), no longer reads the per-preset calibMilkG/duration for scaling.
+    double secPerGram = steamSecondsPerGram();
```
skialpine, line 521: The "pitcher-agnostic" assumption baked in here doesn't hold given this app's own data model: each steam pitcher preset carries its own independent `flow` and `temperature` (`addSteamPitcherPreset`/`updateSteamPitcherPreset`, both editable per-pitcher in `SteamPage.qml`). Heating time depends on power (flow) and ΔT (target temp), not just milk mass — "seconds per gram is the same across pitchers" is only true if flow and target temp are also held constant, which nothing here enforces.

Concretely: a user with a "small pitcher, gentle flow, flat-white temp" preset and a "large pitcher, aggressive flow, cappuccino temp" preset will get one global rate calibrated from whichever preset happened to seed/last-calibrate it, applied incorrectly to the other. The old per-preset `(duration, calibMilkG)` pair at least implicitly captured each preset's own flow/temp context; this throws that away.

Is the assumption "all presets run at the same flow/temp" intentional/enforced elsewhere, or should the rate be scoped per (flow, temp) combination rather than fully global?

src/core/settings_brew.cpp
```diff
@@ -68,6 +68,34 @@ SettingsBrew::SettingsBrew(QObject* parent)
...
+    // old rate every launch and silently undo a user who deliberately uncalibrated
+    // (the new ± control allows 0).
+    if (!m_settings.value("steam/steamRateMigrated", false).toBool()) {
+        if (m_settings.value("steam/steamSecondsPerGram", 0.0).toDouble() <= 0.0) {
+            QByteArray data = m_settings.value("steam/pitcherPresets").toByteArray();
+            QJsonArray arr = QJsonDocument::fromJson(data).array();
+            for (const QJsonValue& v : arr) {
+                QJsonObject preset = v.toObject();
+                double calibMilk = preset.value("calibMilkG").toDouble(0.0);
+                double duration  = preset.value("duration").toDouble(0.0);
```
skialpine: Minor: "first calibrated preset wins" is an arbitrary pick if a user had multiple differently-calibrated presets before upgrading (which, given each preset has its own flow/temp — see comment on `scaledSteamTime` below — is a realistic setup). Worth a code comment acknowledging this is a best-effort seed, not necessarily the "right" one, so a future reader doesn't assume it picks the most representative preset.

skialpine (changes requested):
The redesign direction is right — the old per-pitcher `duration × milk/calibMilkG` scaling genuinely didn't survive switching pitchers, contradicting the section's own help text, and a single calibrated rate is a much simpler mental model. The migration is also careful: sentinel-guarded (not `rate <= 0`-guarded) so a user who deliberately zeroes the rate isn't re-seeded, and it runs after preset seeding so it reads real data.

Two things I'd want resolved before merge:

1. **Physics assumption vs. actual data model** (see inline comment on `scaledSteamTime`): presets have independent `flow`/`temperature`, so "seconds per gram is pitcher-agnostic" isn't generally true in this app. Worth confirming this is an accepted simplification (and ideally noting it in the in-app copy/help text) rather than an oversight.

2. **Stale display code not updated** — `qml/pages/SteamPage.qml` around line 1653–1667 (the "Weight-timed: Xg → Ys" hint under the pitcher-weight field, `steam.hint.weightTimed`) still does `cal.toFixed(0) + "g → " + preset.duration + "s"` directly from the raw `preset.calibMilkG`/`preset.duration` fields. This block isn't touched by the diff, so I couldn't anchor a comment on it directly, but after this PR ships it will show pre-migration math that no longer matches what `effectiveSteamDurationSec()` actually computes (which now uses the global rate, possibly seeded from a *different* preset than the one being displayed). For anyone upgrading from a calibrated install this hint becomes permanently wrong, and there's no UI path left to clear `calibMilkG` since the old "Reference milk" editor is removed. Should be deleted or recomputed via `Settings.brew.effectiveSteamDurationSec(index, cal)`.

Requesting changes for #2 since it's a concrete, user-visible correctness bug; #1 is more of a design question I'd like an explicit answer on.

loganross2001:
You're right that seconds-per-gram isn't strictly pitcher-agnostic — presets carry their own flow/temperature, so the single rate is a deliberate simplification (one mental model, calibrated once) rather than a physical guarantee. I've added that caveat to the calibration copy. I also fixed the weight-timed hint: it now recomputes via `effectiveSteamDurationSec(index, cal)` instead of the raw `calibMilkG`/`duration`, and only renders while weight-timing is actually active (auto-capture on + a calibrated global rate), so it can't show pre-migration fixed-duration math. Since the reference-milk editor is gone, the hint now appears only for presets that still carry a stored reference weight — new presets don't show it, which sidesteps the vestigial-`calibMilkG` concern you raised.

skialpine:
Rebased onto latest `main` and pushed a follow-up commit addressing loose ends from an independent review. Root cause across all of them: the migration converted the *scaling* path but left per-pitcher `calibMilkG` half-alive across import, display, and MCP output.

**Fixes**
- **Backup import re-seeds the global rate (the important one).** The `calibMilkG → steamSecondsPerGram` conversion lived only in the `SettingsBrew` ctor (sentinel-guarded). Restoring a **pre-migration backup** brought back `calibMilkG` + `milkAutoCapture=ON` but no `steamSecondsPerGram`, so on a device where the ctor migration already ran, the rate stayed 0 — weight-timing silently dead with the toggle still lit. `importFromJson` now re-derives the rate from the restored presets when the key is absent. This is the import-time re-seed flagged in the PR's own "Note for reviewers". Shared derivation extracted to `deriveSteamRateFromLegacyPresets()` / `seedSteamRateFromLegacyPresets()`, reused by the ctor migration so both paths stay identical.
- **Stopped reading `calibMilkG` for display.** `SteamPlanText` dropped the stale-reference fallback (uses `lastSteamMilkG`); the SteamPage "Weight-timed: Xg → Ys" hint is now driven by the live `lastSteamMilkG` instead of the frozen per-preset value (which has no UI left to edit). `calibMilkG` is now inert stored data, kept only for backup fidelity + the reseed.
- **MCP `steam_pitcher_list` description** no longer claims `calibMilkG` drives weight-scaling.
- Softened the "(same steam flow)" code comments to name the one-rate model a simplification (matching the in-app note).

**Tests (8 new)** — migration seeding + first-preset-wins, sentinel-prevents-reseed, import re-seed, rate export/import round-trip, `calibrateSteamFromReference` happy-path + non-positive guard, negative clamp, emit-once. Also made a pre-existing uncalibrated-fallback test hermetic (it was relying on ambient store state).

Builds clean; `tst_Settings` 2708/2708 green.

PR #1464, opened 2026-07-10
ShotServer: stop returning raw API keys/passwords over the settings API

loganross2001:
## Problem

The optional on-device LAN web server (ShotServer) exposed stored secrets in plaintext to any client that could reach it:

- `GET /api/settings` returned the **raw values** of the four AI API keys, the Visualizer username/password, and the MQTT username/password.
- `GET /api/backup` (`SettingsSerializer::exportToJson` with `includeSensitive=true`) returned the same secrets in the backup JSON, plus the Visualizer and MQTT usernames unconditionally.

This read path is only gated behind `webSecurityEnabled`, which is **off by default**, so on a typical setup anyone on the local network could read every configured credential.

## Fix (keeps the settings page fully functional)

- **Redact secret values on the wire.** `/api/settings` returns a fixed mask when a secret is configured and `""` when it is not, so the page still shows a field as configured/filled without disclosing the value. The web UI only checks truthiness for provider availability, so it keeps working.
- **Save only real new values.** A posted secret is applied only when it is non-empty and not the mask, so the redacted round-trip (or an untouched field) never overwrites the stored secret; a newly typed value still saves normally.
- **Gate the Visualizer and MQTT usernames** behind `includeSensitive` in the serializer, consistent with the passwords.
- **Force `includeSensitive=false`** on the web backup endpoint, so raw secrets are never emitted over HTTP even if `includeSensitive=true` is requested.
- **Visualizer connection test** substitutes the stored credential when the form still holds the mask, so an already-configured account can be tested without re-entering it.
- **Close the MQTT broker-redirect path.** Broker host/port are not secrets and are applied verbatim, but the stored MQTT password is emitted to whatever broker we connect to — and changing the host also triggers an automatic reconnect. The guard lives in `applyMqttSettings`, the shared chokepoint that both `POST /api/settings` and the connect endpoint funnel through: it refuses to retarget the broker host/port while the password field holds only the mask/empty, so a LAN client can never redirect the broker and have the stored password sent to it in the CONNECT packet. The connect endpoint additionally returns a clear "re-enter the password" error in that case.

## Notes

- Follow-up (non-blocking, out of scope here): on the save path a masked-password host change is silently kept at the stored host rather than surfacing a "re-enter password" message. This is the secure failure mode; a nicer UX would force password re-entry in the web form when host/port change.
- Verified with a clean macOS build.

### DIFF
--- a/src/core/datamigrationclient.cpp
+++ b/src/core/datamigrationclient.cpp
@@ -533,9 +533,12 @@ void DataMigrationClient::onSettingsReply()
 
         QJsonDocument doc = QJsonDocument::fromJson(data);
         if (doc.isObject() && m_settings) {
-            // Exclude machine-specific calibration and broker-specific MQTT password.
-            // Other sensitive keys (API keys, visualizer password) are intentionally
-            // transferred — they belong to the user, not the machine.
+            // The /api/backup/settings response omits all raw credential values
+            // (the endpoint forces includeSensitive=false), so API keys, passwords,
+            // and the visualizer/MQTT usernames are not transferred and must be
+            // re-entered on the target device. Only non-credential settings transfer.
+            // We additionally drop machine-specific flow calibration and defensively
+            // re-exclude any mqttPassword key on import.
             SettingsSerializer::importFromJson(m_settings, doc.object(), {"flowCalibration", "mqttPassword"});
             m_settingsImported = 1;
             qDebug() << "DataMigrationClient: Settings imported successfully";
--- a/src/core/settingsserializer.cpp
+++ b/src/core/settingsserializer.cpp
@@ -251,8 +251,11 @@ QJsonObject SettingsSerializer::exportToJson(Settings* settings, bool includeSen
 
     // Visualizer settings
     QJsonObject visualizer;
-    visualizer["username"] = settings->visualizer()->visualizerUsername();
     if (includeSensitive) {
+        // Username is a credential too — keep it behind includeSensitive so the LAN
+        // web-backup endpoint (which forces includeSensitive=false) never emits it,
+        // consistent with how /api/settings redacts it.
+        visualizer["username"] = settings->visualizer()->visualizerUsername();
         visualizer["password"] = settings->visualizer()->visualizerPassword();
     }
     visualizer["autoUpload"] = settings->visualizer()->visualizerAutoUpload();
@@ -341,8 +344,9 @@ QJsonObject SettingsSerializer::exportToJson(Settings* settings, bool includeSen
     mqtt["enabled"] = mqttSettings->mqttEnabled();
     mqtt["brokerHost"] = mqttSettings->mqttBrokerHost();
     mqtt["brokerPort"] = mqttSettings->mqttBrokerPort();
-    mqtt["username"] = mqttSettings->mqttUsername();
     if (includeSensitive) {
+        // Username is a credential too — gated with the password (see visualizer above).
+        mqtt["username"] = mqttSettings->mqttUsername();
         mqtt["password"] = mqttSettings->mqttPassword();
     }
     mqtt["baseTopic"] = mqttSettings->mqttBaseTopic();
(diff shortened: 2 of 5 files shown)


skialpine:
### Code review

Reviewed the secret-redaction change and pushed a follow-up fix commit (6df1d1158e) addressing four issues found. The core fix (redacting secrets on `/api/settings` and `/api/backup`) is a genuine, worthwhile security improvement — this just closes gaps the first pass left.

**1. Broker-redirect guard leaked the stored MQTT password to a new broker (security).** In `applyMqttSettings`, host/port were applied before the re-entered password. `setMqttBrokerHost()` fires `mqttBrokerHostChanged` → `MqttClient::onSettingsChanged()`, which reconnects **synchronously** and reads the password live from Settings ([mqttclient.cpp#L278](https://github.com/Kulitorum/Decenza/blob/6df1d1158e/src/network/mqttclient.cpp#L276-L285)). So a LAN client posting `mqttBrokerHost=<attacker>` + any non-mask decoy password bypassed the guard and had the **old real password** sent to the attacker's broker before the decoy was applied. Fix: apply credentials first, then host/port ([shotserver_settings.cpp#L178-L197](https://github.com/Kulitorum/Decenza/blob/6df1d1158e/src/network/shotserver_settings.cpp#L178-L197)).

**2. `handleSaveSettings` reported success on a refused broker retarget.** `applyMqttSettings` returns `true` when it refuses a host/port change, and `handleMqttConnect` honors it — but the plain Save path discarded it and always replied `{"success": true}`, silently dropping the change. Now surfaced like the connect path ([shotserver_settings.cpp#L1154-L1176](https://github.com/Kulitorum/Decenza/blob/6df1d1158e/src/network/shotserver_settings.cpp#L1154-L1176)).

**3. Username gating regressed backup/restore and device migration.** Moving `visualizer`/`mqtt` usernames behind `includeSensitive` in the shared `exportToJson` also affected `DatabaseBackupManager::createBackup()` and the migration full-archive (both pass `includeSensitive=false`), silently dropping usernames from those flows. Usernames are un-gated again; the LAN `/api/settings` view still redacts them separately in `handleGetSettings` ([settingsserializer.cpp#L252-L260](https://github.com/Kulitorum/Decenza/blob/6df1d1158e/src/core/settingsserializer.cpp#L252-L260)).

**4. Misleading comments.** The `handleBackupSettings` comment implied the native backup path preserves these secrets (it doesn't — it also uses `includeSensitive=false`), and the migration comment claimed API keys / visualizer password are transferred (they aren't; the endpoint forces `includeSensitive=false`). Both corrected.

skialpine:
### Correction to point 3 above (username gating)

A follow-up security check flagged my un-gating of the usernames: `handleBackupSettings` / `handleBackupFull` are **unauthenticated** over the LAN when `webSecurityEnabled` is off (the default — the auth middleware only runs when security is enabled). Emitting the Visualizer/MQTT usernames there re-exposed them (a Visualizer username is often an email), which is exactly what this PR set out to prevent.

So I reversed course: **usernames are now gated behind `includeSensitive` alongside the passwords** (commit 8b0ae595). Consequence — web backup and device migration carry only non-credential settings; all credentials (API keys, passwords, usernames) are re-entered on the target device. That's consistent with the passwords, which already weren't transferred, so it's a coherent "re-enter credentials after transfer" posture rather than a regression. The comments in `handleBackupSettings` and `datamigrationclient` were updated to describe this accurately.

Open question for a possible follow-up (not in this PR): the on-device local backup (`DatabaseBackupManager`, saved to `Documents/Decenza Backups` and never served over LAN) also passes `includeSensitive=false`, so a local restore re-enters all credentials too. If you'd prefer local backups to be credential-complete, that path could pass `includeSensitive=true` — it's a local file, so no LAN exposure, but it does put secrets on disk. Left as-is pending your call.

loganross2001:
Thank you for the careful follow-up and the fixes — much appreciated. The credential-ordering catch is a real one: applying host/port before the re-entered password meant `setMqttBrokerHost` → synchronous reconnect could read the old stored password, so a decoy non-mask password slipped past the re-entry guard. Applying credentials first and only then host/port is the correct fix, and gating the usernames behind `includeSensitive` (with re-entry on the target device) is a coherent posture consistent with the passwords. Happy with all four changes — thanks for getting it over the line.

PR #1470, opened 2026-07-11
Shot start: recipe pill parity + permissive GHC default

skialpine:
## Problem

A user **without a GHC** reported that starting a shot from the **recipe pill** worked in the simulator but not on real hardware — while the **profile pill** worked in both. Two independent causes.

## 1. Recipe pill start — async selection race

The recipe pill's "tap the selected recipe again to start" gesture tested `Settings.dye.activeRecipeId`, which is only set at the **end** of the async activation pipeline (background DB read + BLE profile upload). On real hardware that echoes back *after* the user's second tap, so the tap re-activated instead of starting. The simulator (no BLE) completed activation instantly and masked the race.

**Fix:** `MainController.selectedRecipeId` — a synchronous, shared selection marker set the instant `activateRecipe()` is called (the recipe analogue of the profile pills' synchronous `Settings.app.selectedFavoriteProfile`). It leads `activeRecipeId` during activation, then tracks it:
- success confirms it,
- external deactivation (ingredient swap) clears it,
- **failed** activation rolls it back — so a second tap can't start a shot for a recipe that never applied.

Both the regular (`IdlePage`) and compact (`RecipesItem`) pill rows now share this one marker and behave identically. The compact popup previously **could not start a shot at all** — it only activated — so this closes a real layout gap.

## 2. GHC gate defaulted to "cannot start"

`m_isHeadless` defaulted to `false` ("app cannot start operations") until the GHC MMR read returned — inverted from de1app, whose `ghc_is_installed` defaults to `0` → `ghc_required()==0` → app can start. A slow or dropped read left **every** start button dead.

**Fix:** default `m_isHeadless = true`, and restore it on disconnect so a GHC machine's `false` can't carry into the next connection. The allowlist **values are unchanged** (`{0,1,2,4}`) — they already match de1app's `ghc_required`; only the default was wrong.

## Parity audit

Every espresso-start affordance was checked. The two pill re-tap paths were the only profile-specific starts with a recipe analogue, and both are now matched. The rest — MiniGHC button, custom `command:startEspresso`, keyboard "E", MCP `machine_start_espresso` — are profile-agnostic (they start the loaded profile, which *is* the active recipe's profile), so they already cover recipes. Profile pages (`ProfileSelectorPage`/`ProfileInfoPage`) never start a shot, exactly like `RecipesPage`.

## Diagnosability

QML `console.log` is captured by `WebDebugLogger`, so a user's debug log now pinpoints failures:
- recipe handlers distinguish the two gates (`app cannot start operations (active GHC?)` vs `machine not ready`),
- `DE1Device::startEspresso()` logs `state/headless/sim` for **every** start surface — the single anchor confirming whether the BLE command was issued.

## Testing

- Build clean (0 errors / 0 warnings).
- `tst_coffeebags`, `tst_recipestorage`, `tst_mcpserver_session`, `tst_mcpserver_protocol` all green (143 tests).
- The race is timing-dependent — **real-hardware verification still needed** (the simulator can't reproduce it): Recipes → tap a recipe → tap it again → shot starts on the second tap, in both regular and compact layouts.

### DIFF
--- a/qml/components/layout/items/RecipesItem.qml
+++ b/qml/components/layout/items/RecipesItem.qml
@@ -50,8 +50,10 @@ Item {
             if (typeof AccessibilityManager !== "undefined" && AccessibilityManager.enabled)
                 AccessibilityManager.announce(TranslationManager.translate("machine.notReady", "Machine is not ready"))
         } else {
-            console.log("[recipe pill/compact] starting espresso — recipe=" + recipe.id + " phase=" + MachineState.phase)
-            DE1Device.startEspresso()
+            // Deferred in MainController until the recipe's profile is applied,
+            // so a fast second tap can't pull a shot on the previous profile.
+            console.log("[recipe pill/compact] requesting start — recipe=" + recipe.id + " phase=" + MachineState.phase)
+            MainController.startSelectedRecipeShotWhenApplied()
         }
     }
 
--- a/qml/components/layout/items/RecipesItem.qml
+++ b/qml/components/layout/items/RecipesItem.qml
@@ -26,6 +26,35 @@ Item {
         return null
     }
 
+    // True when the app is allowed to start machine operations on-screen — same
+    // gate as the profile pills (EspressoItem). An active GHC takes exclusive
+    // control, so on-screen starts are only valid headless or in simulation.
+    readonly property bool canStartOperations: DE1Device.isHeadless || DE1Device.simulationMode
+
+    // Two-tap select-then-start, identical to the regular-layout recipe pill
+    // (IdlePage.tryStartRecipe) so both layouts behave the same. Selection is
+    // the synchronous, shared MainController.selectedRecipeId; the first tap
+    // selects (activates), a tap on the already-selected recipe starts the shot.
+    function tryStartRecipe(recipe) {
+        var alreadySelected = (recipe.id === MainController.selectedRecipeId)
+        if (!alreadySelected) {
+            MainController.activateRecipe(recipe.id)  // sets selectedRecipeId synchronously
+            return
+        }
+        if (!root.canStartOperations) {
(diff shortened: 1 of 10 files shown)


skialpine:
## Multi-agent review round — resolved

Ran a 5-agent review (code / silent-failure / comments / type-design / test-coverage). Outcome:

**Fixed in `37eb69a3`:**
- **A11y parity miss** (code-reviewer): the IdlePage recipes-row screen-reader announcement still read the lagging `activeRecipeId` while the highlight used `selectedRecipeId` — now consistent.
- **Sentinel enforcement** (type-design): non-positive ids normalized to `-1` so a stray `0` can't be seeded.
- **Interleaved-activation ordering** (type-design): documented the self-healing re-convergence.
- **Test gap on the GHC fix** (test-coverage): added `tst_de1device_headless` (default-true + restore-on-disconnect).

**Fixed in `361aa0dc`:**
- **Wrong-profile on fast double-tap** (silent-failure, rated critical): the synchronous marker could start a shot before the recipe's profile finished applying. The pill start now goes through `MainController.startSelectedRecipeShotWhenApplied()`, which **defers** the start until activation confirms (or cancels it on failure / reselection) — never pulling a shot on the not-yet-replaced profile.
- **Untestable state machine** (test-coverage): MainController is deliberately never linked in the suite, so the selection/deferred-start policy was extracted into a pure header-only `RecipeSelectionModel` with 14 direct unit tests (`tst_recipeselectionmodel`).

**Confirmed sound, no action:** comment accuracy (all de1app claims verified against `vars.tcl`), `qint64` type choice, encapsulation (9/10), the `m_isHeadless` fail-open default (machine is the real authority).

All 68 tests pass. The recipe race remains timing-dependent — real-hardware verification of the two-tap start (regular + compact) is still the last step before merge.

PR #1471, opened 2026-07-11
Fix Recipes page not re-applying an already-active recipe after edit

### LINKED ISSUE #1466: [Bug] choosing recipe not updating Shot Plan until Shot Plan is edited separately
### What happened?

See video
https://videoblurb.com/mp/decenza/decenza-shot-preview-not-recipe-updating.mp4
clicking on recipe did not change to the recipe on shot plan.

### Steps to reproduce

1 go to recipe and change the profile, grind size, bean
2 go to home screen, shot plan is the same (expected)
3 tap recipe button and select recipe you've just edited
4 shot plan does not update to new profile, bean, grind setting
5 tap on Profiles button, change the profile
6 tap on Recipes and again tap on the edited recipe
7 Result! The shot plan updates as expected.

### Decenza version

2

### Platform

Android

### Device / OS version

12

### Scale (if applicable)

_No response_

[debug 11july26.zip](https://github.com/user-attachments/files/29918827/debug.11july26.zip)

_No response_

### Screenshots or video

https://videoblurb.com/mp/decenza/decenza-shot-preview-not-recipe-updating.mp4

### Anything else?

_No response_

skialpine:
## Summary
- `RecipesPage.qml` skipped `MainController.activateRecipe()` when the tapped recipe's id already matched `Settings.dye.activeRecipeId`, treating "already selected" as "already applied."
- Editing a recipe's profile/grind/bean while it's the active recipe only refreshes the internal `m_activeRecipe` cache (via `recipeUpdated`/`recipeReady`) — it never re-pushes those values into the live `ProfileManager`/`SettingsDye` state the Shot Plan reads.
- Re-tapping the edited (still-active) recipe on the Recipes page was therefore a silent no-op, so the Shot Plan stayed stale until switching profiles deactivated the recipe first (only then did a re-tap actually run a real activation).
- `activateRecipe()`/`applyActivatedRecipe()` are already fully idempotent — they always re-read the recipe row and re-apply every stage (profile, bag, grind, dose/yield/temp, steam, hot water) — so dropping the guard is safe, and brings `RecipesPage.qml` in line with `RecipesItem.qml`/`IdlePage.qml`, which already call `activateRecipe()` unconditionally.

Fixes #1466

## Test plan
- [ ] In the Recipe editor, edit the profile/grind/bean of the currently active recipe and save.
- [ ] Go to the Recipes page and tap the same (still-active) recipe's plan area.
- [ ] Confirm the Shot Plan on the home screen updates immediately to the new profile/bean/grind, without needing to switch to a different profile first.

### DIFF
--- a/qml/pages/RecipesPage.qml
+++ b/qml/pages/RecipesPage.qml
@@ -212,7 +212,7 @@ Page {
             repointPicker.open()
         }
         onPlanClicked: {
-            if (!card.archivedCard && card.recipe.id !== Settings.dye.activeRecipeId)
+            if (!card.archivedCard)
                 MainController.activateRecipe(card.recipe.id)
         }
 
@@ -268,8 +268,7 @@ Page {
             accessibleName: TranslationManager.translate("recipes.accessible.activate", "Activate recipe %1").arg(card.recipe.name || "")
             accessibleItem: card
             onAccessibleClicked: {
-                if (card.recipe.id !== Settings.dye.activeRecipeId)
-                    MainController.activateRecipe(card.recipe.id)
+                MainController.activateRecipe(card.recipe.id)
             }
         }
 


skialpine:
### Code review

The change is a minimal, correctly-scoped fix that removes the `card.recipe.id !== Settings.dye.activeRecipeId` guard around `MainController.activateRecipe()` in `onPlanClicked`/`onAccessibleClicked` of `qml/pages/RecipesPage.qml`. This brings it in line with `RecipesItem.qml` and `IdlePage.qml`, which already call `activateRecipe()` unconditionally, and matches the "single activation path" contract documented in `docs/CLAUDE_MD/RECIPES.md`. `MainController::applyActivatedRecipe` re-reads and re-applies every stage on each call with no early-return keyed on "already active," so dropping the guard is safe.

---

<details>
<summary>2 below-threshold issues (25-50)</summary>

- [OK to skip] Every tap on an already-active recipe now re-runs the full activation pipeline (profile reload/BLE upload, dose write, steam settings) instead of being a no-op — but this matches the existing unguarded pattern in `RecipesItem.qml`/`IdlePage.qml`, and removing the guard is exactly what fixes the reported bug (score: 40)
- [OK to skip] Overlapping `activateRecipe(sameId)` calls from rapid re-taps could theoretically race and create duplicate steam-pitcher/water-vessel presets — but Qt's queued signal delivery serializes the async DB-read callbacks on the main thread and `QSettings` preset writes are synchronous, so the duplicate-creation window doesn't actually open in practice (score: 25)

</details>

PR #1472, opened 2026-07-11
Recipes: recipe-owned grind + real override flags (fixes #1468)

### LINKED ISSUE #1468: [Bug] oddness in relationship between Recipe, manual changes on Shot Plan & it's colour change
### What happened?

At present there is some "oddness" and possible consistency issues regarding the relationship between the Recipe and maintaining it's integrity as a setting, and all the manual changes you can make on the home screen via Shot Plan, Bean, Profile, History>Load etc...  

It seems that *usually* only Grind setting changes made elsewhere on the interface (outside of Recipes) will make a change to the Recipe grind setting, and this does make some sense, as you may create the Recipe, and still be dialing in, so need that change to be made automatically, also when the bean ages you'll need to grind finer to achieve the same resistance, so auto-grind size changes to Recipes do make sense.

There are a few issues however with making manual changes. 
Video 1 shows the oddness of the shot plan changing colour after doing a manually set shot.
https://videoblurb.com/mp/decenza/Video1_20260711_110005.mp4

Video 2 shows that adjusting grind setting, when the Shot Detail text is orange and then doing a shot, results in the grind setting not  being saved permanently to the Recipe, but actually being reverted to the original grind size after the completed shot.
https://videoblurb.com/mp/decenza/Video2_20260711_112243.mp4

### Steps to reproduce

see both videos

### Decenza version

2

### Platform

Android

### Device / OS version

12

### Scale (if applicable)

_No response_

[debug2 11july26.zip](https://github.com/user-attachments/files/29922879/debug2.11july26.zip)

_No response_

### Screenshots or video

_No response_

### Anything else?

IMO there needs to be some visual indication of why the Shot Plan changes colour, which parameter... if this is the reason the grind setting inadvertently reverts to a previous setting, there really should be some way of knowing this is a problem and what to do about it. It's a mystery to me, but a frustrating one ;)

skialpine:
Fixes #1468 — both reported symptoms (the Shot Plan turning orange for no visible reason, and a grind edit silently reverting after the shot) plus the grind-model simplification that removes the underlying fragility.

## The three defects

**Bug A — the override flags were one-way latches.** `hasTemperatureOverride`/`hasBrewYieldOverride` were armed by every profile load and never cleared, so the Shot Plan highlight was pure coincidence (does a stale frozen value happen to match the current profile?). The flags now mean *"a deliberate override is active"*: profile/recipe switches genuinely clear them, and shot-load / recipe-activation / BrewDialog only arm them when the value differs from the profile default by >0.1. Startup preserves persisted overrides but drops same-as-default noise latched by pre-fix sessions.

**Bug B1 — post-shot grind revert.** The post-shot bag stamp bypassed `SettingsDye`'s self-write counter; its echo looked like an external bag edit, triggering `applyActiveBag()` to reapply the bag's (deliberately stale, under the pin model) grind over the user's just-made edit. Fixed at the root: **grind always lives on the recipe** — the inherit/pin split and the write-through suspension are retired, the bag write-through is unconditional (the bag always mirrors the last dialed grind), so the bag can never be stale and the resync has nothing to revert.

**Bug B2 — re-activation race.** Re-tapping the active recipe (the #1466/#1471 flow) did a fresh uncoordinated DB read that could race the user's own in-flight write and revert it. Same-id re-activation now re-pushes the in-memory cache; a bag-link change since activation falls through to the full fresh read.

## Model changes

- **Creation is the one moment bag grind is read**: the wizard offers the bag's dial as an editable default (shot history for the bean+profile pair beats it); `requestCreateRecipe` adopts it when a bag-linked create map *omits* grind (explicitly empty stays empty); **promote-from-shot uses the shot's own recorded grind/rpm**.
- **Bean-less recipe activation clears the active bag** — the session must not stay attributed to (or write grind into) whatever bag was laying around.
- **Migration 30** backfills inherit-mode rows from their bag's dial (skipping undialed/tea bags); `importRecipesStatic` runs the same backfill for pre-migration transfer/backup sources.
- **Wizard**: inherit toggle removed; equipment reordered above grind so rpm-capability is known when the grind field renders.
- **MCP/web**: `grind.mode`/bag-resolved `effectiveGrind` removed; tool schemas + descriptions updated (clients reload schemas every connection — no compatibility concern).
- CustomItem's brew-settings highlight keeps `brewByRatioActive` deliberately: the compact button has no room for the Shot Plan's yield-override arrow, so it colors for either deliberate override; the temperature clause is now identical to ShotPlanText's.

## Verification

- Full suite: **2684 passed, 0 failed, 0 warnings** (Qt Creator).
- New coverage: migration-30 data pass (adopt/skip/idempotence), import backfill, create-time omitted-vs-explicit-empty default, promotion grind carry, unconditional bag write-through, genuine override clearing, defaults-set-no-override.
- Remaining manual checks (need the machine): the #1468 video repros, restart-survival of a grind edit, and re-tap-while-write-in-flight — listed as the unchecked items in `openspec/changes/fix-recipe-grind-integrity/tasks.md`.

OpenSpec artifacts (proposal/design/specs/tasks) ride in `openspec/changes/fix-recipe-grind-integrity/`. Wiki Manual edits are staged locally, held for release per convention.


skialpine:
## Review pass complete (4 parallel agents: code, tests, comments, silent-failures)

All surviving findings fixed in ef6c1212; suite now **2,691 passed, 0 failed, 0 warnings**.

**Real bugs caught by the review, fixed:**
1. **Same-id re-activation could destroy the bag's bean identity** (found independently by two agents). The cached bag snapshot is empty after a restart and stale after a bag edit; re-applying it wrote blank/old identity fields *through to the bag row*. Fixed by passing a deliberately empty bag map on the fast path — `applyActivatedRecipe` now skips the identity re-apply for it, and the snapshot member is deleted outright.
2. **Profile switch stamped `yieldG=0`/`tempOverrideC=0` into the still-active recipe** before the deactivate watcher ran (pre-existing shape, both sides touched by this PR). The yield/temp stamps are now gated on the loaded profile still being the recipe's own.
3. **The post-import grind backfill ran table-wide**, so a later transfer would stamp the bag's dial back onto a *local* recipe whose grind the user deliberately cleared — violating this PR's own "explicitly empty grind is respected" rule. Now scoped to the rows the import inserted, with a pinning regression test.

**Also fixed:** web-form labels that still taught the retired follow-the-bag model (+ blank grind is now omitted on create so the adopt-bag-dial rule is reachable from the built-in page), the self-contradicting `recipe_activate` description, `DrinkTypes::hasGrind()` extraction, ~10 stale comment remnants, and doc/spec precision issues.

**New coverage:** the headline Bug A scenario (profile switch clears overrides), startup preserves genuine overrides only, `uploadProfile` clears stale overrides, migration-30 wiring through the real `runMigrations()` gate + rpm edges, import-doesn't-clobber, `hasGrind`. MainController-level flows (same-id fast path, bean-less bag clear) are recorded in TESTING.md's Known Coverage Gaps — they need on-device manual verification.

fredphoesh:
too difficult to understand the ai summary.

PR #1490, opened 2026-07-12
build: target-scoped qt_add_resources so resource content changes rebuild without a manual reconfigure

skialpine:
## Problem

Resources for the `Decenza` target were wired via the classic variable form of `qt_add_resources`:

```cmake
qt_add_resources(RESOURCES resources/resources.qrc)
qt_add_resources(RESOURCES resources/emoji.qrc)
qt_add_resources(RESOURCES resources/ai.qrc)
```

This form generates the resource `.cpp` at CMake **configure** time and only tracks the `.qrc` file's own timestamp as a build dependency — **not** the individual files listed inside it. So when a bundled resource's **content changes** (an SVG icon, a sound, a font, a profile JSON, a KB doc), a plain `cmake --build` does **not** regenerate the resource or relink — you have to manually "Run CMake" (reconfigure), and a plain rebuild silently runs the stale binary. This bit us repeatedly during the bundled-fonts / new-SVG-icon work.

## Fix

Migrate `resources.qrc` and `ai.qrc` for the `Decenza` target to the Qt 6 **target-scoped** signature `qt_add_resources(<target> "<name>" PREFIX "/" BASE ... FILES ...)`, mirroring the existing `shot_eval` precedent in the same file. This registers every listed file as a real build dependency, so ninja/make detect content changes and re-run RCC + relink automatically.

- **Paths unchanged.** `BASE=resources` reproduces each `.qrc`'s `prefix="/"` mapping, so every `qrc:/...` path is identical. The three aliased entries (`icons/espresso.svg` → `espresso 8mm.svg`, `icons/flush.svg` → `flush 8mm.svg`, `qrcode.png` → `Logo/qrcode.png`) keep their aliases via the `QT_RESOURCE_ALIAS` source-file property.
- **`emoji.qrc` left on the variable form on purpose.** Its 700+ Twemoji SVGs are a static, machine-generated set whose content never changes — only additions happen, which edit the `.qrc` itself (picked up by the timestamp dep). A FILES list of 700+ entries would be pure noise. Documented inline.
- **No double-registration.** `resources.qrc` / `ai.qrc` are removed from the Decenza target but kept on disk because the **test** targets (`tests/CMakeLists.txt`) still consume them via the variable form. The neighboring `shot_eval` comments are updated to reflect the now-three independent AI-KB file lists (`decenza_ai`, `shot_eval_ai`, `resources/ai.qrc`).

## Verification

Configure + RCC build with the real macOS Qt 6.11.1 kit:

- ✅ Clean `qt-cmake` configure + RCC build succeed, no resource/alias/duplicate errors.
- ✅ Generated `.qrc` confirms every path/alias is unchanged (spot-checked icons, the three aliases, profiles, `ai/*`).
- ✅ **Payoff:** after building the RCC object clean, `touch`-ing an SVG makes ninja mark exactly `qrc_decenza_resources.cpp.o` stale on a plain incremental build — RCC re-runs and the binary relinks **without** a manual "Run CMake".

No runtime behavior change — build-system quality-of-life only.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -576,15 +576,247 @@ if(APPLE)
 endif()
 
 # Resources
-qt_add_resources(RESOURCES resources/resources.qrc)
+#
+# emoji.qrc stays on the classic variable-form qt_add_resources (resolved at
+# CMake *configure* time). Its 700+ Twemoji SVGs are a static, machine-generated
+# set whose CONTENT never changes — only additions happen, and adding an emoji
+# already edits emoji.qrc itself, so the .qrc-timestamp dependency the variable
+# form tracks is enough to trigger a rebuild. Enumerating 700+ files in a
+# target-scoped FILES list would be pure noise for no practical gain.
+#
+# resources.qrc and ai.qrc are migrated to the target-scoped FILES signature
+# below (see "Target-scoped resources", after the Decenza target is created).
+# That form registers every listed file as a real build dependency, so RCC
+# re-runs and the binary relinks automatically when an icon/sound/profile/KB
+# file's CONTENT changes — no manual "Run CMake" reconfigure required. The
+# variable form would only notice a changed *.qrc* timestamp, silently shipping
+# a stale binary after an in-place edit to a listed file (this bit us repeatedly
+# during the bundled-fonts / new-SVG-icons work).
 qt_add_resources(RESOURCES resources/emoji.qrc)
-qt_add_resources(RESOURCES resources/ai.qrc)
 
 # Main executable
 qt_add_executable(Decenza
     ${SOURCES}
     ${HEADERS}
     ${RESOURCES}
+)
+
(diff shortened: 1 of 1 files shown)


skialpine:
Closing as unneeded. On investigation I could not reproduce the stale-resource problem this PR targets: on Qt 6.11.1, the classic variable-form `qt_add_resources(VAR foo.qrc)` already regenerates RCC and relinks on both content changes to listed files AND additions to the .qrc — verified on the Ninja generator (local Qt Creator) and the Xcode generator (macOS CI), in a minimal repro and in the real project on main. So the migration is a no-op for rebuild behavior on the current toolchain while adding a dual-maintenance cost (the FILES list and resources.qrc must be hand-kept in sync since the test targets still consume the .qrc). The PR #1487 pain was likely an older Qt/CMake (since fixed) or a transient stale state.

PR #1492, opened 2026-07-12
Fix recipe yield/temp shown as overrides in Brew Settings (#1485)

### LINKED ISSUE #1485: [Bug] clearing or resetting over-ride values is not recorded in Recipes
### What happened?

See video
https://videoblurb.com/mp/decenza/overridenotrecording_20260712_200303.mp4

I initially was wondering why there were overrides as I had not created them in the first place, but then saw that when I "Cleared" ie Reset to default the parameters, the Recipes were not updated.

### Steps to reproduce

See video
https://videoblurb.com/mp/decenza/overridenotrecording_20260712_200303.mp4

I tried using CLEAR button I also tried changing the value back to zero.

What is noticeable is that when you set it to the default value, the ability to UPDATE RECIPE is removed! This is weird! That's exactly when you might want it... There is something about the over-ride status of a shot plan that breaks (?) the connection between shot plan and Recipe. 

### Decenza version

2

### Platform

Android

### Device / OS version

12

### Scale (if applicable)

_No response_

[12 july 822pm debug.zip](https://github.com/user-attachments/files/29944614/12.july.822pm.debug.zip)

_No response_

### Screenshots or video

https://videoblurb.com/mp/decenza/overridenotrecording_20260712_200303.mp4

### Anything else?

decenza seems to be CREATING the over-rides. I was just testing on my PC and had one Recipe which had ONE override parameter. I zeroed this. Then saw no more override. Then went to HISTORY to create two more Recipes. When I returned to home screen, the original Recipe had TWO over-rides, the original temp back to -4 and the stop at weight changed from 42 to 36!!??
I have a second video which I will upload shortly
SECOND VIDEO showing "creation of over-rides"  https://videoblurb.com/mp/decenza/over-ride-not-recording.mp4

skialpine:
## Why

A recipe's yield and temperature are the recipe's **design** — its baseline — not overrides of the profile. Brew Settings measured "overridden" (highlight, Temp Delta zero-point, Clear target, Update-Recipe enablement) against the **profile default in both modes**, so:

- a recipe's own yield/temp were painted as amber "overrides" the user never created ("why are there overrides I didn't make?");
- **Clear** wiped them back to the profile instead of the recipe;
- **Update Recipe greyed out exactly when you reset to the profile default** — so a recipe's override could never be cleared, and re-activation kept restoring it.

Fixes [#1485](https://github.com/Kulitorum/Decenza/issues/1485).

## What changed

When a recipe is active, everything anchors on the recipe's own `yieldG`/`tempOverrideC` via two new baselines — `recipeTempBaseline` / `recipeYieldBaseline` — that fall back to the profile when the recipe pins none and **collapse to the profile entirely in no-recipe mode (that dialog is byte-for-byte unchanged)**:

- **Override highlight + Temp Delta zero-point + Ratio baseline** anchor on the recipe → a recipe's designed values render un-highlighted (baseline, not phantom override).
- **Clear** reverts to the recipe baseline, not the profile.
- **Both Update-Recipe buttons** gate on the field's `overridden` state — the invariant is **Update enabled ⟺ value highlighted** — so the recipe baseline is movable to any value, including back to the profile default (the reporter's greyed-out bug).

No storage/schema change; activation already applies the recipe's stored yield/temp as the opening overrides, which is what makes them the baseline.

## Verification

Driven end-to-end on the running macOS build (pure-QML change, no `Q_OS_*` branches, so macOS is representative):

- Recipe holding 40g on a profile targeting 36 → Stop-at shows **40 white** (old build: amber).
- **Clear** returns Stop-at to **40 (recipe)**, not 36 (profile).
- Dial back to 36 (profile default) with recipe at 40 → **Update Recipe stays enabled**.
- Temp Delta re-anchors: 0° at the recipe temp, +1° highlights amber.

A bug was caught *by driving the app*: the Temp Update-Recipe button was wrongly enabled at delta 0 when the recipe had no temp override (compared against the raw stored `0` instead of the baseline). Fixed by gating on `overridden`; re-verified in a fresh build.

## OpenSpec

Modifies `recipe-aware-brew-settings` and `brew-overrides` (baseline/Clear/enablement become recipe-relative in recipe mode). Delta specs included; archive will land as the final commit before merge.


fredphoesh:
@skialpine hey Jeff, here is a video 
([13july26android debug.zip](https://github.com/user-attachments/files/29962022/13july26android.debug.zip)
and debug) which shows a couple of things, 
• inconsistent "orange bits" data not the same Recipe VS Shot Plan
• reset temp pressing + and -  revealing a weirdness
• shot plan orange colour not updating without clicking away and back again
https://videoblurb.com/mp/decenza/recipe-shotplan_20260713_094128.mp4

as a side note, and i guess i should make it a feature request, when editing a recipe, should be no need to first press Continue THEN have to press Save too... The first page should say Save/Cancel instead of Continue... I have failed to save MANY times because of this. (as you see on the video)

PR #1496, opened 2026-07-13
Recipe wizard: one-press save in edit mode + unsaved-changes guard

skialpine:
## Problem

Editing a recipe required pressing **Continue** (details step) and then **Save** (summary) — two commits for one edit. Worse, every exit path (Cancel, back arrow) silently discarded edits, and after Continue the summary's WYSIWYG hero card already showed the new values, so backing out read as "saved" while losing everything. Reported after many lost edits; reproduced live on macOS (grind edit → Continue → back → change gone, no warning).

## Changes (all in `RecipeWizardPage.qml`)

- **Details step entered from the summary (edit / clone / promote) now shows Cancel / Save instead of Continue**, and Save commits directly — no second press. The creation walk keeps Continue → summary review, which is where that review has value.
- **Unsaved-changes guard on every exit path.** A baseline snapshot of the exact save map is captured at load (`buildSaveMap()` extracted from `save()`); any exit with a differing map opens the existing `UnsavedChangesDialog` (Discard / Save, Esc keeps editing). Covers summary Cancel, details Cancel, and both back-arrow exits. A half-built creation draft that isn't saveable yet still exits silently.
- The pinned error label now also renders on the details step so a save failure there is never hidden below the scroll.

## Verification (live, macOS build)

- Edit → Details now shows Cancel/Save with "…then Save." caption ✔
- Grind 12→13 saved with one press from the details step, persisted to the list ✔
- Dirty exit (clone, grind changed, back-arrow) intercepted by Unsaved Changes dialog; Discard leaves nothing behind ✔
- Clean exit (no edits) leaves without any prompt ✔

### DIFF
--- a/qml/pages/RecipeWizardPage.qml
+++ b/qml/pages/RecipeWizardPage.qml
@@ -1997,9 +1997,9 @@ Page {
                                 ? TranslationManager.translate("recipes.wizard.detailsOptionalEdit",
                                       "Everything here is optional. "
                                       + "Tap a section to adjust it, then Save.")
-                                : TranslationManager.translate("recipes.wizard.detailsOptional",
+                                : TranslationManager.translate("recipes.wizard.detailsOptionalReview",
                                       "Everything here is optional — it's prefilled and ready to save. "
-                                      + "Tap a section to adjust it, then Continue.")
+                                      + "Tap a section to adjust it, then Review.")
                             font: Theme.captionFont
                             color: Theme.textSecondaryColor
                             wrapMode: Text.WordWrap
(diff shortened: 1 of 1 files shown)


skialpine:
**Live verification of the review-fix commit (macOS debug build, simulation mode):**

- Details step in edit mode: Cancel/Save header, one-press save persisted (grind 12→13→12 round-trip) ✔
- Dirty exit intercepted by Unsaved Changes dialog; Discard leaves the stored recipe untouched; clean exits stay silent ✔
- Back/Escape Keys intercept: instrumented and confirmed the page has activeFocus and its `Keys.onReleased` handler receives key events (tested with a plain key) — the same delivery path Android's `Qt::Key_Back` uses, and the same pattern as RecipeEditorPage/PostShotReviewPage.

**Pre-existing issue found while verifying, out of scope here:** on macOS, the Escape key never reaches the QML item layer at all — Main.qml's global "Escape → go back" handler doesn't fire on any page (verified on the Recipes page with logging: an 'a' key reaches page handlers, Escape doesn't). Android hardware back is unaffected. Worth a separate investigation; should be confirmed with a physical keyboard to rule out synthetic-event artifacts.

skialpine:
Correction to the earlier verification comment: the "Escape never reaches the QML layer on macOS" observation was an artifact of synthetic key events from the automation tooling, not an app bug. Verified with a physical keyboard: Esc navigates back correctly (Beans page → idle), so the global handler and this PR's page-level intercept both receive real Escape presses as designed. No follow-up needed.

PR #1517, opened 2026-07-16
Zone options: honor them on the idle screen, drop the ones that can't apply

loganross2001:
### What

The per-zone options (`distribution` / `alignment` / `style`) are offered for every zone by both editors, persisted in the layout JSON, and rendered by both previews — but the idle screen honored them for only two of nine zones.

This PR started as a fix for `alignment` on the center zones. Reviewing it surfaced that the gap was much wider, so the scope grew to close it properly:

| Zone | Renderer | distribution | alignment | style |
|---|---|---|---|---|
| `statusBar` | `StatusBar.qml` | ✅ | ✅ | ✅ |
| `lowerMidBar` | `IdlePage.qml` | ✅ | ✅ | ✅ |
| `topLeft` / `topRight` | `IdlePage.qml` | ❌ | ❌ | ❌ |
| `bottomLeft` / `bottomRight` | `IdlePage.qml` | ❌ | ❌ | ❌ |
| 3 center zones | `LayoutCenterZone.qml` | n/a | ❌ | ❌ |

`LayoutBarZone` implements all three options correctly — `IdlePage` simply never passed them to the four top/bottom bar zones, so they silently fell back to the component defaults. `LayoutCenterZone` implemented neither `alignment` nor `style`.

The result was a dead control: pick an option, watch both previews honor it, see nothing on the idle screen.

### Changes

- **Wire the four top/bottom bar zones** — `topLeft`, `topRight`, `bottomLeft`, `bottomRight` now receive `distribution` / `alignment` / `style`, as `lowerMidBar` already did. This is the bulk of the fix and adds no new mechanism.
- **Center zones honor `alignment`** — the original change, via the same auto-spacer mechanism `LayoutBarZone` uses.
- **Center zones honor `style`** — item text color and value bold, matching what the top/bottom bars already do. Center zones paint no background of their own; only `statusBar` and `lowerMidBar` do.
- **Center zones drop `distribution`** — hidden in both editors, and the web preview no longer approximates it. A center zone sizes every item from a fixed cell (`buttonWidth`, capped at `Theme.scaled(150) * zoneScale`) so its action buttons never stretch — which is exactly what `equalWidth` and `spaced` require. With the cap kept, the two values collapse into each other and do nothing at all on a button-only row like `centerTop`, so the honest fix is to stop offering them rather than ship a third dead control.
- **One `zoneOpts(zone)` accessor** in `IdlePage` replaces three hand-rolled per-zone lookups.
- Also folded in: the empty `Component.onCompleted` in `LayoutCenterZone`, and two inaccurate comments from the first draft.

### Compatibility

Layouts that never set these options are **unchanged** — every default (`packed` / `center` / `standard`) resolves to exactly today's rendering.

Layouts that **did** set them will change on upgrade: the idle screen starts honoring a choice the user already made and previewed. That's the intent, but it's a visible change rather than a no-op — worth a release-note line.

### Spec & docs

- OpenSpec change `fix-idle-zone-option-wiring`, archived in this PR. It narrows distribution to the bar zones and adds a requirement that the idle screen honor every option the editors offer — and that an option a zone cannot express is offered nowhere.
- Wiki manual updated: Distribution is documented as bar-zones-only, and Style's background is clarified as status-bar/lower-mid-bar only.

### Verification

Build clean (0 errors, 0 warnings). **Not yet verified on-device** — tasks 5.2-5.5 in the change are deliberately left unchecked.

### Notes

Original commit by @loganross2001; the wiring/scope expansion was added on top during review.

One thing left open: right-aligning `centerTop` moves the action buttons, but the inline preset pill row below stays centered (`Layout.alignment: Qt.AlignHCenter`), so the pills detach from their buttons. Recorded as task 5.5 rather than fixed here.


skialpine:
Reviewed and expanded the scope on top of your commit — flagging what changed and why, since this is now quite a bit more than you submitted.

Your fix was correct: `alignment` really was dead for the center zones, settable and previewed but inert. Verifying it turned up that the same gap was wider than the center zones. `IdlePage` never passed `distribution`/`alignment`/`style` to `topLeft`/`topRight`/`bottomLeft`/`bottomRight` either — `LayoutBarZone` implements all three, they were just never wired, so they fell back to component defaults. Only `lowerMidBar` and `statusBar` were fully wired. Rather than fix one third of a dead-option bug, this PR now closes it across the idle screen.

Two things went the other way from your draft:

- **`distribution` is now hidden for center zones** rather than implemented. Their sizing caps items at `buttonWidth` so action buttons never stretch, which is precisely what `equalWidth`/`spaced` need — keep the cap and the two values become indistinguishable and inert on a button-only row like `centerTop`. Offering them would just be a third dead control.
- **Your "same auto-spacer mechanism as LayoutBarZone" comment got reworded.** The mechanism matches, but the gating doesn't: the bar gates on `fillRow` (`fillWidthMode || hasSpacer`), the center zone on `hasSpacer`. They're equivalent today only because center zones have no distribution — worth stating precisely so it isn't a trap later.

Your `"center"`-is-byte-identical claim checked out exactly: both new conditions reduce to `!hasSpacer`.

Credit stays on your commit. Thanks for spotting the original — it was a real bug and the thread you pulled turned out to be attached to a bigger one.

PR #1518, opened 2026-07-16
Visualizer recovery: import missing shots by date range

loganross2001:
### What

Adds a **recovery** flow to the Visualizer settings card: pick a date range and import shots that exist in your Visualizer account but are missing from this device's local history — e.g. after a device swap, a fresh install, or a gap in sync. Shots are fetched, parsed into `ShotRecord`s, de-duplicated against local history, and imported through the existing import path.

- `src/network/visualizerimporter.{cpp,h}` — fetch-by-date-range + import pipeline (loud failures on non-200/parse errors; hollow-shot and phantom-frame guards; idempotent re-runs).
- `src/history/shotfileparser.{cpp,h}` — `parseVisualizerShot`: Visualizer shot JSON → `ShotRecord` (scalars, metadata incl. the standard `barista` who-pulled field, notes, samples). Robust against malformed/partial JSON (sample arrays zipped with `qMin`, string-encoded scalars normalized, empty-timeframe/pressure rejected).
- `qml/pages/settings/SettingsVisualizerTab.qml` — the date-range recovery UI, and a fix so the Upload-Settings card scrolls (the recovery section was clipped below the fold).
- `tests/tst_visualizershotparse.cpp` — parser coverage incl. the malformed-input / value-mapping edge cases.

No new schema; imported shots use the existing `shots` table and the existing `importShotRecord` de-dupe path. Visualizer API usage (endpoints, auth header, DYE field names) mirrors the existing `visualizeruploader`.

### Known limitation (documented in code)

In the **sync-gap** case (some shots already local, recover the rest), a shot pulled on this device has a filename-keyed local uuid, so the recovery's visualizer-id-keyed uuid can't match it and dedupe falls to the existing secondary check (`|Δtimestamp| < 5s AND profile_name == profile_title`). The timestamp is an exact round-trip, so the sole discriminator is the profile title — a shot uploaded by a *different* client with a differently-formatted title, or whose profile was later renamed, can slip past and re-import as a **duplicate row** (not corruption). Fresh-install/device-swap recovery and re-running recovery are unaffected. A fully robust key would persist + compare the Visualizer id on the local shot too; I left that out because it touches the shared import/dedupe path — happy to add it if you'd prefer.

### Follow-up

This is a new user-visible feature, so it wants a wiki manual entry — I'll add that separately (the wiki is a separate repo). Flagging per the contributing guidance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -508,6 +508,7 @@ set(HEADERS
     src/ui/jscanvascontext.h
     src/network/visualizeruploader.h
     src/network/visualizerimporter.h
+    src/network/visualizershotlist.h
     src/network/beanbaseclient.h
     src/network/beanbase_blob.h
     src/ai/aimanager.h
--- a/qml/pages/settings/SettingsVisualizerTab.qml
+++ b/qml/pages/settings/SettingsVisualizerTab.qml
@@ -55,12 +55,17 @@ KeyboardAwareContainer {
                 .arg(imported).arg(total).arg(skipped).arg(failed)
         }
         function onRecoveryComplete(total, imported, skipped, failed) {
-            visualizerTab.recoverStatusError = false
             if (total === 0) {
+                visualizerTab.recoverStatusError = false
                 visualizerTab.recoverStatus = TranslationManager.translate(
                     "settings.visualizer.recoverNothing",
                     "No shots found in that date range.")
             } else {
+                // Style as an error when nothing came through but shots were
+                // found — otherwise a wholesale failure (schema change, a run of
+                // server errors) reads identically to a clean success. Any
+                // non-zero failure count is worth flagging too.
+                visualizerTab.recoverStatusError = failed > 0
                 visualizerTab.recoverStatus = TranslationManager.translate(
                     "settings.visualizer.recoverDone",
                     "Done: %1 imported, %2 already present, %3 failed (of %4).")
@@ -588,7 +593,7 @@ KeyboardAwareContainer {
                             ? TranslationManager.translate("settings.visualizer.recovering", "Retrieving…")
                             : TranslationManager.translate("settings.visualizer.recoverButton", "Retrieve")
                         accessibleName: TranslationManager.translate(
-                            "settings.visualizer.recoverButton", "Retrieve shots from Visualizer")
+                            "settings.visualizer.recoverButtonA11y", "Retrieve shots from Visualizer")
                         primary: true
                         enabled: visualizerTab.visualizerConnected &&
                                  !MainController.visualizerImporter.recovering
--- a/qml/pages/settings/SettingsVisualizerTab.qml
+++ b/qml/pages/settings/SettingsVisualizerTab.qml
@@ -13,6 +13,80 @@ KeyboardAwareContainer {
     property string testResultMessage: ""
     property bool testResultSuccess: false
 
+    // --- Recover shots from Visualizer (date-range history import) ---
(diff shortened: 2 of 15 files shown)


skialpine:
### Code review

Found 4 issues:

1. Recovered shots persist with **zero frame markers**. `parseVisualizerShot` builds each phase marker without a `label`, but `shot_phases.label` is `TEXT NOT NULL`, so every phase INSERT in `importShotRecordStatic` fails the constraint — and the result is ignored (`query.exec(); // Non-critical if markers fail`). Recovered shots draw no frame lines even though the transitions were computed. (The `.shot` path always sets a label.)

https://github.com/Kulitorum/Decenza/blob/cdb4ee78a12e78995fefcb8aca853583c74af968/src/history/shotfileparser.cpp#L403-L411

2. The same markers are numbered from **1** (`++frameNumber`), so no `frameNumber == 0` marker ever exists. Once issue 1 is fixed and markers persist, `detectSkipFirstFrame` never sees frame 0, takes the strict FW-bug branch, and spuriously badges any recovered shot whose first transition is under 2s. `CLAUDE.md` names [`docs/SHOT_REVIEW.md`](https://github.com/Kulitorum/Decenza/blob/cdb4ee78a12e78995fefcb8aca853583c74af968/docs/SHOT_REVIEW.md) as the source of truth for this detector.

https://github.com/Kulitorum/Decenza/blob/cdb4ee78a12e78995fefcb8aca853583c74af968/src/history/shotfileparser.cpp#L404-L407

3. Per-shot recovery has no transient retry: a single network blip on the download or profile fetch permanently counts that shot as `failed` for the run, diverging from the uploader's bounded-retry convention. Only a manual re-run of the range recovers it.

https://github.com/Kulitorum/Decenza/blob/cdb4ee78a12e78995fefcb8aca853583c74af968/src/network/visualizerimporter.cpp#L870-L877

4. New recovery error messages are hardcoded English built in C++ and rendered directly in the settings UI (`recoverStatus`), bypassing i18n. [`CLAUDE.md`](https://github.com/Kulitorum/Decenza/blob/cdb4ee78a12e78995fefcb8aca853583c74af968/CLAUDE.md) says "All user-visible text must be internationalized. Use `TranslationManager.translate(...)`". (Mirrors a pre-existing pattern in the same class.)

https://github.com/Kulitorum/Decenza/blob/cdb4ee78a12e78995fefcb8aca853583c74af968/src/network/visualizerimporter.cpp#L752-L759

Issues 1 and 2 are in the same `espresso_state_change` block and share a fix (give the markers a non-null label and number them so a frame-0 marker exists, matching the live-shot convention).

<details>

- [Worth a look] `importShotRecordStatic`'s INSERT omits columns `saveShotStatic` persists (`beanbase_json`/`bag_id`/`recipe_id`/`steam_json`/`hot_water_json`/`stopped_by`/…); latent today since the recovery parser doesn't populate them, but the PR touched this INSERT (score: 73)
- [Worth a look] `importShotRecordStatic` uses a plain `db.transaction()` instead of the `BEGIN IMMEDIATE` + lock-retry that `saveShotStatic` adopted; `withTempDb`'s `busy_timeout` + idempotent re-run mitigate it (score: 50)
- [OK to skip] No index on `shots(visualizer_id)` for the new per-shot dedupe probe — negligible against a network-bound recovery, and consistent with the existing back-sync scans (score: 50)
- [OK to skip] Wiki manual entry for the feature not added yet — author already flagged it as a separate follow-up in the wiki repo (score: 50)

</details>

PR #1521, opened 2026-07-16
fix(scale): propagate watchdog-forced disconnect so auto-reconnect arms

### LINKED ISSUE #1519: Decenza can’t find the HDS each morning
### What happened?

For the past several mornings that I’ve been using Decenza the app can’t find the HDS scale. There are no other apps using the scale. I usually spend 10 min troubleshooting it by restarting the tablet, the machine, loading a different app (Decent(rea)), closing that app, then restarting again. At some point it reconnects but I never know what fixed it. 

### Steps to reproduce

1. Schedule app to wake up every morning at 6am
2. put app to sleep
3. wait until morning
4. approach machine, look at machine and recognize that the HDS is not connected

### Decenza version

2.0 3484

### Platform

Android

### Device / OS version

Teclast M50Mini

### Scale (if applicable)

HDS

[decenza scale pairing bug 2026-07-15.pdf](https://github.com/user-attachments/files/30090252/decenza.scale.pairing.bug.2026-07-15.pdf)

_No response_

### Screenshots or video

_No response_

### Anything else?

I had Claude pull the debug log. If this isn’t helpful I can pull the log as you describe above.

skialpine:
Fixes #1519.

## The bug (from the raw debug log in the issue)

When the Decent Scale connects but never delivers weight data (degraded tablet BT stack after overnight idle), the watchdog correctly gives up after 10 retries and calls `m_transport->disconnectFromDevice()` to force a reconnect. But `QtScaleBleTransport::disconnectFromDevice()` severs all controller signals before teardown, so **`disconnected()` is never emitted** — despite the `ScaleBleTransport` interface contract saying it is. The connection-priority backoff path already knows this and compensates with a manual `emit disconnected()` (see the comment in `triggerScaleBackoff`); the watchdog path did not.

Consequences, all visible in the reporter's log:

- `setConnected(false)` never runs → the main.cpp auto-reconnect ladder (gated on `connectedChanged`) never arms. The log shows **zero reconnect activity for 12 minutes** after the 84 s watchdog exhaustion, until an unrelated DE1 fault path finally emitted the signal at 816 s.
- The app believes the scale is still connected → the Connections scan finds the scale but filters it out as already known (`items=0`) — the user stares at an empty scan list. That is the literal "Decenza can't find the HDS" symptom.
- Every DE1 wake calls `wake()`, which saw the stale `m_characteristicsReady == true` and **restarted the heartbeat/watchdog on the dead link** (log: `Watchdog started` at 470 s and 627 s with no preceding connect, right after `writeCharacteristic skipped - controller not connected`).

## The fix

- `onWatchdogFired()` exhaustion and `onServicesDiscoveryFinished()` service-not-found: run `onTransportDisconnected()` directly after the silent `disconnectFromDevice()` — the same compensation the backoff path uses. This drives `setConnected(false)` → `connectedChanged` → reconnect ladder arms and retries forever (5 s/30 s/60 s), reconnecting as soon as the stack recovers instead of parking until the user reboots everything.
- `onTransportDisconnected()`: clear `m_serviceFound`/`m_characteristicsReady` — the discovered characteristics don't outlive the link. This blocks zombie writes and keeps `wake()` from reviving a dead link. Both flags are re-armed by `connectToDevice()` on the next connect.

Idempotent by construction: if a transport does deliver a (late) `disconnected()` signal, the second `onTransportDisconnected()` is a no-op (`setConnected` guards on change).

Note: the underlying trigger — the tablet's BT stack degrading overnight (`RemoteHostClosedError`, connects with no notifications, DE1 write cascade) — is environmental; this fix makes the app recover from it automatically instead of amplifying it into a dead morning.

## Tests

- `watchdogExhaustionPropagatesDisconnect` — exhaustion must reach `isConnected() == false` and emit `connectedChanged` (fails on the old code).
- `wakeDoesNotReviveDeadLink` — `wake()` after a forced disconnect must not write to the transport or restart the heartbeat/watchdog (fails on the old code).
- Updated `watchdogDisconnectsAfterMaxRetries` for the new disconnect-handling warning.
- Full suite: 70 passed, 0 failed, 0 warnings (macOS Debug).

## Follow-up: pre-existing issues in the touched file (per CLAUDE.md)

Review agents surfaced three more real defects of the same zombie/supervision family in `decentscale.cpp`; fixed in this PR rather than deferred:

- **`onTransportError()` parked the app on "disconnected" over a live link.** It called `setConnected(false)` on *every* transport error — including transient per-operation failures on a healthy, streaming connection. The scan-based reconnect ladder can't recover that state (a connected peripheral doesn't advertise). Now: dead link (transport reports not connected) → full disconnect handling with propagation; live-link transient → connection kept, watchdog supervises the feed and force-disconnects (with propagation) if data actually stopped.
- **`connectToDevice()` over a live connection created a silent half-dead state**: it cleared `m_characteristicsReady` while leaving the watchdog/heartbeat running — the watchdog guard then stopped supervision mid-session while command writes dropped silently and weight kept flowing. Now stops stale supervision first.
- **`tickleWatchdog()` could resurrect a deliberately stopped watchdog** (a stray notification after disconnect cleanup or `sleep()`), which then exhausted against the silent feed and force-disconnected a sleeping scale. Now requires an active timer — `startWatchdog()` is the only arm point.
- `sendCommand()` drops are now logged instead of silent.

Additional tests: `transientErrorOnLiveLinkKeepsConnection`, `errorOnDeadLinkCleansUp`, `strayDataDoesNotRestartStoppedWatchdog`, `serviceNotFoundPropagatesDisconnect`, `reconnectAfterWatchdogExhaustion`. Suite: 45/45 in tst_scaleprotocol, 69/69 overall, 0 warnings.

## Review follow-up 2 (from the /review pass)

- `BLEManager::onScaleConnectionTimeout` now tears down a transport still holding a link when the scale isn't connected, regardless of how the connect was initiated — closes the setup-phase stall (discovery error before the watchdog is armed) on the background-reconnect path, where the held link kept the peripheral from advertising and made every retry scan come up empty. Gated on the transport actually holding something, so the perpetual 60s retry ladder doesn't churn teardowns while the scale is simply absent.
- Sharpened the `onTransportError` comment: `isConnected()` lags the disconnect callback and is not the dead-link detector — the watchdog is, with the 20s timeout bounding the pre-watchdog window.

### DIFF
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -1399,18 +1399,31 @@ void BLEManager::abortScaleDirectConnectIfPending(const QString& reason) {
 void BLEManager::onScaleConnectionTimeout() {
     m_scaleDirectAbortTimer->stop();
 
-    // If a foreground direct-connect is still parked at the overall timeout (the
-    // ~4s abort cleared the flag, so this only runs when that abort hasn't), tear
-    // its transport controller down before the WiFi/FlowScale fallback so it
-    // can't keep the radio held. Clear the flag FIRST so any signal emitted
-    // during teardown can't re-enter abortScaleDirectConnectIfPending.
-    const bool wasParked = m_directConnectInProgress
-            && !(m_scaleDevice && m_scaleDevice->isConnected());
+    // A transport still holding anything at the overall timeout is stuck and
+    // must be torn down before the retry/fallback below. Two shapes:
+    //  - a parked foreground direct-connect (controller held in Connecting,
+    //    #1303) — the ~4s abort usually clears this, so this only runs when
+    //    that abort hasn't;
+    //  - a connect (foreground or background) whose link came up but whose
+    //    setup stalled (e.g. a discovery error before the scale became ready —
+    //    the driver's watchdog isn't armed pre-ready, so nothing else recovers
+    //    it, and a connected peripheral doesn't advertise, so the retry scans
+    //    below can never find it while the link is held) (#1519).
+    // Clear the direct-connect flag FIRST so any signal emitted during
+    // teardown can't re-enter abortScaleDirectConnectIfPending.
+    const bool notConnected = !(m_scaleDevice && m_scaleDevice->isConnected());
+    const bool wasParked = m_directConnectInProgress && notConnected;
     m_directConnectInProgress = false;
     m_directConnectAddress.clear();
-    if (wasParked) {
-        appendScaleLog("Scale connection timeout — tearing down parked direct-connect controller");
-        if (auto* transport = m_scaleDevice ? m_scaleDevice->bleTransport() : nullptr) {
+    if (notConnected) {
+        auto* transport = m_scaleDevice ? m_scaleDevice->bleTransport() : nullptr;
+        // Gate on the transport actually holding something — the perpetual 60s
+        // retry ladder hits this timeout on every cycle while the scale is
+        // simply absent, and must not log/churn a teardown of nothing.
+        if (transport && (wasParked || transport->isConnected())) {
+            appendScaleLog(wasParked
+                ? QStringLiteral("Scale connection timeout — tearing down parked direct-connect controller")
+                : QStringLiteral("Scale connection timeout — tearing down stuck connection setup"));
             transport->disconnectFromDevice();
         }
     }
--- a/src/ble/scales/decentscale.cpp
+++ b/src/ble/scales/decentscale.cpp
@@ -64,6 +64,12 @@ void DecentScale::onTransportDisconnected() {
     DECENT_WARN("Transport disconnected");
     stopWatchdog();
     stopHeartbeat();
+    // The discovered characteristics don't outlive the link. Clearing these
+    // blocks writes to a dead transport and keeps wake() (DE1 wake path) from
+    // restarting the heartbeat/watchdog on a disconnected scale (#1519);
+    // connectToDevice() re-arms both on the next connect.
+    m_serviceFound = false;
+    m_characteristicsReady = false;
     if (m_checksumDisabled) {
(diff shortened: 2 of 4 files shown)


skialpine:
### Code review

---

<details>

- [Worth a look] A service/characteristic-discovery error during connection setup (link up, watchdog not yet armed) gets no in-driver recovery from the new `onTransportError()` guard — verified pre-existing (the old unconditional `setConnected(false)` was a change-guarded no-op in that window too, so this PR's delta is zero-or-positive), narrow reachability, and backstopped by the BLEManager 20s timeout / scale-side inactivity disconnect; a naive teardown-on-setup-error fix could regress the deliberate de1app double-send redundancy (score: 55)
  https://github.com/Kulitorum/Decenza/blob/47674b61b853930db077c4fde78cb4678daaa802/src/ble/scales/decentscale.cpp#L97-L111
- [OK to skip] `connectToDevice()` on an already-live scale could hit the transport's duplicate-connect debounce, leaving `m_characteristicsReady` false while connected — app-level guards (`main.cpp` scaleDiscovered / tryDirectConnectToScale both early-return on a connected scale) make it practically unreachable, and this PR's stop-timers change is itself the defensive improvement for that path (score: 50)
  https://github.com/Kulitorum/Decenza/blob/47674b61b853930db077c4fde78cb4678daaa802/src/ble/scales/decentscale.cpp#L45-L64
- [OK to skip] The `onTransportError()` comment can read as though `isConnected()` performs meaningful dead-vs-live triage; at error() time both transports' connected flags almost always still read true, so the watchdog is the real detector — comment clarity only, no functional impact (score: 50)
  https://github.com/Kulitorum/Decenza/blob/47674b61b853930db077c4fde78cb4678daaa802/src/ble/scales/decentscale.cpp#L99-L108

</details>

PR #1531, opened 2026-07-16
fix(advisor): flow taste-intake taps back to the review page immediately

skialpine:
## Problem

On Android/DE1: make a shot → tap the AI Advisor → tap taste/body + **Ask** → the AI replies. Go back to the shot review page (e.g. to update TDS) → tap AI Advisor again → it re-shows the **"how did this shot taste?"** intake page instead of resuming the conversation.

## Root cause

The advisor's per-shot taste-intake gate (`ConversationOverlay.openWithShot`) decides whether to show the intake by reading the taste fields off the **shot snapshot the caller hands it**. On `PostShotReviewPage` that snapshot is `editShotData` — the DB-load copy, which is **never refreshed on save** (the page deliberately skips reload to avoid clobbering in-progress edits).

So the taps were persisted to the DB but not mirrored into the snapshot → on reopen the gate read stale-empty taste → intake re-shown. (Shot Detail is immune: it reloads its snapshot on metadata save.)

## Fix

- `ConversationOverlay` emits `tasteIntakeSubmitted(tasteBalance, tasteBody, overall)` on **Ask**, after persisting the taps.
- `PostShotReviewPage`'s AI-Advice click now passes the **live** taste (`editTaste*`/`editEnjoyment`) to `openWithShot`, not the stale snapshot. This also fixes the sibling bug where rating via the page's own taste picker and *then* clicking AI Advice re-asked for taste.
- The page mirrors the emitted taps into its live edit fields via `applyEditState` (re-establishing the rating-slider/chip bindings) and advances the committed baseline instead of re-saving — so **selections in the advisor flow back to the review page immediately** (rating slider + taste chips update) and the next Ask sees the feedback.

## Testing

- Built in Qt Creator: succeeded, 0 errors, 0 warnings.

### DIFF
--- a/docs/CLAUDE_MD/AI_ADVISOR.md
+++ b/docs/CLAUDE_MD/AI_ADVISOR.md
@@ -29,6 +29,16 @@ backed-out-without-entering shot re-shows the intake. **Ask** persists the taps 
 `requestUpdateShotMetadata`, composes a first-person question, and sends it with
 the shot attached; **Skip** drops into the normal text conversation.
 
+The overlay reads only the host page's shot *snapshot* for the gate, and
+`PostShotReviewPage` deliberately does not reload that snapshot on metadata save
+(it would clobber in-progress edits). So on Ask the overlay also emits
+`tasteIntakeSubmitted(tasteBalance, tasteBody, overall)`; the review page mirrors
+those axes into its live `editTaste*` / `editEnjoyment` fields (via
+`applyEditState`, advancing the baseline rather than re-saving) so the rating
+slider and taste chips update at once and the next Ask sees the feedback — the
+review page's AI-Advice click passes the *live* taste, not the stale snapshot, to
+`openWithShot`.
+
 The taps are structured shot columns (`taste_balance`/`taste_body`, migration 33),
 not free text — so the same picker also appears on the post-shot review page (no
 parallel UI), the values flow into the advisor prompt and the dial-in history
--- a/docs/CLAUDE_MD/AI_ADVISOR.md
+++ b/docs/CLAUDE_MD/AI_ADVISOR.md
@@ -34,8 +34,12 @@ The overlay reads only the host page's shot *snapshot* for the gate, and
 (it would clobber in-progress edits). So on Ask the overlay also emits
 `tasteIntakeSubmitted(tasteBalance, tasteBody, overall)`; the review page mirrors
 those axes into its live `editTaste*` / `editEnjoyment` fields (via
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 1 issue:

1. `onTasteIntakeSubmitted` advances the undo baseline (`_committedState`) but not the dirty-check baseline (`editShotData`), so `hasUnsavedChanges` stays stuck `true` after taste flows back — even though the overlay already persisted the taps to the DB.

`hasUnsavedChanges` is computed purely from `editX !== editShotData.X` comparisons ([L474-L497](https://github.com/Kulitorum/Decenza/blob/e17ee0e936ca8d0152de6612c6f30284471ccda2/qml/pages/PostShotReviewPage.qml#L488-L497)); `_committedState` never appears in it. Because the intake gate only fires when the shot has no saved taste yet, `editShotData`'s taste fields are guaranteed blank when the signal arrives, so the live edit fields diverge from `editShotData` and the dirty flag never clears. The next lifecycle flush — `StackView.onDeactivating: autosave()`, i.e. simply backing out after chatting with the advisor — then re-runs `saveEditedShot()` (redundant DB write + Visualizer PATCH for data already synced), re-runs `runStickySync()`, and pushes a phantom no-op frame onto the undo stack (which, if it's the session's first commit, corrupts the "Undo repeatable down to the loaded record" guarantee for the taste/enjoyment fields).

The established sibling pattern for "already persisted elsewhere, just mirror it in" — `ChangeBeansDialog.onBagSelected` — advances **both** baselines via `clonePersistedShot` + field overrides. This handler does only half of it.

https://github.com/Kulitorum/Decenza/blob/e17ee0e936ca8d0152de6612c6f30284471ccda2/qml/pages/PostShotReviewPage.qml#L2550-L2559

Fix: mirror `editShotData` too (as `onBagSelected` does at [L2113-L2120](https://github.com/Kulitorum/Decenza/blob/e17ee0e936ca8d0152de6612c6f30284471ccda2/qml/pages/PostShotReviewPage.qml#L2104-L2122)), without setting `pendingVisualizerUpdate` (the overlay's `requestUpdateShotMetadata` already synced Visualizer):

```qml
var nb = clonePersistedShot(editShotData)
nb.tasteBalance = editTasteBalance
nb.tasteBody = editTasteBody
nb.enjoyment0to100 = editEnjoyment
editShotData = nb
_committedState = captureEditState()
```

The doc paragraph added to `AI_ADVISOR.md` speaks of "the baseline" as one thing — worth updating to note there are two (dirty-check vs. undo) once the fix lands.

PR #1558, opened 2026-07-19
Honour SecRandomCopyBytes on the iOS cert path

skialpine:
## What broke

The v2.0.0 tag push failed the iOS build:

```
src/network/shotserver.cpp:3109: error: ignoring return value of function
declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
    SecRandomCopyBytes(kSecRandomDefault, serialBytes.size(), ...)
** ARCHIVE FAILED **
```

`-Werror=unused-result` came from #1553, where it exists to make a dropped save verdict a compile error rather than a convention. This is the first thing it caught outside the translation code.

## It is a real bug, not flag noise

`serialBytes` is constructed `Qt::Uninitialized`. If the draw fails, the certificate's serial number is whatever was already on that heap block — plausibly the same value across devices, or across regenerations on one device, when RFC 5280 asks for a serial unique per issuer.

The fix fails cert generation, which the caller already handles, and releases `privateKey` on the way out like the three bailouts above it in the same function.

## Why CI caught this and I did not

The code sits inside `#ifdef Q_OS_IOS`. The macOS build never compiles it (macOS uses OpenSSL for this; only iOS uses Security.framework), and the Android CI build I ran specifically to validate the new flag does not compile it either. Verified here by an iOS CI build on this branch.

## Scope

One call site. I scanned for other discarded `warn_unused_result` results across `src/` and found none — the `read`/`write` hits are our own transport methods, not POSIX.

The other five platforms passed on the same tag, since this code is unreachable off iOS. Windows passing also confirms the MSVC half of the flag (`/we4834`).

## After merge

v2.0.0 needs re-tagging to pick this up: `git tag -f v2.0.0 HEAD && git push origin v2.0.0 --force`, then confirm the release is still `--draft=false --prerelease`.

### DIFF
--- a/src/network/shotserver.cpp
+++ b/src/network/shotserver.cpp
@@ -3104,10 +3104,20 @@ bool ShotServer::generateSelfSignedCert(const QString& certPath, const QString& 
         derSet(derSequence(derOid(OID_CN, 3) + derUtf8String("Decenza")))
     );
 
-    // Serial number (16 random bytes)
+    // Serial number (16 random bytes).
+    //
+    // The verdict must be honoured: the buffer is deliberately Qt::Uninitialized, so on a failed
+    // draw the serial would be whatever heap bytes were already there — potentially identical
+    // across devices or across regenerations, when RFC 5280 requires it to be unique per issuer.
+    // Better to fail cert generation, which the caller already handles, than to mint a
+    // certificate whose identity is quietly garbage.
     QByteArray serialBytes(16, Qt::Uninitialized);
-    SecRandomCopyBytes(kSecRandomDefault, serialBytes.size(),
-                        reinterpret_cast<uint8_t*>(serialBytes.data()));
+    if (SecRandomCopyBytes(kSecRandomDefault, serialBytes.size(),
+                           reinterpret_cast<uint8_t*>(serialBytes.data())) != errSecSuccess) {
+        CFRelease(privateKey);
+        qWarning() << "ShotServer: could not draw random bytes for the certificate serial number";
+        return false;
+    }
     // Ensure positive
     serialBytes[0] = serialBytes[0] & 0x7F;
 
--- a/src/network/shotserver.cpp
+++ b/src/network/shotserver.cpp
@@ -3106,11 +3106,11 @@ bool ShotServer::generateSelfSignedCert(const QString& certPath, const QString& 
 
     // Serial number (16 random bytes).
     //
-    // The verdict must be honoured: the buffer is deliberately Qt::Uninitialized, so on a failed
-    // draw the serial would be whatever heap bytes were already there — potentially identical
-    // across devices or across regenerations, when RFC 5280 requires it to be unique per issuer.
-    // Better to fail cert generation, which the caller already handles, than to mint a
-    // certificate whose identity is quietly garbage.
+    // The buffer is deliberately Qt::Uninitialized, so on a failed draw the serial would be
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 4 issues. All four were the same class as the fix under review — an unchecked result on a path that then reports success — and all four are now fixed in d34aa84a.

1. `RAND_bytes` unchecked on the OpenSSL branch: the identical bug to the one this PR fixes, on the branch that Windows, macOS, Linux, Linux ARM64 and Android all build. `serial_bytes` is a bare stack array, so a failed draw stamps stack residue into the certificate serial. Worth stating plainly: `-Werror=unused-result` did **not** catch this, because OpenSSL does not mark `RAND_bytes` `warn_unused_result` the way Apple marks `SecRandomCopyBytes`. The flag caught the one-platform instance and was silent on the five-platform one.

https://github.com/Kulitorum/Decenza/blob/d34aa84a85c4fb4aeb2b13e4a7c694c76176f7e6/src/network/shotserver.cpp#L3265-L3269

2. `PEM_write_X509` and `PEM_write_PrivateKey` unchecked, with `success = true` running regardless — a disk that fills mid-write produced a truncated cert or key and a reported success. `fclose` is now checked too, and before the verdict, since buffered output is flushed there and an early `break` would leak the `FILE*`.

https://github.com/Kulitorum/Decenza/blob/d34aa84a85c4fb4aeb2b13e4a7c694c76176f7e6/src/network/shotserver.cpp#L3316-L3323

3. `QFile::write` unchecked before an unconditional `return true` on the iOS branch — same shape, inside the exact function this PR edits. (CLAUDE.md: "Always fix pre-existing issues in touched files — never dismiss as 'pre-existing'".)

https://github.com/Kulitorum/Decenza/blob/d34aa84a85c4fb4aeb2b13e4a7c694c76176f7e6/src/network/shotserver.cpp#L3198-L3206

4. `QFile::setPermissions` unchecked on both branches, so the private key could silently keep default (possibly group-readable) permissions. This one warns rather than fails: the key still works, and refusing to serve over a permissions quirk is the worse outcome. Moot inside the iOS sandbox, real on a shared desktop account — flagging the judgement call explicitly in case you disagree.

https://github.com/Kulitorum/Decenza/blob/d34aa84a85c4fb4aeb2b13e4a7c694c76176f7e6/src/network/shotserver.cpp#L3219-L3223

Verified clean on the original change: `CFRelease(privateKey)` is the complete cleanup at the new early return (`publicKey`, `pubKeyData`, `privKeyData` are already released; `error` is null; no double-free against the success path), `errSecSuccess` is the right sentinel, and the caller chain fails closed — `setupTls()` and `start()` both propagate, so the server declines to start rather than serving without TLS.

<details>

- [OK to skip] `serialBytes[0] & 0x7F` can leave a leading zero byte ~1 in 256 times, making the DER INTEGER non-minimal. Pre-existing, strict-DER pedantry, no TLS client rejects a self-signed cert over it. (score: 50)
- [OK to skip] `derUtcTime` uses ASN.1 UTCTime, which breaks for validity dates from 2050 (RFC 5280 4.1.2.5 requires GeneralizedTime). Flagged on the original PR #437 and still open, but validity is now+10y = 2036, so it is not reachable yet. (score: 55)

</details>

Also dropped "The verdict must be honoured:" from the comment added in 3f35b719 — reviewer-facing narration rather than a durable constraint.

PR #1568, opened 2026-07-19
Add Maintenance card: relocate Descale wizard + new Transport Mode

### LINKED ISSUE #1152: [Feature] Transport mode in Decenza

When the machine will be turned off for several days or weeks, for example during vacation or storage, it would be useful to fully drain the water from the machine.

As far as I can see, Decenza currently does not have a “Transport Mode” or drain function. This means water may remain inside the machine even when it is not going to be used for a longer period.

### Proposed solution

Please add a Transport Mode function, similar to the original Decent software.

The function should guide the user through draining all water from the machine, so it can be safely shut down, stored, or transported.

Expected behavior:

The user starts Transport Mode from the app.

The machine runs the required steps to empty the internal water system.

The app shows clear instructions during the process.

The app confirms when the machine is fully drained and ready to be powered off.

### Alternatives considered

_No response_

### Additional context

_No response_

skialpine:
## Summary

Adds a **Maintenance** card to Settings → Machine (directly under Shot Map) as the single home for machine-upkeep operations, starting with two actions:

- 🧽 **Descaling Wizard** — relocated here from an obscure button on the Profiles page (which appeared only in the Cleaning/Descale filter view; now removed). The placeholder `descale_wizard.json` *fake profile* that existed solely to surface the wizard in the profile list is deleted, along with its two `.qrc` registrations and the `beverageType === "descale"` special-case tap handler. Real cleaning profiles (Forward Flush, Weber Spring Clean, …) are unaffected.
- 🧳 **Transport Mode** (new — closes #1152) — a guided page (prepare → draining → "safe to power off") that drains the machine's internal water via the firmware **AirPurge** state (`0x14`) so the machine can be safely powered off, stored, or shipped (e.g. before a vacation or freezing transport).

## What's wired

- **Machine layer:** new `MachineState::Phase::Transport` (mapped from `AirPurge`), `DE1Device::startAirPurge()`, and simulator support (drains ~12s → Idle) so the flow runs without hardware.
- **Auto-sleep:** `Phase.Transport` added to `operationActive` so a multi-minute drain can't be cut off by the inactivity timer — matching Descaling/Cleaning.
- **Consistency:** `Phase::Transport` also added to the `MachineStatusItem` status widget, the `machine_stop` MCP guard, the idle-GC `enteringOp` list, the `uploadCurrentProfile` active-phase guard, the phase auto-nav branch, and the TalkBack announcement.
- **Ready gate:** on current firmware (1333/1352) a GHC machine's firmware silently drops an AirPurge request while still heating, so Start is enabled only once the machine reaches ready temperature (bypassed in simulation).

## Deferred (separate on-hold change)

Reliable **cold-machine** maintenance is captured in the sibling `enable-cold-machine-maintenance` change (also in this PR as planning artifacts, marked **on-hold**). It's blocked on a firmware release that honors cold maintenance natively — reaprime/de1app's pre-profile workaround is documented there as the rejected-for-now fallback. No shipped firmware fixes this today (newest bundled stable is 1352), so the ready-gate is the mitigation for now.

## Review

Ran `/pr-review-toolkit:review-pr` (code / comments / silent-failure). Fixes applied:
- STOP mid-drain no longer shows a false "machine is empty — safe to power off" confirmation (could have led to shipping a full machine); completion now latches on any non-Transport landing phase, not just Idle/Ready.
- Stale `ProfileSelectorPage` comment corrected; checkmark switched to `Theme.emojiToImage`.

**Build:** clean, 0 errors / 0 warnings (Qt Creator, Qt 6.11.1).

## Still to do (post-review)
- Runtime verify in the app (drive both flows).
- **Hardware verify** on a real GHC machine (fw 1333/1352): confirm AirPurge self-terminates and the physical GHC stop exits it. If it can hang with no on-screen stop, show the STOP button on GHC too (currently `isHeadless`-gated, matching Descale).
- Wiki manual Maintenance section (held per release convention).


skialpine:
### Code review

Found 3 significant issues (plus 2 minor). All were addressed in f198c71b on this branch.

1. Completion logic reported a false "Machine is empty — safe to power off" on any exit from the Transport phase that wasn't the in-app STOP button. A **BLE disconnect** (`updatePhase()` forces `Phase::Disconnected`) and a **physical GHC stop** (lands in Idle, indistinguishable from a natural end) both fell through to `showComplete = true` — the exact freeze-damage outcome the feature exists to prevent. `machinestate.cpp` warns twice that "a dropped connection is not a completion." Fixed by: adding `transportPage` to the disconnect-navigation whitelist, whitelisting only settled landings (Idle/Ready/Heating) so Disconnected/Sleep are excluded, and rewording the completion screen to conditional guidance instead of asserting the tank is empty.

https://github.com/Kulitorum/Decenza/blob/f198c71b91e4555275d29a43a9f0a4eb70bc48c6/qml/pages/TransportPage.qml#L49-L73

https://github.com/Kulitorum/Decenza/blob/f198c71b91e4555275d29a43a9f0a4eb70bc48c6/qml/main.qml#L3239-L3240

2. Home-screen widget phase labels weren't updated for the new `Transport` phase. `WIDGET_SNAPSHOT.md` mandates a three-place update (doc table + iOS `WidgetPhase.labels` + Android `PHASE_LABELS`) on any `Phase` change; all three were missing it, so the widget would show the raw untranslated string during a drain. Fixed in all three.

https://github.com/Kulitorum/Decenza/blob/f198c71b91e4555275d29a43a9f0a4eb70bc48c6/ios/widget/MachineStatusSnapshot.swift#L30-L33

3. The data-driven `phaseMapping` test has a row per state→phase mapping but was missing the new `AirPurge → Transport` case. Added.

https://github.com/Kulitorum/Decenza/blob/f198c71b91e4555275d29a43a9f0a4eb70bc48c6/tests/tst_machinestate.cpp#L98-L103

---

<details>
<summary>2 lower-priority issues (50-74) — also fixed</summary>

- [Worth a look] A page created while a drain is already running (auto-nav / reopen mid-drain) never latched `wasPurging`, so the confirmation was skipped — fixed with `Component.onCompleted` (score: 55)
- [Worth a look] `isFlowing()`/`wasFlowing` deliberately omit `Transport` but had no comment recording why — added one so a future purge-timer readout isn't a silent 0 (score: 50)

</details>

Build clean, `tst_machinestate` passing (47/0). CLAUDE.md-compliance, deep-bug, history, recent-PR, and comment reviews all otherwise clean.

PR #1574, opened 2026-07-20
Put Purge and Stop on one row on the live steam view

skialpine:
## The problem

The live steaming view overflowed the bottom of the screen and the STOP button was cut in half.

The page is a `ColumnLayout` with `anchors.fill: parent`, no `Flickable` and no clipping. Once its children's minimum heights exceed the parent, Qt lays out at minimum and the rest runs off the bottom. STOP is the last item, so STOP is what got lost — and on a headless machine that is the only in-app way to stop steaming.

Not a window-size bug: `Theme.scale = min(width/960, height/600)` scales everything uniformly, so the ratio of content to available height never changes. The page was simply over budget against its own 960x600 reference — roughly 536 units required against 440 available (600 less two 80-unit page margins). The two `Item { Layout.fillHeight: true }` spacers were already fully collapsed.

## The fix

Purge and Stop now share one row instead of stacking, reclaiming 56 units (a 48-unit button plus 8 of spacing).

The row sits where STOP already lived — a sibling of **both** the timer and chart views — rather than pulling STOP into the timer view, since gating it behind `steamViewMode === "timer"` would have removed the stop control from the chart view entirely.

**Behaviour change:** Purge is now reachable from the chart view, where it was previously timer-view only. This falls out of the row's placement. Documented in the wiki manual.

## What review turned up

Three rounds of review found problems worth recording, since none are visible in the final diff.

**A latch that would have shipped.** The row's first visibility guard read `livePurgeButton.visible || steamStopButton.visible`. A `.visible` read returns *effective* visibility, and `visibleChanged` does not fire when only the explicit value flips — so once the row hid itself, both children read `false` forever and it never came back, taking STOP with it. It only bites when the page instance predates the steam session, so a quick look while already steaming shows nothing wrong. Confirmed against the Qt 6.11.1 `qml` runtime with a reduced case, then fixed to mirror the children's own conditions.

**The keyboard chain was broken in four places.** Adding a focusable control left the loop open on a GHC machine; `decreaseTimeBtn` backtabbed to the flow slider, which sits *after* it, so the chain doubled back and never closed; the -5s/+5s pair was unreachable going forward; and several targets pointed at pills that hide when a preset is "Off" — `KeyNavigation` silently refuses an invisible target, so Tab did nothing at all. All targets now route through visibility-aware helpers.

**A German mistranslation, introduced and then removed here.** Mid-branch the Purge label was moved onto `steam.button.purge` on the reasoning that keys should unify on whatever the shipped translations were made from. Sound principle, but the data was never checked: that key labelled the Stop button's second-press state back in December, when the state meant "stop", so its community translations say exactly that — de "Stopp", da "Stop", ar "إيقاف". A German user would have seen two adjacent buttons both reading "Stopp". The correct translations were already under `steam.label.purge` (de "Reinigen", fr "Purger", da "Rens", ar "تنظيف"), which both purge labels now use; the poisoned key carries a comment so it is not reached for again.

`steam.button.stop` keeps its own key — its translations are correct and it matches the five sibling operation pages. It was hardcoded English before this branch, so translating it is a real gain.

**Accessibility.** Both Purge buttons moved from a raw `MouseArea` to `AccessibleTapHandler`, so under a screen reader an exploratory first touch announces instead of firing the wand purge. All four handler-owning parents on the page now set `Accessible.ignored: true`, the pattern ACCESSIBILITY.md documents and `viewToggleBtn` already used — this also fixes `steamStopButton` and `addPitcherButton`, which had the same gap. The settings-view Purge button was in no tab chain at all and is now wired between the add-preset button and the duration slider.

## Rendered labels

| | Purge | Stop |
|---|---|---|
| en | PURGE | STOP |
| de | REINIGEN | STOPP |
| fr | PURGER | ARRÊTER |
| da | RENS | STOP |
| ar | تنظيف | إيقاف |

Casing is applied with `font.capitalization` rather than baked into the source strings, so one sentence-case purge key can serve both this button and the settings-view one.

## Testing

Full suite 85/85, no warnings. `check_translation_key_conflicts.py` clean across 3271 keys. `qmllint` warnings net down across the branch with no new types. Layout, and the German labels, confirmed in the running app.

Verified by reading rather than on device: the keyboard tab order, and whether VoiceOver announces the four buttons whose accessibility semantics changed.

## Follow-up, deliberately not in this PR

The page is back inside the frame but not with much margin. `anchors.bottomMargin: Theme.pageTopMargin` reserves 80 units "for bottom bar", far more than the bar needs; halving it buys another 40. A user who raises `timerSize` (adjustable to 120) can still push the page out.

## Wiki

Manual and FAQ updated in `Decenza.wiki@362778d`: a new "Live Steaming Screen" section, plus four references to a setting that no longer exists — the docs told users to enable "Headless — Single press to stop & purge" to get behaviour that is now the default, under a name they cannot find.

### DIFF
--- a/qml/pages/SteamPage.qml
+++ b/qml/pages/SteamPage.qml
@@ -1061,15 +1061,12 @@ Page {
 
                 Tr {
                     anchors.centerIn: parent
-                    key: "steam.label.purge"; fallback: "Purge"
+                    // Same key and same type treatment as steamStopButton beside it, so
+                    // the pair reads as one row.
+                    key: "steam.button.purge"; fallback: "PURGE"
                     color: Theme.primaryContrastColor
-                    // Matches steamStopButton: same box, same casing, so the pair reads
-                    // as one row. Capitalised at render time for the same reason it is
-                    // on that button — the key stays sentence-case and shared with the
-                    // settings-view Purge, and case-less scripts are left alone.
                     font.pixelSize: Theme.scaled(24)
                     font.weight: Font.Bold
-                    font.capitalization: Font.AllUppercase
                     Accessible.ignored: true
                 }
 
@@ -1119,21 +1116,16 @@ Page {
                 Text {
                     id: stopButtonText
                     anchors.centerIn: parent
-                    // Reuses steam.label.purge rather than minting a second purge key:
-                    // the registry maps key -> one English string, so the same key with a
-                    // different fallback would oscillate between whichever site rendered
-                    // last (see scripts/check_translation_key_conflicts.py).
+                    // steam.button.* carries the shouted label, matching espresso/flush/
+                    // hotwater/descaling/transport. The sentence-case steam.label.purge
+                    // stays on the settings-view Purge button — one key holds one English
+                    // string, so a shared key could not serve both casings.
                     text: (steamSoftStopped && Settings.hardware.steamTwoTapStop)
-                          ? TranslationManager.translate("steam.label.purge", "Purge")
-                          : TranslationManager.translate("steam.label.stop", "Stop")
+                          ? TranslationManager.translate("steam.button.purge", "PURGE")
+                          : TranslationManager.translate("steam.button.stop", "STOP")
                     color: Theme.primaryContrastColor
                     font.pixelSize: Theme.scaled(24)
                     font.weight: Font.Bold
-                    // The all-caps look is applied at render time, not baked into the
-                    // source strings. Translators supply natural case, scripts without a
-                    // case distinction are left alone, and the key stays shareable with
-                    // the sentence-case Purge button in the same row.
-                    font.capitalization: Font.AllUppercase
                     Accessible.ignored: true
                 }
 
--- a/qml/pages/SteamPage.qml
+++ b/qml/pages/SteamPage.qml
@@ -1016,6 +1016,7 @@ Page {
             // at the shorter end of the supported aspect range. The row sits outside
             // the timer/chart view switch, so Purge is available in both views.
             RowLayout {
+                id: liveActionRow
                 Layout.alignment: Qt.AlignHCenter
                 // Mirrors the two children's own conditions rather than reading their
                 // .visible. A `visible` read returns EFFECTIVE visibility, so once this
@@ -1025,13 +1026,19 @@ Page {
                 visible: isSteaming || DE1Device.isHeadless
                 spacing: Theme.spacingMedium
 
+                // One source of truth for the button box, so the two can never drift
+                // apart again. They are peer actions in a shared row and read as
+                // mismatched the moment their boxes differ.
+                readonly property int buttonWidth: Theme.scaled(200)
+                readonly property int buttonHeight: Theme.scaled(60)
+
             // Purge button on the live steaming view — stops steam and triggers
             // the DE1 steam-wand purge.
             Rectangle {
                 id: livePurgeButton
                 visible: isSteaming
-                Layout.preferredWidth: Theme.scaled(150)
-                Layout.preferredHeight: Theme.scaled(60)  // matches steamStopButton so the row aligns
+                Layout.preferredWidth: liveActionRow.buttonWidth
+                Layout.preferredHeight: liveActionRow.buttonHeight
                 radius: Theme.buttonRadius
                 color: livePurgeMa.pressed ? Qt.darker(Theme.primaryColor, 1.2) : Theme.primaryColor
 
@@ -1055,7 +1062,14 @@ Page {
                 Tr {
                     anchors.centerIn: parent
                     key: "steam.label.purge"; fallback: "Purge"
-                    color: Theme.primaryContrastColor; font: Theme.bodyFont
+                    color: Theme.primaryContrastColor
+                    // Matches steamStopButton: same box, same casing, so the pair reads
+                    // as one row. Capitalised at render time for the same reason it is
+                    // on that button — the key stays sentence-case and shared with the
+                    // settings-view Purge, and case-less scripts are left alone.
+                    font.pixelSize: Theme.scaled(24)
+                    font.weight: Font.Bold
+                    font.capitalization: Font.AllUppercase
                     Accessible.ignored: true
                 }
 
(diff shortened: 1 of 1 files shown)


skialpine:
## Automated review round — one real bug found and fixed

Ran `code-reviewer` and `comment-analyzer` over the diff. Pushed d21d28e5 in response.

### Critical: the row guard was a one-way latch (fixed)

`visible: livePurgeButton.visible || steamStopButton.visible` reads **effective** visibility, not the explicitly-assigned value. A child of a hidden parent reports `visible === false` regardless of its own binding, and `visibleChanged` does not fire when the explicit value flips but the effective one does not.

So: page exists while not steaming → row hides itself → both children now read `false` forever → binding can never re-arm → **the row never comes back**, taking the STOP button this PR exists to un-clip with it.

Confirmed against the Qt 6.11.1 `qml` runtime with a reduced case rather than reasoned about:

```
--- at construction (not steaming) ---
actionRow.visible = false
--- after steaming starts ---
liveColumn.visible = true
purge explicit cond = true | purge.visible reads = false
actionRow.visible   = false   <-- latched off
```

The guard now mirrors the children's own conditions (`isSteaming || DE1Device.isHeadless`), which is latch-free. Worth noting this would most likely have shipped: it only bites when the page instance predates the steam session, so a quick look while already steaming shows nothing wrong.

### Also fixed

- **Focus loop was left open.** The preset pills and view-toggle branched only on `steamStopButton`, so on a GHC machine (Stop hidden) Purge was unreachable except from the flow slider — which is itself timer-view only, making Purge mouse-only in chart view with a backtab pointing into a hidden subtree. Both entry points now route through `firstLiveActionButton()` / `lastLiveActionButton()`.
- **Pre-existing focus trap on STOP:** `Keys.onTabPressed` accepted the event unconditionally, so with no preset pills focus went nowhere.
- **Comment fixes:** the rationale now cites the absent `Flickable` (checkable in ten seconds) instead of a "600-unit budget" (not a real ceiling); dropped transitional now/used-to-be phrasing; dropped "to the left", which inverts under RTL mirroring.

### Finding I did not accept

The comment reviewer argued the un-indented `RowLayout` body should be re-indented and the `} // end` marker dropped, on the grounds that the comparable block at line 941 has "a correctly indented body". It does not — that block opens at indent 12 with its children also at indent 12 and closes with `} // end timer view ColumnLayout`. The new block matches the file's existing convention exactly.

### Still open — not addressed here

- `"PURGE"` / `"STOP"` (`SteamPage.qml:1109`) are hardcoded English while the `Accessible.name` beside them is translated. Pre-existing; needs new translation keys, so it wants its own change.
- Purge uses a raw `MouseArea` with `Accessible.*` on the Rectangle; STOP uses `AccessibleTapHandler`. Both compliant, but two adjacent buttons in one row now announce differently under TalkBack.
- Behavioural asymmetry, pre-existing and worth a second opinion: STOP calls `requestIdle()` **and** `root.goToIdle()`, Purge calls only `requestIdle()` and stays on the page. Intended?

skialpine:
### Code review

Found 3 issues:

1. The settings-view Purge button is not in the keyboard tab chain. It has `activeFocusOnTab: true` but no `KeyNavigation.tab`/`backtab`, and nothing in the file targets it — `durationSlider.KeyNavigation.backtab` routes past it to `addPitcherButton`. The gap predates this PR, but this PR modifies exactly this Rectangle for an accessibility change (adds the id, converts MouseArea to AccessibleTapHandler, rewrites the Keys handlers). [ACCESSIBILITY.md](https://github.com/Kulitorum/Decenza/blob/c26d3fb6b94408a00578c5c32729256496ecdabd/docs/CLAUDE_MD/ACCESSIBILITY.md) Rule 3 says "Wire `KeyNavigation.tab` and `KeyNavigation.backtab` on all interactive controls in logical reading order, forming a **closed loop**", and CLAUDE.md says "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'."

https://github.com/Kulitorum/Decenza/blob/c26d3fb6b94408a00578c5c32729256496ecdabd/qml/pages/SteamPage.qml#L1553-L1562

2. Both converted Purge buttons lost their accessible name from the node that actually receives keyboard focus. The conversion moved `Accessible.role`/`name`/`focusable`/`onPressAction` off each Rectangle onto the handler via `accessibleItem`, but `accessibleItem` only drives the announce-first state tracking — it does not cascade the name back to the parent. Each Rectangle keeps `activeFocusOnTab: true` while now carrying no `Accessible` attached property, so keyboard focus lands on an unnamed, roleless node where before the diff it announced "Purge the steam wand, button". Note `steamStopButton` and `addPitcherButton` have the same shape already, so this matches local convention — but for these two specific controls it is still a step down from the pre-diff state. Affects `livePurgeButton` (linked) and `settingsPurgeButton` at L1553-L1562. Worth confirming on-device before acting, since the practical impact depends on Qt's fallback behaviour.

https://github.com/Kulitorum/Decenza/blob/c26d3fb6b94408a00578c5c32729256496ecdabd/qml/pages/SteamPage.qml#L1037-L1048

3. The -5s / +5s buttons are unreachable by forward Tab. Nothing tabs into `decreaseTimeBtn` (L775) — `increaseTimeBtn.KeyNavigation.tab` goes to the flow slider (L853), and `viewToggleBtn.Keys.onTabPressed` skips the whole timer-view Column. They are reachable only by Shift+Tab from the slider, so the loop is asymmetric. Pre-existing and the button definitions themselves are not in this diff, but this PR reworked the surrounding focus chain wholesale (the two new helper functions, the preset-pill and view-toggle handlers), which is what makes the omission visible. Same closed-loop rule as issue 1.

https://github.com/Kulitorum/Decenza/blob/c26d3fb6b94408a00578c5c32729256496ecdabd/qml/pages/SteamPage.qml#L936-L939

---

<details>

- [Worth a look] `livePurgeButton.KeyNavigation.tab/backtab` target `livePresetRepeater.itemAt(0)` / `itemAt(count-1)` directly, without the visibility check the rest of the PR uses. An "Off" preset is hidden while steaming, and Qt will not focus an invisible target, so Tab silently does nothing when the first/last pill is "Off". One-line fix: route through the `firstLiveActionButton()`/`lastLiveActionButton()` helpers this PR already introduced. Not reachable with the shipped Small/Large defaults. (score: 72)
- [Worth a look] The two announce-first comments assert the protection unconditionally, but `AccessibleTapHandler` only announces-then-activates when `AccessibilityManager.enabled`; in normal mode it fires on release exactly like a raw MouseArea. Easy comment fix in lines this PR is already writing. (score: 65)
- [OK to skip] Purge and Stop still differ in `radius` (`Theme.buttonRadius` 12 vs `Theme.cardRadius` 16) and border, despite the new shared-property comment claiming they "can never drift apart again". The difference reads as deliberate hierarchy for a stop control; the overstated comment is the only real defect. (score: 65)

</details>

The Purge-stays-on-page vs Stop-navigates-to-idle asymmetry was raised and reviewed, and is intentional. Not listed above.

PR #1582, opened 2026-07-21
Idle-page polish, picker a11y, and block duplicate active names

skialpine:
Two related threads from the UI audit: finishing the idle-page polish, and stopping duplicate names on active items.

## Idle-page polish

**Brew bar no longer paints over the bottom nav.** The lower-mid band is anchored to `bottomBar.top`, so the previous "slide DOWN out of the carousel's way" had nowhere to go and drove the band straight over the action bar. It now **fades out** when the center carousel expands into it, and the hidden band is `enabled: false` so it can't swallow taps meant for the bar underneath.

**Picker accessibility.** The last four idle pickers (Espresso/Steam/Flush/HotWater) move from `Popup` to `Dialog` with null header/footer, matching Beans/Equipment/Recipes, so TalkBack/VoiceOver trap focus in the pill list. `Dialog` subclasses `Popup` and `modal`/`dim` were already set, so **appearance is unchanged** — the only added effect is focus-trapping, which is inert without a screen reader.

**MQTT.** "Publish Discovery Now" now sets `Layout.fillWidth` like every other child of its column instead of hugging its content.

## Block duplicate active names

An equipment package or recipe can no longer be saved under a name another **active** item already holds — two same-named items are indistinguishable in the pills, switcher lists and readout bars. Save is blocked with an inline message; nothing is auto-renamed or label-decorated.

- **Scope:** equipment = in-inventory packages; recipe = non-archived. Removing/archiving frees a name.
- **Comparison** is trimmed + case-insensitive and excludes the item being edited, so re-saving under your own name (or a casing variant) works.
- **Enforced twice:** a client-side gate in `SwitchEquipmentDialog` / `RecipeWizardPage` for immediate feedback, and a storage guard so MCP and ShotServer can't slip duplicates past it. Storage rejects; it never auto-renames.

### The app, MCP and ShotServer behave identically

| Surface | Create | Rename |
|---|---|---|
| App form | blocked, inline message | blocked, inline message |
| MCP | specific name-conflict error | specific name-conflict error |
| ShotServer | 409 + message | 409 + message |

Renames needed a reason channel, so `packageUpdateFailed` / `recipeUpdateFailed` were added **alongside** the existing terminal signals — purely additive, so no existing caller or test stub changed.

### Deliberately out of scope

- **Bean bags.** A bag's identity is roaster + coffee rather than a user-typed name, and the existing behaviour is correct. All bag work was reverted; the artifacts record why.
- **Import paths.** They restore existing user data rather than accept a newly typed name.
- **No migration.** Pre-existing duplicates are left exactly as they are. Note this also means nameless auto-created packages (which display the grinder) are unaffected — the rule prevents new duplicate names, it does not retroactively separate ones already there.

## Testing

- `tst_Equipment::activeNameUniqueness` and `tst_RecipeStorage::activeNameUniqueness` cover the match matrix (exact / case / whitespace / blank), self-exclusion, and that removing-from-inventory / archiving frees a name and unarchiving re-collides.
- Full suite green via Qt Creator: **85 binaries, 0 failed, 0 skipped, no warnings.**
- Both forms smoke-tested in the running app; brew-bar fade and picker toggle behaviour verified.

## Docs

Wiki manual updated and pushed (`Decenza.wiki` c294d8e) — Equipment Packages and Creating a Recipe, including how this differs from the existing identical-*gear* rule.

OpenSpec change `block-duplicate-active-names` is archived in the final commit, promoting the new `active-name-uniqueness` capability.


skialpine:
### Code review

Found 6 issues. All have been fixed in 5d22b4df — findings are recorded below against the commit they were found in (39a079fe).

**1. `requestInventory()` queued ahead of each entry path's own load, reintroducing a regression a prior commit deliberately fixed**

`runAsync` dispatches to a single FIFO `SerialDbWorker`, so a full non-archived-recipe scan delayed the very recipe/shot the page was opened to show. Commit `e3b7e43` had gated this call to the blank-create walk for exactly that reason. The adjacent comment also claimed the opposite of what the code did ("queued after each path's own load below", while sitting above it).

https://github.com/Kulitorum/Decenza/blob/39a079feabe6f64097dce3188453ec3f182eaa04/qml/pages/RecipeWizardPage.qml#L466-L470

**2. Shipped behaviour contradicts the live `idle-page-panel-clearance` spec**

The spec promoted one commit earlier requires that content below a center-zone panel "SHALL slide down", and that room is made "by the transient content offset alone". This PR replaced the slide with a fade to `opacity: 0` and did not update the spec. The behaviour is correct (the band is anchored to `bottomBar.top`, so "down" only ever meant "onto the bottom bar") — the spec was stale, and has been amended to cover content that has nowhere to yield.

https://github.com/Kulitorum/Decenza/blob/39a079feabe6f64097dce3188453ec3f182eaa04/openspec/specs/idle-page-panel-clearance/spec.md#L42-L44

**3. `_existingRecipeNames` documented as cosmetic when it now gates Save**

Comment said "used only to disambiguate an auto-name collision". `nameInUse` also reads it to disable the Save button, which changes whether it matters that the list is briefly empty.

https://github.com/Kulitorum/Decenza/blob/39a079feabe6f64097dce3188453ec3f182eaa04/qml/pages/RecipeWizardPage.qml#L269-L272

**4. `onRecipeUpdateFailed` not gated on `_submitting`**

Unlike the terminal handler beside it. Another surface failing a rename of the same recipe while the wizard sat open would latch `"nameInUse"`, so a later save failing for an unrelated cause reported the wrong message.

https://github.com/Kulitorum/Decenza/blob/39a079feabe6f64097dce3188453ec3f182eaa04/qml/pages/RecipeWizardPage.qml#L1748-L1753

**5. Pre-existing invariant comment left contradicting the new rule**

`createPackageWithGrinderStatic` asserts "two packages may share a display name" beside the name derivation, while this PR forbids exactly that. Both are true at different levels; the comment now states which — derived names are deliberately not unique and this function carries no guard, uniqueness is enforced one level up and only for a user-entered name, and migration/import are exempt.

https://github.com/Kulitorum/Decenza/blob/39a079feabe6f64097dce3188453ec3f182eaa04/src/history/equipmentstorage.cpp#L610-L613

**6. Signal doc comments overstate cross-surface parity**

Claimed the app, MCP and ShotServer "describe a rename collision identically". They share the cause code and emission ordering; the user-facing wording differs per surface, as it should.

https://github.com/Kulitorum/Decenza/blob/39a079feabe6f64097dce3188453ec3f182eaa04/src/history/equipmentstorage.h#L263-L267

---

Checked and clean: CLAUDE.md compliance (Theme, i18n, accessibility, DB-off-main-thread, MCP conventions); no critical bugs in the name-guard logic, transaction/rollback, guard-before-copy-on-write, terminal-signal guarantees, or the IdlePage binding; no prior-PR feedback repeated. Full suite green (85 binaries, 0 failed, no warnings); specs validate 106/106.

PR #1585, opened 2026-07-21
Last shot's chart as an app background

skialpine:
## What

Two more entries in the background chooser: **Last Shot** and **Last Shot (Advanced)**, drawing the most recent completed shot's chart behind every page.

- **Per-line visibility is inherited** from the review page's existing `graph/show*` toggles — including live: `HistoryShotGraph` now re-reads them on change (its eleven `showX` properties were one-shot `Q_INVOKABLE` initialisers, so every instance except the one the legend touched froze at startup values; SteamGraph already had the fix).
- **The two entries are the basic/advanced split**, deliberately not a mirror of `shotReview/advancedMode` — that toggle inspects one shot and must not repaint the app.
- **Drawn dimmed at draw time** over the theme colour, axis scale intact, inset clear of the bars and readouts.

## The load-bearing idea

The chart is static between shots, so it is rendered **once** into an image and blitted. Steady-state cost is one texture; everything downstream of the grab is an ordinary `Image`, making inertness true by construction. The renderer waits on `FrameAnimation` frames, not a timer — a fixed delay cleared a 129-sample shot and not a 292-sample one, and `frameSwapped` is emitted on the render thread, so both obvious alternatives were wrong. The cache is keyed on shot, advanced flag, theme identity, visibility set and size; a mismatch re-renders. Construction is behind a `Loader` and the shot loads only when selected (or when the chooser needs it), so users who never choose this pay nothing.

## The background axis is now three-way

`backgroundSource` records the kind explicitly; the old pairwise clearing was correct for two sources and quadratic for three. Existing installs migrate by derivation on read; a stored source its parameter cannot back is not believed, making migration and self-healing one code path. Notification is driven by the derived value against a snapshot — the stored key and the derived value move independently, and guarding on the stored key made the clear paths never notify, which would have latched ~70 chrome call sites in glass mode after deleting the photo they showed. `glassChrome`/`hasBackgroundImage` require a **render to exist**, so a fresh install choosing Last Shot keeps opaque chrome instead of scrimming over a flat page, and the chrome appears in the same binding pass as the picture.

## Review

Three rounds. Round one (4 agents) found the notify bug, the frozen legend, the load path's silent failures, restore asymmetry, and the chooser tiles inheriting the live shot chart. Round two (5 agents, /review) found ten more, the best of them being that round one's `-1`/`0` storage-sentinel split was built on a sentinel that never arrives — the storage layer emits -1 for an empty history too, so the "fix" pinned fresh installs at loading forever. Also: an instance-level `onPressureDataChanged` silently REPLACING the component's own reload handler, the Loader making the renderer's cleanup unreachable, the "Choose…" button testing the two legacy properties, and `FlowCalibrationPage` as the graph page that forgot to opt out. Round three (1 agent, scoped to the fix commit) came back clean on all five interactions it was pointed at.

## Tests

`tst_backgroundpresets` at 48: the three-way exclusivity matrix asserting the STORED key (the derived getter reconstructs the answer, so getter-only assertions passed with every source write deleted), notify-on-clear, bindings-see-final-value-at-emit, the advanced flip resting on a single emit, restart persistence, pattern survival, migration fall-throughs including the deliberate "shot is believed as stored" asymmetry, serializer round-trip, and a source-scraping cache-key completeness check that caught `showWeightAxis` missing on its first run. The contrast floors now call `BackgroundPresets::derive`/`liftFrom` rather than a private copy of the arithmetic.

**Full suite green: 86/86.** No PR CI gate; the local run is the gate.

## Shipped complete

Wiki manual updated (Backgrounds + Glass Chrome sections). Delta specs re-read against shipped code before archiving — two claims had gone stale during review (the scrim alpha, and glass-on-active vs glass-on-render-exists) and were corrected first. `openspec archive` promoted the specs: `last-shot-chart-background` (8 requirements) created, two capabilities updated.

## Deferred to the beta, recorded in the archived tasks

- Tablet timing of the one-off grab at shot end (steady state is a texture; only the moment of render is in question).
- Composited light-theme eyeball (the render itself is verified light-capable via image dump).


skialpine:
### Code review

Five independent review agents (CLAUDE.md compliance, deep bug scan, historical context vs #1584, prior review comments, comment compliance). 10 verified findings, all fixed in c0b0bcbe; the wiki manual remains open.

1. Empty shot history pinned the loader at "loading" forever — the storage layer initialises its query result to -1 and only overwrites it with a found row, so an empty history emits -1, not 0; the -1/0 split in the handler was built on a sentinel that never arrives. Every fresh install that picked "Last Shot" got a blank picker caption and no fallback.

https://github.com/Kulitorum/Decenza/blob/c0b0bcbebe8ad258649b08bc00619f95494dfb43/qml/components/LastShotChartSource.qml#L156-L175

2. `glassChrome`/`hasBackgroundImage` were true with no picture drawn — "shot" is believed as stored (it has no parameter), so a fresh install choosing Last Shot scrimmed ~70 chrome sites over a flat colour, the elevation-cancelling failure `chromeFill()`'s own comment documents.

https://github.com/Kulitorum/Decenza/blob/c0b0bcbebe8ad258649b08bc00619f95494dfb43/qml/Theme.qml#L504-L516

3. Re-selecting the shot background never refreshed — shots pulled while another background was active were dropped by the selection gate, and the "only if not loaded" guard then re-grabbed the days-old shot on the way back.

https://github.com/Kulitorum/Decenza/blob/c0b0bcbebe8ad258649b08bc00619f95494dfb43/qml/components/LastShotChartSource.qml#L236-L250

4. The renderer's cleanup branch was unreachable — the Loader keyed on the same selection destroys the renderer before any handler on it runs, leaving `_renderedUrl` holding a dead image-provider handle that `BackgroundSurface` bound on the next selection.

https://github.com/Kulitorum/Decenza/blob/c0b0bcbebe8ad258649b08bc00619f95494dfb43/qml/components/LastShotChartRenderer.qml#L36-L40

5. Declaring `onPressureDataChanged` on the `HistoryShotGraph` instantiation REPLACED the component's own `onPressureDataChanged: Qt.callLater(doReload)` rather than adding to it, silently disabling the chart's reload-on-pressure. Moved to a `Connections`, which appends.

https://github.com/Kulitorum/Decenza/blob/c0b0bcbebe8ad258649b08bc00619f95494dfb43/qml/components/LastShotChartRenderer.qml#L155-L166

6. Both picker shot tiles were blank for exactly the users deciding whether to apply — the singleton only loaded while the background was already selected, so `ready` stayed false and neither caption could appear. The picker now calls `ensureLoaded()` on open.

https://github.com/Kulitorum/Decenza/blob/c0b0bcbebe8ad258649b08bc00619f95494dfb43/qml/components/BackgroundPickerDialog.qml#L59-L66

7. An "image"-sourced backup left the device half-applied — the unconditional preset restore wiped the colour while a shot source survived untouched. The image branch now clears to a defined state matching its own log message.

https://github.com/Kulitorum/Decenza/blob/c0b0bcbebe8ad258649b08bc00619f95494dfb43/src/core/settingsserializer.cpp#L752-L766

8. Settings > Machine said "Choose…" while the shot background was active — the button tested the two legacy properties, the last raw two-way check in the codebase (the anti-pattern #1584's shared predicate exists to eliminate). It asks `backgroundSource` now.

https://github.com/Kulitorum/Decenza/blob/c0b0bcbebe8ad258649b08bc00619f95494dfb43/qml/pages/settings/SettingsMachineTab.qml#L851-L861

9. The glass-chrome hint only named the image case — that string exists to explain the forced-on state, and the shot chart now forces it too.

https://github.com/Kulitorum/Decenza/blob/c0b0bcbebe8ad258649b08bc00619f95494dfb43/qml/pages/settings/SettingsMachineTab.qml#L829-L832

10. `FlowCalibrationPage` draws its own `GraphsView` but never opted out of the shot-chart background — the "page that forgot to say so" `ThemedPageBackground`'s opt-in comment predicts.

https://github.com/Kulitorum/Decenza/blob/c0b0bcbebe8ad258649b08bc00619f95494dfb43/qml/pages/FlowCalibrationPage.qml#L15-L19

Still open (not code): the wiki manual update (CLAUDE.md: "a shipped feature with no manual entry is incomplete") — tracked as task 9.1 and to be done before merge, with the OpenSpec archive.

---

<details>

- [OK to skip] Renderer header comment said "main.qml instantiates this unconditionally" after the Loader made that false — fixed in the same commit while there (score: 60)
- [OK to skip] DecenzaGraphsTheme's "page shows through" comment overstates the effect since the plot area is opaque; the shipped look was approved on-screen, so the comment wording is the only gap (score: 50)
- [OK to skip] `theme["backgroundShotAdvanced"].toBool()` on a missing key defaults to basic silently — only reachable with a hand-edited backup carrying "shot" but not the flag (score: 50)

</details>

PR #1591, opened 2026-07-21
Make the equipment/recipes idle-button injection actually one-time (#1586)

### LINKED ISSUE #1586: [Bug] Update today caused return of equipment button, pushing Settings off the screen
### What happened?

Odd thing. After the update, "Equipment" came back. I had removed it when I added the Grind Setting Widget. The problem is that it cramed the bottom row so much that settings disappeared completely off the page... so I couldn't access settings at all.
I also got Service Error 5 a few times, and the DE-1 would not connect to Decenza. I tried restarting decenza, clearing cache, but eventually restarted DE1 and Tablet and normality resumed WRT connection

### Steps to reproduce

updated to sunday build, see attached log.

apologies, for mixing two issues... too tired
the second issue about the Equipment button reappearing I guess you could just add  what I have and then add Equipment back to the left bottom side.

### Decenza version

2

### Platform

iOS, Android

### Device / OS version

12

### Scale (if applicable)

_No response_

[21 july android debug.zip](https://github.com/user-attachments/files/30244081/21.july.android.debug.zip)

_No response_

### Screenshots or video

https://github.com/user-attachments/assets/35a18a8e-1b21-4cfc-b827-93f5ac900c9e

### Anything else?

_No response_

skialpine:
Fixes #1586.

## What happened

A user removed the Equipment button, and an update brought it back. On his already-packed bottom bar the extra widget pushed **Settings clean off the screen edge** — and Settings is the only route to the layout editor that could fix it, so he was locked out of his own layout.

## Why

The injection was never one-time. Its gate asked *"is an equipment widget present right now?"* from inside `getLayoutObject()`, which runs on **every launch** (the cache is per-process). Presence is user-editable state, so removing the widget flipped the gate back to `false` and the next cold start re-injected and re-persisted it.

The comment above it claimed *"Idempotent + persisted once so it never duplicates"* — true, but that only ever guaranteed no **duplicates**, not a single **run**. It was coming back on every restart, not just the update.

Both PRs that added these buttons shipped the correct one-time ledger and didn't use it: `schema_version`, a monotonic counter the app writes and the user cannot reset by editing a layout. Equipment crossed schema **22** (#1341), recipes **25** (#1441).

## The fix

- `ShotHistoryStorage` records the schema version **as opened, before migrations run**, and exposes `crossedSchemaVersion(n)` — true only on the single launch that advanced the DB past `n`.
- The two injections move out of the every-launch read path into `injectEquipmentButtonIfMissing()` / `injectRecipesButtonIfMissing()`, with identical placement logic (after beans; left of espresso, else beside equipment, else append).
- `MainController::init` calls each **only when its version was crossed this run**, still AND-ed with the existing "if missing" check.

| Case | Behaviour |
|---|---|
| Already past 22/25 (the reporter, and everyone current) | Never touched again — **a removal sticks** |
| Genuine first-time upgrade | Injected exactly once, never again even if removed |
| Fresh install | Default layout ships both; crossing-gate + present-check → **no double-add** |

Recipes had copied the identical defect (its own commit message says *"same pattern as the equipment inject"*) and is fixed the same way.

## Escape hatch

Long-press the page title on the idle screen to open Settings — a route that never depends on the customizable layout. Idle-only (the status bar is a global overlay shown on every page) and a short tap does nothing, so it can't fire by accident.

`main.qml` gains `currentPageObjectName` because a global-overlay widget isn't a child of any page, so it can't walk the visual parent chain the way in-page widgets (`BeansItem`, `EquipmentItem`) do.

## Notes

- **No space guard.** The lockout only recurred *because* the button kept coming back; the one-time gate removes the cause. A C++-side "does it fit" check could only ever be an estimate — at startup, before the window lays out, there are no widget geometries to measure.
- `ensureSettingsAccessible()` is untouched. It guarantees a Settings widget is *present*, which never triggered here — the widget was present the whole time, just off-screen.
- Wiki manual §10 gains a recovery note; held locally per the release-hold convention, not pushed.

## Testing

- 6 new tests: removal survives a reload, both placements + fallbacks, idempotency, no-op-when-already-present.
- Full suite **89/89 green**, 0 warnings, verified against a binary actually containing these changes.
- Long-press gesture manually confirmed on device by @Kulitorum.

### DIFF
--- a/qml/components/layout/items/PageTitleItem.qml
+++ b/qml/components/layout/items/PageTitleItem.qml
@@ -15,6 +15,17 @@ Item {
         return win ? (win.currentPageTitle || "") : ""
     }
 
+    // True only when the idle/home page is on top. The long-press-to-Settings
+    // rescue gesture below is gated on this: the status bar (and this title) is a
+    // global overlay shown on every page, but the gesture must only fire on Idle,
+    // where a user whose bottom bar overflowed can no longer reach the Settings
+    // widget (issue #1586). Reads the window's currentPageObjectName (this widget
+    // is not a child of any page, so it can't walk the parent chain to find it).
+    readonly property bool onIdlePage: {
+        var win = Window.window
+        return win ? win.currentPageObjectName === "idlePage" : false
+    }
+
     implicitWidth: isCompact ? compactContent.implicitWidth : fullContent.implicitWidth
     implicitHeight: isCompact ? compactContent.implicitHeight : fullContent.implicitHeight
 
@@ -50,6 +61,32 @@ Item {
                 font: Theme.bodyFont
                 visible: MachineState.isFlowing
                 Accessible.ignored: true
+            }
+        }
+
+        // Escape hatch: long-press the Idle page title to open Settings. A user
+        // whose bottom bar overflowed can lose the Settings widget off the screen
+        // edge (issue #1586) with no other route into Settings — this gives one
+        // that never depends on the customizable layout. Idle-only (onIdlePage);
+        // a short tap does nothing so it can't be triggered by accident.
+        //
+        // Accessible.ignored on purpose: screen-reader users are not the ones
+        // locked out — an off-screen Settings widget is unclipped and still in the
+        // a11y tree, so TalkBack/VoiceOver can focus it by swipe navigation. This
+        // gesture is a rescue for sighted touch users, who cannot see or reach an
+        // off-screen button. Announcing it as a second interactive node over the
+        // title would only duplicate the title readout for AT.
+        AccessibleTapHandler {
+            anchors.fill: parent
+            enabled: root.onIdlePage
+            visible: root.onIdlePage
+            supportLongPress: true
+            Accessible.ignored: true
+            accessibleName: root.pageTitle
+            onAccessibleLongPressed: {
+                var win = Window.window
+                if (win && win.goToSettings)
+                    win.goToSettings()
             }
         }
     }
--- a/qml/components/layout/items/PageTitleItem.qml
+++ b/qml/components/layout/items/PageTitleItem.qml
@@ -67,15 +67,27 @@ Item {
         // Escape hatch: long-press the Idle page title to open Settings. A user
         // whose bottom bar overflowed can lose the Settings widget off the screen
         // edge (issue #1586) with no other route into Settings — this gives one
-        // that never depends on the customizable layout. Idle-only (onIdlePage);
-        // a short tap does nothing so it can't be triggered by accident.
+        // that does not depend on where the Settings widget ended up. Idle-only
+        // (onIdlePage); a short tap does NOT open Settings, so it cannot be
+        // triggered by accident.
         //
-        // Accessible.ignored on purpose: screen-reader users are not the ones
-        // locked out — an off-screen Settings widget is unclipped and still in the
-        // a11y tree, so TalkBack/VoiceOver can focus it by swipe navigation. This
-        // gesture is a rescue for sighted touch users, who cannot see or reach an
-        // off-screen button. Announcing it as a second interactive node over the
-        // title would only duplicate the title readout for AT.
+        // Two limits worth knowing before relying on this as THE rescue:
+        //  - It lives in the compact (status-bar) rendering only. The pageTitle
+        //    widget is itself layout-editable, so a user who moved it to a center
+        //    zone or removed it has no hatch — this narrows #1586, it does not
+        //    make Settings unconditionally reachable.
+        //  - Accessible.ignored keeps it out of the a11y TREE, but not out of the
+        //    touch path: with a screen reader active a short tap still runs
+        //    AccessibleTapHandler's announce(). It is ignored because the root
+        //    Item already exposes the title as StaticText, and a second node over
+        //    the same text would just duplicate that readout.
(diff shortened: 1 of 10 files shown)


skialpine:
### Code review

Five review passes (CLAUDE.md compliance, deep bug scan, git history, prior-PR reconciliation, comment accuracy), plus a narrow follow-up round. No critical bugs in the core mechanism — the crossing arithmetic was specifically attacked against deferred migrations, constructor ordering, the zone insert logic, `QSettings::status()` semantics and the QML gating, and held up in each case.

**Correction (edited):** finding #1 below originally said the fix landed in c7d7bcd, and linked there. It did not — the `Accessible.ignored` line that makes the claim true was missing from that commit, so the permalink showed code contradicting its own adjacent comment. It landed in fd76f97d. Links now point there. Details in the follow-up round at the bottom.

Found 5 issues:

1. The long-press rescue gesture was `Accessible.ignored`, making it invisible to TalkBack/VoiceOver (`docs/CLAUDE_MD/ACCESSIBILITY.md` says "Any interactive element with secondary actions (long-press, double-tap) **must** set `Accessible.description`"). The justification was that an overflowed Settings widget stays in the a11y tree since nothing sets `clip` — but that is an unverified OS-level assumption, and it hid the only rescue path from the users least able to see the layout.

https://github.com/Kulitorum/Decenza/blob/fd76f97d6c770677858f844f4640cc6952fa5849/qml/components/layout/items/PageTitleItem.qml#L36-L43

2. `currentPageObjectName` duplicated the existing `Theme.currentPageObjectName` under the same name — already set on every page change by `updateCurrentPageScale()` and already consumed by `ShotPlanItem`. Two independently derived copies of one fact. The window property is gone; the widget reads the singleton.

https://github.com/Kulitorum/Decenza/blob/fd76f97d6c770677858f844f4640cc6952fa5849/qml/components/layout/items/PageTitleItem.qml#L24-L26

3. `crossedSchemaVersion()` named `SettingsNetwork::injectEquipmentButtonIfMissing` as a caller. It cannot be one — `SettingsNetwork` has no access to `ShotHistoryStorage`. `MainController` checks the gate and invokes the injection as the consequence.

https://github.com/Kulitorum/Decenza/blob/fd76f97d6c770677858f844f4640cc6952fa5849/src/history/shothistorystorage.h#L44-L50

4. A test comment claimed reading through a second `Settings` proves the write reached storage. Both share Qt's `QConfFile` cache in-process, so it proves the value escaped `SettingsNetwork`'s own layout cache — the distinction that matters, since that cache is what a restart discards — but not a disk write.

https://github.com/Kulitorum/Decenza/blob/fd76f97d6c770677858f844f4640cc6952fa5849/tests/tst_settings.cpp#L1300-L1308

5. No test covered a deferred migration against the gate — the case the header comment spends most of its words justifying. If a first upgrade hit a failed migration, the gate could be spent on a launch that did nothing, losing the button permanently.

https://github.com/Kulitorum/Decenza/blob/fd76f97d6c770677858f844f4640cc6952fa5849/tests/tst_dbmigration.cpp#L1015-L1027

<details>
<summary>Follow-up round (3 further issues, all fixed in fd76f97d)</summary>

6. The claim in #1 shipped ahead of the code — wrong in the commit message, in the QML comment inside that commit, and in this comment's permalink. Fourth consecutive round where a fix asserted a guarantee the code did not make, which is the same defect as the bug this PR fixes.

7. `QSettings::status()` is sticky: `setStatus` only overwrites a stored error when the current value is `NoError`, so a `FormatError` at load latches for the process and made every later write look failed. `saveLayoutObjectVerified` now samples status before the sync and reports an already-broken store as "cannot verify" rather than blaming this write.

https://github.com/Kulitorum/Decenza/blob/fd76f97d6c770677858f844f4640cc6952fa5849/src/core/settings_network.cpp#L1210-L1222

8. `IdlePage`'s lower-mid band credited `Accessible.ignored` with keeping screen readers out of the faded-out band. It does not: Qt flattens an ignored node rather than pruning its subtree (`unignoredChildren()` recurses through it and promotes descendants carrying their own role), so every handler in the band stays reachable. What actually neutralises them is the adjacent `enabled: false`. Comment corrected; behaviour deliberately unchanged.

https://github.com/Kulitorum/Decenza/blob/fd76f97d6c770677858f844f4640cc6952fa5849/qml/pages/IdlePage.qml#L1447-L1463

</details>

<details>
<summary>1 known trade-off, left as-is</summary>

- [OK to skip] The gate is committed in SQLite but acted on in QSettings, with no "already injected" marker. A crash between the two permanently skips the button. Narrow window (a handful of synchronous calls in one constructor) and deliberate — it trades guaranteed delivery for guaranteed no-resurrection, which is the point of the fix.

</details>

Known limits, stated rather than papered over: the rescue gesture is compact-mode only, so moving or deleting the `pageTitle` widget removes it — this narrows #1586 rather than making Settings unconditionally reachable. And band widgets in the faded lower-mid band remain reachable by screen reader (announced disabled); the only real fix is `visible: false`, which costs the fade animation.

Also confirmed: neither #1341 nor #1441 ever discussed the one-time-ness question in review, so this was never raised and dismissed. The wiki manual has a matching recovery note in the Layout Editor section, held locally per the release convention.

Full suite 89/89 green, 0 warnings. QML is manually verified — long-press confirmed working on device.

PR #1592, opened 2026-07-21
Fix web bean editor carrying the previous bag's product URL (#1588)

### LINKED ISSUE #1588: [Bug] Add new beans on Web Interface, product URL has previous bean URL
### What happened?

I added a bean including it's product url
i added another bean and saw that the same url was automatically put in the product url field of the new bean...  should not be there, different url for different bean!

### Steps to reproduce

as above

### Decenza version

2

### Platform

Windows

### Device / OS version

11

### Scale (if applicable)

_No response_

_No response_

_No response_

### Screenshots or video

_No response_

### Anything else?

_No response_

skialpine:
Closes #1588.

## The bug

The web `/beans` page had **two** URL fields:

- `fInfoUrl`, inside the collapsed "Get info from a roaster page (AI)" section — used only as the extraction source, never stored;
- `dLink`, "Product link" under Bean details — the bag's stored product link.

`openEditor()` repopulates every field from the record it opens, but `fInfoUrl` was not in that list. So it survived from one edit to the next, and adding a second bean showed the first bean's URL sitting in the Product URL box — exactly what the reporter saw.

## The fix

Collapse the two into one **Product URL** field, matching the app's bag form, where a single URL field is both the stored link and the page "Get info from page" reads. It is repopulated per bag like every other field, and now applies to tea bags too — they have a product page as well, and the extraction already understood the tea vocabulary while the section stayed hidden for them.

Two more on the way past:

- clear the status line on editor open, so an "Extraction failed" from the previous bag doesn't hang over a fresh form;
- restore `margin: auto` on `dialog` in the shared management CSS. The `* { margin: 0 }` reset was overriding the UA rule, so the beans, recipes **and** equipment dialogs all opened pinned to the top-left corner instead of centred.

## Verification

Rebuilt the page's HTML/CSS/JS standalone from the C++ raw strings and drove it in a browser with canned API data:

| | pre-fix | post-fix |
|---|---|---|
| URL typed on bag 1, then "+ Add Coffee" | `https://roaster-one.example/bean-one` | *(empty)* |
| Existing bag with a stored link | — | loads into the field |
| Tea bag | field hidden | field shown and populated |
| Save a new tea with a URL | — | posts `beanBaseData: {"link":"…"}` |

Dialog centring confirmed by screenshot before/after. Full local suite: 89/89 passing; clean build, no warnings.

No wiki change needed — the manual already documents this as a single "product URL + Get info from page" step, which is what the web page now does.


skialpine:
### Code review

Found 3 issues. All three were in the bean editor this PR reshapes, and all are fixed in 0d47b6a6.

1. Clearing the last blob field silently did not save. `saveEditor()` omitted `beanBaseData` when the working blob came out empty, and `bagFieldsFromBody` is a sparse whitelist — an absent key means "leave the column alone" — so the deletion was discarded and the old value returned on reopen. The same symptom class as #1588, and this PR made it the common shape: a bag whose blob holds only `{"link": …}` is now ordinary, where before the link field was buried in a collapsed coffee-only section.

https://github.com/Kulitorum/Decenza/blob/89f48f4392e4f9e6dd6c99b3e5c11f52f5ca7c86/src/network/shotserver_bags.cpp#L963-L965

2. Nothing raised while the editor was open could be read. `#status` lives in the page container and the modal covers it — `elementFromPoint` over the status line returns the dialog. So the extraction progress message, the 90s timeout, "Extraction failed: …", the server's `400 "No AI provider configured"`, and "Roaster or coffee name is required" were all invisible. With "Get info from page" promoted by this PR to a primary always-visible control on every bag, a user without an AI provider got a button that appeared to do nothing at all.

https://github.com/Kulitorum/Decenza/blob/89f48f4392e4f9e6dd6c99b3e5c11f52f5ca7c86/src/network/shotserver_bags.cpp#L506-L510

3. Extraction overwrote fields the user or a picked Bean Base entry had already filled. [BEAN_BASE.md](https://github.com/Kulitorum/Decenza/blob/89f48f4392e4f9e6dd6c99b3e5c11f52f5ca7c86/docs/CLAUDE_MD/BEAN_BASE.md#L77) specifies extraction "fills **empty** detail fields only", and the app's `take()` in `ChangeBeansDialog.qml` enforces exactly that; the web page assigned unconditionally. Since search and extraction now sit side by side and both uncollapsed, pick-then-extract is an ordinary two-click sequence that silently replaced canonical origin/process/variety with shop-page text.

https://github.com/Kulitorum/Decenza/blob/89f48f4392e4f9e6dd6c99b3e5c11f52f5ca7c86/src/network/shotserver_bags.cpp#L915-L921

Each was reproduced and re-verified against the real generated page driven in a browser. Also dropped the dead `f.roasterName`/`f.roastName`/`f.coffeeName` writes — like the `f.link` line, `AIManager::parseBagExtraction`'s key whitelist has never contained them — and corrected two comments that stated a rationale the extraction contract does not support.

Checked and clean: no dangling `fInfoUrl`/`getInfoSection` references, `link` round-trips for both kinds, the tea extraction path is genuinely supported server-side, `showInfo`'s `^https?://` scheme gate still guards the only place `link` becomes an anchor, and the shared `margin: auto` fix helps all three management dialogs rather than breaking them. Git history confirms none of the changed behaviours were deliberate earlier fixes — all four originated in #1542's from-scratch parity build.

<details>

- [Worth a look] Changing a bag's product URL on the web doesn't invalidate the cached bag photo, so a linked bag keeps the old page's image; the app's `confirmForm()` calls `refreshBagImage()` for exactly this. Pre-existing feature gap, not a regression from this diff — the `/api/bag/<id>/image` route has no invalidation hook at all (score: 65)

</details>

skialpine:
### Code review

Third round, covering the five commits added since the [first review](https://github.com/Kulitorum/Decenza/pull/1592#issuecomment-5039891240). Found 4 issues. The two most serious were introduced by this PR's own later commits and are fixed in bfd9c976; two remain open.

**Fixed in bfd9c976:**

1. MCP `bag_update` fired `refreshBagImage` before the write, which is line-for-line the defect the web route was fixed for one commit earlier. `proceed()` only learns success via `bagUpdated`, so a failed or nonexistent update still swapped a photo — and the cache key is shared by every bag on the same canonical bean, so it changed what *other* bags display, for a link change never persisted.

https://github.com/Kulitorum/Decenza/blob/5b99978b3ec3ba49b6d3a76eb0b1e9d5d6a5b2d6/src/mcp/mcptools_write.cpp#L1986-L1993

2. The unparseable-blob guard meant "never send a blob" where it should have meant "don't send an *empty* blob over one we couldn't read". On exactly the bags needing repair it silently discarded details the user had just typed, and a Bean Base pick saved as linked with the corrupt blob still in the column — unfixable from the web editor. Reproduced before and after in a browser.

https://github.com/Kulitorum/Decenza/blob/5b99978b3ec3ba49b6d3a76eb0b1e9d5d6a5b2d6/src/network/shotserver_bags.cpp#L1178-L1186

**Still open:**

3. The update route derives the image cache key from the request body rather than the stored row, so a linked bag whose update omits `beanBaseId` is keyed `bag-<rowid>` — a file nothing ever reads, while the reply says `imageRefreshed: true`. The browser always resends the id, but this route is documented for MCP/AI callers, and the MCP twin gets it right by reading `curBeanBaseId` off the loaded row.

https://github.com/Kulitorum/Decenza/blob/bfd9c976a2c76b37a0762565c5ad027ac3155cc6/src/network/shotserver_bags.cpp#L496-L500

4. A refreshed photo is not visible in the app until restart. The cache filename is the key, so a refresh writes new bytes to the same path and emits `bagImageReady` with an unchanged string — no property change, and Qt's pixmap cache is keyed by URL with no mtime check. Pre-existing, but it is precisely the user-visible outcome this PR's shared-code change exists to deliver. (CLAUDE.md: "Keep the two surfaces in sync ... update the counterpart in the same change".)

https://github.com/Kulitorum/Decenza/blob/bfd9c976a2c76b37a0762565c5ad027ac3155cc6/qml/components/BeanThumbnail.qml#L44-L49

Also verified clean this round, with commit evidence: the `refreshBagImage` redesign preserves the original safety invariant rather than regressing it (#1440/#1437 introduced evict-first and `.part`+rename defensively, not as incident fixes); `QSaveFile` is a strict superset of the old atomicity guarantee; `force=true` cannot loop and re-arms the once-per-session guard; the cross-thread captures in the MCP hop are correct; and all four duplicated `registerWriteTools` stubs were updated.

<details>

- [Worth a look] `BeanBaseClient::replaceBagImageFromUrl` has no test, unlike every sibling entry point — and its distinguishing behaviour is precisely the guard it omits (score: 65)
- [Worth a look] The app stashes a stage-2 photo under `bag-<rowid>` even when a Bean Base pick has since set a canonical id, so extract-then-pick caches it where nothing reads it; the web create route gets this right (score: 60)

</details>

PR #1596, opened 2026-07-22
Two fixes from the 2.0.0 release log: an hourly Qt SSL warning, and lost saves under DB lock contention

skialpine:
Two unrelated defects the 2.0.0 (build 3496) debug log turned up. Bundled rather than split; they touch different subsystems and can be reviewed independently.

---

# 1. `QIODevice::read (QSslSocket): device not open`, once an hour

Every update check was followed ~30 s later by that warning, in every user's log. Seven occurrences in a single 6.7-hour session, each ~29.9 s after an `UpdateChecker` reply — 7 for 7.

### Why

Not our reply handling — that is correct and the reply is disposed of properly. The chain is inside Qt:

1. The finished connection is parked in Qt's connection cache for the default **120 s** (`ExpiryTime`, `qnetworkaccesscache.cpp`).
2. GitHub closes it after **~30 s** idle.
3. `QSslSocket` raises `RemoteHostClosedError`; the channel hits `qhttpnetworkconnectionchannel.cpp:1025`, whose own comment reads *"Not actually an error, it is normal for Keep-Alive connections to close after some time"* — and then, **for HTTP/2 only**, calls `handleConnectionClosure()`.
4. That runs `QHttp2ProtocolHandler::handleConnectionClosure()` → `_q_receiveReply()` → `QHttp2Connection::handleReadyRead()` → `frameReader.read(*socket)` on an already-closed socket, with no `isOpen()` check.

The GitHub API serves HTTP/2 (`curl -w %{http_version}` → `2`), which is why it produces this and our HTTP/1.1 endpoints never have.

### Change

`ConnectionCacheExpiryTimeoutSecondsAttribute = 0` — we poll hourly and never reuse the connection, so drop it as soon as the reply is done. We close it before GitHub can, so the closure path never runs. Costs nothing: the connection was never going to survive the hour.

`checkForUpdates()` and `onPeriodicCheck()` were building the same request by hand; the policy now lives in one `releaseInfoRequest()` helper both call.

### Test

New `tst_updatechecker` pins the attribute, the headers/URL, and that both call paths produce an identical request — because **the attribute has no local symptom**. Drop it and everything still works; the only evidence is a warning in shipped logs an hour at a time.

---

# 2. Lost saves under SQLite lock contention

Twice in three days a recipe save failed and rolled back, showing the user *"Couldn't save to the recipe"* (`BrewDialog.qml:378`, on the path its own comment calls "the ONLY path a yield/temp value reaches the recipe") or *"Could not save the recipe"* in the wizard:

```
WARN  RecipeStorage: update failed: "database is locked Unable to fetch row"
```

The second landed in a post-shot cluster, where the shot insert and bag stamp are writing.

### Why

`withTempDb` does set `busy_timeout = 5000` — it just never applied, because of the transaction's shape:

```
db.transaction()              → Qt's SQLite driver issues a plain "BEGIN"  (DEFERRED)
loadRecipeStatic(db, id)      → SELECT: takes a SHARED read lock
updateRecipeFieldsStatic(...) → UPDATE: must UPGRADE to a write lock
```

If another connection wrote since the `BEGIN`, SQLite refuses the upgrade **immediately** without consulting `busy_timeout` — waiting could deadlock, so it doesn't. Confirmed `db.transaction()` is a bare `BEGIN`: `qsql_sqlite.cpp:894`.

The contention is by design: four storages (shots, bags, equipment, recipes) share `shots.db` and **each has its own worker thread**, so FIFO ordering is per-storage, not global.

`BEGIN IMMEDIATE` takes the write lock up front, which is what lets `busy_timeout` absorb it. `ShotHistoryStorage::saveShotStatic` already did exactly this for exactly this reason; nothing else did.

### The audit

Reviewing every `.transaction()` site for the same read-then-write shape. **Fixed** — all on the live `shots.db` while other storages' workers are writing:

| Site | Reads before writing |
|---|---|
| `RecipeStorage::requestUpdateRecipe` | `loadRecipeStatic` → UPDATE — **the observed failure** |
| `EquipmentStorage::supersedeOrEditStatic` | identity lookup → repoint bags + retire package |
| `ShotHistoryStorage::importShotRecordStatic` | grinder-package lookup → write shot |
| CoffeeBagStorage legacy preset import | duplicate probe → INSERT (one-time path) |

**Left alone, with reasons:**

- Schema migrations (`shothistorystorage` 473/596/906/1219/1297) run inside `initialize()`, which `MainController` calls *before* it constructs the bag, equipment and recipe storages — no other writer exists yet.
- The export/import paths at 3568/3882 operate on a separate destination file, not `shots.db`.
- `deleteShots`, the package delete, and the FTS rebuild write as their first statement, so `busy_timeout` already covers them.

The BEGIN IMMEDIATE + retry now lives in one `beginImmediateTransaction()` helper in `dbutils.h` carrying the explanation once, rather than a fourth hand-rolled copy; `isLockError` moved there too as `isSqliteLockError`, and `saveShotStatic` now uses the shared one.

Two behavioural notes worth a look in review:

- **Equipment**: a lock now aborts to a no-op instead of running the edit unwrapped. Previously a failed `BEGIN` fell through to a non-atomic multi-statement edit, which could repoint bags without retiring the old package — the duplicate-live-package state its own comment warns about. A *nested* begin still proceeds unwrapped, which is correct: the outer transaction supplies atomicity.
- **Recipes**: nested is treated as failure, since the rejection paths roll back and would discard an outer caller's work. No such caller exists today.

### Not tested, deliberately

Faithfully reproducing this needs a concurrent commit interleaved *inside* the storage's transaction, between its SELECT and its UPDATE. There is no hook for that, and the approximations I considered were timing-based and would be flaky. This rests on the mechanism above plus the existing `saveShotStatic` precedent, not on a new test. Flagging that rather than burying it.

The existing suite does cover that normal operation is unaffected: `tst_recipestorage`, `tst_equipment`, `tst_coffeebags`, `tst_dbmigration` and `tst_shotimportdedupe` all exercise these paths.

---

## Verification

Full suite via Qt Creator: **91 passed, 0 failed, 0 skipped, no warnings.**

Neither fix is confirmed on-device. Both remove a log line, so confirming means watching a running install — and for the lock fix, absence is weak evidence either way, since it only fired twice in three days.

### DIFF
--- a/qml/components/BrewDialog.qml
+++ b/qml/components/BrewDialog.qml
@@ -111,6 +111,10 @@ Dialog {
     // recipeUpdated(recipeId, success) result (RecipesPage's pending-id
     // pattern) so a failure never looks like the tap was ignored.
     property int _pendingRecipeUpdateId: -1
+    // Cause from recipeUpdateFailed, consumed by the recipeUpdated handler that
+    // follows it. Cleared on every read so one failure's reason cannot colour
+    // the next attempt's message.
+    property string _recipeUpdateFailReason: ""
     // Failure banner for the fire-and-forget recipe operations (update /
     // switch). Auto-dismisses — the allowed toast use of a timer.
     property string recipeErrorText: ""
@@ -368,6 +372,14 @@ Dialog {
             if (root.visible)
                 MainController.recipeStorage.requestInventory()
         }
+        // Lands just before recipeUpdated(false) and names the cause. "busy" is
+        // the database being locked by another write — the one failure here the
+        // user can act on, and this is the path the lock bug was reported from,
+        // so it must not fall back to the generic wording.
+        function onRecipeUpdateFailed(recipeId, reason) {
+            if (recipeId === root._pendingRecipeUpdateId)
+                root._recipeUpdateFailReason = reason
+        }
         // "Update Recipe" is fire-and-forget and, with the auto-stamp gone,
         // the ONLY path a yield/temp value reaches the recipe — a swallowed
         // failure would silently lose the user's dialed value.
@@ -375,8 +387,14 @@ Dialog {
             if (recipeId !== root._pendingRecipeUpdateId)
                 return
             root._pendingRecipeUpdateId = -1
-            if (!success && root.visible)
-                root.showRecipeError(TranslationManager.translate("brewDialog.recipeUpdateFailed", "Couldn't save to the recipe"))
+            const reason = root._recipeUpdateFailReason
+            root._recipeUpdateFailReason = ""
+            if (!success && root.visible) {
+                root.showRecipeError(reason === "busy"
+                    ? TranslationManager.translate("brewDialog.recipeUpdateBusy",
+                          "The database was busy — tap Update Recipe again.")
+                    : TranslationManager.translate("brewDialog.recipeUpdateFailed", "Couldn't save to the recipe"))
+            }
         }
     }
 
--- a/qml/pages/RecipeWizardPage.qml
+++ b/qml/pages/RecipeWizardPage.qml
@@ -1773,9 +1773,15 @@ Page {
                     wizardPage._saveFailReason = ""
                     pageStack.pop()
                 } else {
+                    // "busy" means the database was locked by another write — the
+                    // one cause here where trying again actually works, so say so
+                    // rather than showing the generic failure.
                     wizardPage.errorMessage = wizardPage._saveFailReason === "nameInUse"
                         ? TranslationManager.translate("recipes.dialog.nameInUse",
                               "That name is already in use — choose a different name.")
+                        : wizardPage._saveFailReason === "busy"
+                        ? TranslationManager.translate("recipes.wizard.errorSaveBusy",
+                              "The database was busy — tap Save again.")
                         : TranslationManager.translate("recipes.wizard.errorSave", "Could not save the recipe")
                     wizardPage._saveFailReason = ""
                     wizardPage.showSaveError()
--- a/src/core/dbutils.h
+++ b/src/core/dbutils.h
@@ -119,12 +119,28 @@ public:
     }
 
     ~DbWriteTxn() {
-        if (m_db)
-            m_db->rollback();
+        // A failing rollback here means the transaction we thought we owned was
+        // already terminated by someone else — in practice, a caller that
+        // committed with the raw QSqlDatabase::commit() instead of commit()
+        // below, which leaves this guard armed. That happened at two sites
+        // within a day of the class existing, silently, because the return was
+        // discarded. Logging it makes the next occurrence fail loudly instead:
+        // the test suite runs with QTest::failOnWarning(), so any storage test
+        // touching such a path now breaks, at every present and future site,
+        // without anyone having to write a test for it.
+        //
+        // It also catches the genuinely dangerous case: on a long-lived
+        // connection a failed ROLLBACK leaves the write transaction open, and
+        // every later write on that connection silently joins it.
+        if (m_db && !m_db->rollback())
+            qWarning() << "DbWriteTxn: ROLLBACK failed — the transaction was already ended by "
+                          "someone else (a raw db.commit()?), or is still open:"
+                       << m_db->lastError().text();
     }
 
     DbWriteTxn(DbWriteTxn&& other) noexcept
-        : m_db(other.m_db), m_lockTimedOut(other.m_lockTimedOut) { other.m_db = nullptr; }
+        : m_db(other.m_db), m_lockTimedOut(other.m_lockTimedOut),
+          m_commitError(std::move(other.m_commitError)) { other.m_db = nullptr; }
     DbWriteTxn(const DbWriteTxn&) = delete;
     DbWriteTxn& operator=(const DbWriteTxn&) = delete;
     DbWriteTxn& operator=(DbWriteTxn&&) = delete;
@@ -142,21 +158,27 @@ public:
     // back in that case) or if we never held one. After this the guard is inert,
     // so the destructor does nothing.
     bool commit() {
-        if (!m_db)
+        if (!m_db) {
+            m_commitError = QStringLiteral("no transaction was held");
             return false;
+        }
         QSqlDatabase* db = m_db;
         m_db = nullptr;
         if (db->commit())
             return true;
-        // Capture before the rollback below replaces it — otherwise a caller
-        // logging db.lastError() after a failed commit reports the ROLLBACK's
-        // outcome (often success) instead of the reason the COMMIT failed.
+        // Capture now. A SUCCESSFUL rollback leaves lastError() alone — Qt's
+        // driver only calls setLastError on failure — but a FAILING one
+        // overwrites it, and it fails exactly when SQLite has already unwound
+        // the transaction ("cannot rollback - no transaction is active"). In
+        // that case a caller logging db.lastError() would name the rollback
+        // instead of the commit, which is the case worth diagnosing.
         m_commitError = db->lastError().text();
         db->rollback();
         return false;
     }
 
-    // Why the last commit() failed. Empty unless commit() returned false.
+    // Why the last commit() failed. Always set when commit() returns false,
+    // including when the guard held no transaction. Empty after a success.
     QString commitError() const { return m_commitError; }
 
 private:
--- a/src/core/dbutils.h
+++ b/src/core/dbutils.h
@@ -12,13 +12,20 @@
 #include <functional>
 #include <memory>
 
-// True when a failed statement failed because someone else holds the lock
-// (SQLITE_BUSY = 5, SQLITE_LOCKED = 6) rather than for a reason retrying cannot
-// fix. The text checks are a backstop: the native code is empty on some driver
-// paths, and Qt appends its own prefix to the message.
+// True when a statement failed because someone else holds the lock rather than
+// for a reason retrying cannot fix. This is what separates "wait and try again"
+// from "give up": a constraint violation must never be retried.
 //
-// This distinction is what separates "wait and try again" from "give up": a
-// constraint violation or a nested-transaction error must never be retried.
+// The TEXT match is the primary detector, not a backstop — do not delete it as
+// "fragile string matching". Qt enables SQLite's EXTENDED result codes by default
+// (qsql_sqlite.cpp, `useExtendedResultCodes`, cleared only by a connect option
+// this app never sets), and the failure this whole helper exists to catch — the
+// WAL read→write upgrade — reports SQLITE_BUSY_SNAPSHOT = 517, not 5. So
+// nativeErrorCode() is "517" and the numeric checks below miss it entirely; what
+// catches it is sqlite3_errmsg's "database is locked". The numeric checks cover
+// the plain SQLITE_BUSY (5) / SQLITE_LOCKED (6) cases and driver paths that
+// report no text. QSqlError::text() is the database message with Qt's driver text
+// appended.
 inline bool isSqliteLockError(const QSqlError& e) {
     const QString code = e.nativeErrorCode();
     return code == QLatin1String("5") || code == QLatin1String("6")
@@ -27,24 +34,36 @@ inline bool isSqliteLockError(const QSqlError& e) {
 }
 
 // Outcome of beginImmediateTransaction().
+//
+// Only `Started` means "you own a transaction and must commit or roll it back
+// exactly once". Every caller here refuses everything else, and new ones should
+// too: test `!= TxnBegin::Started`, not `== TxnBegin::Locked`. Handling one
+// failure state and falling through on the others runs the body with no
+// transaction at all, which turns its rollbacks into silent no-ops.
+//
+// `Nested` is unreachable today — every caller runs on a fresh withTempDb
+// connection in autocommit. It is distinguished anyway because an outer
+// transaction is the one case where proceeding could be legitimate, and a caller
+// that adopts one must NOT commit or roll back. If you find yourself wanting
+// that, prefer restructuring so the transaction has a single owner.
 enum class TxnBegin {
-    Started,  // we own a fresh write transaction
-    Nested,   // already inside one; the outer transaction provides atomicity
+    Started,  // we own a fresh write transaction — commit or roll back exactly once
+    Nested,   // already inside one, owned by an outer frame — do not terminate it
     Locked,   // another writer held the lock through every attempt
     Failed,   // anything else — see the logged error
 };
 
-// Starts a top-level write transaction, retrying briefly while another writer
-// holds the lock.
+// Starts a write transaction, retrying while another writer holds the lock.
 //
 // Use this, NOT QSqlDatabase::transaction(), for any transaction that reads
-// before it writes. Qt's SQLite driver issues a plain DEFERRED BEGIN
-// (qsql_sqlite.cpp: `q.exec("BEGIN")`), which takes a read lock for the first
+// before it writes. Qt's SQLite driver issues a plain DEFERRED BEGIN (see
+// QSQLiteDriver::beginTransaction()), which takes a read lock for the first
 // SELECT and then has to UPGRADE it for the first write — and if another
 // connection wrote in between, SQLite fails that upgrade IMMEDIATELY rather than
-// waiting out busy_timeout, because waiting could deadlock. The PRAGMA
-// busy_timeout that withTempDb sets is simply not consulted, so the write fails
-// on the first brush with contention.
+// waiting out busy_timeout. (In WAL, which this database uses, the read snapshot
+// is stale and waiting could never succeed, so the busy handler is not consulted
+// at all.) The PRAGMA busy_timeout that withTempDb sets is simply not reached, so
+// the write fails on the first brush with contention.
 //
 // That is not theoretical here: four storages (shots, bags, equipment, recipes)
 // share shots.db, each on its own worker thread, so their writes overlap by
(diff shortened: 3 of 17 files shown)


skialpine:
## Review pass — three defects fixed, two claims retracted

Ran the multi-agent review over the branch. It found three real defects in the first two commits (one a regression I introduced), and disproved two things I asserted in this PR description. Fixed in `a477f4e9`..HEAD.

### Defects

**1. An equipment identity edit reported as saved (regression).** `supersedeOrEditStatic` returned `packageId` on a lock — which is *also* what a successful edit-in-place returns, and callers only have `> 0`. The dialog closed, MCP answered `{"success": true}` with the old identity, and a combined rename+burrs save persisted the rename while dropping the burrs. `SwitchEquipmentDialog.qml:132` carries a comment warning against precisely this. Failure returns `-1` now; the caller stops rather than applying the rest. The pre-existing `fail()` lambda had the same overload for every SQL error, so that went too.

**2. The fix was a guaranteed no-op at two sites.** `BEGIN IMMEDIATE` cannot take the lock while a statement is still active on the connection — SQLite skips the busy handler while the connection sits in a read transaction, so it fails instantly and retrying cannot help. The legacy bag import left a `COUNT(*)` probe open (a count *always* returns a row, so it is never auto-reset) and the shot import left its dedupe probe open. Both `finish()` now, and the precondition is documented on the helper.

**3. Shot import fell through with no transaction.** It checked only for `Locked`. On `Failed` it ran unwrapped, so its rollbacks became no-ops and a failed samples insert left a permanently graphless shot row while reporting failure. Now refuses anything but `Started`, and checks its commit.

### Two claims retracted

- **"The export/import paths operate on a separate destination file."** Wrong: `destDb` is `destDbPath` — the live database. `srcDb` is the imported file. The backfill is read-then-write on the live DB from a background thread. Converted.
- **"Not testable without a flaky timing hook."** Also wrong, and the more useful correction. The bug is not a race: SQLite refuses the stale-snapshot upgrade *immediately*, so it reproduces from two statements in a fixed order on one thread. `tst_dbtxn` now covers the deferred-vs-immediate interleave, the active-statement precondition, `Locked`, `Nested`, and that a constraint violation is never mistaken for contention — no `qWait`, no timing assertions.

### Comment corrections

- `isSqliteLockError`'s comment had it backwards. Qt enables extended result codes, so the upgrade reports `SQLITE_BUSY_SNAPSHOT` (**517**) — the numeric checks miss it and the **text** match is the primary detector. Anyone "tightening away the fragile string matching" would have silently broken the whole change.
- "No statement inside the transaction can block" is contradicted by `saveShotStatic`'s own COMMIT retry, in the same file. Hedged.
- "Retrying briefly" was really **~20 s** — each attempt can burn the full `busy_timeout` first. Default is 2 attempts now, the bound is stated, and the GUI-thread caller passes 1.

Also: a locked recipe save now sets `failReason "busy"` so the user is told to try again (MCP was reporting *"Recipe N not found"*); `Nested` returns are logged; `[[nodiscard]]` makes ignoring the result a build error.

### Not done

The type-design review argued `TxnBegin` should be an RAII guard — the ownership obligation is caller discipline today, and it had already been got wrong at one of four sites. Its cost table shows three of the four getting shorter. I think the argument is right, but it rewrites four commit/rollback flows and this PR is large enough; flagging it as a follow-up rather than folding it in.

**Suite: 92 passed, 0 failed, no warnings.**

skialpine:
## RAII guard folded in

`TxnBegin` + `beginImmediateTransaction()` are gone, replaced by `DbWriteTxn`, which owns the transaction for its scope and rolls back on any exit that is not an explicit `commit()`.

The argument that decided it: the enum handed back a code and walked away, leaving "terminate this exactly once, and only if you started it" as caller discipline — and that discipline was already wrong at one of four sites *within a day of the enum existing*. Shot import checked for one failure value and fell through on the others, running the body with no transaction and silently turning its rollbacks into no-ops. With the guard that shape is unrepresentable: an early return, a missed branch and a thrown exception all unwind, and `[[nodiscard]]` on the class makes a dropped guard a compile error.

It is also **less** code where the guard had been hand-rolled:

| Site | Removed |
|---|---|
| equipment identity edit | the `inTxn` bool and the `fail()` lambda — a guard built out of a flag and a closure; ten `return fail()` became `return -1` |
| recipe update | six `db.rollback()` calls before six early returns |
| shot import | three more |
| legacy bag import, import backfill | one each |

**A nested begin is now a failure everywhere, not something to adopt.** Adopting looks harmless and isn't: an inner rejection path would leave its partial writes staged in the outer caller's transaction, because the rollback is not ours to perform. It stays distinguishable in the log, and `lockTimedOut()` still drives the recipe *"the database was busy — tap Save again"* message.

Two behaviour changes fall out, both toward refusing rather than half-doing:

- **The equipment package delete** used to fall through and delete unwrapped when its BEGIN failed — the orphaned-items state its transaction exists to prevent. It now aborts.
- **`saveShotStatic` keeps its own hand-rolled BEGIN IMMEDIATE**, deliberately: it retries the whole INSERT *and* the COMMIT, which the guard doesn't model, and it's covered by the shot corpus. Forcing it into the guard would have been the riskiest edit in the change for no gain.

`tst_dbtxn` grows the two cases that only exist because of the guard: scope exit without commit rolls back, and commit releases so the destructor cannot undo it.

**Suite: 92 passed, 0 failed, no warnings.**

skialpine:
## Live-tested against the running macOS build

Exercised on the Mac instance (ASan+UBSan debug build, so every path below ran instrumented). Verified with SQL against the live database and diffed against the pre-test daily backup.

### Verified working

- **Rollback on early return** — the guard's headline property. A rename colliding with an existing active recipe, sent together with `doseG: 19`: request rejected, and the dose was still 18 afterwards. The dose was written inside the transaction and thrown away on the rejection path, which is now the destructor rather than an explicit rollback.
- **Commit path** — recipe grind 9 → 9.1 → 9, persisted and read back each time.
- **Concurrent writers across storages** — bursts of 6–8 simultaneous writes spanning the recipe, bag and equipment workers (separate threads on the shared `shots.db`). All succeeded, no lock warnings, no failures.
- **Equipment copy-on-write, all three branches** — fork (package with shots → new id, old retired), edit-in-place (0-shot package keeps its id), and merge (identity retargeted onto an existing package → returns that package, empty source deleted).
- **Integrity** — `PRAGMA integrity_check` ok, `foreign_key_check` clean, zero orphaned equipment items, zero shots or bags pointing at a missing package. Every value I touched restored byte-for-byte against the backup.

### Two bugs it found (fixed in 9bf9e1c9)

**Dangling lineage pointer after a merge.** Merging a package that was itself a fork target left `superseded_by` pointing at a hard-deleted id — the older package renders as "older" with no successor to resolve. `requestDeletePackage` refuses a delete outright for exactly this reason; the merge had no equivalent. It now re-points to the merge target, which is the natural successor.

Caught by running `superseded_by NOT IN (SELECT id FROM equipment_packages)` after a fork-then-merge; it came back non-empty.

**`equipment_update` reported `shotCount: 0` for every package it edited**, regardless of history — `shotCount` is a per-query aggregate, not a package column, and this path only loaded the package row. An assistant reading that would conclude the package is disposable, which is the exact judgement copy-on-write exists to protect.

Both pre-existing, neither caused by the transaction work, both in files this PR already touches.

### Still not verified

- **The SSL fix.** No update check has run: `onPeriodicCheck()` early-returns unless the app is frontmost. Needs the window focused when the hourly timer fires.
- **The `Locked` path in production.** Parallel MCP writes never collided hard enough to trigger it. Covered by `tst_dbtxn` only.

**Suite: 92 passed, 0 failed, no warnings.**

skialpine:
### Code review

Found 6 issues:

1. Two converted call sites terminate the transaction with the raw `db.commit()` instead of `txn.commit()`, so the `DbWriteTxn` guard is never released and its destructor fires a `ROLLBACK` on scope exit after the commit already succeeded. Currently harmless (the data is already committed and SQLite rejects the stray ROLLBACK), but it defeats the "terminate exactly once" contract this PR introduces, and it is the exact bug `tst_dbtxn.cpp::commitReleasesSoTheDestructorIsInert` exists to catch. The other four sites do it correctly.

https://github.com/Kulitorum/Decenza/blob/9bf9e1c9c7ad25b0a99a24a45b059560e349878e/src/history/recipestorage.cpp#L657-L663

https://github.com/Kulitorum/Decenza/blob/9bf9e1c9c7ad25b0a99a24a45b059560e349878e/src/history/shothistorystorage.cpp#L4210-L4216

2. Data loss on overwrite import. The dedupe probes call `deleteShotStatic()` in autocommit mode, so the existing shot is deleted and committed before any transaction opens. Previously the following `db.transaction()` return was ignored and a DEFERRED BEGIN essentially never failed, so the replacement INSERT always ran. Now a failed `begin()` returns -1 with the old shot already destroyed and nothing written in its place. Fix: open the transaction before the probes so the deletes roll back with everything else.

https://github.com/Kulitorum/Decenza/blob/9bf9e1c9c7ad25b0a99a24a45b059560e349878e/src/history/shothistorystorage.cpp#L4012-L4018

https://github.com/Kulitorum/Decenza/blob/9bf9e1c9c7ad25b0a99a24a45b059560e349878e/src/history/shothistorystorage.cpp#L4048-L4056

3. `m_db` never sets `PRAGMA busy_timeout`, unlike every `withTempDb` connection. `DbWriteTxn`'s stated premise is that "BEGIN IMMEDIATE takes the write lock up front, which is what lets busy_timeout absorb the contention" - on this connection there is no busy handler at all, so SQLITE_BUSY returns instantly and the only tolerance is the guard's 50 ms backoff. This is what makes issue 2 reachable.

https://github.com/Kulitorum/Decenza/blob/9bf9e1c9c7ad25b0a99a24a45b059560e349878e/src/history/shothistorystorage.cpp#L118-L122

4. The new `failReason = "busy"` channel is only consumed by `RecipeWizardPage.qml`. `BrewDialog.qml` subscribes to `recipeUpdated` but never to `recipeUpdateFailed`, so it still shows the generic "Couldn't save to the recipe" - and that dialog is the path the PR description cites as the original field symptom, on what its own comment calls "the ONLY path a yield/temp value reaches the recipe".

https://github.com/Kulitorum/Decenza/blob/9bf9e1c9c7ad25b0a99a24a45b059560e349878e/qml/components/BrewDialog.qml#L372-L381

5. Stale comment naming a type that no longer exists: `TxnBegin::Started` was the enum design this PR replaced with `DbWriteTxn`. `grep -rn "TxnBegin"` finds only this comment.

https://github.com/Kulitorum/Decenza/blob/9bf9e1c9c7ad25b0a99a24a45b059560e349878e/tests/CMakeLists.txt#L427-L431

6. A failed non-identity update (e.g. a rename) is masked by `|| any` whenever the identity edit above it succeeded, so half the update is dropped and the whole thing is reported as a success. That is the same outcome this PR's own new comment at `equipmentstorage.cpp:420-423` calls "the worst of the three". Pre-existing, but CLAUDE.md says "Fix pre-existing violations in any file you touch - do not dismiss them as 'pre-existing'".

https://github.com/Kulitorum/Decenza/blob/9bf9e1c9c7ad25b0a99a24a45b059560e349878e/src/history/equipmentstorage.cpp#L441-L445

Checked and clear: the guard's own ownership semantics (move ctor nulls the source, `commit()` nulls before committing, copy/move-assign deleted, `[[nodiscard]]` on the class); all ~15 removed `db.rollback()` calls; the `-1` return change against both callers; the lineage re-point branch placement and chain behaviour; the three new `finish()` precondition calls; and every falsifiable claim in the `dbutils.h` comment block, re-verified against Qt 6.11.1 sources including the `SQLITE_BUSY_SNAPSHOT` = 517 point.

skialpine:
All six review issues fixed in `f0a3d0e`.

**1. Data loss on overwrite import** — the transaction now opens at the top of `importShotRecordStatic`, before the dedupe probes, so the `deleteShotStatic()` calls that replace an existing shot are inside it and roll back with everything else. This was the real one: my own change created it. Ignoring the old `db.transaction()` return meant a DEFERRED BEGIN essentially never failed and the replacement INSERT always ran; turning that into "abort unless the transaction starts" converted a case that used to limp through into one that destroys a shot and writes nothing back.

**2. Guard bypassed at two sites** — both now use `txn.commit()`. Also fixed something the review didn't flag: `commit()` rolls back internally when the COMMIT fails, which overwrote `db.lastError()` before the caller logged it, so both warnings were about to report the ROLLBACK's outcome instead of the reason the commit failed. `commit()` now captures the error first and exposes it as `commitError()`.

**3. `m_db` busy_timeout** — added, matching every `withTempDb` connection. This is what made issue 1 reachable rather than theoretical, and I had argued it down earlier as nearly redundant. That was wrong: it is the only contention tolerance on the synchronous import path.

**4. BrewDialog** — now subscribes to `recipeUpdateFailed` and shows "The database was busy — tap Update Recipe again." for `busy`. The reason channel existed for exactly this dialog and only reached the wizard.

**5. Stale comment** — `TxnBegin::Started` replaced with the shipped `DbWriteTxn::begin()` wording.

**6. Masked rename** — a failed non-identity update now fails the whole update instead of being masked by `|| any`. The identity edit commits in its own transaction and cannot be undone from there, so the honest report is failure with the partial state visible.

`tst_dbtxn` gains the ordering case the data-loss bug turned on: a delete issued before the transaction is not undone, one issued inside it is.

Suite: 92 passed, 0 failed, no warnings — including `tst_shotimportdedupe`, which exercises the restructured import path.

On process: the two bypassed commits came from a scripted find-and-replace whose pattern no longer matched after an earlier pass had already stripped part of it. It silently changed nothing and I didn't verify. The two I edited by hand were both correct.

skialpine:
Third review round. All findings fixed, and where a mechanism existed I used it instead of care.

## A claim I had wrong, and repeated

The `busy_timeout` pragma I added last round was **redundant**: Qt's SQLite driver already calls `sqlite3_busy_timeout(5000)` on every connection it opens (`qsql_sqlite.cpp:751,828`), overridable only by a connect option this app never sets. So `m_db` always had one — and the story that its absence made the import failure reachable, which two review agents and I all asserted, was never true. Nobody had checked the driver source.

It was also actively harmful: `PRAGMA busy_timeout = N` **returns a row**, and Qt leaves a row-returning statement un-reset, so the reused `QSqlQuery` held a read transaction open across `createTables()`, `runMigrations()` and the startup WAL checkpoint — the exact active-statement hazard `DbWriteTxn`'s own precondition documents. The line is gone.

## The import is restructured, not patched again

Both previous shapes were wrong in opposite directions. Deleting outside a transaction lost shots. Opening the transaction first fixed that but made every import take the write lock — including pure duplicates that write nothing — turning a "skipped" result into a "failed" one under contention, which both importers show the user in their tally.

The probes are now read-only and only record which shots they replace; the deletes run inside the transaction. That gets both properties.

## Structural, so the next mistake is caught by a machine

- **`deleteShotStatic` is `[[nodiscard]]`.** Its return was dropped at all three probes, so a failed delete fell through to the INSERT and either duplicated the shot or failed on the uuid constraint and blamed the constraint. With `-Werror=unused-result` that is now a build error.
- **`~DbWriteTxn` logs a failed ROLLBACK.** That is what a raw `db.commit()` inside a guard's scope looks like from the inside — and the suite runs with `QTest::failOnWarning()`, so the mistake that shipped twice now breaks existing storage tests at every present and future site, with no new test needed.
- **The GUI-thread import passes `beginAttempts = 1`**, which the guard's COST note asks for, rather than leaving it to be remembered.

## The rest

The equipment early return left `*resultId` as the forked id while the ShotServer handler filters on `updatedId == packageId`, so **the HTTP request got no response at all**; it now reports the id the caller asked about plus a `failReason` distinguishing a half-applied save from a no-op. `equipment_update` still carried the `|| ok` mask this branch fixed in `requestUpdatePackage`, because it reimplements it. Two sites still logged `db.lastError()` after `txn.commit()`. And `commitError()`'s comment had the mechanism backwards — a *successful* rollback leaves `lastError()` alone; only a failing one overwrites it.

## The test now actually holds the line

The case I added last round asserted that transactions roll back — true of every SQL database, and green before the fix. Replaced with one that drives the real importer, injecting a deterministic post-delete failure via a `BEFORE INSERT ... RAISE(ABORT)` trigger. (Lock contention would have been false confidence: it would also block the old autocommit DELETE, so the row survives under the buggy code and the test passes against the bug.)

**Verified by reintroducing the bug** — moving the deletes back outside the transaction — and confirming it fails:

```
FAIL!: failed_overwrite_import_keeps_the_original_shot()
       'shots == 1' returned FALSE. (the import failed but destroyed the shot it was replacing)
```

Plus a second case pinning that a skipped duplicate needs no write lock.

Suite: 92 passed, 0 failed, no warnings. Every edit verified present afterwards, which is what I failed to do the round the scripted replaces silently no-op'd.

skialpine:
### Code review

Found 3 issues:

1. `equipment_update` commits a name-only rename and then reports it as a failure. The `|| ok` removed in 596c7dd2 was carrying the SUCCESS assignment as well as the failure mask; the replacement only ever sets `ok = false`. `ok` starts false and becomes true only inside the identity-edit block, so a call with just `{"packageId": N, "name": "..."}` skips that block, writes the rename, and then answers `{"error": "Package not found or update failed"}`. Retrying gives the same false error, since the second attempt no longer differs from the stored name. Fix: set `ok = true` on the success path.

https://github.com/Kulitorum/Decenza/blob/596c7dd2ff1006a309122c5834e2a0d458441cf4/src/mcp/mcptools_write.cpp#L2417-L2430

2. The ShotServer equipment handler still hangs on the SUCCESS fork/merge path — the same defect this PR fixed one branch of. `supersedeOrEditStatic` returns a different id whenever the package is forked (it has shots) or merged, so `packageUpdated(<forked id>, true)` fails the `updatedId != packageId` filter: no `respondJson`, no `disconnect`, and the HTTP request never gets an answer. The PR's own comment at `equipmentstorage.cpp:453-457` names this exact mechanism when fixing the failure path. `src/history/equipmentstorage.cpp` is a file this PR modifies, and CLAUDE.md says "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'".

https://github.com/Kulitorum/Decenza/blob/596c7dd2ff1006a309122c5834e2a0d458441cf4/src/network/shotserver_equipment.cpp#L169-L175

3. The new `duplicate_skip_needs_no_write_lock` test skips the `close()` + `drain()` the file documents as mandatory. Its own helper comment says "initialize() spawns a distinct-cache thread; close() + drain before the storage destructs or it can crash"; every other test in the file, including the sibling added in the same commit, does it. Flaky-crash risk in the nightly sanitizer job rather than a deterministic failure.

https://github.com/Kulitorum/Decenza/blob/596c7dd2ff1006a309122c5834e2a0d458441cf4/tests/tst_shotimportdedupe.cpp#L231-L236

Checked and clean: the `importShotRecordStatic` restructure (probe de-duplication, `query.finish()` placement, all `return 0` paths write-free, rollback on a failed delete); `~DbWriteTxn`'s new warning does not fire on any routine path; `beginAttempts` is consistent across header default, definition and both callers; the MCP early `return` exits the intended scope; `commitError()`'s new inert-branch message is unreachable from any logging site; and the `PRAGMA busy_timeout` removal is correct — Qt's driver sets it unconditionally (`qsql_sqlite.cpp:751,828`) and the pragma returns a row Qt leaves un-reset, which is what made the added line harmful. Every falsifiable claim in the new comments was verified against Qt sources and against SQLite empirically; none was wrong this round.

<details>

- [Worth a look] `shotserver_equipment.cpp` answers the new `partiallyApplied` reason with HTTP 404 "Package not found", for a package that exists and was half-modified — a 500 with the reason would read better (score: 60)

</details>

skialpine:
All four fixed (the three reported plus the 60-scored one).

**1. `equipment_update` name-only rename.** The `ok = updatePackageFieldsStatic(...) || ok` I removed was doing two jobs — masking a failed rename behind a successful identity edit (the bug) and setting the success flag for a successful rename (not the bug). Removing the expression removed both, so a name-only call committed the rename and answered "Package not found or update failed". Both halves are spelled out now.

New regression test `equipmentUpdateNameOnlyReportsSuccess` in `tests/tst_mcptools_write.cpp`, verified by removing the fix again:

```
FAIL!: equipmentUpdateNameOnlyReportsSuccess()
       'result["success"].toBool()' returned FALSE.
       ({"error": "Package not found or update failed: 1", "success": null})
```

It also asserts the rename is on disk, so a green result cannot mean "reported success without writing".

**2. ShotServer success-path hang.** The `updatedId != packageId` filter is gone from the update route. Since failures always carry the requested id (`requestUpdatePackage` resets it), only a success can differ — so dropping the filter is safe, and the response now returns the RESULTING id, which is what a client must address after a fork or merge. The `/remove` route keeps its filter; it has no copy-on-write path.

**3. `partiallyApplied` no longer answers 404.** A package that exists and was half-modified is not "not found", and re-sending the same body is not a no-op — 500 with which half landed.

**4. The duplicate-skip test** now closes and drains the storage, as its own file documents.

Suite: 92 passed, 0 failed, no warnings.

---

On the pattern, since it is now four rounds: every one of my fix commits has introduced a defect, and three in a row have been the same mistake — changing the shape of an expression or control flow without re-deriving everything the old shape did. `|| ok`, `|| any`, and the transaction ordering were all this. The countermeasures I have added (`[[nodiscard]]`, the destructor warning, and now two revert-verified regression tests) each close one instance rather than the class.

I would not treat another clean automated round as evidence this is finished. This wants human review before merge.

PR #1598, opened 2026-07-22
Add custom endpoint URL support for OpenAI and Anthropic

csader:
## Summary

- Adds optional custom endpoint URL fields for the OpenAI and Anthropic providers, letting users point either provider at any compatible API (proxy, gateway, or self-hosted endpoint)
- When empty (default), the hardcoded upstream URLs are used — no behavior change for existing users
- Follows the same pattern already used by the Ollama provider's configurable endpoint

## Changes

- `src/core/settings_ai.h/.cpp` — new `openaiEndpoint` and `anthropicEndpoint` Q_PROPERTYs
- `src/ai/aiprovider.h/.cpp` — `setBaseUrl()` on OpenAI/Anthropic providers; uses custom base when set for all API calls (chat completions, responses, test connection)
- `src/ai/aimanager.cpp` — plumbs settings to providers on create and on settings change
- `qml/pages/settings/SettingsAITab.qml` — "Custom Endpoint (optional)" text field shown when OpenAI or Anthropic is selected

## Test plan

- [x] Full test suite passes (90/90 tests, macOS)
- [x] Set a custom endpoint, hit Test Connection — verify request goes to custom URL
- [x] Clear the endpoint field — verify default behavior restored
- [x] Confirm no UI regression on Gemini, OpenRouter, Ollama provider tabs

### DIFF
--- a/qml/pages/settings/SettingsAITab.qml
+++ b/qml/pages/settings/SettingsAITab.qml
@@ -6,7 +6,7 @@ import "../../components"
 
 KeyboardAwareContainer {
     id: aiTab
-    textFields: [apiKeyField, ollamaEndpointField, openrouterModelField, customUrlField, claudeRcUrlField, providerEndpointField]
+    textFields: [apiKeyField, ollamaEndpointField, openrouterModelField, customUrlField, claudeRcUrlField]
     targetFlickable: aiFlickable
 
     property string testResultMessage: ""
--- a/qml/pages/settings/SettingsAITab.qml
+++ b/qml/pages/settings/SettingsAITab.qml
@@ -243,51 +243,6 @@ KeyboardAwareContainer {
                     }
                 }
 
-                // Custom endpoint (OpenAI / Anthropic only)
-                ColumnLayout {
-                    visible: Settings.ai.aiProvider === "openai" || Settings.ai.aiProvider === "anthropic"
-                    Layout.fillWidth: true
-                    spacing: Theme.scaled(8)
-
-                    Tr {
-                        key: "settings.ai.customEndpoint"
-                        fallback: "Custom Endpoint (optional)"
-                        color: Theme.textColor
-                        font.pixelSize: Theme.scaled(14)
-                        font.bold: true
-                    }
-
-                    StyledTextField {
-                        id: providerEndpointField
-                        Layout.fillWidth: true
-                        placeholderText: Settings.ai.aiProvider === "openai"
-                            ? "https://api.openai.com"
-                            : "https://api.anthropic.com"
-                        inputMethodHints: Qt.ImhNoPredictiveText | Qt.ImhNoAutoUppercase | Qt.ImhUrlCharactersOnly
-                        text: {
-                            switch(Settings.ai.aiProvider) {
-                                case "openai": return Settings.ai.openaiEndpoint
-                                case "anthropic": return Settings.ai.anthropicEndpoint
-                                default: return ""
-                            }
-                        }
-                        onTextChanged: {
-                            switch(Settings.ai.aiProvider) {
-                                case "openai": Settings.ai.openaiEndpoint = text; break
-                                case "anthropic": Settings.ai.anthropicEndpoint = text; break
-                            }
-                        }
-                    }
-
-                    Text {
-                        text: TranslationManager.translate("settings.ai.customEndpointHint", "Leave empty for default. Set to use an OpenAI/Anthropic-compatible API endpoint.")
-                        color: Theme.textSecondaryColor
-                        font.pixelSize: Theme.scaled(11)
-                        wrapMode: Text.Wrap
-                        Layout.fillWidth: true
-                    }
-                }
-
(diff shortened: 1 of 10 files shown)


skialpine:
### Code review

Found 4 issues:

1. Custom endpoint URL is missing trailing-slash normalization (unlike the Ollama precedent it's modeled on)

All 5 new URL-construction sites concatenate `m_baseUrl` directly with a fixed path (`/v1/chat/completions`, `/v1/responses`, `/v1/models`, `/v1/messages`), with no trailing-slash handling. `OllamaProvider` in the same file already normalizes this exact case (`if (!urlStr.endsWith("/")) urlStr += "/";`) before appending its path. A user entering `https://my-proxy.com/` (a common way to copy a proxy/gateway base URL) gets a double-slash path that many servers reject.

https://github.com/Kulitorum/Decenza/blob/56129c9aad774e84396996312b744af07b1d839c/src/ai/aiprovider.cpp#L198-L204

(same pattern repeats in `OpenAIProvider::sendResponsesRequest`, `OpenAIProvider::testConnection`, `AnthropicProvider::sendRequest`, `AnthropicProvider::testConnection`)

2. New endpoint settings aren't wired into the ShotServer web settings page (CLAUDE.md says ShotServer pages must "aim for feature parity: every field and action the app offers on that screen should be reachable from the web" and "Keep the two surfaces in sync... update the counterpart in the same change")

`openaiEndpoint`/`anthropicEndpoint` are added to `SettingsAI` and exposed in the in-app QML, but `shotserver_settings.cpp`'s GET serializer and POST handler (and the web HTML form) still only round-trip `ollamaEndpoint` — the new fields have no web equivalent, so a user configuring a custom endpoint from the ShotServer web UI has no way to do so.

https://github.com/Kulitorum/Decenza/blob/56129c9aad774e84396996312b744af07b1d839c/src/network/shotserver_settings.cpp#L1379-L1386

(POST handler has the matching gap at [L103-L110](https://github.com/Kulitorum/Decenza/blob/56129c9aad774e84396996312b744af07b1d839c/src/network/shotserver_settings.cpp#L103-L110))

3. New endpoint settings are dropped by backup/restore (device migration)

`settingsserializer.cpp`'s export/import round-trips `ollamaEndpoint` but not the new `openaiEndpoint`/`anthropicEndpoint` fields. A user migrating from device A to device B (per `docs/CLAUDE_MD/DATA_MIGRATION.md`) would silently lose a configured custom endpoint, with the restore still reporting success.

https://github.com/Kulitorum/Decenza/blob/56129c9aad774e84396996312b744af07b1d839c/src/core/settingsserializer.cpp#L297-L304

(import side has the matching gap around [L820](https://github.com/Kulitorum/Decenza/blob/56129c9aad774e84396996312b744af07b1d839c/src/core/settingsserializer.cpp#L817-L823))

4. New endpoint settings aren't exposed via MCP settings tools

`mcptools_settings.cpp` (settings_get) and `mcptools_write.cpp` (settings_set) expose `ollamaEndpoint` but not the new `openaiEndpoint`/`anthropicEndpoint` fields, so they can't be read or set through the MCP surface the way the equivalent Ollama setting can.

https://github.com/Kulitorum/Decenza/blob/56129c9aad774e84396996312b744af07b1d839c/src/mcp/mcptools_settings.cpp#L188-L193

---

5. Please move the custom endpoint field behind an "Advanced" button/dialog next to Test Connection, rather than showing it inline whenever OpenAI or Anthropic is selected. Since Anthropic is the in-app recommended provider, most users who pick either of these two providers will see this field even though the large majority never need it — an Advanced disclosure would keep the common path clean without losing the feature for people who want it.

https://github.com/Kulitorum/Decenza/blob/56129c9aad774e84396996312b744af07b1d839c/qml/pages/settings/SettingsAITab.qml#L492-L501

---

<details>

- [Worth a look] `setBaseUrl()` lacks the `// empty → keeps default` comment that the adjacent `setModel()` calls carry for the same fallback behavior, in `src/ai/aiprovider.h` / `src/ai/aimanager.cpp` (score: 50)

</details>

skialpine:
Thanks for your contribution. Happy to include it in the app if you address the review items above.

csader:
Thanks for the thorough review! All 5 items addressed in e2e61ff7:

1. **Trailing-slash normalization** — `setBaseUrl()` now strips a trailing `/` at set-time (matching the Ollama pattern), so all 5 URL-construction sites are safe.
2. **ShotServer web settings** — GET serializer, POST handler (save + test), HTML form inputs, and JS load all round-trip `openaiEndpoint`/`anthropicEndpoint`.
3. **Backup/restore** — both fields added to export and import in `settingsserializer.cpp`.
4. **MCP tools** — `settings_get` and `settings_set` both expose the new fields with descriptions.
5. **Advanced dialog** — endpoint field moved behind an "Advanced" button next to Test Connection; opens a modal dialog with save/cancel.

Also added the `// empty → keeps default` comment on both `setBaseUrl()` declarations (the below-threshold item).

PR #1618, opened 2026-07-23
Scope R2 to review page and drop the DE1 AuthorizationError dialog

skialpine:
## Why

Two BLE surfaces produce noise that helps no one, both observed on real hardware:

- **DE1 `AuthorizationError` modal** — an idle DE1 whose encrypted link is torn down under BLE contention (dual-HIGH signature; the scale is left at HIGH priority in observe mode *by design*) raised a modal **"Connection Error / Authorization error"** every 10–25 min. On these PIN-less devices `AuthorizationError` is never a user-actionable pairing failure, and the link auto-reconnects seconds later — there is nothing for the user to do.
- **R2 refractometer connecting everywhere** — the R2 is only used to capture TDS/EY on the post-shot review page, but an app-wide reconnect ladder pursued it *forever, off the page*: endless failed connect attempts (~once/60s for 30+ min in the log) contending with the DE1/scale links and spamming the log.

## What changed

- **DE1 dialog** — `BleTransport::onControllerError` no longer emits the `errorOccurred` dialog for `AuthorizationError`; the WARN log and `de1LinkFault` wedge signal are unchanged.
- **R2 scoping** — refractometer auto-reconnect is gated on the review-page "hunt" at the single chokepoint `tryDirectConnectToRefractometer()`; the app-wide reconnect tick self-stops off-page; the hunt re-kicks its scan chain on a mid-page drop; leaving the page disconnects the R2.
- **Scale reconnect is deliberately untouched** — the scale is needed everywhere and keeps its own separate always-on path. Gating at the R2 chokepoint specifically avoids editing the scale-adjacent app-resume/screensaver lambdas.

## Files

- `src/ble/bletransport.cpp` — suppress `AuthorizationError` dialog
- `src/ble/blemanager.{h,cpp}` — `isRefractometerHunt()`, chokepoint gate, mid-page-drop re-kick
- `src/main.cpp` — reconnect tick self-stops off-page (scale lambdas untouched)
- `qml/pages/PostShotReviewPage.qml` — disconnect R2 on deactivation
- `openspec/changes/quiet-unhelpful-ble-noise/` — proposal, design, specs (deltas to `ble-error-surfacing` + `refractometer-review-page-discovery`), tasks

## Testing

Full suite via Qt Creator: **92/92 passing, 0 warnings** on the sanitizer-instrumented debug build (scale tests included: `tst_scaleblepriority`, `tst_scalefeedliveness`, `tst_wifiscalediscovery`, `tst_difluidr2`).


skialpine:
### Code review

Found 5 issues (all now addressed in 83de0056):

1. Hunt can't recover from a mid-page scan error (functional). `onScanFinished` chains the hunt back-to-back, but `onScanError` deliberately does not re-chain — it relied on the always-on background reconnect tick, which this PR makes self-stop off-page and does not re-arm on hunt activation. Opening the review page for an R2 that never connected this session (so no disconnect transition armed the tick) leaves the hunt dependent solely on the scan-finished chain: a transient scan error breaks it with nothing to recover until the page is reopened.

https://github.com/Kulitorum/Decenza/blob/7d88ad6fd407e2346343a0e39f6e82bed0749a53/src/ble/blemanager.cpp#L1235-L1246

2. Debug diagnostic now backwards. Turning the hunt OFF logs "background reconnect cadence resumes", but this PR makes turning the hunt off *stop* the R2 reconnect. This string ships to the debug log attached to bug reports, so it misleads triage.

https://github.com/Kulitorum/Decenza/blob/7d88ad6fd407e2346343a0e39f6e82bed0749a53/src/ble/blemanager.cpp#L1853-L1857

3. Stale timer-declaration comment: "it simply keeps trying whenever it is disconnected and an address is saved" — no longer true; it only tries while the review-page hunt is active.

https://github.com/Kulitorum/Decenza/blob/7d88ad6fd407e2346343a0e39f6e82bed0749a53/src/main.cpp#L1944-L1950

4. Startup-arm comment documents dead code: it claims the armed timer picks up an R2 powered on later, but at startup the hunt is off, so `tryDirectConnectToRefractometer()` no-ops and the armed tick self-stops on first fire. Removed the dead arm.

https://github.com/Kulitorum/Decenza/blob/7d88ad6fd407e2346343a0e39f6e82bed0749a53/src/main.cpp#L2913-L2925

5. App-resume / screensaver-exit comments overstate off-page behavior ("(re)start its persistent reconnect sequence (mirrors the scale path)"): off the review page, arming the R2 tick is inert (fires once, self-stops), so it does not mirror the scale's genuine retry.

https://github.com/Kulitorum/Decenza/blob/7d88ad6fd407e2346343a0e39f6e82bed0749a53/src/main.cpp#L3722-L3731

<details>

- [Worth a look] `bletransport.cpp` contention-family comment lists only `AuthorizationError`/`ConnectionError` but the `if` below checks `RemoteHostClosedError` too — fixed in 83de0056 (score: 50)

https://github.com/Kulitorum/Decenza/blob/7d88ad6fd407e2346343a0e39f6e82bed0749a53/src/ble/bletransport.cpp#L486-L496

</details>

Not changed (intended per design.md, score 35): the R2 no longer auto-connects off the review page, so the R2 status dot / Read-TDS control on the Settings Connections tab won't auto-reconnect there — an acknowledged Option A trade-off worth a look if that surface should stay live.

PR #1621, opened 2026-07-23
Fix TypeError when reordering shot-plan chips by drag

skialpine:
## Problem

Dragging a shot-plan chip to reorder it logged:

```
ScreensaverEditorPopup.qml:573: TypeError: Value is undefined and could not be converted to an object
```

## Cause

The drag `onReleased` handler called `popup.planMoveItem(...)` **before** resetting `_planDragging`. `planMoveItem()` reassigns `shotPlanItems`, which is the model behind the `DelegateModel` (`planVisualModel`). Reassigning it synchronously rebuilds every chip delegate — including the one that owns this `MouseArea`. With the delegate torn down mid-handler, the trailing `popup._planDragging = false` could no longer resolve the `popup` id through the dead delegate context, throwing the TypeError.

## Fix

Read the drag state out of the delegate context and reset the flag first, so the delegate-destroying `planMoveItem()` call is the handler's **last** statement — nothing touches `popup` or the delegate after the model mutation.

## Testing

QML change, verified manually (no QML test harness). Dragging a shot-plan chip to reorder no longer logs the TypeError.

### DIFF
--- a/qml/components/layout/ScreensaverEditorPopup.qml
+++ b/qml/components/layout/ScreensaverEditorPopup.qml
@@ -565,13 +565,18 @@ Dialog {
                                                 if (drag.active) popup._planDragging = true
                                             }
                                             onReleased: {
-                                                if (popup._planDragging) {
-                                                    var endIndex = planChip.liveIndex
-                                                    if (_startIndex >= 0 && endIndex !== _startIndex)
-                                                        popup.planMoveItem(_startIndex, endIndex)
-                                                }
+                                                // Read state out of the delegate context and reset
+                                                // BEFORE the move: planMoveItem() reassigns
+                                                // shotPlanItems, which rebuilds the DelegateModel and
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

The fix correctly reorders the `onReleased` handler so the delegate-destroying `planMoveItem()` call (which reassigns `shotPlanItems`, resetting the `DelegateModel` and tearing down the delegate) is the last statement, with all `popup`/delegate-scoped reads and resets performed first. Move behavior is preserved — `from`/`to` are captured at the same control-flow point as before, and the `to = _planDragging ? liveIndex : -1` ternary faithfully encodes the original "only commit if actually dragging" guard. The adjacent `onCanceled` handler correctly needs no change: it mutates only the `DelegateModel` items group (`planVisualModel.items.move`), which reorders delegates in place rather than destroying them.

PR #1624, opened 2026-07-23
Recover a briefly-unreachable WiFi scale without a manual rescan

skialpine:
A WiFi Half Decent Scale that dropped off the network for a few seconds stayed disconnected **until the user manually rescanned**. Three independent defects combined to make that permanent.

## 1. A transient transport error evicted a known-good cached IP

`DecentScaleWifi::onError` treated *any* error on a cached-IP attempt as proof the IP was wrong. Its reasoning was sound for the case it named — a router answering `ws://<ip>/snapshot` with HTTP 403 really does mean the address was reassigned — but it generalised from "some errors prove the IP is wrong" to "any error proves the IP is wrong".

So an unreachable scale got its correct cache entry thrown away, the hostname fallback dialled ~45 ms later (inside the same unreachability window), that failed identically, the single fallback was consumed, and the driver gave up.

Errors are now classified by **what they prove about the address**:

| Class | Errors | Behaviour |
|---|---|---|
| Nothing answered | `NetworkError`, `SocketTimeoutError`, `HostNotFoundError` | Retain cache, don't consume the fallback, defer the retry to the app-level loop |
| A peer answered | `ConnectionRefusedError`, handshake/protocol failures, recognition timeout | Unchanged — evict and fall back |

`ConnectionRefusedError` sits on the "answered" side deliberately: a TCP RST proves a host is up at that address and refused the port, which is exactly the DHCP-reassignment evidence the fallback exists to act on.

## 2. The recognition timer was armed after `open()`

`open()` fails **synchronously** for an address the OS rejects at the routing layer, so `onError`'s `stop()` ran against a timer that had not started yet. `attemptTarget` then armed it anyway, and `onRecognitionTimeout` fired 5 s later — evicting the very cache entry the error handler had just decided to keep.

Found from a device log, not from the tests: the first version of the new cache test waited 600 ms and passed with the bug live. It now waits past the recognition window.

## 3. The DE1 simulator flag blocked real scale connectivity

`switchScale` and `tryDirectConnectToScale` both early-returned on the flag set from `simulationMode` — "no DE1 attached" — which is a *different* setting from `simulatedScaleEnabled` ("Simulated Scale"). With the DE1 simulator on and the scale simulator off, real scales could not connect at all, and the reconnect loop that fix 1 defers to never ran.

Scale scanning was already exempt (`blemanager.cpp` states "suppress DE1 scanning but allow scale/refractometer scans"); these two sites were the outliers. They now gate on the scale simulator, and enabling DE1 simulation no longer tears down a connected physical scale.

This matters most for the WiFi scale, which reaches the scale over the network and shares no radio with the DE1 at all.

## Test fixture fix

`init()` unconditionally *required* a DISCONNECTED warning via `ignoreMessage`, but `setConnected()` only warns on a true→false transition — so any test whose driver never connects failed with "Not all expected messages were received", for a reason unrelated to what it checks. Replaced with an allow-not-require filter, declared **ahead of** the driver it covers so it outlives the driver's teardown warnings.

## Verification

- Full suite green: **93/93**, no warnings.
- Fixes 1–3 all confirmed firing correctly on real hardware from device logs.
- **End-to-end reconnect is not yet verified.** The dev machine turned out to have macOS silently dropping all LAN packets from the ad-hoc-signed debug build — a packet capture showed zero packets from the app's source port while concurrent shell connects to the same host:port succeeded in 30–155 ms. That is an environment issue, not a code one, but it means tasks 4.1–4.4 in `tasks.md` remain open.

## Notes for review

- `tasks.md` 5.5/5.6 (automated review, archive) and 7.5 (regression coverage for the gate split — there is no BLEManager test target today) are still open.
- `design.md` carries one open question: whether the mDNS re-resolve measurably helps or whether the backoff delay alone accounts for recovery. It's recorded as inference, not established fact, with a task to settle it.


skialpine:
### Code review

Found 4 issues:

1. The falling-edge re-arm does not exclude the simulator's synthetic `sim:` primary, so unchecking Simulated Scale starts the reconnect ladder against it. `tryDirectConnectToScale` only special-cases `wifi:`/`usb:`, so a `sim:` address falls through to the BLE path, builds an invalid `QBluetoothAddress("SIM:00:00:00:00:00:00")`, times out after 20 s, and emits `flowScaleFallback` (a "No Scale Found" dialog) plus `scaleRetryNeeded` — repeating on the 60 s ladder indefinitely. Pre-PR this was unreachable because `m_disabled` short-circuited that path. `isBluetoothAvailable()` returns true unconditionally in simulator mode, so that guard cannot stop it either. Debug builds only.

https://github.com/Kulitorum/Decenza/blob/8373d4194a783269cee15103637198e80ad93756/src/main.cpp#L2070-L2077

2. `dialCachedIpAfterResolveFailure()` is inert on the eviction path it most needs to cover. `onError`'s wrong-host branch evicts the cache (`m_ipCacheUpdate(m_hostname, QString())`) *before* handing off to `attemptHostname()`, so when resolution then fails the helper reads an already-empty cache and returns false — the cycle dials nothing, and so does every cycle after it. Reachable with a power-cycled ESP32 that refuses the port (`ConnectionRefusedError`, classified non-transient) while mDNS is deaf under BLE-coexistence load.

https://github.com/Kulitorum/Decenza/blob/8373d4194a783269cee15103637198e80ad93756/src/ble/scales/decentscalewifi.cpp#L988-L994

3. The header comment on `m_retryShouldReresolve` contradicts the code in the same PR. It claims the flag makes the next `connectToHost()` bypass "both the preferredIp and cached-IP shortcuts", but `preferredIp` is checked first and wins outright — the flag is only cleared as a side effect and is never consulted when `preferredIp` is present. The adjacent `.cpp` comment states the opposite ("preferredIp OUTRANKS m_retryShouldReresolve"). A maintainer working from the header would believe a set flag suppresses a caller-supplied `preferredIp`, which is the behaviour the change deliberately removed.

https://github.com/Kulitorum/Decenza/blob/8373d4194a783269cee15103637198e80ad93756/src/ble/scales/decentscalewifi.h#L219-L224

4. `attemptHostname()` sets `m_triedHostnameFallback = true` unconditionally at its top, so the cached-IP dial started by `dialCachedIpAfterResolveFailure()` runs with that flag already latched. Both eviction gates then evaluate false — `onError`'s `cachedIpAttempt` and `onRecognitionTimeout`'s fallback branch. With the re-resolve flag set and resolution persistently failing, a cached IP that DHCP reassigned to a live host (a router answering 403) is dialled every reconnect cycle and can never be evicted. Self-heals once resolution recovers; until then it repeatedly hits an unrelated LAN host.

https://github.com/Kulitorum/Decenza/blob/8373d4194a783269cee15103637198e80ad93756/src/ble/scales/decentscalewifi.cpp#L246-L250

---

<details>

- [Worth a look] Changing the pre-shot check to `isScaleSimulated()` aborts a stop-at-weight shot at preheat in debug simulator builds with Simulated Scale off, because the synthetic `sim:` entry makes `hasSavedScale()` true while no real scale is connected — same `sim:` blind spot as issue 1, and a documented intentional tradeoff (score: 72)
- [Worth a look] The `m_userInitiatedShutdown` enumeration reads as unconditional for both `attemptHostname` resolution-failure branches, but the new early return means the flag is now set only when there is no cached IP to fall back on; the comment's own preamble claims exhaustive accuracy is its whole value (score: 60)

</details>

PR #1625, opened 2026-07-23
Consolidate settings onto one app-named store

skialpine:
## Why

Decenza wrote **two** QSettings stores on every platform: an explicit `("DecentEspresso", "DE1Qt")` handle in the `Settings` domain objects, and a default-constructed handle everywhere else that resolved to a second, app-named store. Users saw both — two plists on macOS, two registry keys on Windows, two `.conf` files on Linux and Android — one of them named after an app Decenza has never shipped as.

Internally it was a standing footgun: a bare `QSettings` in new code landed in whichever store its author happened to construct. Four source files carried warning comments about exactly that, and it had already produced an isolated *third* store (`"Decenza"/"DE1"`) that needed migrating out of. The ~37 bare call sites also bypassed `DECENZA_TESTING` isolation, so tests wrote into the developer's real preferences.

## What changed

- **`AppSettings`** (`src/core/appsettings.h`) is now the only way to open the store. It names the canonical `("DecentEspresso", "Decenza")` identity in one place and applies the test redirect centrally, so no call site can bypass it. A subclass rather than a factory because `QSettings` is a `QObject` and non-copyable, and a shared instance would race the background-thread reads in `ShotHistoryStorage`.
- **A one-time migration** (`src/core/settingsstoremigration.cpp`) copies every legacy `DE1Qt` key across, verifies each write by reading it back, and only then destroys the legacy store — deferring without stamping its guard on any error, since it deletes its own source. It also removes the abandoned `("Decenza", "DE1")` store.
- **~71 call sites converted**; `factoryReset()` now clears both legacy identities.
- Reference docs updated (`SETTINGS.md`, `TESTING.md`, factory-reset design).

## Ordering is load-bearing

The migration must run before `Settings` and `AccessibilityManager` are constructed. `AccessibilityManager`'s own migration guard lived in the store being drained — construct it first and it sees an unstamped guard, re-runs its legacy migration, and overwrites the user's *current* accessibility settings with values from the abandoned third store. `factoryReset()` clears the legacy stores for the mirror-image reason: their guards live in the store the reset erases, so leaving them populated lets the next launch silently undo the reset.

## Downgrade is explicitly out of scope

A build older than this change finds no legacy store and starts from defaults. Accepted deliberately — it is what makes deleting the legacy store possible, and the deletion is the point: leaving it behind would still show two files.

## Verification

Suite **96/96 green**. Two new test targets: `tst_appsettings` (asserts the store-identity assumption the change rests on, and scans `src/` for handles built any other way) and `tst_settingsstoremigration` (7 cases over temp stores, including defer-on-error and the accessibility-guard ordering hazard).

The macOS upgrade path was then exercised on real data, from a restored pre-migration backup — and **found two defects that no test would have**:

1. **cfprefsd rewrote an empty `DE1Qt.plist` husk** after the in-migration unlink, so the user still saw two files. The unlink cannot win that race from inside the migration; `removeEmptyLegacyStoreFile()` now runs every launch, and by then the daemon has dropped the domain. No timer — the next launch is the event.
2. **`QSettings` merges the platform fallback search list** into `allKeys()`/`contains()`, so the first run reported `copied 44 of 118 legacy key(s)` for a store holding 44. The 74 extras were system preferences, skipped only because `contains()` consulted the same fallbacks on the canonical side. Both handles now call `setFallbacksEnabled(false)`, with a regression test.

Re-verified after both fixes: run 1 logged `copied 44 of 44` and left the husk; run 2 logged `removed leftover empty store file` and **one plist remains**, cfprefsd domain deregistered, all 75 keys and both migration guards intact.

## Still outstanding

Android CI test build (`QSettings` uses INI there, no cfprefsd) — not yet run.


skialpine:
### Code review

Found 7 issues. All are fixed in 4947611 — the links below point at the reviewed commit `ee59aefc` where each issue existed.

1. `removeEmptyLegacyStoreFile()` unlinks the legacy store whenever it enumerates no keys, without checking `status()`. An unreadable store (FormatError/AccessError) reports empty exactly like an absent one — and this runs unconditionally right after `migrateLegacySettingsStore()` may have returned `deferredOnError` *because* the store was unreadable and deliberately preserved it. The defer path keeps the data; the next statement deletes it. Entire settings namespace lost, guard unstamped, next launch migrates an absent store and stamps success.

https://github.com/Kulitorum/Decenza/blob/ee59aefcf453f717d97163cf959308120f5ca90f/src/core/settingsstoremigration.cpp#L42-L49

2. `factoryReset()` clears the canonical store and two legacy identities, but not `("DecentEspresso", "Decenza DE1")` — the pre-rename store read by `migrateDefaultQSettingsFromOldAppName()`. This PR moved the app-name migration's guard into the canonical store, so a reset erases the guard and the next launch copies that store's entire contents back into the now-empty canonical store. A "wipe all data" that restores the user's pre-rename preferences.

https://github.com/Kulitorum/Decenza/blob/ee59aefcf453f717d97163cf959308120f5ca90f/src/core/settings.cpp#L763-L769

3. `freshInstall` is now always false. The migrations stamp a guard into the canonical store before `Settings` is constructed — including on a brand-new install, where there is nothing to migrate (asserted by this PR's own `emptyLegacyStoreStampsTheGuardWithoutCopying`). Every install looks like an upgrade, so new users get seeded the legacy two-tap steam default instead of the current single-tap one.

https://github.com/Kulitorum/Decenza/blob/ee59aefcf453f717d97163cf959308120f5ca90f/src/core/settings.cpp#L73-L75

4. Skip-if-present resolves `localization/language` the wrong way. Its premise — "a key already in the canonical store was written by the running build" — fails for keys the pre-consolidation code wrote to *both* stores from different places. `TranslationManager` read this one through `Settings` (legacy store) while backup-restore and device-to-device wrote it to the default (now canonical) store. A user who set a language and later restored a backup has the live value in the legacy store and a stale one in canonical, so their UI language changes on upgrade.

https://github.com/Kulitorum/Decenza/blob/ee59aefcf453f717d97163cf959308120f5ca90f/src/core/settingsstoremigration.cpp#L90-L95

5. Duplicate same-store handle with a now-false comment in the restore path. `AppSettings accessStore` sits alongside the in-scope `AppSettings settings` under a comment claiming writing to `settings` "would be a silent no-op" — true before consolidation, false after. The sibling read path `buildExtraSettingsObject()` was collapsed in this same diff, and tasks.md 5.4 claims this cleanup as done. (CLAUDE.md says "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'.")

https://github.com/Kulitorum/Decenza/blob/ee59aefcf453f717d97163cf959308120f5ca90f/src/network/shotserver_backup.cpp#L1074-L1081

6. `AccessibilityManager::migrateLegacyStore()`'s comment says the legacy store "is intentionally left intact ... factoryReset() wipes it explicitly". This PR's `runSettingsStoreMigrationOnce()` now destroys it on an ordinary launch, so the file whose whole job is that migration documents the opposite of what happens. (Same CLAUDE.md rule; the file is modified by this PR.)

https://github.com/Kulitorum/Decenza/blob/ee59aefcf453f717d97163cf959308120f5ca90f/src/core/accessibilitymanager.cpp#L147-L151

7. design.md D3 — a section titled "Ordering is load-bearing" — and the completed tasks.md 4.2 both state the app-name migration runs first. The shipped code runs the store migration first, with an inline comment explaining why that order is required. The code is right; two authoritative artifacts in the same PR describe the reverse.

https://github.com/Kulitorum/Decenza/blob/ee59aefcf453f717d97163cf959308120f5ca90f/openspec/changes/consolidate-qsettings-stores/design.md#L93-L96

<details>

- [Worth a look] The abandoned `("Decenza","DE1")` cleanup sits behind the `if (r.alreadyDone) return;` early return, so it can only fire on the migrating launch — stranding the store forever for anyone whose accessibility guard is stamped later in that same launch (upgrades from builds predating that migration). Narrow cohort, cosmetic impact. Fixed anyway: the cleanups now run every launch and are idempotent. (score: 65)
- [OK to skip] `runAppNameMigrationOnce()` now depends on the store migration having carried its guard across, so a `deferredOnError` causes it to redundantly re-run. Both paths are idempotent (skip-if-present copy, `mergeDirectoryContents` fast-exits), so this is harmless churn on an error path. (score: 50)

</details>

Verified on macOS against real data before and after the fixes: `copied 44 of 118` (the fallback-search defect, since fixed) → `copied 44 of 44` → `removed leftover empty store file`, ending with a single plist, 75 keys, and both migration guards intact. Suite is 96/96 green. Android CI build still outstanding.

PR #1636, opened 2026-07-25
Fetch model list from custom endpoints (#1598 follow-up)

csader:
## Summary

Follow-up to #1598 (custom endpoint URLs). When a custom endpoint is configured, the hardcoded model catalogs (e.g. `claude-sonnet-4-6`, `gpt-5.4`) might not match what the endpoint serves. This adds dynamic model fetching via `/v1/models`, replicating the existing Ollama pattern.

- OpenAI/Anthropic providers gain `refreshModels()` that GETs `/v1/models` and emits the model list
- `setModel()` skips catalog validation when a custom endpoint is set (accepts any model string)
- AIManager exposes `customEndpointModels` Q_PROPERTY, auto-refreshes on endpoint change
- QML shows a ComboBox + Refresh button (like the Ollama picker) when a custom endpoint is active

### DIFF
--- a/qml/pages/settings/SettingsAITab.qml
+++ b/qml/pages/settings/SettingsAITab.qml
@@ -246,6 +246,7 @@ KeyboardAwareContainer {
                 // Model selection (providers exposing a fixed catalog of >1 model).
                 // Generic: any provider whose availableModels() returns multiple
                 // entries lights this up automatically — no per-provider wiring.
+                // Hidden when a custom endpoint is set (dynamic picker shown instead).
                 ColumnLayout {
                     id: modelSelect
                     // Re-evaluated when the active provider changes (the binding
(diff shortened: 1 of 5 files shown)


skialpine:
Chris I am not sure how well this will work. For each model change I have had to change the prompt. Things like caching and level level of thinking have had to change each time to either make it work or get good results. 

On a related note have you tried the MCP. On my Android tablet I mostly am using Claude desktop on Android using the discus button. Not sure if you have a chat app running on your tablet that works with an MCP, but the MCP is a lot more powerful then the AI assistant.

csader:
Fair enough, though without this change or something else I don't think my custom endpoint setting I built will work for me or anyone else. The endpoint needs to have the model call structured specifically, so the models list you've built in isn't compatible. Happy to try another approach if you have any ideas.

skialpine:
can you say more about what models you have available? I thought you had a proxy, so not sure why on your side you could not make the Decenza available ones or at least one of them available.

Not being able to tune the prompt is why open router and ollama do not work very well and why I push people back to the big 3 when they complain as those are the ones I spent time and money to test.

Happy to work on a solution with you, still not sure I fully understand your needs.

skialpine:
Did not hear back from you for more then a month so closing. feel free to ask me to reopen.

PR #1639, opened 2026-07-25
Reconcile built-in profiles with reaprime, add 7 shared profiles, upgrade stored encoding on load

skialpine:
## What this found

The audit this change called for has been run, and it inverted much of the plan.

Measured against `tadelv/reaprime@25bf9f25`: of the **63 profile titles the two apps shared, 52 were already machine-equivalent** and 11 diverged — and **every one of the 11 resolved in Decenza's favour**. Nothing in `resources/profiles/` needed correcting.

The proposal's "~50 have real value differences" was an artefact of comparing encodings. Three conventions accounted for it and none changes the shot: 402 rows where we omit a zero `weight`, 231 where we emit a no-op zero `limiter`, ~27 where reaprime writes `""` on the axis the pump doesn't drive.

The 11 traced to two upstream mechanisms, both de1app's:

- **`save_profile` writes `advanced_shot` out of the global `::settings` array**, so a `settings_2a`/`2b` file ships frames belonging to whichever profile was last loaded. reaprime's `Default` ran frames at **75 °C and 54 °C** against a declared `espresso_temperature` of 90.0.
- **de1app issue [#350](https://github.com/decentespresso/de1app/issues/350)** shadows the A-Flow profiles with a stale 6-frame snapshot.

Both are fixed on reaprime's side now. Re-running the comparison returns **71/71 equivalent, zero reaprime-only titles**.

The check on that conclusion is the two `settings_2c` cases — advanced profiles involve no frame regeneration, so Decenza has no mechanism there to be accidentally right. It reproduces de1app's frames exactly on both.

## What was cut

No built-in needed fixing, so the planned 3-way `profile_sync` leg and the reaprime regression corpus were **cut rather than built**. The comparison is recorded in `audit.md` with the script that reproduces it, and our own drift is already gated by `tests/data/profiles_legacy/`. A fixture asserting "these 11 diverge" would have gone red the moment reaprime fixed them.

## What this does

**Seven profiles reaprime shipped and we didn't are now built-ins** — the four Baseline contact levels, `I Can't Believe It's Not Filter`, `PSPH` and `Soup 58`. Each gained a dialing knowledge-base entry (a gate, not optional metadata). PSPH and Soup 58 arrived with notes containing only their own title; both now describe what the profile does, read from its frames. Soup 58's 10 °C final frame is **explained rather than "fixed"** — the DE1 cannot chill, so a target below tank temperature means no heating, which is what the 58 refers to.

**User profiles are upgraded to the canonical encoding when loaded, not in a one-time pass.** A pass that completes cannot be correct: profiles arrive in the folder at any time — sideloaded, restored from a backup, synced — and those are the copies most likely to be legacy-encoded. Writes are parity-gated against the bytes actually on disk and skipped with a warning when conversion would drift a value. Built-ins are never written. This matters because `DatabaseBackupManager` copies the directory verbatim, so a legacy-encoded file travels into a backup and onto another device where a stricter reader rejects it.

**`Profile::saveToFile` is now atomic** (`QSaveFile`). It truncated in place — survivable for a git-tracked built-in, not for rewriting a user's profile purely to reformat it.

**`profile_sync` reports a `SOURCE CONFLICT`** when a plugin and base `.tcl` disagree, naming both paths and the divergence. It previously applied precedence silently, which is why our own tooling held both A-Flow copies and could not see #350. Non-gating: the precedence is a decision, not a guess, and #350 resolving either way removes the conflict.

## Corrections to the record

- **`isAdvancedType()` treating `settings_2c2` as advanced is correct.** I flagged it as a divergence from de1app's stop-target switches; it isn't. `fix_profile_type` rewrites `2c2` to `2c` on load and save, so those switches never see it. Recorded as a comment so the misreading isn't repeated.
- **The float-noise argument in #350 doesn't hold.** de1app's own repo profiles carry the same noise (`Cleaning_forward_flush_x5.tcl` ships `pressure 9.999999999999993`). The `water_temperature` argument still stands. Recorded in `audit.md`.

## Verification

- Full suite via Qt Creator: **98 passed, 0 failed, 0 skipped**, `tests_with_warnings` empty.
- Build clean, 0 warnings.
- `profile_sync --rewrite-format` touches **only** the 7 new files — the existing 93 were already byte-canonical.
- New tests in `tst_profilemanager`: upgrade on load, no rewrite when already canonical, built-in never written, lossy fixture left byte-for-byte with an asserted warning, backup round-trip stays reaprime-readable.


skialpine:
### Code review

Found 9 issues. The first three are one story: the new parity gate does not actually guarantee what its comment claims, because two other write paths reach the same files ungated and the gate itself is one-directional.

1. `jsonParityErrors` cannot detect an **added** key, so `upgradeStoredEncoding` silently changes content while claiming to change only encoding. `collectParityErrors` iterates keys in `before` only, so a key absent on disk but present in the canonical output is never visited. `Profile::fromJson` derives a missing `espresso_temperature` from the frames — the documented Visualizer-import case — so for that population the gate returns empty and the derived value is written under the "upgraded stored encoding" label. That directly contradicts the new comment: "This pass converts ENCODING and must never change content".

https://github.com/Kulitorum/Decenza/blob/0398bb1054bda6003c57daeede754b3b24057ff5/src/profile/profile.cpp#L770-L775

2. The pre-existing `espresso_temperature` repair rewrites the same file with **no parity check at all**, about 25 lines after the gate refuses — and it runs *after* the notes backfill, so it writes the built-in's notes into the user's file. That is precisely what the new comment above the upgrade says must never happen ("would put text into the user's file that the file never had"). When `espresso_temperature` was absent, both writers fire and the file is rewritten twice per activation.

https://github.com/Kulitorum/Decenza/blob/0398bb1054bda6003c57daeede754b3b24057ff5/src/controllers/profilemanager.cpp#L1459-L1472

3. `migrateProfileFormat()` runs from the constructor, before any `loadProfile()`, and unconditionally re-serializes every profile lacking a `"version"` key with only a title/steps sanity check — no `jsonParityErrors`, no `reaprimeReadabilityErrors`. It is gated by a settings flag, so on a fresh install or a restore that does not carry the flag, the unaudited rewrite reaches the legacy population first — the same population the new guarded path was added to protect.

https://github.com/Kulitorum/Decenza/blob/0398bb1054bda6003c57daeede754b3b24057ff5/src/controllers/profilemanager.cpp#L290-L294

4. `upgradeStoredEncoding` does a synchronous file read and a synchronous write inline in `loadProfile`, which is a `public slots:` method called directly from QML (`BrewDialog.qml`, `CustomItem.qml`, `EspressoItem.qml`) and therefore runs on the main thread. CLAUDE.md says "**Never run database or disk I/O on the main thread.** Use `QThread::create()` with a `QMetaObject::invokeMethod(..., Qt::QueuedConnection)` callback". A justification exists in the archived design doc but not in the code, and the PR adds a second instance of the pattern rather than addressing the existing one — CLAUDE.md also says "Fix pre-existing violations in any file you touch".

https://github.com/Kulitorum/Decenza/blob/0398bb1054bda6003c57daeede754b3b24057ff5/src/controllers/profilemanager.cpp#L1284-L1292

5. The storage branch can read one file and write another. `ProfileStorage::readProfile` tries external then falls back to internal; `writeProfile` tries external **first**, creating the directory if needed. A profile living only in the internal fallback is therefore read from internal and written to the user's shared `de1plus/profiles` folder — a file that was never there — while the copy actually examined stays legacy-encoded. Conversely, if the external write fails, `writeProfile` falls back to internal and returns `true`, so the upgrade logs success while the file it read is untouched, and repeats on every activation. Android-specific (`externalProfilesPath()` is empty on desktop).

https://github.com/Kulitorum/Decenza/blob/0398bb1054bda6003c57daeede754b3b24057ff5/src/core/profilestorage.cpp#L253-L270

6. `upgradeStoredEncoding` rewrites a profile without checking `isReadOnly()`, while the adjacent repair in the same function does check it ("Skip read-only/built-in profiles"). Built-ins are excluded via the `Origin` enum, so this concerns a `read_only: 1` profile sitting in a writable tier.

https://github.com/Kulitorum/Decenza/blob/0398bb1054bda6003c57daeede754b3b24057ff5/src/controllers/profilemanager.cpp#L1434-L1440

7. The test comment states the wrong precision. `target_weight` serializes with `ProfileJson::TargetMass`, which is **1** decimal, so `36.1234` becomes `36.1` and the drift is `0.0234` — not "two decimals", "36.12" and "0.0034". The test passes either way, but the comment misdocuments the exact mechanism it exists to explain.

https://github.com/Kulitorum/Decenza/blob/0398bb1054bda6003c57daeede754b3b24057ff5/tests/tst_profilemanager.cpp#L2348-L2353

8. The `jsonParityErrors` doc comment says it is "Shared by profile_sync's rewrite audit and the built-in parity tests". This PR adds a third caller, `ProfileManager::upgradeStoredEncoding`, which is the most safety-critical of the three since it runs against real user profiles.

https://github.com/Kulitorum/Decenza/blob/0398bb1054bda6003c57daeede754b3b24057ff5/src/profile/profile.h#L262-L266

9. The fixture comment attributes the avoided heal to the wrong mechanism: it says the value is "consistent with the frame", but the heal fires only when the key is absent, or equals the bare `93.0` default and falls outside the frame range. `92.0` avoids it because it is not `93.0`, not because it matches the frame.

https://github.com/Kulitorum/Decenza/blob/0398bb1054bda6003c57daeede754b3b24057ff5/tests/tst_profilemanager.cpp#L2272-L2277

<details>

- [Worth a look] `Profile::saveToFile` uses `QSaveFile` without `setDirectWriteFallback(true)`; Qt's docs recommend it "to save documents edited by the user", and `open()` fails outright when the parent directory disallows creating files even if the target itself is writable. All current call sites are app-owned directories and the failure mode is a refused write rather than data loss (score: 50)
- [Worth a look] The new `settings_2c2` comment says `fix_profile_type` rewrites the type "on both load and save"; de1app's save path uses its own inline check rather than that function, so the outcome is right but the mechanism attribution is not (score: 50)
- [OK to skip] The `SOURCE CONFLICT` report hard-labels the losing side "B = base (discarded)", which would mislabel a collision between two plugin directories; only one plugin directory contributes colliding profiles today (score: 50)

</details>

PR #1647, opened 2026-07-26
Ship the simulator in non-debug builds (fixes nightly + the next release build)

skialpine:
## What broke

Nightly sanitizers [#536](https://github.com/Kulitorum/Decenza/actions/runs/30192207389) failed at **Build**, not on a sanitizer finding. Both the asan and ubsan jobs died identically:

```
tests/tst_simulatedscale.cpp:80:20: error: 'class DE1Device' has no member named
  'setSimulator'; did you mean 'setSimulationMode'?
```

`DE1Device::setSimulator()` was `#ifdef QT_DEBUG`. The nightly configures `-DCMAKE_BUILD_TYPE=RelWithDebInfo`, so `NDEBUG` is set, `QT_DEBUG` is not, and the member does not exist. Local Debug builds define it, which is why the pre-PR suite passed.

**The same PR broke release builds of the app itself.** #1629 also added an unguarded `de1Device.setSimulator(nullptr)` teardown call in `main.cpp`, outside the `#ifdef QT_DEBUG` block that ends 640 lines earlier. Every release workflow builds `-DCMAKE_BUILD_TYPE=Release`; the last release build ran 2026-07-24, before #1629 landed, so the next tag push would have failed to compile.

## Why the guard was wrong in the first place

Simulation Mode is not a debug affordance — it is a shipped user feature. Only its wiring was debug-only:

| Piece | In a release build? |
|---|---|
| `DE1Simulator` + `SimulatedScale` compiled and linked | **Yes** — unconditional, `CMakeLists.txt:767` |
| `developer/simulationMode` setting | **Yes** — `QT_DEBUG` only changes the *default* |
| "Simulation Mode" card in Settings → Machine | **Yes** — no guard, translated title/description, "Restart required" note |
| Ctrl+D shortcut | **Yes** — no guard |
| SIMULATION MODE badge, `canStartOperations`, status "Simulation" | **Yes** |
| `setSimulator` / `m_simulator` / every sim branch in `DE1Device` | **No** |
| `main.cpp` block that constructs the simulator | **No** |

So a user on a shipped build could enable Simulation Mode, be told to restart, restart, see the SIMULATION MODE badge and enabled Espresso/Steam/Flush buttons — and have every button do nothing, because `requestState()`'s simulator branch was compiled away and it fell through to `if (!m_transport) return;`.

## The fix

Remove the guards so the code that ships matches the UI that ships:

- **`de1device.h` / `de1device.cpp`** — the `DE1Simulator` forward declaration, the include, `setSimulator()`, `m_simulator`, and the five `m_simulationMode && m_simulator` branches (`requestState`, `stopOperationUrgent`, `goToSleep`, `uploadProfile`, `setShotSettings`). Every one is already runtime-gated on a non-null `m_simulator`, so they stay inert until `main.cpp` wires one up. Also drops a comment that claimed "Release builds have no simulator at all".
- **`main.cpp`** — the simulator includes and the construction/wiring block, which is already inside `if (settings.app()->simulationMode())`. The GHC simulator window (a desktop `QWidget` dev window) stays `(Q_OS_WIN || Q_OS_MACOS) && QT_DEBUG`.
- **`main.cpp`** — the `BLEManager` gate that disables DE1 BLE in simulation mode. Without it a release user in Simulation Mode would keep scanning for a real machine.

No binary-size cost: the simulator was already being linked into every build.

## Testing

- Full local suite via Qt Creator: **101 passed, 0 failed, 0 skipped, no warnings**.
- Debug builds only prove the path that already worked, so the nightly-sanitizers workflow was dispatched against this branch to compile and run the exact RelWithDebInfo configuration that failed. Result linked in a comment below.

## Follow-up for the maintainer

Now that Simulation Mode actually functions for end users, the wiki manual may want an entry for it. Not done here — flagging per the CLAUDE.md rule rather than pushing wiki edits unasked.

### DIFF
--- a/src/ble/de1device.cpp
+++ b/src/ble/de1device.cpp
@@ -6,9 +6,7 @@
 #include "profile/profile.h"
 #include "../core/settings_hardware.h"
 
-#ifdef QT_DEBUG
 #include "../simulator/de1simulator.h"
-#endif
 #include <QBluetoothAddress>
 #include <QDateTime>
 #include <cmath>
@@ -867,7 +865,6 @@ void DE1Device::parseMMRResponse(const QByteArray& data) {
 // -- Machine control methods (delegate through transport) --
 
 void DE1Device::requestState(DE1::State state) {
-#ifdef QT_DEBUG
     if (m_simulationMode && m_simulator) {
         switch (state) {
         case DE1::State::Espresso:
@@ -909,7 +906,6 @@ void DE1Device::requestState(DE1::State state) {
         }
         return;
     }
-#endif
 
     if (!m_transport) return;
     if (dropDeviceWriteIfFirmwareFlash("requestState")) return;
@@ -1029,12 +1025,10 @@ void DE1Device::customEvent(QEvent* event) {
 }
 
 void DE1Device::stopOperationUrgent(qint64 sawTriggerMs) {
-#ifdef QT_DEBUG
     if (m_simulationMode && m_simulator) {
         m_simulator->stop();
         return;
     }
-#endif
     if (!m_transport) return;
     if (dropDeviceWriteIfFirmwareFlash("stopOperationUrgent")) return;
     clearCommandQueue();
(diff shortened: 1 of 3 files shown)


skialpine:
Verified against the exact configuration that failed: nightly-sanitizers dispatched on this branch — [run 30206701031](https://github.com/Kulitorum/Decenza/actions/runs/30206701031).

Both jobs green. `tst_simulatedscale` compiles and passes under RelWithDebInfo in each:

- **asan** — 8m27s, `100% tests passed, 0 tests failed out of 100`, `tst_simulatedscale ... Passed 16.57 sec`
- **ubsan** — 6m51s, same

Local Debug suite via Qt Creator: 101 passed, 0 failed, no warnings.

PR #1648, opened 2026-07-26
Drop the simulator from tablet release builds, UI included

skialpine:
Simulation Mode is a development aid on a tablet — the machine is right there — so tablet **release** builds now ship without it: no engine, no wiring, and no Settings card. Debug builds keep it on every platform, and desktop keeps it in every configuration (a laptop with no DE1 attached is a legitimate way to run the app).

The previous removal was half a removal. The engine was compiled out, but the fully translated "Simulation Mode" card, its Ctrl+D shortcut, and its settings-search entry all still shipped — so a user could switch on a mode with nothing behind it and no way to tell why nothing worked. This removes the feature, not just its implementation.

## How it works

`DECENZA_SIMULATOR` is the single source of truth.

- **`CMakeLists.txt`** defines it via a generator expression — `$<CONFIG:Debug>` on Android/iOS, always on desktop. It **must** be a genex: the iOS release build uses the Xcode generator, where `CMAKE_BUILD_TYPE` is empty at configure time, so a configure-time test reads as "not Debug" for every config including Debug. That is exactly what defeated the old conditional and led #487 to link the engine into every build.
- **`src/simulator/*.{h,cpp}`** guard their whole contents on it, so those translation units compile to nothing when it is off. The sources stay listed in the target, which keeps AUTOMOC's view identical across configurations.
- **`de1device.{h,cpp}`** guards only the five branches that *dereference* `m_simulator`. `setSimulator()` and the member need just the forward declaration and stay unguarded, so callers compile everywhere — guarding the seam itself is what let #1629's teardown call break the release build and the nightly while every local debug build stayed green.
- **`SettingsApp`** gains `simulatorAvailable` (CONSTANT), and `simulationMode()` returns false outright where there is no simulator. The stored value is deliberately **not** cleared: settings travel between devices, and a tablet must not erase the preference belonging to the desktop it came from.
- **QML** hides the Settings card (height collapsed so it leaves no gap), disables Ctrl+D, and filters the card out of settings search so no result scrolls to a card that is not there.
- **`tests/CMakeLists.txt`** defines `DECENZA_SIMULATOR` unconditionally, so the suite can never regress the way #1629 did.

## Two pre-existing platform bugs fixed

Both are the same mistake: `if(WIN32 OR APPLE)` — and `APPLE` is true for iOS.

- `ghcsimulator.cpp`, a desktop-only QWidget helper, has been compiled into every iPad build despite its own "desktop only" comment. It drives `DE1Simulator`'s signals, so with the simulator absent it would not even compile there.
- `GHCSimulatorWindow.qml` shipped in every IPA — 30 `QmlCacheGeneratedCode` symbols for a window iOS never loads, whose `GHCSimulator` context property does not exist there at all.

Both gates are now `WIN32 OR (APPLE AND NOT IOS)`. The **Mini GHC** layout widget is unrelated and stays — it is enabled for headless machines (`DE1Device.isHeadless`), not just simulation.

## Verification

Every configuration this touches was built, and the simulator-off builds were verified against the shipped binary rather than trusted because they were green.

| Run | Config | Simulator | Result |
|---|---|---|---|
| Local (Qt Creator) | macOS Debug | on | 101 passed, 0 failed, no warnings |
| [Android APK](https://github.com/Kulitorum/Decenza/actions/runs/30208390965) | Ninja, Release | **off** | green, 5m18s |
| [Nightly asan](https://github.com/Kulitorum/Decenza/actions/runs/30208363253) | Linux, RelWithDebInfo | on | green, 100/100 |
| [Nightly ubsan](https://github.com/Kulitorum/Decenza/actions/runs/30208363253) | Linux, RelWithDebInfo | on | green, 100/100 |
| [iOS App Store](https://github.com/Kulitorum/Decenza/actions/runs/30208999917) | **Xcode multi-config**, Release | **off** | green, 14m |

The iOS run is the one that proves the mechanism: it is the only place `$<CONFIG:Debug>` is evaluated on a multi-config generator.

Strings in the shipped binaries, with a control to show the grep works and the binary is from this branch:

| String | Android APK | iOS IPA |
|---|---|---|
| `DE1Simulator: State ->` | 0 | 0 |
| `DE1Simulator: Dose set to` | 0 | 0 |
| `Creating DE1 Simulator` | 0 | 0 |
| `GHCSimulatorWindow_qml` | — | 0 (was 30) |
| `Simulation mode:` (control, unguarded line in `main.cpp`) | 1 | 1 |

The iOS binary is 64 KB smaller than the first build of this branch, which is the QML removal alone.

### DIFF
--- a/.github/workflows/ios-release.yml
+++ b/.github/workflows/ios-release.yml
@@ -209,9 +209,18 @@ jobs:
         if: ${{ github.event_name == 'workflow_dispatch' }}
         run: |
           cd build/ios
+          # -configuration Release is load-bearing, not decoration. Without it
+          # `xcodebuild build` uses the scheme's Run action, which is Debug for a
+          # CMake-generated project, while the archive below uses the Archive
+          # action (Release). This check would then compile a different arm than
+          # the one that ships — and since DECENZA_SIMULATOR is on in Debug and
+          # off in Release on iOS, the simulator-absent configuration would first
+          # be compiled at a release tag, which is exactly the pre-tag surprise
+          # this step exists to prevent.
           xcodebuild build \
             -project Decenza.xcodeproj \
             -scheme Decenza \
+            -configuration Release \
             -destination "generic/platform=iOS" \
             CODE_SIGNING_ALLOWED=NO \
             CODE_SIGNING_REQUIRED=NO \
@@ -271,9 +280,13 @@ jobs:
           # command-line specifier applies to ALL targets and would clobber
           # the extension's own profile. Both targets carry their per-target
           # profile in-project (set by inject_widget_extension.rb).
+          # Release is already what the Archive action selects; stating it keeps
+          # the two xcodebuild calls above and here reading the same, so neither
+          # can drift onto a different configuration via a scheme change.
           xcodebuild archive \
             -project Decenza.xcodeproj \
             -scheme Decenza \
+            -configuration Release \
             -destination "generic/platform=iOS" \
             -archivePath $RUNNER_TEMP/Decenza.xcarchive \
             CODE_SIGN_STYLE=Manual \
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1342,8 +1342,11 @@ set(QML_FILES
     qml/pages/CommunityBrowserPage.qml
 )
 
-# Simulator QML - desktop only (GHC window uses QWidget)
-if(WIN32 OR APPLE)
+# Simulator QML - desktop only (GHC window uses QWidget).
+# `WIN32 OR APPLE` would include iOS, which shipped this window's compiled QML
+# in every IPA — 30 QmlCacheGeneratedCode symbols for a window iOS never loads,
+# and whose GHCSimulator context property does not exist there at all.
+if(WIN32 OR (APPLE AND NOT IOS))
     list(APPEND QML_FILES
         qml/simulator/GHCSimulatorWindow.qml
     )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -764,16 +764,20 @@ if(NOT IOS)
     )
 endif()
 
-# Simulator engine - always included for multi-config generators (Xcode, Visual Studio)
-# CMAKE_BUILD_TYPE is empty at configure time, so conditional inclusion won't work.
-# The simulator code is small and has no extra dependencies, so it's safe to always link.
+# Simulator engine - listed unconditionally so AUTOMOC sees the same files in
+# every configuration. Whether any of it actually compiles is decided by
+# DECENZA_SIMULATOR (see the target_compile_definitions block below), which the
+# files themselves guard on.
 list(APPEND SOURCES
     src/simulator/de1simulator.cpp
     src/simulator/simulatedscale.cpp
 )
 
-# GHC simulator window - desktop only (uses QWidget window)
-if(WIN32 OR APPLE)
+# GHC simulator window - desktop only (uses QWidget window).
+# NOT `if(WIN32 OR APPLE)`: APPLE is true for iOS too, so that spelling compiled
+# this into every iPad build despite the comment. It drives DE1Simulator's
+# signals, so with the simulator absent it would not even compile there.
+if(WIN32 OR (APPLE AND NOT IOS))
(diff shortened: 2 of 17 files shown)


skialpine:
## Review round applied (`6bff69cf`)

Four review agents ran over this branch. Each finding was verified independently before acting on it; two were wrong on their facts and are noted at the bottom.

### One real bug — three agents found it independently

`setSimulationMode()` compared against a getter this PR hard-wires to `false`, making it asymmetric on a build with no simulator:

- `set(true)` → `false != true` → **persisted `true`** while the getter kept reading `false`
- `set(false)` → `false != false` → **no write**, so the stored `true` could never be cleared

Reachable through MCP `settings_set`, which reported `updated: ["simulationMode"]` for a write with no effect while `settings_get` returned `false` in the same breath. The stored `true` was then unreachable through every API on that device, and would take effect the next time the same install ran a build that *has* the simulator — booting it into simulation with DE1 BLE disabled and no way back.

Fixed: the setter refuses to persist a value the build cannot honour, falls through on `false` so a stale `true` still clears, and both accessors now compare against a new `storedSimulationMode()` helper. That helper is not cosmetic — the stored default is not uniform (debug desktop starts in simulation), so comparing against a raw `value(…, false)` would have broken "switch it off" while the key is unset.

### The comment justifying the old behaviour was factually wrong

It claimed the stored value must be preserved because "settings travel between devices (device-to-device migration, restored backups)". They do not, for this key: `SettingsSerializer` never *exports* a `developer` object at all, and the only developer key on the import path is `translationUpload`. The scenario that actually matters is same-device build switching — one install running both a Debug and a Release build — which is precisely what the latch broke. Comment replaced with the reason that holds.

### Also fixed

- **MCP** — `settings_set` rejects `simulationMode` with an explicit error instead of reporting a no-op as success; `settings_get` exposes `simulatorAvailable`; the schema description says so.
- **`DE1Device::setSimulationMode()`** refuses `true` when no simulator is compiled in. No caller can reach it today, but it is a public `Q_PROPERTY WRITE` that fabricates a connected machine and compiles identically in both arms, so a future writer would have reopened the bug with no compile-time signal.
- **`simulatorAvailable`** is a plain const member now, matching `isDebugBuild()` on the same class. A `static constexpr` READ does work — one agent ran `moc` to confirm — but a property QML fails to resolve is falsy and fails *closed*, which would strip Simulation Mode from desktop builds with nothing logged. Not worth the deviation.
- **Comment corrections** — the GHC window is not a QWidget (`GHCSimulator` is a `QObject`; the window is a QtQuick `Window`); #1629 broke the nightly but the release break was **latent, never observed**, and required *both* a guarded seam and an unguarded teardown call; dropped a brittle symbol count; resolved a contradiction between the CMake and `de1device.h` accounts of what tablet builds contain; deduplicated one rationale that had been copied verbatim into four files.
- **Dropped a redundant `implicitHeight` ternary** — `ColumnLayout` already excludes invisible items — and documented the three-way `objectName` / `cardId` / search-filter coupling.
- **`-DDECENZA_SIMULATOR_OVERRIDE=0`** builds the simulator-absent shape on desktop. Without it *no routine CI job compiles that arm*: the nightly is Linux (always on), and the mobile workflows run only on tag push or manual dispatch.
- **`ios-release.yml` now passes `-configuration Release`.** The unsigned compile check specified no configuration, so it used the scheme's Run action (Debug) while the archive uses Archive (Release) — the check validated the arm *with* the simulator, and the shipping arm would first have been compiled at a release tag.

### Where the review was wrong

- The "30 QmlCacheGeneratedCode symbols" figure was called unmeasured. It came from `strings` on the actual IPA, not the `.aotstats` file the agent assumed; its counter-measurement was taken on a macOS Debug object file. The number is still brittle and not load-bearing, so it is gone — but not for the stated reason.
- A `static constexpr` Q_PROPERTY READ was suspected of being invalid. It is valid; the change was made for fail-closed safety and consistency, not correctness.

Full suite after the changes: **101 passed, 0 failed, 0 skipped, no warnings.**

The wiki manual described Simulation Mode with no platform qualifier, which would send an iPad user hunting for a card that is not there. That is fixed and committed in the wiki repo, held unpushed pending release.

Issue #1423, opened 2026-07-06
[Bug] message when starting decenza


fredphoesh:
### What happened?

error popup

### Steps to reproduce

When opening decenza from standby each morning, this error has shown for the past 10 or 15 versions of decenza. Scale is connected and working...

<img width="2000" height="1200" alt="Image" src="https://github.com/user-attachments/assets/1edb2157-dca9-415c-be2d-66fd5304861d" />

### Decenza version

1.8.1

### Platform

Android

### Device / OS version

12

### Scale (if applicable)

Skale2 (connected and working)

skialpine:
Can try please arrangh a debug log to this issue so I can see what is going on with your system.

PR #1659, opened 2026-07-27
Stop DE1 controller errors from popping a dialog that says nothing (#1658)

### LINKED ISSUE #1658: [Bug] Connection error message on startup
### What happened?

When I power on my DE1 V1. 0 with Samsung A8 tablet with Decenza, I get a blue error box pop up on the screen that just says "connection error". If I just close that pop up all is well - no connection issues. This has been happening through many build updates and even a complete reinstall at one point. (Always using the latest beta build). I vaguely remember someone posting about this before, but don't know what the answer was. If it's relevant, I'm using a HDS on wifi network with firmware 3.1.11.

### Steps to reproduce

Open app and connect to DE and HDS after overnight shutdown. But only if I switch off power to DE completely, which I usually do overnight. 

### Decenza version

Build  3504

### Platform

Android

### Device / OS version

Samsung A8  Android 14

### Scale (if applicable)

HDS on 3.1.12

[debug-1.zip](https://github.com/user-attachments/files/30396135/debug-1.zip)

_No response_

### Screenshots or video

Image

### Anything else?

_No response_

skialpine:
Closes #1658.

## The reported bug

A reporter who switches their DE1 off overnight got a blue "connection error" box on every app start. The app direct-wakes the saved BLE address at launch, the machine hasn't finished powering up, Android returns `ConnectionError`, and the transport raised a modal for it:

```
[  5.78] BLEManager: DE1 direct wake - connecting to "DE1" at "FD:3D:C3:72:F8:8A"
[ 15.82] WARN [BLE DE1] !!! CONTROLLER ERROR: ConnectionError (state=Unconnected) !!!   ← modal
[ 20.60] DE1 reconnect: attempt 1 of 12
[ 36.51] [BLE DE1] Controller state: Connected                                          ← fine all along
```

This happened in 7 of the 15 sessions in their log, and in **every one** the reconnect ladder had the DE1 back 26–118 s later. The box only ever interrupted a recovery already under way, with a single OK button and nothing the offline `ConnectionIndicator` wasn't already showing.

**Controller errors are now log-only.** `AuthorizationError` had already been carved out for exactly this reason (#1093), as had write-retry exhaustion (#1423); this finishes the job for the rest of the enum. The errors a user *can* act on are untouched, because they come from the layers that can actually name them — `BLEManager` for permissions and for an adapter that won't come back up, and the service-discovery path for a link that connects but yields no DE1 service ("try toggling Bluetooth off/on"). The link-teardown family still feeds `de1LinkFault`, so the connection-priority coordinator and the BLE-stack-wedge detector are unchanged.

Also folds the two duplicated `QLowEnergyController` name switches — the DE1 transport and the scale transport each carried their own, so the same radio failure could be described two ways in one debug log — into a shared `blecontrollererror.h`, mirroring what `bleserviceerror.h` already does for service errors. That mapping picks up `RssiReadError` and `AdvertisingState`, which both copies were missing.

## Two more findings from the same log

Both verified still present on build 3504 (the log spans 3493→3504; the reported sessions are 3504, which is what `main` builds today).

**The Home Screen widget's last-shot tile never updated.** It was wired to `ShotDataModel::stopTime()`, written by exactly one caller — `markStopAt()`, on the stop-at-weight path. A shot ending any other way (manual stop, profile end, volume stop, or SAW blocked by an oscillating scale, which is what happened here) finishes with a full data set and `stopTime` still at its `-1` sentinel. `WidgetLastShot::make()` rejected it as non-finalized and the tile kept whatever it last accepted — in that log, every shot in the file:

```
[1575.330] ShotHistoryStorage: Saved shot 605 — Duration: 62.64 s
[1575.333] WARN [widget] setLastShot rejected non-finalized shot: yieldG 37.81 durationSec -1
```

`MainController` now emits `shotPersisted(shotId, yieldG, durationSec)` from inside the save callback, where the finalized pair is captured by value — `extractionDuration()` excluding SAW settling, and the settled 0.1 g-rounded yield, the same two numbers that went into the stored row.

**128 `jsonToDouble: failed to parse string "" - using default 2` warnings** from a single library browse. An empty string is not corruption — it is how de1app and the Visualizer write a setpoint that doesn't apply to a frame, and all 128 were a pressure frame's `flow`. The return value was always the default and still is; only the warning goes. A non-empty string that won't parse still warns, because that one really is a value the author meant being silently replaced.

## Deliberately not changed

- **`[SettingsDrift] DE1-dropped-write` (×12).** The detector working: the DE1 dropped a settings write, the resend fired, all 12 resolved on attempt 1. Quieting it would hide the evidence that the mechanism earns its keep.
- **The WiFi-scale IP cache.** The log shows a correct cached IP being evicted while the scale was merely asleep — but the last occurrence is at 2026-07-24 11:33 on **build 3502**, and every session from **3503** onward is clean across three days. That is #1624 arriving, not an open bug.
- **The `_imported_imported_imported4` profile chains** (152 of 234 profiles). Suggests re-import duplicates rather than replaces, but no import event appears anywhere in the log — a lead, not a diagnosis.

## Testing

Full suite green: **103/103, 0 failures, 0 warnings.**

- `tst_blecontrollererror` (new): pins the error/state name mappings against Qt's numeric enum values, and both sides of the teardown-family predicate that drives the wedge detector.
- `tst_settling`: characterizes the `stopTime` trap — a fully-recorded non-SAW shot with `rawTime() > 0` and real weight data still reports `stopTime() == -1`.
- `tst_profile`: an empty string takes the default *silently* (enforced by `failOnWarning`), while `"garbage"` still warns.

No manual update needed — the removed dialog was never documented, and the wiki's troubleshooting table doesn't mention it.


skialpine:
### Code review

Reviewed at f7adaefc. Found 3 issues, all since fixed in 4c97cfa1.

1. `tst_bletransporterror` fails deterministically on Linux CI. `init()` calls `QTest::failOnWarning()` with no argument, but on Linux the `UnknownRemoteDeviceError` rows emit two warnings that no `ignoreMessage` covers: the setcap hint, and the `CAP_NET_ADMIN missing` warning that `BleCapability::linuxMissing()` itself triggers on first call. Unprivileged CI runners have no `CAP_NET_ADMIN`, so both fire. Invisible on macOS because the paths are `#if defined(Q_OS_LINUX)` guarded — and with no PR CI gate this lands on main and breaks `nightly-sanitizers.yml`.

https://github.com/Kulitorum/Decenza/blob/f7adaefcf504f638513cddc8a1047b35898c26e7/src/ble/bletransport.cpp#L606-L617

2. The same test leaks a detached `QThread` under the nightly Linux ASan job. `onControllerError` unconditionally calls `BleCapability::logLinuxBtDiagnosticsOnce()`, which on Linux spawns a thread running up to three 2-second `QProcess` calls, cleaned up via a queued `deleteLater`. `QTEST_MAIN` returns straight after `qExec`, so that `deleteLater` never gets an event-loop turn. CLAUDE.md notes the nightly Linux ASan job is "the only place leaks are detected" since LSan does not exist on macOS.

https://github.com/Kulitorum/Decenza/blob/f7adaefcf504f638513cddc8a1047b35898c26e7/src/ble/blecapability.cpp#L114-L128

3. Directional pointer is backwards: the `shotPersisted` comment says `shotDiscarded()` is "above", but it is declared at line 430, below `shotPersisted` at 382.

https://github.com/Kulitorum/Decenza/blob/f7adaefcf504f638513cddc8a1047b35898c26e7/src/controllers/maincontroller.h#L376-L383

Fix for 1 and 2: both Linux diagnostics now sit under the `#ifndef DECENZA_TESTING` guard this function already uses for its `BLEManager` dependency. They are diagnostics for a human reading the debug log, not behaviour, so a test build loses nothing — and the test keeps all 11 enum rows rather than dropping the inconvenient one.

Also checked and clean: CLAUDE.md compliance (no timer-as-guard, no main-thread I/O, no Settings changes, naming, i18n — this PR removes untranslated raw strings from modals rather than adding any); git history (does not undo #1309, whose substance is the Android adapter power-cycle plus the actionable auto-recovery-failed message, both untouched; #1093 and #1423 carve-outs cited accurately; no prior commit deliberately made USB errors user-visible); no unaddressed reviewer feedback on this or the 3 most recent PRs touching these files; and `discardUnopenedTransport` lifetime — the delete is safe only because `de1Discovered` has not fired, and that ordering holds in both call sites.

PR #1661, opened 2026-07-27
Fix two background-source defects: the shot chart's scrim, and a stored key that never converges

skialpine:
Three defects in the background-source code, found by reading a tablet debug log and then a macOS one. All are invisible on screen in the sense that nothing looks obviously broken, which is why they shipped.

---

## 1. The shot-chart background never scrimmed the chrome over it

`Theme.hasBackgroundImage` read `LastShotChartSource.imageSource` directly. That singleton lives in `qml/components/`; `Theme.qml` lives in `qml/` and imports nothing from the module, so the type was never in scope there. The binding compiled clean and threw at runtime on **every** evaluation:

```
qrc:/qt/qml/Decenza/qml/Theme.qml:549:
ReferenceError: LastShotChartSource is not defined
```

A throwing binding leaves the property at its default, so `hasBackgroundImage` was pinned `false` for the entire `"shot"` case — the chart drew as wallpaper and all ~70 call sites reading it painted their chrome **opaque** over the picture, the inverse of the intent.

Introduced by #1585, released in 2.0.1.

## 2. ...and once it evaluated, it was asking the wrong question

Making the arm evaluate exposed that it was app-wide, while the shot chart is the one background a **page** can refuse. `ThemedPageBackground.suppressShotChart` turns it off on the eleven pages that draw a graph of their own — `EspressoPage`, `SteamPage`, `PostShotReviewPage`, `ShotDetailPage`, `ShotComparisonPage`, `FlowCalibrationPage`, `RecipeEditorPage`, `SimpleProfileEditorPage`, `ProfileEditorPage`, `ProfileInfoPage`, `AutoFavoriteInfoPage` — and those pages then paint a flat colour.

So "a render exists" told the chrome a picture was behind it where there was none. `glassChrome` went true, `chromeFill()` scrimmed over a flat page (the elevation-cancelling failure its own comment documents), and `insetBackgroundColor` took the `scrimColor(backgroundColor)` branch — 40% of a colour composited over itself, which is that colour exactly. Text-field boxes, switch tracks and unselected pills drawn straight on those pages disappeared, and the `_flatInsetTint` branch written for this case was skipped.

**Fix:** bind to what the surface is actually *drawing*. `BackgroundSurface` publishes `paintsShotChart` (its existing `_hasShotChart` under a public name), and `main.qml` binds `Theme.shotChartOnCurrentPage` to it for the item on top of the stack. That is false while a render is pending, false under a preset, and false on a suppressing page — the three cases where chrome must stay opaque.

A binding rather than a push from the surface: several `ThemedPageBackground`s are alive at once in the stack, so whichever wrote last would win regardless of which page you were looking at. It costs nothing extra either — `paintsShotChart` is state the surface computes anyway, and its `shotChart &&` short-circuits before touching `LastShotChartSource`, so the singleton's startup load guard still holds.

## 3. The stored background source never converged after a parameter was released

`setBackgroundPreset("")` — what the chooser's "no colour" does, and what `clearBackgroundPreset()` does on behalf of other actions such as `ScreensaverVideoManager` deleting a backing file — wrote `theme/backgroundPreset` but left `theme/backgroundSource` at `"colour"`. The guard only ever wrote the key when a colour was **set**, so the pair went inconsistent and stayed that way.

Nothing looked wrong — `backgroundSource()` derives `"none"` from the values actually set, so the right background drew. But the stored key never converged, and the derivation logs when it overrides a stored value, so every launch reported:

```
[Theme] Stored background source "colour" is not backed by a value this build can draw - using "none"
```

**152 times in one machine's log.** That message exists to make an override diagnosable; one that fires forever on a state that can never resolve trains the reader to skip it.

`setBackgroundImagePath("")` had the identical defect, leaving `"image"` pointing at a path that is gone. Fixed both.

Fixed at the write site, not the read site: `backgroundSource()` still derives and still rewrites nothing, which is what keeps an install predating the key working. Both new branches release the key only when it still names the parameter being cleared, and both document the ordering they depend on — every caller stores the new source kind *before* clearing the old parameter, so a nested clear finds a key that no longer names it.

### Test

`clearBackground()` already wrote the key explicitly as its last step, which is why `clearingReturnsToNone()` passed throughout — the gap was clearing a parameter *without* going through it. The new case asserts the **stored** key after releasing a parameter on its own; asserting the derived getter would pass against the bug.

Defect 2 has no automated test: the failure is QML-runtime-only and the suite is C++.

---

## On the guard that didn't catch #1

`qmllint` **did** flag that line — as one of 88 identical `Unqualified access` warnings in `Theme.qml`. The other 87 are context-property false positives, so the one real error was indistinguishable from the noise. Worth a separate think about declaring context properties to qmllint so that signal becomes usable; not attempted here.

## Verification

- Build: 0 errors, 0 warnings.
- Full suite via Qt Creator: **104 passed, 0 failed, 0 skipped, 0 with warnings.**
- Defect 1 confirmed fixed against a running app (chart grab logged, no ReferenceError); defect 3 covered by the new test case.

### DIFF
--- a/qml/Theme.qml
+++ b/qml/Theme.qml
@@ -544,9 +544,24 @@ QtObject {
     // (it has no parameter), so on a fresh install with no shots the source says shot while
     // BackgroundSurface paints a flat colour — and scrimming chrome over a flat colour is
     // the exact elevation-cancelling failure chromeFill()'s own comment documents.
+    //
+    // That fact is PUSHED in by LastShotChartSource rather than read from it. Everything
+    // else this file reads — Settings, TemperatureDisplay, EmojiAssets — is a context
+    // property, visible everywhere without an import; LastShotChartSource is a QML singleton
+    // in qml/components/, and a file in qml/ cannot see a type from another directory. This
+    // condition used to reference it directly, which compiled clean and threw
+    // "ReferenceError: LastShotChartSource is not defined" on every evaluation, pinning
+    // hasBackgroundImage to false for the whole shot case: the chart drew, and all ~70 call
+    // sites below painted chrome opaque over it — the exact inverse of the intent.
+    //
+    // Keep it a push. Theme is the lowest-level singleton in the app and deliberately
+    // depends on nothing but context properties; importing the module here to reach one
+    // component would trade a runtime error for a dependency cycle waiting to happen.
+    property bool shotChartRendered: false
+
     readonly property bool hasBackgroundImage: Settings.theme.backgroundSource === "image"
                                                || (Settings.theme.backgroundSource === "shot"
-                                                   && LastShotChartSource.imageSource.length > 0)
+                                                   && shotChartRendered)
 
     // The fill a piece of chrome should actually paint.
     //
--- a/qml/Theme.qml
+++ b/qml/Theme.qml
@@ -545,23 +545,25 @@ QtObject {
     // BackgroundSurface paints a flat colour — and scrimming chrome over a flat colour is
     // the exact elevation-cancelling failure chromeFill()'s own comment documents.
     //
-    // That fact is PUSHED in by LastShotChartSource rather than read from it. Everything
-    // else this file reads — Settings, TemperatureDisplay, EmojiAssets — is a context
-    // property, visible everywhere without an import; LastShotChartSource is a QML singleton
-    // in qml/components/, and a file in qml/ cannot see a type from another directory. This
-    // condition used to reference it directly, which compiled clean and threw
-    // "ReferenceError: LastShotChartSource is not defined" on every evaluation, pinning
-    // hasBackgroundImage to false for the whole shot case: the chart drew, and all ~70 call
-    // sites below painted chrome opaque over it — the exact inverse of the intent.
+    // Set by main.qml from the background of the page on TOP of the stack — see the note
+    // there. Not read from LastShotChartSource: that singleton lives in qml/components/ and
+    // a file in qml/ cannot see a type from another directory, so referencing it here
+    // compiled clean and threw "ReferenceError: LastShotChartSource is not defined" on
+    // every evaluation, pinning hasBackgroundImage false for the whole shot case. Theme is
+    // also the lowest-level singleton in the app and deliberately depends on nothing but
+    // context properties; importing the module here to reach one component would trade a
+    // runtime error for a dependency cycle waiting to happen.
     //
-    // Keep it a push. Theme is the lowest-level singleton in the app and deliberately
-    // depends on nothing but context properties; importing the module here to reach one
-    // component would trade a runtime error for a dependency cycle waiting to happen.
-    property bool shotChartRendered: false
+    // Per-PAGE, unlike the image below, because the chart is the one background a page can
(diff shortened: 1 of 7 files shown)


skialpine:
### Code review

Found 2 issues:

1. `hasBackgroundImage` is global, but the shot chart is suppressed per-page — so 11 pages now get glass chrome over a flat background, and inset controls drawn straight on the page become invisible. Fixing the ReferenceError flips the shot arm to `true` app-wide, but `ThemedPageBackground.suppressShotChart` means `EspressoPage`, `SteamPage`, `PostShotReviewPage`, `ShotDetailPage`, `ShotComparisonPage`, `FlowCalibrationPage`, `RecipeEditorPage`, `SimpleProfileEditorPage`, `ProfileEditorPage`, `ProfileInfoPage` and `AutoFavoriteInfoPage` paint the flat `Theme.backgroundColor` instead. On those pages `_hasShotChart` is false and `imagePath` is empty (`selectShotChartBackground` clears it), so `_hasImage` is false — yet `glassChrome` is true, `chromeFill()` scrims over nothing, and `insetBackgroundColor` takes the `scrimColor(backgroundColor)` branch. That expression's own comment says it "is 40% of a colour composited over itself, which is that colour exactly, and the control disappears" — the `_flatInsetTint` branch written for exactly this case is skipped. This was masked on `main` only because the ReferenceError pinned the branch false; the shot case is now wrong on the suppressed pages instead of the unsuppressed ones.

https://github.com/Kulitorum/Decenza/blob/2ef90f6ccafcc16cd288f07385ef31240f4398d6/qml/Theme.qml#L561-L565

https://github.com/Kulitorum/Decenza/blob/2ef90f6ccafcc16cd288f07385ef31240f4398d6/qml/Theme.qml#L645-L648

2. The new `else if` branches depend on a load-bearing ordering that is not documented where it matters. They are only safe because each setter calls `storeBackgroundSource(<new kind>)` *before* clearing the other parameter — `setBackgroundImagePath(nonEmpty)` stores `"image"` before `clearBackgroundPreset()`, and `selectShotChartBackground()` stores `"shot"` before both clears — so the branch tests a key that has already moved and correctly declines to fire. Reordering those two lines would silently reintroduce the stomp, emitting an intermediate `"none"` through `notifyBackgroundSourceChanged` that QML bindings would latch. `clearingOneSourceDoesNotDisturbAnother` would catch it, but the constraint is invisible at the call site. (CLAUDE.md's conventions lean on comments that "record deliberate design decisions and warn against specific mistakes.")

https://github.com/Kulitorum/Decenza/blob/2ef90f6ccafcc16cd288f07385ef31240f4398d6/src/core/settings_theme.cpp#L75-L83

https://github.com/Kulitorum/Decenza/blob/2ef90f6ccafcc16cd288f07385ef31240f4398d6/src/core/settings_theme.cpp#L126-L131

Checked and clean: the C++ call orderings all trace correctly (no branch fires when a different source is already stored); the QML push has no missed-clear path, since every write to `_renderedUrl` goes through the bound `imageSource`; the initial `""`/`false` pair is correct by construction for a late- or never-instantiated singleton; no binding loop; `backgroundSource()` still derives without writing, so the "nothing is rewritten on read" invariant holds; and the new test correctly asserts the raw stored key rather than the derived getter.

Issue #877, opened 2026-04-27
[Bug] long press sleep - app freezes


GCDE-VER1:
### What happened?

After updating to latest beta I have added a "quit" button next to "sleep" button, but same result with either. Long pressing sleep or pressing quit button results in DE shutting down and HDS scale turning off, but Decenza screen remains and is frozen. I have to force close the app. Tried restarting tablet (A9+), and clearing Decenza cache, but no change. Log attached. 

[debug-2.zip](https://github.com/user-attachments/files/27107755/debug-2.zip)

### Steps to reproduce

Hit quit or sleep button. 

### Decenza version

1.7.2  Build 3325

### Platform

Android

### Device / OS version

Android 15

### Scale (if applicable)

HDS

skialpine:
fixed in new beta

PR #1680, opened 2026-07-27
Migrate 14 context properties to QML singletons, and fix what the gate caught

skialpine:
Continues `fix-qmllint-usability`. Three strands: the first CI run of the qmllint gate, the missing-property triage, and section 2 of the change (context properties).

## The gate's first CI run found a real gap

Nightly run 30310401195 failed on `qml/simulator/GHCSimulatorWindow.qml: exists but is not in the qt_add_qml_module list`.

**Not a runtime bug.** `DECENZA_SIMULATOR` is defined on Linux, which invites the conclusion that Linux loads an unbundled QML file — it does not, because the load site is guarded by `(Q_OS_WIN || Q_OS_MACOS) && QT_DEBUG`. I said otherwise first and was wrong.

What was broken is lint **coverage**. `QML_FILES` appended the file under `WIN32 OR (APPLE AND NOT IOS)`, so the module file set — and therefore the generated `Decenza.rsp` the gate lints — varied by platform. macOS lints 218 files, Linux linted 216. The missing one carries a ceiling of 57 diagnostics that the only automated lint run had never evaluated. Guard is now `NOT ANDROID AND NOT IOS`.

Generalisable: a per-file gate is only as trustworthy as the invariance of its file set. The remaining conditional append (Quick3D screensavers) has the same shape and passes only because the Linux runner finds Quick3D.

Also answered: `install-qt-action` **does** ship `qmllint` on Linux.

## One real bug from the missing-property triage

`Theme.dangerColor` does not exist. Four uses in `SettingsDebugTab.qml` styled error dialogs with `undefined`. Of 326 findings, 4 were real; the rest are structural false positives, written up in `bugs-found.md`, including 32 (`DrinkType`, `SettingsTabs`) that remain **unexplained** and are deliberately not exempted.

## Section 2: context properties

Classification corrected three premises the task list was written on:

- Only **2 of 7** "set more than once" names are genuine runtime swaps (`ScaleDevice`, `Refractometer`). `DE1Device`/`GHCSimulator` publish the same object to a second engine — easier as singletons, not harder.
- **Three names are read by nothing** — `FlowScale`, `DE1Simulator`, `IsDebugBuild`. Verified by grep, not inferred from a zero count. Deleted.
- **Three objects are declared after `engine`** and migrating them as-is would reintroduce the refractometer teardown crash `main.cpp` already documents. QML drops a context property when its object emits `destroyed()`; a static singleton pointer has no equivalent. Excluded, with the rule recorded in the new header.

**14 names migrated: 661 of 943 warnings removed.** Clean files 85 → 89 of 218.

Registration lives in a new `src/core/contextsingletons_qml.h`, not in the class headers — `add_decenza_test()` links no `Qt6::Qml`, and CLAUDE.md records that breaking `saw_parity`. This is `settings_qml.h`'s pattern generalised. The `Q_GADGET`/`QML_*` macros are written out per struct because moc does not expand macros when scanning for them; only the publish logic is shared.

## Verification

- Full suite **105/105**, no warnings. `tst_qmlregistration` failed first on its own negative assertion (`SteamDataModel is now registered — delete its entry`) — that allowlist did its job and is removed.
- Gate passing, 218/218 linted, baseline updated.
- Runtime: app launched, six of the fourteen exercised through `main.qml`, no TypeErrors and no `create()` criticals in 794 log lines. All 14 confirmed exported as `isSingleton: true`.
- **Not verified:** `CrashReporter` (needs a crash), and page-level bindings in `SettingsAITab`/`LibraryPanel`/`ShotGraph`/`SteamGraph`/`WeatherItem`.

## Left for follow-up

282 warnings, each blocked on something specific: façade for the 2 runtime swaps (155), duplicate-registration conflicts for `SteamHealthTracker`/`ProfileStorage`/`McpServer` (50), declaration hoists (52), value-holder singleton needing call-site renames (16), `#ifndef Q_OS_IOS` guards (10).


skialpine:
### Code review

Five reviewers (CLAUDE.md compliance, deep bug scan, git-history context, prior-PR feedback, comment accuracy). **No large bugs found.** Six issues found and acted on in [`b5af1651`](https://github.com/Kulitorum/Decenza/commit/b5af1651ca38382b99aa348d11d031b3878c1349).

The deep scan independently verified the load-bearing claims against Qt's own source: all 14 objects declared before `engine`, no runtime reassignment of any migrated name, and `QJSEngine::setObjectOwnership(CppOwnership)` confirmed necessary — without it `qqmlengine.cpp:1875` hands a stack object to the GC.

Fixed:

1. `qmllint_report.py` silently resolved Qt's own qmllint, which cannot finish `CustomItem.qml` and is OOM-killed after ~10 min. The correct invocation existed only in a developer's private notes, and someone lost ten minutes to the hang. Now refuses in a second with both correct commands.

2. Eight hardcoded English strings in the two dialogs this branch already edited (CLAUDE.md: "Fix pre-existing violations in any file you touch").

3. Stale teardown comment justifying the #877 Android deadlock workaround with "de1Device is exposed to QML via setContextProperty" — false after this branch. Hazard unchanged; premise wasn't.

4. "943 warnings across these names" attributed to the 14 in that file; 943 is all 30 candidates, the 14 are 660.

5. `src/mcp/mcpremoteaccess.h` missing from HEADERS — only one of 14. AUTOMOC covered it, nothing broken.

6. Stale allowlist rationale left above its own replacement in `tst_qmlregistration.cpp`.

Recorded, not fixed — migrating a context property makes `typeof X !== "undefined"` permanently true, and the failure mode worsens: an unpublished singleton is `null`, `typeof null === "object"`, so the guard passes and the access throws instead of falling through. Flagged by two reviewers independently; ~19 such guards join ~11 already known in `CustomItem.qml`. Low practical risk (all three publish before `engine.load()`). Written up in `bugs-found.md` with the reason **not** to sweep by deletion: `Theme.qml`'s `EmojiAssets` guard looks identically dead and is not.

Full suite: 105 passed, 0 failed, 0 warnings.

skialpine:
### Follow-up: the `typeof` guard issue is now fixed, not deferred

My earlier comment said this was recorded rather than fixed. That is stale — done in [`7c81f586`](https://github.com/Kulitorum/Decenza/commit/7c81f5865c252097f038ccd1479867fa31777957), 153 guards across 54 files.

**Counting first changed the job.** The review described ~19 guards from this branch plus ~11 already known, in two files. The real distribution was dominated by something neither reviewer had looked at:

| | |
|---|---|
| `AccessibilityManager` | 106 guards, 50 files |
| `DE1Device` / `BLEManager` / `BatteryManager` | 19 (this branch) |
| `MachineState`, `Settings`, `ProfileManager`, `MainController`, `TranslationManager` | 28 |

**The obvious fix is a regression.** Rewriting `typeof X !== "undefined"` to a plain `if (X)` throws **ReferenceError** when the identifier is genuinely undeclared — which is exactly the wholesale-registration failure the guard exists for, and which this project has had (1,081 ReferenceErrors against a green build). `typeof` is the only form safe on an undeclared name, so it stays and gains a null check:

```js
typeof X !== "undefined" && X !== null
```

`undefined` false, `null` false, live object true — all three states.

**Scope came from the type system, not grep.** Of 169 such guards in `qml/`, only the 153 naming a type exported in `Decenza.qmltypes` were rewritten. The other 16 name `ScaleDevice`, `Refractometer`, `USBManager`, `GHCSimulator`, `McpServer`, `pageStack`, `Window` — still context properties or plain JS, where the unmodified `typeof` is correct. Rewriting those would have been wrong in the other direction; they join the set when their own migration lands.

Verified: suite 105/105 0 warnings; gate passed 218/218 linted, 89 clean, no diagnostic movement; app launched clean (only unrelated MQTT/funnel network warnings).

Worth stating the limit: the smoke test cannot exercise the `null` branch, because reaching it requires a singleton that failed to publish and `main()` assigns all 14 before `engine.load()`. The branch is insurance against a registration failure, not a reachable state. The check that mattered was the lint gate, since `CustomItem.qml` holds 32 of the 153 and only the patched qmllint can read that file at all.

skialpine:
### Second review round — five reviewers, six issues, all fixed in [`74d56628`](https://github.com/Kulitorum/Decenza/commit/74d566280117a58ed1200abcd2e86c3cecdfb53f)

Re-reviewed because the first round only covered through `f0df165e`; the 55-file guard sweep had never been reviewed by anything.

**Two corrections to my own earlier comment:**

The "169 total / 16 remaining" figures were wrong — actual **186 / 33**. I transcribed them wrong from my own script output and repeated them here before two reviewers caught it independently. The list of 7 untouched names was right. Corrected in `bugs-found.md` and recorded there rather than quietly fixed, since that file warns that a plausible count is not a correct one.

The sweep was **one-sided**. The regex matched `!== "undefined"` and missed the inverted form, which is the identical defect:

```js
if (typeof X === "undefined" || !X.enabled) return   // typeof null === "object" -> !null.enabled throws
```

10 sites, 9 files, all `AccessibilityManager`, all in files the sweep had already edited. Fixed.

**Also fixed:**

- `Theme.qml` **reverted** — I had added a null test to the `EmojiAssets` guard. That singleton has no `create()` and no published instance, so it can never *be* null; the addition was dead code contradicting the comment explaining the guard.
- The qmllint guard was **dead where it mattered**. It tested "was `--qmllint` passed" as a proxy for intent, but both CMake targets always pass it from `find_program()`. Now tests where the binary lives.
- Its printed remedy was **unrunnable** — suggested `--update-baseline --skip-unlintable`, which the script refuses. Now `python3 scripts/…`, shlex-quoted.
- `settings.debug.importFailed` duplicated `shotimporter.title.importFailed`; reuses the existing key.

**Reported regressions do not reproduce.** A separate agent reported unqualified rising in `PresetPillRow`/`HotWaterPage`/`SteamPage`. Gate passes against a current build (218/218 linted, 89 clean) and ceilings may only fall, so a rise fails it. Two of the three quoted starting numbers also disagree with the committed baseline (135 vs 131, 223 vs 211), and `HotWaterPage.qml` has no `typeof` guards and is not in this diff — that measurement was against a different tree or a stale build.

Suite 105/105, 0 warnings. All four per-PR text-invariant checks pass.

PR #1683, opened 2026-07-28
Migrate four more context properties, and fix the baseline #1680 shipped

skialpine:
Second batch after #1680. Picked as the set whose blocker was mechanical rather than a design question: **103 of the 283 remaining warnings**, no forwarding façade, no new lifetime concept.

| Name | Warnings | What it needed |
|---|---|---|
| `SteamHealthTracker` | 40 | Drop the `…Type` workaround |
| `FlowCalibrationModel` | 37 | Declaration hoist |
| `ProfileStorage`, `McpServer` | 10 | Nothing — plain migrations |
| `AppVersion`, `AppVersionCode`, `PreviousCrashLog`, `PreviousDebugLogTail` | 16 | A holder object |

### `main` is currently red, and this fixes it

The baseline shipped in `db97314d` records three ceilings **below what the tree produces**, so the nightly's ubsan leg has been failing at the `QML diagnostics gate` step.

It is **not a code regression.** The unqualified identifiers in `PresetPillRow.qml` are `modelData` (30) and `root` (17) — no name #1680 touched — and `HotWaterPage.qml` is not in that commit's diff at all. The baseline was generated against a build predating part of that PR's own C++ changes, so qmllint resolved less deeply and counted fewer warnings. That is the stale-build trap this change documents as methodology error #2, committed again in the change that exists to prevent it.

CI's **stock** qmllint (217 files, Linux) and the **patched** one (218 files, macOS) report identical counts. That also settles a question left open earlier: a baseline generated with the patched binary *is* enforceable by a stock one.

The three ceilings rise to their true values — a ratchet in the forbidden direction, so it is written up in `bugs-found.md` rather than quietly applied. The ~8 `modelData`/`root` delegate-scoping defects underneath are pre-existing and now visible; fixing them needs `required property` declarations in three files unrelated to this batch, left as follow-up.

### Notable in the migration

- **`SteamHealthTracker` loses `SteamHealthTrackerType`.** The suffix existed only because a context property of the plain name resolves ahead of a type of that name. The singleton removes the collision instead of routing around it, exactly as `MachineState` did. The header claimed `tst_qmlregistration` pinned the exported name — it did not; nothing under `tests/` ever referenced it, so the rename that comment warned against would have gone green. Replaced with the guarantee that does hold.
- **`DE1DeviceType` removed.** Same shadowing workaround, `DE1Device` became a singleton in #1680, nothing referenced it.
- **`GHCSimulator` deliberately excluded.** Its instance legitimately does not exist off Windows/macOS debug builds, so registering it would `qCritical` on every Android, iOS and Linux launch. That needs an explicit optional-singleton path in the publish helper — a real decision, not worth taking for 15 warnings.
- **Two real defects surfaced**, both in `FlowCalibrationPage.qml`, the one file whose singleton this batch added, and neither reachable before because a context property has no type to check a member against: a redundant `?.` on a non-nullable `QString`, and `.toFixed()` on a `??` union that types as `QJSPrimitiveValue`.

### Verification

- Full suite **105 passed, 0 failed, 0 warnings**
- Gate **passes**, 218/218 linted, clean list **89 → 90**
- `SettingsCalibrationTab.qml` 40 → 1, `FlowCalibrationPage.qml` 39 → 2, `main.qml` 159 → 141, `SettingsUpdateTab.qml` → clean
- All four `check_*.py` pass


skialpine:
### Review round applied

Five review agents (CLAUDE.md compliance, deep bug scan, comment accuracy, silent-failure, type design). Every finding was verified against the code before acting.

**`AppInfo` deleted — the PR now adds no new types instead of two.**

Three of its four values already had an owner:
- `AppVersion`/`AppVersionCode` duplicated `UpdateChecker::currentVersion`/`currentVersionCode` — same `VERSION_STRING` and `versionCode()`, already `CONSTANT`, already registered, already reached as `MainController.updateChecker` **in the same file that displayed them**.
- `PreviousCrashLog`/`PreviousDebugLogTail` moved to `CrashReporter`, which is where QML already goes to submit them.

That left `qmlsingletonpublish.h` with one caller — its whole justification was `AppInfo` as a second — so it folded back into `contextsingletons_qml.h`.

**Three false comments, all written in the first commit of this PR:**

| Claim | Reality |
|---|---|
| `appinfo.h`: QML clears the crash log "through `CrashReporter`" | `CrashReporter` never touches the file. It's `MainController::clearCrashLog()` → `CrashHandler` |
| `main.cpp`: the three storage `…Type` registrations were context-property workarounds | `git log -S` finds no such context property ever. The comment also contradicted its own next paragraph |
| `main.cpp` teardown: `FlowCalibrationModel` declared after `engine`, protected by QML dropping a context property | Both false since the hoist — and this is the documentation for the exact use-after-free hazard, so a reader could have moved the declaration back down on its authority |

**Closed the test gap the review exposed.** `SettingsCalibrationTab.qml` compares against `SteamHealthTracker.EstablishingAfterReset`; a renamed enumerator makes that `=== undefined` — silently false, wrong wording forever, no error, no log. The header had claimed `tst_qmlregistration` pinned that name; it did not, and deleting the false claim didn't close the hole.

The new check is general, not written for one name: every `Singleton.UpperCaseMember` access in `qml/` must resolve in `Decenza.qmltypes`. It covers the unscoped form the existing `MachineState.Phase` test cannot, since there's no enum name to anchor on. **Verified by negative control** — injecting a renamed enumerator fails it; reverting passes.

**Declined, with reasons:**
- `try/catch` around `Component.onCompleted`. Real observation (a null singleton throws where a missing context property was falsy, and 47 lines of startup ride behind `main.qml:900`), but the only route to null is deleting a publish line that `tst_qmlregistration` asserts exists.
- `qFatal` instead of `qCritical` in the publish helper — contradicts that code's deliberate "checked, not asserted" design, which exists because `Q_ASSERT` compiles out of Release.
- Stripping `?.` from `FlowCalibrationPage` for call-site consistency — churn with no defect behind it.

Suite 105 passed, gate passes at 90/218.

PR #1684, opened 2026-07-28
Make the QML gate detect an unachievable ceiling and a silent relaxation

skialpine:
Follow-up to #1683. The baseline defect corrected there was not interesting because a number was wrong — it was interesting because **every check in the script was pointed the wrong way to catch it.** Both holes are closed here. No numbers change; no C++ or QML is touched.

### Hole 1 — the staleness check could not see the type registry

`check_fresh()` compares each QML file against the build's own copy. That is the right check for "you edited QML and forgot to rebuild", and structurally blind to what actually happened: qmllint resolves module types through `Decenza.qmltypes`, generated by `qmltyperegistrar` from the C++ `QML_*` macros, so a changed **registration** leaves every QML file byte-identical while the registry describes the previous generation.

It fails in the safe-looking direction. Names that do not resolve make qmllint give up early on expressions it cannot type, so it reports **fewer** warnings — which reads as an improvement, and `--update-baseline` writes it in as a ceiling the tree cannot meet. That is how #1680 shipped three of them and left the nightly red for a day.

`check_registry_fresh()` asserts every `QML_SINGLETON` under `src/` appears in the registry's exports. Content-based like its sibling, because any git operation rewrites source mtimes and a timestamp check would call a freshly built tree stale.

### Hole 2 — `--update-baseline` relaxed the gate silently

Every enforcement path asked whether counts *rose*. Nothing asked whether a recorded ceiling was *achievable*, and nothing distinguished the two things `--update-baseline` does: locking in an improvement, and relaxing the gate. Same keystroke, identical output.

A rise now costs `--allow-ceiling-rise`. The refusal names each file with before/after; permitting one prints what was raised, so the relaxation lands in the log rather than only in a diff nobody reads. Leaving the clean list counts as a rise from zero. Raising is sometimes correct — correcting #1680's ceilings meant raising them — so this is a speed bump, not a ban, and the message points at the likelier cause first: a stale build.

### Verification — by negative control, not by observing a pass

| Case | Result |
|---|---|
| Header registering a singleton absent from the qmltypes | Refused, names it |
| Same, header removed | Clean pass, 218/218 |
| Faked unachievable ceiling in the baseline | Write refused |
| Same, with `--allow-ceiling-rise` | Written, prints `SteamPage.qml 200->212` |
| Gate, patched qmllint | Passes, 90/218 clean |
| Gate, CI path (stock + `--skip-unlintable`) | Passes, 90/217 |
| Full suite | 105 passed, 0 warnings |

Neither guard fires on a good tree, on either qmllint.

### The general shape

A ratchet that only measures one direction will accept a bad measurement pointing the other way, indefinitely and without complaint. "The tree improved" and "the measurement under-reported" produced identical evidence here, and only one of the two was ever considered.


skialpine:
### Targeted review applied — the guard was itself blind

Three agents (bug scan, silent-failure, comment accuracy). Two independently found the same critical defect, and it is the one worth reading:

**`check_registry_fresh()` covered 20 of the tree's 27 singletons and reported success.**

It resolved a singleton's QML name from `QML_NAMED_ELEMENT` or `QML_FOREIGN` and had no third branch, so a plain `QML_ELEMENT` contributed nothing — the `re.split` discarded the class name, leaving nowhere to get it. The seven invisible ones were `MainController`, `MachineState`, `ProfileManager`, `AccessibilityManager`, `TranslationManager`, `EmojiAssets`, `MarkdownRenderer` — most of the largest, and the same population as the incident the guard cites.

**It passed its own negative control**, because that control used the `QML_FOREIGN` shape and exercised the branch that worked. A guard verified only on the path that works is not verified. The `if not declared` backstop couldn't help: twenty names is a healthy-looking non-empty set, and "found some" reads exactly like "found all" — the shape of the truncated qmllint run that once counted never-analysed files as clean.

Three fixes, all the same anti-pattern — a lookup whose not-found case defaults to *harmless*, inside a check whose whole job is noticing absence:

| Hole | Fix |
|---|---|
| `QML_ELEMENT` singletons invisible | Capture the class name, add the `else` branch. 27/27 |
| Presence ≠ singleton — a type mid-migration **to** a singleton matches a stale registry that still lists it uncreatable | Require the export and `isSingleton: true` in the same component |
| `.get(key, {})` on the baseline — wrong-schema JSON made every file read as no-history, so no rise detectable, write proceeds silently | Require the schema; refuse unparseable JSON by name; unseen file = rise from zero |

Also now refuses `--update-baseline --allow-stale`. The write path already refused `--from-raw` and `--skip-unlintable`; `--allow-stale` was missed, and it's the one that disables *both* freshness checks — exactly how the bad baseline was written. Reading a report off a knowingly-stale build stays allowed.

**Re-verified against each shape separately**, which is what the first attempt skipped:

| Control | Result |
|---|---|
| Bare `QML_ELEMENT` + `QML_SINGLETON`, unbuilt | `ZzProbeElement: absent from the registry` |
| Type present but not as a singleton | `CoffeeBagStorage: present, but NOT as a singleton` |
| Wrong-schema baseline | Refused, names the missing keys |
| Unparseable baseline | Refused, quotes the JSON error |
| Unseen file with warnings | `SteamPage.qml: 0 -> 212` blocked as a rise |
| Fault removed, each case | Clean pass, 218/218, 90 clean |

Suite 105 passed. `check_*.py` ×4 pass.

Confirmed by review and worth stating: CI **does** run this guard — `qmllint_check` passes `--check --skip-unlintable --qmllint --import-path`, never `--allow-stale`. And the doc claim that it checks "every `QML_SINGLETON`" was false as originally shipped; `bugs-found.md` now records the 20-of-27 miss rather than the claim.

PR #1687, opened 2026-07-28
Make QML identifiers resolvable: clean list 92 -> 126, and the bugs that were hiding behind the noise

skialpine:
Continues `openspec/changes/fix-qmllint-usability`. **The change is deliberately NOT archived** —
tiers of the expanded scope remain, and we agreed to review this first and then decide whether to
keep going.

## What it does

Every QML identifier that was resolved by runtime name lookup — and therefore checked by nothing —
now resolves at compile time.

| | before | after |
|---|---:|---:|
| clean files (zero `unqualified`) | 92 | **126** / 219 |
| `unqualified` | 4938 | **2738** |
| `missing-property` | 322 | **145** |
| `import` | 4 | **0 — category cleared** |
| `setContextProperty()` calls in main.cpp | 5 names | **0** |

Full suite green throughout: **105 passed, 0 failed, 0 warnings.**

## The bugs this surfaced

The point of the exercise, and the reason the numbers matter:

1. **The screensaver's RSS ceiling had never once fired.** `ScreensaverPage.qml` declared
   `videoRssCeilingMB` on its own root, then read it back as `root.videoRssCeilingMB` — but `root`
   is main.qml's id, which has no such property. So the guard was `liveRss > undefined`, always
   false, and the mitigation for the Qt FFmpeg/VideoToolbox leak was inert.
2. **`SimpleProfileEditorPage`'s title was always `undefined`** when no profile name was set. The
   binding referenced `editorTitle`, which existed only as a `var` inside a function — assigned,
   never read. Two mistakes propping each other up.
3. **Our own singletons were being shadowed.** `DrinkType.shortLabel`, `DrinkType.icon`,
   `DrinkType.fromRecipeMap`, `SettingsTabs.indexOf` all reported as missing members while plainly
   declared. Cause: 106 directory imports (`import "../components"` alongside `import Decenza`)
   re-resolving singleton files as plain component types. Deleting them cleared the whole `import`
   category.
4. **An iOS null-deref prevented.** Registering a `#ifdef`-guarded context property silently breaks
   every `typeof X !== "undefined"` guard around it — the type then always exists. Three bindings in
   `SettingsConnectionsTab.qml` would have thrown on iOS.
5. **One I introduced, caught by the linter.** A bulk pass prefixed `ValueInput`'s delegate-local
   `gear` with the component root; every gear tap would have assigned `undefined`. It survived
   exactly one relint. That is the payoff working.

## Architecture

- **`AppShell` singleton** — 120 cross-file reaches into main.qml's root (`goBack`, `goToIdle`,
  `stopReason`, `sessionMeasuredMilkG`…) become a declared contract. Navigation is signals;
  main.qml keeps every line of navigation logic. Every cross-file `pageStack` use is gone, along
  with all `Qt.resolvedUrl` navigation including a path built by string concatenation.
- **`ScaleDeviceProxy` / `RefractometerProxy`** — the last context properties. A singleton cannot
  be re-pointed, so the singleton is a proxy and the proxy moves. **Hardware-verified**: Decent
  Scale and DiFluid R2, both directions, three connect/disconnect round trips, zero QML errors.
- **`Keyboard` singleton** — 108 `Qt.inputMethod.*` calls were unchecked (`Qt.inputMethod` types as
  bare `QObject`). Also gives CLAUDE.md's IME-commit rule an actual home in code.
- **`AccessibilityManager::cleanForSpeech()`** — moved from QML to C++ and unit-tested for the
  first time, 14 data-driven cases.
- **`pragma ComponentBehavior: Bound` + `required property`** across 30 files. This is Qt's own
  prescription (qtdeclarative's "Exposing State from C++ to QML").

## Behaviour change — one, deliberate

A `CustomItem` widget configured to navigate to Steam/Espresso/Hot Water/Flush now **pushes**
instead of `replace(null, …)`, so back returns to idle. The rule is that replace is for
machine-driven phase changes, not for a destination; the old code also left `pageStack.depth` at 1,
which made `goBack()`'s `depth > 1` test fail and the back control silently dead.

## Review focus / what is NOT verified

- **Runtime.** Navigation was click-through verified (Steam, Settings, PostShotReview end-to-end,
  no mis-wired `Connections` handlers). Tier 3 — both profile editors, ShotHistory/ShotDetail/
  PostShotReview, and `ValueInput` (on nearly every settings and editor screen) — has **not** been
  run.
- **`Keyboard.commit()` touched 27 files** including every save/send/search dialog. Its failure
  mode is mobile-only and silent, so a desktop pass cannot prove it — this wants an Android test
  build.
- **`-DDECENZA_SIMULATOR_OVERRIDE=0`** (the tablet-production shape) is reasoned about, not built;
  no routine CI job builds it.
- USB, WiFi and simulated-scale re-point paths are compile-checked only.

Docs updated: `QML_NAVIGATION.md` rewritten (it described a `pageStack` usage that no longer
exists), `QML_GOTCHAS.md` +5 sections, `CLAUDE.md` +4 one-liners.


skialpine:
### Code review

Found 5 issues.

1. **A null singleton instance is not `undefined` in QML — the `GHCSimulator` guard does not guard, and throws a TypeError on every window activation in Linux, Windows/macOS Release, and Android/iOS Debug builds.** Verified against the Qt 6.11.1 sources: `qv4qmlcontext.cpp:229` resolves a singleton *name* with `QQmlTypeWrapper::create(v4, nullptr, r.type)`, discarding the result of `singletonInstance<QObject*>()` — the wrapper is created either way, and `QQmlTypeWrapper` has no `virtualToBoolean` override, so the name is a truthy Object with `typeof === "object"`. Only the *member read* degrades: `qqmltypewrapper.cpp:319` fails its `if (QObject *qobjectSingleton = ...)` and falls through to `Object::virtualGet`, giving `undefined`. So `typeof GHCSimulator !== "undefined" && GHCSimulator` passes and `GHCSimulator.mainWindowActivated()` throws. The type is registered whenever `DECENZA_SIMULATOR` is defined (every desktop config; `$<CONFIG:Debug>` on mobile — `CMakeLists.txt:1057-1067`) but the instance is published only under `(Q_OS_WIN || Q_OS_MACOS) && QT_DEBUG` (`main.cpp:3514`), which is exactly where it gets tested locally. `Connections { target: typeof GHCSimulator !== "undefined" ? GHCSimulator : null }` on the next line has the same problem. Test the member, not the name.

https://github.com/Kulitorum/Decenza/blob/dbf226a573d9756b0e8610aae7e5c0a2b4c9a8da/qml/main.qml#L846-L857

2. **`main.qml` still calls `cleanForSpeech()` bare after it moved to C++.** The function was `main.qml`'s own and is now `AccessibilityManager::cleanForSpeech`. Every other call site was updated (`ProfileEditorPage.qml:1725`, `ProfileSelectorPage.qml:758/763/774/799/806`); this one was not, and no local definition remains. It is the global "announce any tapped text" handler, so it is a ReferenceError on every text tap with a screen reader active.

https://github.com/Kulitorum/Decenza/blob/dbf226a573d9756b0e8610aae7e5c0a2b4c9a8da/qml/main.qml#L988-L992

3. **`contextsingletons_qml.h`'s "READ BEFORE ADDING AN ENTRY" block is contradicted by the rest of the same file, and by the document it cites.** It says three context properties remain, that `ScaleDevice`/`Refractometer` "need a forwarding façade — genuinely the hardest thing left, and section 4 defers it deliberately", that "ghcSimulator is still NOT here", and "do not register without the hoist". The same file registers `GHCSimulatorForeign` (:519), `ScaleDeviceForeign` (:541) and `RefractometerForeign` (:563), the hoist is done in `main.cpp`, and this PR rewrites the cited `openspec/changes/fix-qmllint-usability/tasks.md` §4 from "Deferred" to "DONE, after being deferred". This compounds issue 1: a reader following this paragraph concludes the type is unregistered, which is the assumption that makes the `typeof` guard look correct. The same block instructs "Update this figure when you add an entry; it is the one number in this file that goes stale silently" — and the figure is now stale by ~91 references.

https://github.com/Kulitorum/Decenza/blob/dbf226a573d9756b0e8610aae7e5c0a2b4c9a8da/src/core/contextsingletons_qml.h#L38-L44

4. **Widget-driven navigation can stack a duplicate of the page you are already on.** `goToEspresso`/`goToSteam`/`goToHotWater` each no-op when already on that page; `goToSettings`, `goToFlush`, `goToRecipes`, `goToShotHistory`, `goToBeanInfo`, `goToEquipment`, `goToAutoFavorites` and `goToCommunityBrowser` push unconditionally. `StatusBar` is inside `pageStack` at `z: 600` and visible on every page but the screensaver (`main.qml:1292-1301`), so its widgets are tappable from the destination itself — tap Settings while on Settings and Back returns to a second `SettingsPage`, not to idle. This is new behaviour: the same call sites previously replaced, as the comment above these functions records. `startNavigation()` does not cover it — it clears via `Qt.callLater`, so it only blocks re-entry inside one event-loop turn.

https://github.com/Kulitorum/Decenza/blob/dbf226a573d9756b0e8610aae7e5c0a2b4c9a8da/qml/main.qml#L3574-L3583

5. **`ScaleDeviceProxy`'s "WHY EVERY SLOT IS FORWARDED" comment is false, and `QML_GOTCHAS.md` restates it as a rule for future proxies.** `ScaleDevice` declares 12 public slots; the proxy forwards 9. `hasIndependentTimerReset()`, `resetFlowCalculation()` and `addFlowSample()` are omitted. No functional bug today — all three are called only from C++ through direct `ScaleDevice*` pointers, never from QML — but the comment claims "The surface is preserved deliberately", and `docs/CLAUDE_MD/QML_GOTCHAS.md` tells the next author to "Forward **every** public slot, not the ones QML calls today", citing this class as the worked example. `RefractometerProxy`'s equivalent comment is accurate.

https://github.com/Kulitorum/Decenza/blob/dbf226a573d9756b0e8610aae7e5c0a2b4c9a8da/src/ble/scaledeviceproxy.h#L38-L45

A sixth issue — the identifier-qualification pass inserting a prefix mid-chain, turning `results.length` into `results.shotHistoryPage.length` and breaking Shot History infinite scroll with a NaN offset — was found independently by two reviewers and is fixed in `dbf226a5`. The detector for that class is paired-line, not textual: if removing one segment from a new dotted chain yields a chain present in the old line, the insertion landed mid-chain. Run over the whole branch diff it finds exactly that one occurrence.

<details>

- [Worth a look] `contextsingletons_qml.h:535` says `ScaleDevice` has "79 references across 18 QML files"; an exhaustive `git grep -lw` at the base commit returns 11 files. The reference count is right, the file count is not. (score: 65)
- [Worth a look] `keyboard.h:13` and `qmllint_report.py:107-108` both say "108 call sites across 27 files". The 108 is exact; the file count is 36, and the PR's own migration diff touches 36. Propagated to two files. (score: 65)
- [Worth a look] `contextsingletons_qml.h:556` says Refractometer is "12 references in PostShotReviewPage.qml. The last context property." — `SettingsConnectionsTab.qml` also references it. (score: 60)
- [Worth a look] `qmllint_report.py`'s `missing-property` ceiling history ends one step at 303 and begins the next at 300, with no bullet for the gap — either a missing step or a typo. (score: 60)
- [OK to skip] `ScreensaverPage.qml:828,863` still test `if (ScaleDevice && ScaleDevice.connected)`. Redundant now that the proxy is never null, harmless, and in a file this PR touches. (score: 55)
- [OK to skip] `LibraryItemCard.qml` keeps a namespaced directory import. Namespaced imports do not shadow the module's singletons, so it is not the defect the other 106 deletions addressed — but it does give those types a second identity. (score: 50)

</details>

skialpine:
All five review findings fixed, plus the below-threshold items. Build clean, full suite passes.

| # | issue | fix |
|---|---|---|
| 1 | truthy null-instance singleton — `typeof X !== "undefined" && X` passed, then threw on every window activation in Linux / Release desktop / mobile Debug | guard the member (`GHCSimulator.mainWindowActivated !== undefined`); the comment that asserted the opposite rewritten with the Qt citations |
| 2 | `main.qml` called `cleanForSpeech()` bare after it moved to C++ — ReferenceError on every screen-reader text tap | qualified |
| 3 | `contextsingletons_qml.h` header contradicted its own file and the `tasks.md` it cites | rewritten to match |
| 4 | status-bar widgets stacked a duplicate of the page you were already on | `pushUnlessCurrent()`; `SettingsPage.showTab()` for the tab case, since `requestedTabId` is consumed only in `StackView.onActivated` |
| 5 | `ScaleDeviceProxy` forwarded 9 of 12 public slots while claiming all, and `QML_GOTCHAS.md` cites it as the worked example | three forwarded |

Below-threshold, also done: the four stale counts (`ScaleDevice` 18 to 11 files, `Qt.inputMethod` 27 to 36 in both copies, Refractometer's second call site, the 303/300 gap), the redundant object-existence guards on the never-null proxies, and `LibraryItemCard`'s namespaced directory import.

Two lessons landed in the docs, neither previously written down:

- **The relint is not a backstop for mid-chain qualifier insertion.** `QML_GOTCHAS.md` cited `ValueInput`'s `gear` as proof the linter catches these — it does, onto a *typed* receiver. `results.shotHistoryPage.length` drew no warning at all, because `results` is a signal-handler parameter typed `var`. Added the column-shift rule (insert right-to-left) and a paired-line detector that finds the class exhaustively on a diff: if removing any one segment from a dotted chain in the new line yields a chain present in the old line, the insertion landed mid-chain.
- **A comment asserting Qt behaviour needs a file-and-line citation.** Issue 1 is the worked example of the compounding failure: an un-sourced claim went into a header as settled fact, the comment then licensed a guard that could not work, and both survived until someone opened the Qt source.

Separately trimmed CLAUDE.md 260 to 222 lines — exposition moved to the docs that already hold it, every one-line tripwire verified still resident. The emoji block was a genuine move (`EMOJI_SYSTEM.md` did not have that guidance); the rest were true trims against content already duplicated.

Still unverified and called out in the PR body: the Tier 3 runtime click-through, `Keyboard.commit()` on Android, and the `-DDECENZA_SIMULATOR_OVERRIDE=0` shape. The navigation changes above are compile- and test-clean but have not been clicked through either.

PR #1688, opened 2026-07-28
Clear four of five qmllint category exemptions, and the bugs they were hiding

skialpine:
Batch 1 of driving qmllint to zero. Tree goes **92 -> 131 of 219 clean**, and **`CATEGORY_EXEMPTIONS` is now empty**.

## Read this first: the nightly on `main` will be red

`missing-property` is at **92 and deliberately unexempted**. The qmllint gate does not run on PRs, but it does run in `nightly-sanitizers.yml` on `main`, so merging this turns the nightly red until batch 2 clears the remaining 92.

That is the intended trade. A category exemption is **global**: at `missing-property: 145` a brand-new missing-property bug anywhere in the tree was invisible. Red and honest beats green and blind — but it only stays defensible if batch 2 follows promptly.

## Exemptions deleted by fixing what they hid

| category | was | now |
|---|---|---|
| `equality-type-coercion` | 1 | **0** |
| `duplicate-property-binding` | 2 | **0** |
| `Quick.property-changes-parsed` | 5 | **0** |
| `Quick.layout-positioning` | 21 | **0** |
| `missing-property` | 145 | 92, unexempted |

`Quick.layout-positioning` were all genuine qmllint false positives — its check tests whether an object is *declared lexically* inside a Layout, never whether a Layout manages it, and `Dialog`/`Popup` derive from `QObject` while `parent: Overlay.overlay` reparents out at runtime. They are now suppressed **at each site** with the reason inline (13 suppressions), not behind one global number, so the category is enforced and the next real one fails.

## Bugs found, not lint noise

1. **The two-finger accessibility back gesture has never worked.** `main.qml` called `stackView.depth`/`stackView.pop()`; no such id exists (it is `pageStack`). ReferenceError killed the handler including its `else`. Now `depth > 1 ? goBack() : goToIdle()` — both branches needed, because `replace(null, X)` clears the stack and that is how every machine-driven page and the screensaver are entered, leaving them at depth 1.
2. **Two frozen clocks.** `Timer { onTriggered: parent.currentTime = ... }` in `ScreensaverPage` and `ShotMapScreensaver`. `Timer` is a `QObject` with no `parent`, so the write went to the file root's parent and the clock never ticked.
3. **Two dead initialisers.** `gradientHue: 0.6` and CrtShaderEffect's `time: 0`, both overridden by a `NumberAnimation` value source with an explicit `from`.
4. **`CustomItem.qml` had 3 diagnostics, not the 122 on record** — that number was never a measurement, because no run had ever completed on it. The patched qmllint does the whole tree in ~9s.

## scripts/qml_qualify.py

Encodes the rules that each cost a bug: never infer a prefix (qmllint prints the correct one), insert right-to-left, split qmllint's glued diagnostics, and refuse to edit when the parsed count disagrees with qmllint's own count for the file.

That last one is not hypothetical — the first version silently lost **20 of 144** sites in `RecipeWizardPage` to the glued-line bug and still reported a plausible number.

## What review caught, and how

Four review agents ran against this branch. Everything they found was mine and is fixed. Two are worth recording because of *how* they surfaced:

- The tool's 20 lost diagnostics printed a normal-looking report.
- Two of 38 inserted ids bound to the **wrong object** (a control 500 lines away, inside a delegate whose `parent` is the row). Nothing detected a wrong reference — the baseline writer **refused** because one file's ceiling rose 54 -> 60, and that refusal is the only reason anyone looked. Reverted; the containment check is now in `QML_GOTCHAS.md`.

Both failures still produced plausible numbers. That is the pattern to distrust.

## Verification

Build and full suite clean. `tst_decentscalewifi` failed once and passed on re-run; this branch contains no C++.

**`FlushPage` is the one thing wanting eyes rather than a compile** — its delegates now declare `modelData`/`index` with `required property`, the class that compiles and tests green while being broken. The preset pills and their arrow-key navigation should be looked at in a running app.

### DIFF
--- a/docs/CLAUDE_MD/QML_GOTCHAS.md
+++ b/docs/CLAUDE_MD/QML_GOTCHAS.md
@@ -394,6 +394,32 @@ paired-line, not textual:
 Run over a whole branch diff that is a few seconds and it is exhaustive; it found this occurrence
 and confirmed there were no others.
 
+## Introducing an id? Verify every use falls inside that object's own block
+
+Giving a control an id to replace an untyped `parent.` hop is a good fix, and it is easy to bind
+the id to the WRONG object. An automated pass over
+`Button { background: Rectangle { color: parent.down ... } }` searched upward for the enclosing
+`background:`/`contentItem:` with no bound, found one hundreds of lines above, and attached six
+references to a button in a different part of the file:
+
+```qml
+// line 273
+AccessibleButton { id: createNewProfileButton; ... }
+// line 774, inside `trailingActionDelegate: Component { StyledIconButton { ... } }`
+icon.color: createNewProfileButton.selected ? ...   // WRONG: `parent` here is the delegate's row
+```
+
+It compiles, and both objects exist, so nothing complains — it simply reads the wrong object.
+
+The check that catches it is cheap and does not need a linter: **for every id you introduce, find
+its declaration line and its block's extent by indentation, then assert every `<id>.` use in the
+file falls inside that range.** Run over a 38-id pass this found exactly the 2 that were wrong.
+
+Note the ordinary gate does NOT catch this. It surfaced only because one file's `unqualified`
+ceiling ROSE (54 -> 60) and the tool refused to write a baseline that relaxes a ceiling — the
+rise was the symptom, not the defect. A gate that silently accepted the new number would have
+shipped it.
+
 ## `Window` is an attached property — do not write it through an id
 
 ```qml
--- a/qml-diagnostics-baseline.json
+++ b/qml-diagnostics-baseline.json
@@ -60,7 +60,6 @@
     "qml/components/layout/items/WeatherItem.qml": 32,
     "qml/components/library/LibraryItemCard.qml": 66,
     "qml/components/library/LibraryPanel.qml": 95,
-    "qml/main.qml": 110,
     "qml/pages/AutoFavoriteInfoPage.qml": 77,
     "qml/pages/AutoFavoritesPage.qml": 71,
     "qml/pages/BeanInfoPage.qml": 4,
@@ -211,6 +210,7 @@
     "qml/components/layout/items/TemperatureItem.qml",
     "qml/components/layout/items/WaterLevelItem.qml",
     "qml/components/qrcode.js",
+    "qml/main.qml",
     "qml/pages/AISettingsPage.qml",
     "qml/pages/FlowEditorPage.qml",
     "qml/pages/HotWaterPage.qml",
@@ -234,6 +234,6 @@
     "duplicate-property-binding": 2,
     "equality-type-coercion": 1,
     "missing-property": 146,
-    "unqualified": 2734
+    "unqualified": 2624
   }
 }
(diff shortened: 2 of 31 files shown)


skialpine:
### Second review pass — on the half the first review never saw

The four agents that reviewed this branch earlier saw it up to `d5bf03ac`. Five commits landed after that (the exemption removals, ~70 Control `parent.`-hop rewrites across 18 files, 13 inline suppressions, a revert) — 27 files, 316 insertions, unreviewed. Two agents were run against exactly that range.

**They found six defects. Five would have thrown at runtime. All were mine.**

**Segment substitution, not insertion.** The Control pass replaces the seven characters `parent.` with `<id>.`. On a chain that was already `parent.parent.X` it rewrites the *second* segment and leaves the first, so the expression reads a member off the wrong object:

| site | was | became | effect |
|---|---|---|---|
| `SettingsLanguageTab:291,774` | `parent.parent.visualPosition` | `parent.progressBar.visualPosition` | download progress fill never binds |
| `ShotHistoryPage:713` | `parent.parent.checked` | `parent.checkBox.checked` | shot-selection tick never renders |
| `ValueInput:878` | `parent.parent.getContrastColor` | `parent.scrubberPopup.getContrastColor` | drag bubble text colour dead |
| `SteamItem:205` | `popupPillRow.parent.popupSuffixVersion++` | `popupPillRow.presetPopup...` | **repeating TypeError on every scale-weight change** while the steam popup is open |

`verify()` could not see any of them. Its mid-chain check drops one segment from each new chain and looks for the result in the old line — that detects **insertion** and nothing else, and nothing was inserted here. It now has a substitution case (equal length, exactly one differing segment, not the first). That is the same blind spot twice — insert-into-chain, then substitute-within-chain — so it is written against the shape rather than the instance.

**The `!=` -> `!==` "cleanup" was a functional regression.** `customScanPath` is a `property url`. QML hands a url to JS as a `UrlObject`, whose default `virtualIsEqualTo` returns false unconditionally (`qv4managed.cpp`), and strict equality never coerces — so `customScanPath !== ""` is **always true** and `rescan()` stopped falling back to auto-detect. Reachable from the Rescan button and from the list refresh after every successful import, for any user who never set a custom folder. `!=` coerced via `ToPrimitive`, which is why it worked.

That one is worth dwelling on: it was the change used to justify deleting an exemption, it made a lint category green, and it broke a real path lint cannot see. **Lint-clean and wrong is worse than lint-dirty and right.**

Also fixed: all 20 newly-qualified `<id>.down` reads are on `AccessibleButton`, whose own header says `down` stays false because its `touchArea` accepts the press — so every one of those custom backgrounds had a dead pressed state. Now `.down || .isPressed`.

**Verified clean by the same pass:** the other 36 id rewrites, all 13 layout-positioning suppressions (each wraps a `Dialog`/`Popup` or an item reparented to `Overlay.overlay`), the `PropertyChanges` rewrite (Qt's own documented replacement), and both value-source initialiser removals.

`missing-property` 92 -> 87. Every other category 0. Build and full suite clean.

PR #1694, opened 2026-07-29
Fix #1691: Anthropic thinking off explicitly; stop silently truncating every provider

### LINKED ISSUE #1691: [Bug] Claude returned empty response
### What happened?

While attempting "AI Advice" on a shot history, I entered a prompt that resulted in a message, "Anthropic returned empty response content,."  

### Steps to reproduce

1. Select any History
2. Select 'AI Advice'
3. Enter prompt

### Decenza version

2.0.1 Build 3506

### Platform

Windows

### Device / OS version

Windows 11 Home Insider Preview, v.25H2

### Scale (if applicable)

_No response_

[CCable_Debug_28JUL2026_1925PDT.zip](https://github.com/user-attachments/files/30488252/CCable_Debug_28JUL2026_1925PDT.zip)

_No response_

### Screenshots or video

Image

### Anything else?

Keep up the great work, Jeff!

skialpine:
Fixes #1691.

## The bug

`AnthropicProvider` sends `max_tokens` with **no `thinking` field**. That default is not stable across models:

| model | `thinking` omitted |
|---|---|
| `claude-sonnet-4-6` | no thinking |
| `claude-sonnet-5` | **adaptive thinking** |

`max_tokens` bounds thinking *and* response text together, so on Sonnet 5 the whole 1024-token budget went to a thinking block. The reply came back **HTTP 200, `content` non-empty, `stop_reason: "max_tokens"`, zero `text` blocks** — landing on the `Anthropic returned empty response content` path.

Every request failed, for every user who picked Sonnet 5 (added in #1445). Sonnet 4.6 users were unaffected, which is why it took a user report to surface.

The reporter's log shows 12–15 s between request and failure across three separate days — tokens *were* generated, just not as text.

`testConnection()` has the same gap and only checks for an HTTP error, so **it passed while every real request failed** — the key tested fine and the advisor was unusable.

## Changes

**1. Turn thinking off explicitly** on all three Anthropic request builders plus `testConnection()`. Same reason the OpenAI and Gemini paths already force reasoning/thinking off: dial-in advice needs little chain-of-thought, and hidden thinking tokens bill at the output rate. Anthropic was the one that never got the equivalent, then Sonnet 5 flipped the default under it.

**2. Raise the output cap 1024 → 4096** across all providers (shared `AIProvider::MAX_OUTPUT_TOKENS`). 1024 was too tight regardless of thinking — a dial-in reply plus the trailing fenced `nextShot` JSON block (#1054) lands right at that edge. The cap is a ceiling, not a reservation, so this costs nothing unless the model actually writes more.

**3. Report truncation instead of hiding it.** No provider inspected its stop/finish reason, so a capped reply was emitted *as if complete* — silently dropping the `nextShot` block with nothing in the log. All four cloud providers now check (Anthropic `stop_reason`, OpenAI `finish_reason` + `incomplete_details`, Gemini `finishReason`, OpenRouter `finish_reason`) and surface a shared message. This matters most on OpenRouter, whose model is a free-text user string and can point at a reasoning model that eats the cap the same way.

The Anthropic path also logs the **block types** on any text-less reply: `content` can be non-empty while carrying no text block, and the generic message alone made that indistinguishable from a refusal.

## Tests

`tst_aiproviders` gains a canned-response server (same shape as `FakeBeanBaseServer`) that captures the posted **body** — an absent field in the request JSON is invisible to the existing pure catalog tests.

- request shape: `thinking.type == "disabled"` and `max_tokens == 4096` on both `analyze()` and `analyzeConversation()` (the path the in-app advisor actually uses)
- a thinking-only reply is reported as truncated, not as the generic empty-content error
- a complete reply still succeeds — the truncation branch doesn't swallow good replies

Full suite via Qt Creator: **108 passed, 0 failed, 0 skipped, no warnings**. Build clean, 0 warnings.

Note `get_test_details` reports per-CTest-binary, not per-function, so the three new functions don't resolve by name — they ran inside `tst_aiproviders`, which rebuilt after the edits and passed.

## Not covered

Ollama sets no cap at all (unbounded) and is left alone.

No user-visible feature change, so no wiki manual edit.

### DIFF
--- a/src/ai/aiprovider.cpp
+++ b/src/ai/aiprovider.cpp
@@ -4,6 +4,7 @@
 #include <QJsonObject>
 #include <QJsonArray>
 #include <QNetworkRequest>
+#include <QStringList>
 #include <QTimer>
 #include <QUrl>
 #include <QVariant>
@@ -32,6 +33,13 @@ QString AIProvider::tr_(const char* key, const char* fallback) const
         return m_translationManager->translateString(QString::fromUtf8(key),
                                                QString::fromUtf8(fallback));
     return QString::fromUtf8(fallback);
+}
+
+QString AIProvider::truncatedResponseError() const
+{
+    return tr_("ai.error.truncated",
+               "The AI's reply was cut off before it finished. Please try again, "
+               "or ask a more specific question.");
 }
 
 QString AIProvider::friendlyNetworkError(QNetworkReply* reply) const
@@ -243,9 +251,9 @@ void OpenAIProvider::analyze(const QString& systemPrompt, const QString& userPro
     // chat/completions ("Unsupported parameter") — max_completion_tokens is
     // the accepted cap. Live-caught July 2026: stage-1 extraction and the
     // advisor both 400'd on gpt-5.4/gpt-5.4-mini.
-    requestBody["max_completion_tokens"] = 1024;
+    requestBody["max_completion_tokens"] = MAX_OUTPUT_TOKENS;
     // GPT-5 family are reasoning models; keep reasoning off so hidden
-    // reasoning tokens don't count against the 1024-token output cap (which would
+    // reasoning tokens don't count against the output cap (which would
     // risk truncating the trailing nextShot JSON block) and to keep latency/cost
     // low. Dial-in advice needs little chain-of-thought. Mirrors Gemini's
     // thinking=off. The 5.4 generation REPLACED the value "minimal" with "none"
@@ -284,7 +292,7 @@ void OpenAIProvider::analyzeUrl(const QString& systemPrompt, const QString& user
     QJsonObject reasoning;
     reasoning["effort"] = QString("low");
     requestBody["reasoning"] = reasoning;
-    requestBody["max_output_tokens"] = 2048;
+    requestBody["max_output_tokens"] = MAX_OUTPUT_TOKENS;
 
     sendResponsesRequest(requestBody);
 }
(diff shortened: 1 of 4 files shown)


skialpine:
## Review pass applied (802233a3)

Four review agents. Three things they caught were defects in the original commit, not polish:

**1. The Gemini truncation check was unreachable in exactly the case #1691 generalizes.** `finishReason` was read eighteen lines after an `parts.isEmpty()` early return — and a candidate that stops on `MAX_TOKENS` *while still thinking* comes back carrying a `finishReason` and no `content` key at all. So it exited with the generic "Gemini returned empty content" and the new branch never ran. `gemini-3.5-flash` runs `thinkingLevel: "minimal"`, which is on, not off, so this is the live shape. Anthropic had the same ordering; both hoisted. `SAFETY`/`RECITATION`/`PROHIBITED_CONTENT` arrive the same way, and a prompt-level block returns `promptFeedback.blockReason` with no candidates — that reason is the whole explanation and was being thrown away.

**2. My truncation test could not fail.** The fixture was *both* text-empty and `stop_reason: max_tokens`, so it couldn't discriminate: reducing the guard to `if (text.isEmpty())` left all three tests green. The unpinned half was the dangerous one — a **partial** reply emitted as if complete, silently missing the `nextShot` block. Added a partial-text case and verified by mutation that it now fails when the guard is weakened (107/108 with the guard reduced, 108/108 restored).

**3. Provider error bodies were logged verbatim at six sites.** Provider 4xx bodies echo request fragments back, our prompts carry shot history, bean names and tasting notes, and these logs get attached to public issues here. Now logs the error type/code with a bounded fallback.

Also in this pass:

- `INVARIANT` on `disableAnthropicThinking` — newer Anthropic models reject the `disabled` form outright, or accept it only at/below a given effort, so one catalog entry could 400 every request. Mirrors the existing OpenAI note ~300 lines up.
- **Truncation on the conversation path now shows the partial reply with a notice** rather than discarding it — that reply is prose the user reads. `analyze()`/`analyzeUrl()` still fail hard; their result is machine-parsed and a half-parsed result is worse than none.
- Anthropic allow-lists `end_turn`/`stop_sequence` instead of enumerating bad reasons, which fails safe as Anthropic adds more — and catches `pause_turn`, reachable on `analyzeUrl` (it attaches `web_fetch`) and partial by definition.
- OpenAI Responses gated on `status != "completed"` rather than one incomplete reason; `content_filter` and `failed`/`cancelled` were shipping partial text as complete. `message.refusal` / refusal parts surfaced instead of the generic message.
- OpenRouter: choice-level error object (a 200 the top-level check misses) and `content_filter`/`error` finish reasons.
- Ollama: reads `done_reason`. This app sets no cap, but `num_predict`/`num_ctx` live in the *user's* config and Ollama reports `"length"` when they truncate. Its empty-response branch logged nothing at all.
- `m_model` in every one of these warnings — the root cause was that the default differs *between* models, and none of them said which.

**Tests:** `FakeAnthropicServer` → `FakeProviderServer`, generalized; `buf` is a `shared_ptr` captured by value (the raw new/delete across two lambdas was a use-after-free *and* a leak that only the nightly Linux ASan job would have caught); `listen()` failure is loud in Release. Added `analyzeUrl` and `testConnection` to the request-shape test, the partial-vs-fail policy split, a `pause_turn` case, and a data-driven truncation case for the other four providers — including the exact Gemini shape that was broken. Every failure case now also asserts the opposite signal stayed silent.

Suite: **108 passed, 0 failed, no warnings.** Build clean.

PR #1695, opened 2026-07-29
Clear qmllint to ZERO: unqualified 2285 -> 0, and the seven dead empty-state labels it was hiding

skialpine:
Batch 3, the last category. `unqualified` goes 2285 -> **0** across 87 files, clean 135 -> **222 / 222**. The baseline now records 222 clean, 0 ceilings, 0 observed categories, and `CATEGORY_EXEMPTIONS` is still empty — it has never held an entry through any of the three batches.

## The one ceiling that was ever raised is gone

`SettingsConnectionsTab.qml` held the project's only RAISED ceiling (39 -> 41, from #1689's outer-id-in-a-delegate lines). It is at zero, so that rise no longer stands as precedent.

## Bugs the noise was hiding

**Five dead empty-state labels.** A visual child of a `ListView` is reparented to its `contentItem` (`qquickflickable.cpp:2442`), so `parent` in those blocks is the contentItem, not the ListView. `parent.count` was therefore `undefined`, `undefined === 0` is `false`, and the placeholder never appeared:

- SettingsConnectionsTab: "No devices found", twice (DE1 scan list and discovered-devices list)
- SettingsScreensaverTab: "Loading…" and "No categories" — the category list looked empty and silent for the whole time it was fetching

Two more had the same centring mistake without the count bug (BeanBaseSearchBar's "No matches", SettingsSearchDialog's "No settings found"): centring on a zero-high contentItem put them half above a clipped edge. All seven now read geometry off the list by id.

**Six unchecked `Keys` handlers.** `EspressoPage` read a bare `event` in six `onReturnPressed` / `onSpacePressed` blocks — the deprecated injected-parameter path. It works, but nothing can check it, so a typo inside one is invisible until a key is pressed. They now declare `function(event)`.

## What the work actually was

Three shapes, in rough proportion:

1. **`pragma ComponentBehavior: Bound`** on 60-odd files, so a delegate can resolve its own file's ids. Every delegate in those files declares each injected model role it uses as a `required property` in the same edit — without that, Bound stops role injection and the delegate breaks at runtime, silently, with nothing in the build, the linter or the tests to say so.
2. **Qualifying a same-file identifier** read from a nested scope (`root.foo`, `delegateId.modelData`).
3. **An id on a delegate or inline component** that had none, so its nested children could name its properties.

`FlipClockScreensaver` established something worth recording: the pragma resolves file-scope ids inside an **inline component** too, not just inside a delegate. Its `FlipDigitCard` read `root` for all twelve of its style inputs, and those resolve at runtime only because an inline component shares the file's QML context. 33 -> 0 on the pragma alone.

## Three defects I introduced, all caught by the gate refusing

None of these was caught by reading the diff:

- `CommunityBrowserPage`: a scripted pass rewrote **a function-local `var totalPages`** into `communityBrowser.totalPages`, a member the page does not have — infinite scroll would have compared against `undefined` and stopped loading. Same family as the `parent.parent.X` rewrite from batch 1.
- `StringBrowserPage` / `ProfileSelectorPage`: sibling-object ids (`stringModel`, `centerTimer`, `viewFilter`) and layout-local properties (`col1/2/3Width`) qualified against the page as if they were its members.
- `LibraryItemCard`: `LayoutItemDelegate` already declares `modelData` required, so redeclaring it on a delegate of that type is a `property-override` — and unnecessary, since a base type's required property is still bound by the view.

The qualification pass is now driven by the `(line, flagged-token)` pairs qmllint actually reported, not by a name list applied to whole lines. A line flagged only for an outer id is deliberately left untouched for the pragma to resolve, so "did this line change?" stays a meaningful assertion rather than one that silently passes.

## Verification

- `--check` (not `--update-baseline`) run after every batch, against a freshly built tree — the staleness guard refused two runs and both were real.
- Per-file delta computed after every batch using the report script's **own** parser: a hand-rolled `^Warning:` regex undercounts by 275, because qmllint glues a following diagnostic onto a line with no newline. `REGRESSED: none` on every batch.
- Full suite: **107 passed, 0 failed, 0 skipped, no warnings.**
- Both gate invocations pass: full-tree, and the CI-shaped `--skip-unlintable`.

## What the tools cannot check — please open these screens

Role injection breaks at runtime only. The build, qmllint and all 107 tests are blind to it.

| Screen | What to look for |
|---|---|
| Settings → **Connections** | Scale-picker dropdown rows (star, name, WiFi/BLE/USB badge); discovered devices list; **"No devices found" should now appear on an empty list** |
| **Recipe wizard**, every step | Drink-type, bag, profile, equipment, pitcher and vessel tiles all render — seven delegates in this one file |
| **Recipes** page | Recipe cards, archived section, re-point-bag picker |
| **Library** panel | Tabs, save menu, entry grid, card previews |
| **Layout editor** | Zone chips, add-item list, background picker, readout options, colour picker, emoji picker |
| Settings → **Themes / AI / Machine / Screensaver / Language / Debug / Firmware** | Every option row and colour swatch |
| **Profile selector**, **Profile import**, **Visualizer multi-import**, **Community browser** | List rows render with text |
| **Shot graphs** (live, history, profile, comparison) | Traces, phase markers, axis tick labels |
| **Screensavers**: Flip clock, Pipes, Shot map (**flat and globe**) | Digits flip; pipes animate; markers appear. The globe uses `Repeater3D` — the one delegate type I am least sure takes required properties |
| Idle-screen widgets | Weather forecast strip, last-shot graph, tinted icons |

Not merging without your review.


skialpine:
### Code review

Reviewed at `7934751d`, after the rebase onto `main`. The mechanical core is sound: five independent passes found **no defect in the `pragma ComponentBehavior: Bound` / role-injection class** — every delegate that gained the pragma declares each injected role it reads, every `required property` names a role its view actually supplies (`modelData` for JS arrays/`QVariantList`, `model` for `ListModel`/C++ models), and no shared base type picked up a conflicting declaration. That is the failure mode neither the compiler, qmllint, nor the 108 tests can see, so it was the main thing worth checking.

Nine inaccurate `pragma` comments were found during the review and fixed in `995256ea` and `7934751d` — each named several delegates as reading a file id when only one of them does, or credited `layer.effect` blocks that read only `Theme`.

Found 4 issues, all pre-existing violations in files this PR modifies:

1. Untranslated user-visible strings (CLAUDE.md: "All user-visible text must be internationalized")

https://github.com/Kulitorum/Decenza/blob/7934751d82306a75b99d9fc21773e60525e85360/qml/components/AIConversationPanel.qml#L55-L58

Also `ShotMapScreensaver.qml:488`, `CommunityBrowserPage.qml`, `SettingsMachineTab.qml:187-188,1834`, `SettingsConnectionsTab.qml:87,984`, `SettingsDebugTab.qml:266,282`, `SettingsLanguageTab.qml:817`, `VisualizerBrowserPage.qml:146`, `RecipeEditorPage.qml:420`, `ProfileSelectorPage.qml:694,706`, `WeatherItem.qml`. Several are `accessibleName` strings whose paired visible `text:` is already translated.

2. `compileActionType()` returns bare English while `LayoutItemDelegate` translates the same labels (CLAUDE.md: "All user-visible text must be internationalized")

https://github.com/Kulitorum/Decenza/blob/7934751d82306a75b99d9fc21773e60525e85360/qml/components/library/LibraryItemCard.qml#L148-L153

`LayoutItemDelegate.qml` wraps every one of these labels in `TranslationManager.translate("idle.button.*", …)`. The result reaches the screen for library items with no saved thumbnail, so in a non-English locale the library card reads "Steam" where the real widget reads the translation.

3. Interactive elements with no `Accessible.*` properties (CLAUDE.md: "every interactive element needs `Accessible.role`, `Accessible.name`, `Accessible.focusable: true`, and `Accessible.onPressAction`")

https://github.com/Kulitorum/Decenza/blob/7934751d82306a75b99d9fc21773e60525e85360/qml/pages/settings/StringBrowserPage.qml#L299-L317

That one has role/name/focusable but no `onPressAction`. Fully missing on: `CrashReportDialog.qml:372-384` (View issue on GitHub), `ColorEditor.qml:180-224` (RGB/HLS slider), `VisualizerMultiImportPage.qml:538-548` (shot row — while the sibling deselect target below it has the full set), `SettingsConnectionsTab.qml:153-163` (Cancel) and `:781-795` (device row, which also has no `text:` for Qt to derive a name from), `SettingsMachineTab.qml:479-490`, `SettingsScreensaverTab.qml:737-766`.

4. Hardcoded colours instead of Theme tokens (CLAUDE.md: "never hardcode colors, font sizes, spacing, or radii")

https://github.com/Kulitorum/Decenza/blob/7934751d82306a75b99d9fc21773e60525e85360/qml/pages/settings/LayoutEditorZone.qml#L212-L214

Also `RatingInput.qml:121` (`"#ffffff"` for `Theme.primaryContrastColor`), `SettingsAITab.qml:123,134,165`, `AddLanguagePage.qml:117,154` — in the last two, the sibling `Text` uses a Theme token for the identical condition.

These four are all consequences of one judgement call: this PR touches 87 files mechanically, and CLAUDE.md's ["Fix pre-existing violations in any file you touch"](https://github.com/Kulitorum/Decenza/blob/7934751d82306a75b99d9fc21773e60525e85360/CLAUDE.md#L233) puts every pre-existing violation in all 87 in scope. Fixing them here would roughly double the diff and mix i18n/a11y work into a lint cleanup. Worth deciding deliberately rather than by default.

<details>

- [Worth a look] `Qt.rgba(255, 255, 255, a)` passes 0-255 arguments to an API that takes 0.0-1.0 — `ChangeBeansDialog.qml:1451`, `ShotGraph.qml:390`, `HistoryShotGraph.qml:655`. Qt clamps (`qqmlbuiltinfunctions.cpp:342-350`), so it renders as intended white; the code just misstates the contract (score: 60)

</details>

<details>
<summary>2 discarded (below 50)</summary>

- Bare `modelData` in `SettingsConnectionsTab.qml:795`'s `onClicked` while a sibling binding is qualified — resolves against the delegate's own scope, so the pragma doesn't affect it and qmllint accepts it; style only (score: 40)
- "Zero is unverified for `CustomItem.qml` because CI skips it" — the PR does not modify that file, and the script refuses `--update-baseline` when skipping (score: 25)

</details>

skialpine:
All five review findings are fixed, in four commits on top of the rebase.

**Colours** (`5a3762e4`) — the `Qt.rgba()` finding turned out to be more than cosmetic. `StyledComboBox` reads:

```qml
color: control.enabled ? Qt.rgba(255, 255, 255, 0.1) : Qt.rgba(128, 128, 128, 0.1)
```

Both triples clamp to `(1, 1, 1)` (`qqmlbuiltinfunctions.cpp:342-350`), so **a disabled combo box has rendered identically to an enabled one everywhere in the app**. Translated faithfully — 255→1.0, 128→0.5 — across all seven out-of-range sites. Hardcoded colours swapped for the tokens they were approximating, including LayoutEditorZone's `"orange"`/`"#64B5F6"` widget-category literals, which now follow the theme via `sourceBadgeOrangeColor`/`sourceBadgeBlueColor`.

**Accessibility** (`b2807e7d`, `cda10ff1`) — all eight elements now carry role, name, focusable and an action. The ColorEditor channel sliders are declared `Accessible.Slider` with increase/decrease actions rather than buttons, since that is what they are. The two list rows set `contentItem` rather than `text`, so Qt had no string to derive a name from; both now name themselves from their model data and expose checked state.

My first attempt used `Accessible.value`, which does not exist — `QQuickAccessibleAttached` has only role/name/description/id/ignored/labelledBy/labelFor plus the bool states (`qquickaccessibleattached_p.h:55-71`). The gate caught it as a `missing-property`; the value is in the accessible name now.

**i18n** (`e580c9af`) — every flagged string, plus CommunityBrowserPage's four filter combo models (49 entries). Those keep their separate value arrays, so the index-to-value mapping is untouched. WeatherItem's spoken announcement was built by concatenation and now uses one %-placeholder string so translators can reorder it.

LibraryItemCard was the only one with present-tense user impact: `compileActionType()` returned bare English, so a library card read "Steam" in every locale while the identical widget rendered by `LayoutItemDelegate` read the translation. It now uses the **same `idle.button.*` keys**, fallback for fallback, so English is unchanged.

Where an existing key covered the exact string it is reused; where wording genuinely differed (the community variable/action lists) a new page-local key carries the current English rather than silently adopting different text.

Verified: qmllint 222/222, all four text-invariant checks pass, 108 tests pass.

Two notes on scope. `StyledComboBox` and `PresetButton` were not previously in this branch — pulled in because they carry the same `Qt.rgba` defect and one of them is the actual disabled-state bug. And the colour/i18n substitutions change pixels: the disabled combo box, the AI provider tile's green tint, the language-card border, and the layout-editor chip colours all shift slightly. Worth a look alongside the screen list.

PR #1698, opened 2026-07-29
Recover AOT compilation in Settings, Theme and IdlePage

skialpine:
## What

qmlcachegen was refusing to compile chained lookups through `Settings` and `MainController`. A non-final property can be shadowed by a subclass, so `Settings.theme.x` degraded the base to `var` and bailed with *"Cannot use shadowable base type for further lookups"*. Per [`qqmljsshadowcheck.cpp:196-198`](https://code.qt.io/cgit/qt/qtdeclarative.git/tree/src/qmlcompiler/qqmljsshadowcheck.cpp), a final property is the **only** escape. Nothing subclasses either class and no QML redeclares these names, so `FINAL` states a fact that was already true.

Then annotated the untyped functions in the two hottest QML files — an unannotated function is not compiled, and neither is any *call* to it, so a handful of definitions poison many call sites.

| | Before | After |
|---|---|---|
| Global AOT | 44.7 % | **60.6 %** |
| `Theme.qml` | 11.5 % | **34.5 %** |
| `IdlePage.qml` | 43.4 % | **51.9 %** |
| shadowable-base skips | 574 | **72** |

> **Correction.** This table first read `46.4 %`. That was measured against a build which had recompiled only the two edited QML files, leaving 213 units' `.aotstats` describing the previous state. The real figure is **60.6 %** — annotating `Theme.qml` unblocked ~4,200 call sites tree-wide (`call to untyped JavaScript function` went **4,681 -> 499**), and that only became visible once a `Q_OBJECT` header edit forced all 215 units to regenerate. `BUILD_PERFORMANCE.md` now carries the check that catches this.

`Theme.qml` is the one that matters — it is read on every page construction and was the worst-compiled file in the app.

## Three findings worth more than the percentages

**1. Removing a skip reason does not mean the binding compiles.**

| | skips removed | AOT gained |
|---|---|---|
| Pass 1 (sub-object accessors) | 429 | **394** (92 %) |
| Pass 2 (remaining 218 settings props) | 73 | **15** (21 %) |

58 of pass 2 bindings simply failed for a *different* reason instead. Pass 2 is kept anyway — `FINAL` is correct on all of them and stops the class returning — but anyone extending this work will over-predict without knowing that.

**2. Type annotations coerce, and that is a live hazard.** `maxLen: int` would have turned `maxLen ?? 2000` into `0 ?? 2000` = `0`, truncating **every accessible string to an ellipsis**. The four optional parameters take `var`. Being conservative with `var` elsewhere costs 67 remaining "ambiguous conversion" skips in `Theme.qml`; tightening them is the same silent-coercion risk and **nothing tests QML**.

**3. One annotation caused a real runtime bug, and I do not understand why.** Annotating the seven `TemperatureDisplay` wrappers made qmlcachegen compile `tempUnitSuffix()` into something returning `undefined`, which surfaced as `Unable to assign [undefined] to QString` at `RecipeEditorPage.qml:435` and `:495` in the running app. The C++ signatures are fine (`unitSuffix(bool) -> QString`). I reverted the smallest set that certainly contains it rather than guess at a mechanism. **If someone re-annotates those seven later, it will come back.**

## Test change

`tst_textescaping` extracts these function bodies out of `Theme.qml` and evaluates them in a plain `QJSEngine`, where `function f(s: string)` is a JavaScript `SyntaxError`. Fixed in the test — strip annotations from the signature only, since bodies legitimately contain `:` in ternaries. Otherwise a test dictates that the hottest QML in the app stay un-annotated.

## Verification

- **108 tests pass**, 0 failed, 0 with warnings
- qmllint gate **222/222**
- App launched, idle page reached clean, `RecipeEditorPage` errors confirmed gone with the same code path exercised

## What a reviewer should push back on

- **Blast radius.** 254 property declarations across 14 headers for a 1.7-point global AOT gain. Pass 2 alone was 218 declarations for 15 bindings.
- **Desktop only.** Nothing here is verified on Android or iOS. Android has the V4 JIT (`configure.cmake:96-100`); **iOS does not** — Apple forbids it — so AOT is the only compiled path there and this is worth strictly more on iOS than the macOS numbers suggest.
- **The startup harness cannot measure this.** Three launches of identical code gave **3.4 s, 6.4 s and 13.4 s** for the AOT-sensitive window. The variance tracks window-focus churn and swamps any AOT effect by an order of magnitude. `BUILD_PERFORMANCE.md` does not warn about this; relevant to anyone about to run that A/B on a tablet and trust one number.

## Not done, deliberately

`IdlePage.qml` still has 125 `Cannot retrieve a non-object type by ID: idlePage` skips. It is systemic — 81 files, weak correlation with `pragma ComponentBehavior: Bound` (51 % vs 24 %) — and I could not explain it, so I left it alone.

### DIFF
--- a/docs/CLAUDE_MD/BUILD_PERFORMANCE.md
+++ b/docs/CLAUDE_MD/BUILD_PERFORMANCE.md
@@ -185,13 +185,18 @@ earlier draft said the swappable device handles "realistically cannot" be typed,
 because `ScaleDevice` and `Refractometer` are reassigned at 11 sites as hardware
 connects and disconnects, while the stable globals were the fixable ones. Measured:
 
-| | draft prediction | actual, now |
-|---|---|---|
-| `ScaleDevice`, `Refractometer` | unfixable | **0 skips** |
-| `AccessibilityManager` | fixable | **0 skips** |
-| `MainController` | fixable | 313 -> 152 |
-| `Settings` | fixable | 595 -> **539** |
-| `TranslationManager` | fixable | 1831 -> **1790** |
+| | draft prediction | after the cleanup | after #1698 |
+|---|---|---|---|
+| `ScaleDevice`, `Refractometer` | unfixable | **0 skips** | 0 |
+| `AccessibilityManager` | fixable | **0 skips** | 0 |
+| `MainController` | fixable | 313 -> 152 | **shadowable-base share fixed** |
+| `Settings` | fixable | 595 -> 539 | **shadowable-base share fixed** |
+| `TranslationManager` | fixable | 1831 -> **1790** | 1787 — permanent, see below |
+
+The middle column is the state the prediction was judged against; do not read it as
+current. #1698 then took the *shadowable-base* cause from 574 skips to 72 project-wide,
+which is the part of `Settings`/`MainController` that was fixable at all. Their residual
+skips now sit in other causes (untyped calls, unresolved ids), counted in the table above.
 
 The swappable ones went to zero because swapping moved *behind* a stable singleton
 that re-points its internals — mutability was never the obstacle, dynamic scoping
--- a/docs/CLAUDE_MD/BUILD_PERFORMANCE.md
+++ b/docs/CLAUDE_MD/BUILD_PERFORMANCE.md
@@ -110,38 +110,71 @@ It is not qmlcachegen — it is compiling what qmlcachegen emits: **2.0M lines**
 generated C++ across 216 files, every one containing AOT-compiled functions. Any
 lever aimed at this cost should target the generated C++, not the tool.
 
-## AOT coverage, and why it is only 45 %
+## AOT coverage, and the two levers that moved it
 
-**Re-measured 2026-07-29**, after the QML cleanup (#1665, #1688, #1690, #1695)
-took `setContextProperty` to zero and qmllint to 222/222. The previous numbers,
-and the prediction built on them, are kept below because the prediction turned
-out to be **wrong in both directions** and that is the useful part.
+**Re-measured 2026-07-29** after the QML cleanup (#1665, #1688, #1690, #1695) and
+again after #1698 (`FINAL` on the settings/controller accessors, type annotations
+in `Theme.qml` and `IdlePage.qml`).
 
 ```
-total bindings/functions  29097
-  AOT compiled            13017  (44.7%)
-  skipped -> interpreter  14665  (50.4%)
-  partial                  1415  ( 4.9%)
+                          post-cleanup      post-#1698
+total bindings/functions       29097           29097
+  AOT compiled            13017 (44.7%)   17631 (60.6%)
+  skipped -> interpreter  14665 (50.4%)   10051 (34.5%)
+  partial                  1415 ( 4.9%)    1415 ( 4.9%)
 ```
+
+**Read the measurement warning below before trusting any number you take
+yourself.** #1698 reported +1.7 points for days because every intermediate
+measurement was taken against a build that had recompiled only the two edited
+QML files. The other 213 units' `.aotstats` still described the *previous* state,
+so the change looked local. The real figure only appeared once a `Q_OBJECT`
+header edit invalidated `Decenza.qmltypes` and forced all 215 units to
+regenerate. **Before believing an AOT number, check that the `.aotstats` files
+are newer than your edit:**
+
+```bash
+find <build>/.rcc/qmlcache -name '*.aotstats' -newer <the file you changed> | wc -l
+```
+
+If that is not ~216, you are reading the last build, not this one.
+
+**The untyped-function cascade is the whole story.** Annotating ~32 functions in
+`Theme.qml` alone took `call to untyped JavaScript function` from **4,681 skips to
+499** — because `Theme.*` is called from every file in the app, so typing one
+file unblocked roughly 4,200 call sites elsewhere. This is the 8:1 ratio
+described below, paying off in the direction the ratio predicted. The `FINAL`
+work, by contrast, was worth 429 skips: real, but an order of magnitude smaller.
 
 Grouped by root cause:
 
 | Skips | Share | Cause |
 |---|---|---|
-| 4681 | 31.9 % | call to untyped JS function |
-| 2751 | 18.8 % | unresolved id / model role |
-| 2694 | 18.4 % | member on unresolved type |
-| 2186 | 14.9 % | other |
-| 1787 | 12.2 % | `TranslationManager.translate` — callable `Q_PROPERTY` |
-| 566 | 3.9 % | untyped function definition |
+| 2753 | 27.4 % | unresolved id / model role |
+| 2681 | 26.7 % | member on unresolved type |
+| 1787 | 17.8 % | `TranslationManager.translate` — callable `Q_PROPERTY` |
+| 1729 | 17.2 % | other |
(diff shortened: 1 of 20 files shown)


skialpine:
## Review pass applied — and the headline number in this PR was wrong

Four agents reviewed this. Two of them **contradicted each other** on the coercion mechanism, so I read the Qt source myself. Resolving that changed the code, not just the comments.

### The number was wrong: 44.7 % → **60.6 %**, not 46.4 %

Every intermediate measurement was taken against a build that had recompiled only the two edited QML files. The other 213 units' `.aotstats` still described the *previous* state, so the change looked local. The real figure appeared only when a `Q_OBJECT` header edit invalidated `Decenza.qmltypes` and forced all 215 units to regenerate.

Annotating `Theme.qml` unblocked ~4,200 call sites **tree-wide** — `call to untyped JavaScript function` went **4,681 → 499** — because `Theme.*` is called from everywhere. That is the 8:1 cascade paying off.

`BUILD_PERFORMANCE.md` now carries the corrected numbers plus the one-line check that catches this trap:
```bash
find <build>/.rcc/qmlcache -name '*.aotstats' -newer <file you changed> | wc -l
```

### Correctness fix

A `string` annotation does **not** turn `undefined` into an empty string — it becomes the literal `"undefined"` (`qv4jscall_p.h:337` → `qv4runtime.cpp:618`), and this applies in the plain interpreter too, not only under AOT (`qv4function.cpp:68-73` routes `JsTyped` through `coerceAndCall`).

That silently killed the `if (!text) return ""` guard in six `Theme.qml` functions across 46 call sites. Not live today, but one future binding onto a `var` model role would render the word "undefined" on screen, or announce it via TalkBack through `toAccessibleText`. Those six parameters are now `var`, which coerces nothing — behaviour identical to `main`, functions still compiled.

### Docs that told the next person to do the wrong thing

- **`SETTINGS.md` steps 1 and 3** specified `Q_PROPERTY` without `FINAL`. `CLAUDE.md` mandates reading that checklist before adding a domain, so following it verbatim silently cost AOT. Both now require it.
- **`BUILD_PERFORMANCE.md`** described this PR's work as "the largest remaining self-inflicted bucket… genuinely fixable" — an open action item after it was closed.
- `settings.h` cited an openspec path that has since been archived.

### Over-claims in my own comments, corrected

"a final property is the only escape" is **false** — `qqmljsshadowcheck.cpp` returns `NotShadowable` at nine points. "cost 574 AOT skips, 97 on `theme`" sat above twelve lines worth at most **429**, and conflated `Settings.theme` with the unrelated `qml/Theme.qml`. "the one warm bucket" asserted a performance claim this PR never profiled.

### Landmines now signposted in source

`Theme.qml:406` — do not annotate the seven temperature wrappers, with the exact symptom and an admission the mechanism is unexplained. `Theme.qml:352` — `cap: int` is safe only because both callers resolve `?? 2000` first. `tst_textescaping.cpp` — states plainly that stripping annotations means coercion is untested there, so a green run proves nothing about it.

### Test hardening

The strip regex could silently corrupt a default value containing `: ident` inside a string or regex literal (`sep = "a: b"` → `sep = "a"`), still parse, and leave the test green while exercising a function `Theme.qml` does not have. It now refuses to strip when parameter text contains a quote, slash, brace or paren, and verifies the comma count survives.

**108 tests pass, gate 222/222.** Still unverified: Android and iOS.

skialpine:
### Code review

Found 6 issues. All are in this PR's own comments, docs and description — the code itself audited clean across five reviewers.

1. `BUILD_PERFORMANCE.md` contradicts itself. The "actual, now" table still shows the pre-fix `Settings 595 -> 539` and `MainController 313 -> 152`, while a bullet below says the same bucket was FIXED at 574 -> 72. CLAUDE.md names this file as the source of truth for AOT coverage, so a reader cannot tell which is current.

https://github.com/Kulitorum/Decenza/blob/1e084fdb05b7564e408fa907e56a9c2ef4091718/docs/CLAUDE_MD/BUILD_PERFORMANCE.md#L187-L194

2. Three irreconcilable numbers for the same measurement. This comment claims the twelve accessors accounted for 429 skips and recovered 394 bindings; `BUILD_PERFORMANCE.md` says shadowable-base went 574 -> 72, a recovery of 502. Nothing states why they differ, and 394 is not reproducible from the repo. (CLAUDE.local.md: "If a count looks plausible, that is not evidence it is right.")

https://github.com/Kulitorum/Decenza/blob/1e084fdb05b7564e408fa907e56a9c2ef4091718/src/core/settings.h#L84-L89

3. "Theme.qml holds 85 of the repo's 174 `Settings.theme.` reads" mixes counting methods — 85 is a line count, 174 an occurrence count. A consistent method gives 88/174 or 85/163; the stated fraction is not reproducible either way.

https://github.com/Kulitorum/Decenza/blob/1e084fdb05b7564e408fa907e56a9c2ef4091718/src/core/settings.h#L87-L90

4. The stated mechanism for leaving scalars non-final is wrong. Scalar properties DO reach the shadow check — every lookup goes through `checkShadowing()` and its `isFinal()` test. What they never reach is `checkBaseType()`'s hard error at `qqmljsshadowcheck.cpp:248`, because nothing chains a further lookup off them. As written it implies scalars are structurally exempt from the pass.

https://github.com/Kulitorum/Decenza/blob/1e084fdb05b7564e408fa907e56a9c2ef4091718/src/controllers/maincontroller.h#L100-L103

5. The test's founding invariant is eroded with no assertion guarding it. The header says it runs the REAL functions, "not a C++ reimplementation… A copy would drift from the shipping code and then assert that the copy is correct, which is worse than no test." Stripping annotations means the QJSEngine now evaluates a transformed copy, and argument coercion — the exact bug class this PR hit — is structurally invisible. The gap is documented honestly, but the follow-up QML-engine test it names does not exist.

https://github.com/Kulitorum/Decenza/blob/1e084fdb05b7564e408fa907e56a9c2ef4091718/tests/tst_textescaping.cpp#L3-L6

6. The PR description still reports `Global AOT 44.7% -> 46.4%`. A later comment corrects this to 60.6% and the branch's `BUILD_PERFORMANCE.md` carries 60.6%, but the description was never edited — so reviewers read a number that understates the change by an order of magnitude before reaching the correction.

<details>

- [Worth a look] `tst_textescaping.cpp` comment says a parameter default would "fail loudly"; a simple `pixelSize: var = 16` contains none of the guarded characters and strips silently (score: 72)
- [Worth a look] `Theme.qml` claims the seven temperature wrappers are "the only unannotated functions left in this file" — nested `h(x)` and `linearise` also are (score: 68)
- [Worth a look] `settings.h` marks its own scalar leaves FINAL while `maincontroller.h` deliberately does not, and `SETTINGS.md` step 1 carries no carve-out — inert, but the commit contradicts its own stated rule (score: 65)
- [OK to skip] `QML_GOTCHAS.md` "never redeclare Qt FINAL properties" vs `SETTINGS.md` "always add FINAL"; different senses, no cross-reference (score: 55)

</details>

Discarded below 50: a theory that `_emojiAssetPath` could hit the same unexplained miscompile as `tempUnitSuffix` and emit `` for every emoji (score: 33 — that failure would be unmissable in normal use, so its absence is evidence against).

Verified clean by multiple reviewers: every Qt source citation is exact; the `FINAL` rollout is complete with no property missed and no bulk-edit damage; nothing subclasses `Settings`/`MainController`; all `int`/`real` annotated params provably receive live numeric values; the extract lambda has no index or regex defect for the six functions it handles.

PR #1702, opened 2026-07-29
docs(openspec): propose upgrade-qt-6-12

skialpine:
Creates the `upgrade-qt-6-12` change — the precondition every item in `charts-qt-6-12-polish` waits on. Qt 6.12 GA is **2026-09-22** and its branch is feature-frozen, so the delta was verified against `qt/qtbase` and `qt/qtgraphs` at `ref=6.12` rather than guessed from roadmaps. Docs-only.

**The change is marked BLOCKED on a product decision**, because one finding is not a build detail.

## Qt 6.12 drops iOS 17

Qt 6.12's iOS page: **"iOS 18 or higher (including iOS 26)"**, **"Xcode 16 (iOS 18 SDK) or higher"**. Qt 6.11 was iOS 17. So:

- `CMakeLists.txt:160` must go from `CMAKE_OSX_DEPLOYMENT_TARGET "17.0"` to `"18.0"`.
- **The iPad7,4 used for iOS testing here cannot run iPadOS 18 at all** (A10X; iPadOS 18 needs A12+). The `devicectl install` deploy path stops working, and the iOS Simulator is not an option on this Mac.
- Existing iOS 17 users stop being offered updates.
- macOS minimum also rises to 14.4.

The proposal lays out three options — take iOS 18, hold iOS on 6.11.1 while everything else moves, or defer the upgrade — and is written for the first, with each divergent task marked `[(b) differs]`. Nothing else in the change depends on which is chosen. **This needs your call before work starts.**

## The Android accessibility overrides mostly go away — but not the crash patch

Verified on the **`6.12` branch**, not `dev` and not a changelog:

| Bug | On 6.12? | Evidence |
|---|---|---|
| QTBUG-145786 keyboard opens on a11y focus | **Yes** | `41e6aecb7cd3` (2026-06-23); dev commit carries `Pick-to: 6.12 6.11` |
| QTBUG-118858 typed characters not spoken | **Yes** | `505853d02184` (2026-07-16), touching **both** the platform plugin and the jar — the same two artifacts we patch |
| QTBUG-140490 / QTBUG-144207 `AndroidDeadlockProtector` abort | **No** | `qandroidplatformopenglwindow.cpp:68` on `6.12` still reads `qFatal("Failed to acquire deadlock protector for %s.", …)`. Gerrit 735089 has no `Pick-to:` footer, so it targets `dev` = 6.13 |

So: **delete the `Qt6Android.jar` override outright** (it existed only for the a11y Java classes), and rebuild the `.so` from 6.12 carrying the crash patch alone — dropping the 12 a11y commits. The tasks first try the better outcome: ask for 735089 to be picked to 6.12, in which case `android/qt-overrides/` is deleted entirely, which is the README's own stated preference over rebuilding. The crash is worth the remaining patch: 9 of 42 Android reports since 2026-01-18, and 2 of 2 on 6.11.1 builds.

The tasks also require verifying both TalkBack behaviours on device *after* going stock — an upstream regression restores the patch — plus re-checking 6.12's new a11y behaviour we never had (scrolled-viewport announcements, expandable state, `ButtonDropDown` classname).

## Also captured

- **JDK 17 → 21** (`android-release.yml:69`) — a 6.12 requirement.
- **`nightly-sanitizers.yml`** is a 7th workflow to bump; "the six platform workflows" misses it.
- Re-test the Windows `aqtsource` pin (`TODO(qt-6.11)`) against 6.12 and drop it if PyPI head works.
- **Qt Bluetooth**: no new API, but 6.12 changes Android scan-record UUID/length parsing and defers `canceled()` to match classic scan — our scan-stop path orders on that signal, so on-device discovery verification is a task, not an assumption. Nothing about CCCD, so the native-CoreBluetooth iOS scale transport still can't be retired.
- Record whether the shipped Qt has `graphs-2d-high-performance-backend` compiled in, which is what `charts-qt-6-12-polish` §3 is gated on. Default answer to "build qtgraphs from source for it" is no.
- MQTT is unaffected — Decenza uses Eclipse Paho C, and Qt ships no open-source `qtmqtt` binaries anyway.
- CMake: our `cmake_minimum_required(VERSION 3.21)` stays valid (6.12 needs ≥3.16 there), but 3.25 is the recommended *configuring tool* version — a task confirms what each toolchain actually invokes.

## build-config delta

`RENAMED` the framework requirement to Qt 6.12, `MODIFIED` it (adding a scenario for the Android JDK/NDK toolchain and one for `nightly-sanitizers.yml`), `MODIFIED` the iOS deployment target to 18.0 with a scenario saying a green iOS build proves compilation and not device coverage, and `ADDED` **Patched Qt Platform Artifacts Are Version-Locked**: fail the build on a `BUILT_AGAINST_QT` mismatch rather than falling back to stock, delete rather than rebuild when upstream fixes a bug, justify that deletion from the release branch rather than `dev`, and verify the behaviour on device after dropping a patch.

`openspec validate --all` → **136 passed, 0 failed**.

## One incidental fix

`.gitignore`'s unanchored `build-*/` also matched `openspec/**/specs/build-config/`, so `git add` of a change directory **silently skipped its build-config delta spec** — which is how this PR nearly shipped without its spec file. The pattern protected nothing (the only `build-*` directories in the tree are those three spec dirs), so it is now anchored to the repo root as `/build-*/`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,9 +5,11 @@
 *.jks
 *.keystore
 
-# Build directories
+# Build directories. Anchored to the repo root on purpose: unanchored `build-*/`
+# also matched `openspec/**/specs/build-config/`, silently ignoring delta specs for
+# the build-config capability (a `git add` of the change directory just skipped them).
 build/
-build-*/
+/build-*/
 cmake-build-*/
 
 # CTest artifacts (regenerated on every test run; a stray top-level Testing/
--- a/openspec/changes/charts-qt-6-12-polish/proposal.md
+++ b/openspec/changes/charts-qt-6-12-polish/proposal.md
@@ -10,11 +10,10 @@ This change collects every visual / API / performance item that was deferred dur
 
 Decenza will upgrade to Qt 6.12 in a separate `upgrade-qt-6-12` change after 6.12 GA (2026-09-22). This change runs after that upgrade lands.
 
-**`upgrade-qt-6-12` now exists** (`../upgrade-qt-6-12/`) and is blocked on a product decision of its
-own: Qt 6.12 raises the iOS minimum from 17.0 to iOS 18, dropping a device class including the iPad
-used for iOS testing here. Whichever way that goes, it decides when this change can start — and if it
-goes to "hold iOS on 6.11.1", the charts items still proceed, since every graph surface is shared and
-would be built by the 6.12 toolchain on Android and desktop.
+**`upgrade-qt-6-12` now exists** (`../upgrade-qt-6-12/`) and is ready to run at GA. It raised one
+product question — Qt 6.12 lifts the iOS minimum from 17.0 to iOS 18, dropping pre-A12 devices —
+which was decided on 2026-07-29: take iOS 18, one Qt version on every platform, and explain the
+inherited floor in the release notes. So the only thing gating this change is GA itself.
 
 ## Verification against Qt 6.12 (done 2026-07-29, Beta 2 tree)
 
--- a/openspec/changes/charts-qt-6-12-polish/proposal.md
+++ b/openspec/changes/charts-qt-6-12-polish/proposal.md
@@ -10,6 +10,12 @@ This change collects every visual / API / performance item that was deferred dur
 
 Decenza will upgrade to Qt 6.12 in a separate `upgrade-qt-6-12` change after 6.12 GA (2026-09-22). This change runs after that upgrade lands.
 
+**`upgrade-qt-6-12` now exists** (`../upgrade-qt-6-12/`) and is blocked on a product decision of its
+own: Qt 6.12 raises the iOS minimum from 17.0 to iOS 18, dropping a device class including the iPad
+used for iOS testing here. Whichever way that goes, it decides when this change can start — and if it
+goes to "hold iOS on 6.11.1", the charts items still proceed, since every graph surface is shared and
+would be built by the 6.12 toolchain on Android and desktop.
+
 ## Verification against Qt 6.12 (done 2026-07-29, Beta 2 tree)
 
 Every item below was checked against the actual `qt/qtgraphs` **`6.12` branch** (feature freeze
--- a/openspec/changes/charts-qt-6-12-polish/tasks.md
+++ b/openspec/changes/charts-qt-6-12-polish/tasks.md
@@ -1,7 +1,15 @@
 # Tasks: Qt 6.12 polish for charts migration
 
 **Precondition**: `upgrade-qt-6-12` change archived and Decenza building cleanly on Qt 6.12 GA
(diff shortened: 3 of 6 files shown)


skialpine:
Added a second commit: **`57205bb4` docs(openspec): point charts-qt-6-12-polish at upgrade-qt-6-12**.

`charts-qt-6-12-polish/tasks.md` said its precondition *"does not exist yet"*. It does now, so the polish change records what that means:

- Its precondition is blocked on the **iOS-18 decision** in this PR's change. If that lands on "hold iOS on 6.11.1", the charts items still proceed — every graph surface is shared and would be built by the 6.12 toolchain on Android and desktop.
- **Removed a duplicated instruction.** Both changes were about to own the "is `graphs-2d-high-performance-backend` compiled in?" check. `upgrade-qt-6-12` §0 owns it; the polish change's §3 gate now reads the answer from there instead of repeating the procedure, with the fallback procedure kept only for the case where it hasn't been recorded yet. Same for the "build qtgraphs from source" question — the upgrade proposal already carries it with a default answer of **no**, so §3 closes against that answer rather than re-litigating it.

`openspec validate --strict` passes on both changes.

skialpine:
Third commit: **`462f3484` docs(openspec): decide the iOS floor — take iOS 18, explain it in the notes**.

**Decision recorded: take iOS 18, one Qt version on every platform.** Status goes BLOCKED → **READY**; the only thing gating the change now is 6.12 GA. The two rejected options (hold iOS on 6.11.1, defer the upgrade) are written down with their reasons so they don't get re-proposed, and the `[(b) differs]` task branches are deleted rather than left as dead instructions.

The obligation that comes with it is tracked in three places instead of being prose:

- **`tasks.md` §2** — a release-notes task with the actual requirement: state the new minimum, state that the cause is the **Qt 6.12 upgrade** (Qt dropped iOS 17; every Qt 6.12 app inherits the floor), name the pre-A12 devices that can't go to iOS 18, and say Android/desktop/Linux are unaffected so a platform floor isn't read as an app-wide one. Factual and short — the cause is external, so say so without apologising.
- **`tasks.md` §0** — marks it not-shippable-without, so it can't quietly slip.
- **`specs/build-config/spec.md`** — new scenario *"A raised platform minimum is explained to users"*, which generalises it: any release raising any platform's minimum OS must give the new minimum, the cause, the affected devices, and — where the cause is upstream rather than ours — say so plainly. That way this isn't a one-off note about 6.12.

Also added: check whether the wiki manual states a minimum iOS version (the other place a user would look), asking before pushing the wiki edit per the usual convention.

`charts-qt-6-12-polish` updated in turn — its precondition is no longer described as blocked, only GA gates it. `openspec validate --all` → 136 passed, 0 failed.

PR #1711, opened 2026-07-30
Log weight-timed steaming's scale-vs-fallback decision at steam-start

skialpine:
## Summary
- Diagnostics-only: adds debug logging at both weight-timed-steaming scaling decision points (`SteamPage.qml`'s `onIsSteamingChanged` and `syncSteamTimeout()`) — captured milk, master toggle, `steamSecondsPerGram` calibration, selected pitcher, and which duration was applied and why.
- Fixes nothing yet. Reported bug: capture milk on idle → brew via GHC → Shot Review → press GHC Steam → steaming falls back to fixed duration instead of scaling. Two rounds of static investigation ruled out the capture being cleared during the shot or by Shot Review's lifecycle, and recipe-driven pitcher reassignment (user confirmed no recipe/pitcher reselection happens in between). Remaining candidate — missing/zero global calibration, or milk not reaching the scaling call for an unconfirmed reason — needs a real reproduction's log to confirm.
- Full investigation trail and follow-up tasks in `openspec/changes/keep-milk-weight-across-shot/`.

## Test plan
- [x] Built clean via Qt Creator MCP — 223/223 qmllint, no new warnings
- [ ] Reproduce the GHC-from-Shot-Review sequence on beta, pull the debug log, confirm which gate fired
- [ ] Follow-up PR with the actual fix once confirmed

### DIFF
--- a/openspec/changes/keep-milk-weight-across-shot/.openspec.yaml
+++ b/openspec/changes/keep-milk-weight-across-shot/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-30
--- a/openspec/changes/keep-milk-weight-across-shot/design.md
+++ b/openspec/changes/keep-milk-weight-across-shot/design.md
@@ -1,0 +1,65 @@
+## Context
+
+Weight-timed steaming decides between a scaled steam duration and the recipe/preset's fixed duration in `SettingsBrew::scaledSteamTime(index, milkG)` (`src/core/settings_brew.cpp:614-624`):
+
+```cpp
+int SettingsBrew::scaledSteamTime(int index, double milkG) const {
+    if (!milkAutoCaptureEnabled()) return 0;  // toggle gates ALL weight scaling, not just auto-capture
+    QVariantMap p = getSteamPitcherPreset(index);
+    if (p.isEmpty() || p.value("disabled").toBool()) return 0;
+    if (milkG <= 0.0) return 0;
+    double secPerGram = steamSecondsPerGram();
+    if (secPerGram <= 0.0) return 0;  // uncalibrated
+    return qBound(5, qRound(secPerGram * milkG), 120);
+}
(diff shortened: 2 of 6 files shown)


skialpine:
### Code review

Found 2 issues:

1. The new logging can't distinguish "no milk captured" from "milk present but silently gated to zero upstream" — `SettingsBrew::netMilkForPitcher()` has its own zero-out gates (no saved `pitcherWeightG`, or reading outside `[50, 1500]` g) that feed `sessionMeasuredMilkG`/`lastOnScaleMilk` before this PR's logging ever sees them. If the real GHC reproduction logs `sessionMeasuredMilkG=0`, these two causes are indistinguishable from "user never placed a pitcher" — one level short of the PR's own stated goal (`design.md`'s Goals: "which of the four gates blocked scaling... what the actual captured-milk/rate/pitcher values were at that instant").

https://github.com/Kulitorum/Decenza/blob/712b1acdd42f76968ab8300474c69a58f802e4da/src/core/settings_brew.cpp#L603-L612

2. The unchanged comment above the steam-start scaling logic describes a single milk fallback ("falling back to the last on-scale reading"), but the new log line right below it prints two distinct, separately-prioritized values (`sessionMeasuredMilkG` checked first, `lastOnScaleMilk` as fallback per `capturedMilkForScaling()`). A reader diagnosing a captured log against this comment won't know why two values appear or which one actually won.

https://github.com/Kulitorum/Decenza/blob/712b1acdd42f76968ab8300474c69a58f802e4da/qml/pages/SteamPage.qml#L158-L178

---

<details>

- [Worth a look] The two new log blocks (steam-start and `syncSteamTimeout()`) hand-copy the same five-field format and pitcher lookup instead of a shared helper — CLAUDE.md's centralization principle calls out exactly this drift pattern, though impact here is modest (diagnostics only, two call sites) (score: 72)

</details>

PR #1712, opened 2026-07-30
Fix nightly sanitizer cache always going cold on manual dispatch, bump size

skialpine:
## Summary
- `MANUAL` relied on `inputs.use_cache`'s truthiness for the `workflow_dispatch` branch. A manual dispatch via `gh workflow run` / the REST API with no `-f use_cache=...` sends the input unset, which reads as an empty string — falsy — regardless of the declared `default: true`, which only applies to the GitHub web UI's "Run workflow" form. Run 30502511739 built stone cold from exactly this. Now compares against the literal string `'false'`, so only an explicit opt-out goes cold — nightly cron and every manual dispatch default to using the cache.
- Bump `max-size` from `600M` to `1.5G` per leg. The 600M cap kept the cache permanently at 100% full (measured 2026-07-29: 0.6/0.6 GB, 869 cacheable calls, only 56.62% hit rate) — too small for ~870 sanitizer-instrumented compilation units. ASan and UBSan keep separate caches (their objects never hash-match on sanitizer flags, and sharing one key would race between the two parallel matrix legs saving at different times). Repo sits at ~6/10 GB, so the extra ~1.8G total fits comfortably.

## Test plan
- [x] YAML validated (`ruby -ryaml`)
- [ ] Watch the next scheduled nightly run — should restore + save under the new size cap, and a future manual dispatch without `-f` should stay warm

### DIFF
--- a/.github/workflows/nightly-sanitizers.yml
+++ b/.github/workflows/nightly-sanitizers.yml
@@ -158,16 +158,18 @@ jobs:
           # dead code. Written as an implication instead, which has no falsy-B
           # hole.
           #
-          # Also NOT a bare `inputs.use_cache` on the workflow_dispatch side.
-          # A manual dispatch via `gh workflow run` / the REST API with no `-f
-          # use_cache=...` sends the input unset, and an unset boolean input
-          # reads as an empty string here — falsy — regardless of the `default:
-          # true` declared below. That default is a UI-form convenience only;
-          # 30502511739 (2026-07-30) built stone cold from exactly this, with
-          # no `-f` on the dispatch. Comparing against the literal string
-          # 'false' means only an EXPLICIT opt-out goes cold; missing, empty,
-          # or 'true' all mean "use the cache" — nightly and every manual
-          # dispatch, unless someone deliberately asks for a cold build.
+          # `github.event.inputs.use_cache` rather than the bare `inputs.use_cache`
+          # used before — belt-and-braces, not a fix for an observed bug. Run
+          # 30502511739 (2026-07-30) was checked as the suspected case of a
+          # workflow_dispatch losing the declared `default: true` above when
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 2 issues, both fixed in 970a5ef1:

1. The `MANUAL` fix's justification was wrong. Checked run 30502511739's own step-env dump directly — `MANUAL: true`. The workflow_dispatch `default: true` was applied correctly; that run went cold from the space-based yield further down the same step (free space was under `CACHE_HEADROOM_MIN_GB` at the time), which is the safety valve working as intended, not the input-parsing bug the original comment claimed. Kept `github.event.inputs.use_cache != 'false'` as a harmless defensive improvement, but rewrote the comment to state the verified story instead of a disproven one.

https://github.com/Kulitorum/Decenza/blob/970a5ef17de58b59591d318a0f0e851a1d105a77/.github/workflows/nightly-sanitizers.yml#L154-L171

2. `max-size: 1.5G` ignored this job's own 2GB yield floor, not just the 10GB cap. Measured 2026-07-30: repo at 7.15/10 GB with the two sanitizer entries at their old ~0.6G each, so non-sanitizer usage is ~6.1G — leaving only ~1.9G of headroom above the yield floor for both legs combined. 1.5G/leg (2.8G combined) would go negative on that budget: the first night both caches fill up, the yield fires, and it deletes what it's holding to hand the space back — the exact cold-cache cycle this PR exists to fix. Dropped to 800M/leg (1.6G combined), leaving a real ~300M margin.

https://github.com/Kulitorum/Decenza/blob/970a5ef17de58b59591d318a0f0e851a1d105a77/.github/workflows/nightly-sanitizers.yml#L296-L314

---

<details>

- [Worth a look] "declared below" pointed the wrong direction (the default is declared above, line 46) — fixed in the same commit as part of the comment rewrite (score: 60)

</details>

PR #1715, opened 2026-07-30
Root pages at QtQuick.Templates.Page to restore AOT compilation

skialpine:
Referencing an object by `id` fails AOT compilation whenever that object's type comes from **QtQuick.Controls**. qmlcachegen resolves the id, then `genericType()` walks the base chain looking for a scope named `QObject`, does not find one, and returns `m_jsValueType` — a value type — which `qqmljstypepropagator.cpp:637` rejects with "Cannot retrieve a non-object type by ID". `QtQuick/Controls/qmldir` declares no types at all; every concrete type arrives from a style module chosen at run time, which is why the base chain is unresolvable at compile time.

Rooting a page at `QtQuick.Templates.Page` gives the same C++ `QQuickPage` with no style indirection, so the id resolves. 31 pages migrated. `AddLanguagePage.qml` is deliberately excluded: it uses the `StackView` attached property, which qmllint does not count as a use of the Controls import, so the gate fails whether the import is kept or dropped.

Acts on the RCA documented in #1714.

## Measured

Both sweeps on a cache forced consistent (touch every `.qml`, then a full build):

```
                        post-#1698      post-#1715
AOT compiled          17631 (60.6%)   18168 (62.5%)
skipped -> interpreter 10051 (34.5%)    9504 (32.7%)
partial                1415 ( 4.9%)    1415 ( 4.9%)
```

id skips 1,464 → 510. `root` is now the entire remaining id class at 563 skips, all of it `main.qml`'s `ApplicationWindow`.

## Bugs this surfaced

**The migration silently broke an `as Page` cast in a file it never touched.** Under a style, `QtQuick.Controls.Page` is the style's `Page.qml` — a **composite** type — and a composite matches only instances whose own metaobject chain contains it (`qqmltypewrapper.cpp:513-516`: "Rectangle{} is never an instance of CustomRectangle"). `as` is `doInstanceof`, and a failed object cast returns `null` (`qv4runtime.cpp:394-406`). Re-rooting the pages dropped the style composite out of their chain, so `main.qml`'s `pageStack.currentItem as Page` returned null for every page, pinning `Theme.shotChartOnCurrentPage` to false and silently killing the shot-chart background. Fixed to `as T.Page`, which matches Templates-rooted and Controls-rooted pages alike. It compiles, qmllint is silent, and the tests do not instantiate the stack — nothing catches this but reading the Qt source.

**A qmlcachegen narrowing defect in generated code.** `currentIndex: [...].indexOf(step)` on a `StackLayout` began AOT-compiling once the page rooted at `T.Page`, and `QJSList::indexOf`'s `qsizetype` is assigned straight into an `int` with no cast — `-Werror=shorten-64-to-32`. Not fixable at the call site; replaced with a typed `switch`.

## Also in this PR

- **Three unreachable QML files deleted** (604 lines): `DialingAssistantPage.qml` (orphaned 2026-01-10 when its only entry point was removed), `ColorWheel.qml` (dead since the 2026-03-16 theme overhaul), `ProfileDisplayButton.qml` (**never instantiated in any commit**). Found by a reachability sweep; `CMakeLists.txt` and the qmllint baseline updated in step. Nothing in the build detects this — `qt_add_qml_module` membership is enough to compile and ship a file.
- **A dead accessibility map entry** in `main.qml` orphaned by the first of those deletions.
- **A security-inventory correction** in `tst_markdownrenderer.cpp`: it claimed three call sites feeding untrusted text with escaping off and listed bullets summing to four; there are five. `SettingsUpdateTab` has two and never carried the `x2` its neighbour got.
- **A stale empty `StackView.onActivated`** in `ShotDetailPage.qml` whose comment described behaviour `main.qml` has done declaratively for some time.

## Doc corrections in `BUILD_PERFORMANCE.md`

- **Retracts "a mixed QML cache is benign".** Running the app on one logged `Unable to assign [undefined] to QString` at `EspressoPage.qml:882` and `SteamGraph.qml:213`, both on `Theme.tempUnitSuffix()`, in units generated before #1714's annotations while `Theme_qml.cpp` was generated after. That is the same function and the same message as the original report which banned annotations on those wrappers for a release — so the mixed cache reproduces exactly the failure the ban was founded on. Forcing all 213 units to regenerate cleared it, confirmed by re-running espresso and steam.
- **Records that an ordinary rebuild does not clear it.** Merging #1714 here regenerated exactly one of 217 `.aotstats`. qmlcachegen reads `Theme.qml` via `-I <builddir>` but does not declare it as an input and emits no depfile, so ninja cannot know the edge exists.
- **Documents the `as`-cast hazard** as a precondition for re-rooting anything else at Templates: grep the whole tree for `as <Type>`, not just the files you are editing.
- **Re-derives the per-cause table** from the consistent cache, replacing hand-maintained counts.

## Verification

qmllint 213/213 clean. Full suite passes. Visual pass over idle, settings (switch tracks — the documented symptom of the cast failure), shot history, shot detail, a dialog, and push/pop title restore; plus espresso, steam, flush, hot water and the recipe wizard against the simulator. Zero `qrc:` QML errors on a consistent cache, confirmed by re-running the exact actions that previously produced two.


skialpine:
### Code review

Found 5 issues, all fixed in e04dbcca.

1. **The migration silently broke an `as Page` cast in an untouched file** — the one real bug. Under the Material style `QtQuick.Controls.Page` is a *composite* type, and a composite matches only instances whose own metaobject chain contains it (`qqmltypewrapper.cpp:513-516` — "Rectangle{} is never an instance of CustomRectangle"). `as` is `doInstanceof`, and a failed object cast returns `null` (`qv4runtime.cpp:394-406`). Re-rooting the pages at `T.Page` drops the style composite out of their chain, so `pageStack.currentItem as Page` returned null for every page, pinning `Theme.shotChartOnCurrentPage` to `false` and killing the shot-chart background. Nothing catches it: compiles clean, qmllint silent, tests do not instantiate the stack. Fixed to `as T.Page`, which matches Templates-rooted and Controls-rooted pages alike.

   Commit `8a783700` asserted *"No file declares a property typed Page or casts to it."* It was one grep from false. `BUILD_PERFORMANCE.md` now carries that grep as a precondition for re-rooting anything else at Templates.

https://github.com/Kulitorum/Decenza/blob/e04dbccaee4c695c8f1d2c927eb623a4a7866075/qml/main.qml#L1168-L1184

2. **Dead accessibility map entry orphaned by this PR** — deleting `DialingAssistantPage.qml` removed the only `objectName` that `"dialingAssistantPage"` could ever match, stranding it and its `main.pageAiAssistant` translation key.

3. **Security inventory undercount** in `tst_markdownrenderer.cpp`. The header said three call sites and listed bullets summing to four; there are **five** — `SettingsUpdateTab` has two (inline and full-screen release notes) and never carried the `x2` its neighbour got. The file's own docstring calls this "a SECURITY claim". Header now states how to re-derive the count instead of inviting the next hand-adjustment.

https://github.com/Kulitorum/Decenza/blob/e04dbccaee4c695c8f1d2c927eb623a4a7866075/tests/tst_markdownrenderer.cpp#L1-L15

4. **Empty `StackView.onActivated`** in `ShotDetailPage.qml` whose comment described re-asserting the header title. `main.qml` has bound that declaratively off `currentItem.pageTitle` for some time — handler inert, comment describing a mechanism that no longer exists. Removed.

5. **`BUILD_PERFORMANCE.md` left at the pre-migration numbers.** Updated to post-#1715, and recorded that the id-skip column is a *subset* of the hard-skip column — summing them across both sweeps invents a 954-binding discrepancy, which a reviewer did and reported as a measurement bug.

---

Two agent findings rejected rather than fixed:

- **"AOT stats do not sum"** — arithmetic error, not a measurement bug. `17,626 + 10,046 = 18,127 + 9,545 = 27,672` exactly; id skips are counted inside hard skips.
- **Other `indexOf()` -> `currentIndex` sites in newly-AOT files** — a clean build settles it; no further narrowing errors exist.

Verified: qmllint gate 213/213 with zero diagnostics; build clean; full test suite passes; running app log shows zero `qrc:` QML errors. Remaining gap is navigation — 30 of the 31 migrated roots are `Component`s instantiated on push, so they need a visual pass.

PR #1717, opened 2026-07-30
Root the button family, ApplicationWindow and every dialog at QtQuick.Templates; fix the QML cache dependency gap

skialpine:
Referencing a member on an object whose type comes from **QtQuick.Controls** fails AOT compilation: `QtQuick/Controls/qmldir` declares no types, every concrete type arrives from a style module chosen at run time, and qmlcachegen cannot walk that base chain at build time. #1715 fixed it for pages; this does the button family, `main.qml`'s `ApplicationWindow`, and every dialog.

## Measured

On a cache forced consistent:

```
                        post-#1715      this PR
AOT compiled          18168 (62.5%)   20459 (70.3%)
skipped                9504 (32.7%)    7223 (24.8%)
partial                1415 ( 4.9%)    1415 ( 4.9%)
```

`root` and `popup` id skips are gone; the missing-member class fell 2,653 → 676.

**The total moved by less than the parts, and both numbers are right.** Resolving a type lets qmlcachegen reach further into bindings it used to abandon at the first unresolvable member, where it hits the next blocker. `TranslationManager.translate` went 1,833 → 2,253 and is now 31% of everything left — it has grown at every step, and it is the program's floor, not its next target. **Diff per-cause, never on the total.**

## What each part needed

**Buttons** (`AccessibleButton`, `StyledIconButton`, `ActionButton`, `StyledSwitch`) were near-drop-ins: all four already replaced `contentItem`/`background`/`indicator` and use no Material attached property, so nothing of the style's QML reached the screen. What did — the `implicitWidth` formula (`qquickcontrol.cpp:1749-1757`: it lives only in style QML, so a Templates root without it is 0 wide), the insets, and the padding — is re-declared with its Qt source cited.

**`ApplicationWindow`** was one line for 583 skips: Material's `ApplicationWindow.qml` is three lines whose only content is `color: Material.backgroundColor`, already overridden.

**Dialogs were the case where per-file re-rooting would have been wrong.** Unlike the buttons, the style's `Dialog.qml` contributed things genuinely on screen that no dialog overrode — the grow-fade transitions, the modal dim, and the header that renders `title`. Not one of the 131 sites declared `enter`/`exit`. `DecenzaDialog.qml` carries them once. Two facts made a single base sufficient, both checked rather than assumed: all 104 inline blocks override `background` (established by walking each block's braces), and the tree's only `Dialog`-specific API is `onRejected`, which `T.Dialog` has.

`ColoredIcon` is deliberately **not** migrated — its mechanism is Material's internal `IconLabel` doing `icon.color` tinting. Three inline `Popup {}` blocks are left alone too: the base carries `Dialog`'s `modal: true`, which would start dimming the screen behind them.

## Also fixed

**The cross-file QML cache staleness this repo has documented for months.** qmlcachegen resolves QML-declared types by *reading* other files through `-I`, does not declare them as inputs, and has no `--depfile`. Editing `Theme.qml` changed how ~214 units should compile and rebuilt none of them — a build that was quietly wrong, not merely stale. `CMakeLists.txt` now makes every unit depend on every `.qml`. Affordable because of `restat`: qmlcachegen is 205 ms median per file against 3387 ms to compile what it emits, so ninja prunes every unit whose output is byte-identical.

**A real threading defect.** `isDbWorkIdle()` claimed to cover all background DB work but only saw the FIFO worker; eleven read queries spawn detached threads that bypass it, and `initialize()` starts one — so every user of `ShotHistoryStorage` had an uncounted thread running against the DB file. Now centralized behind `runDetachedDbThread()`, counted, with an RAII decrement.

**A latent test flake.** Three `tst_settling` tests budgeted `qWait(50)` loops at 700-750 ms against a 1000 ms ceiling; under the parallel ASan suite one measured 1015 ms, settling completed on the timer, and the path under test never ran.

## Review

Five defects found and fixed, two of which would have shipped: `StyledSwitch` had lost Material's `padding: 8` (switches 12 px shorter, undoing a deliberate touch-target fix), and three dialogs with `title:` rendered no heading. Both traced to the same habit — verifying the 104 inline blocks for one property and then writing conclusions about all 131 into comments as fact. The dependency-wiring gate was also found to be incapable of passing (221 expected vs 214 matchable) and its stated rationale false: `add_custom_command(APPEND)` against a non-matching OUTPUT is a hard configure error, not a silent no-op, established by reproduction.

## Verification

Full suite 110/110. qmllint 221/221. Visual pass over idle, settings, shot history, shot detail and dialogs; transitions measured running at 220-224 ms; the dialog header confirmed rendering by temporarily setting a title, because the log could only show the binding evaluated, not that it was true.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1858,10 +1858,11 @@ foreach(_qml_file IN LISTS QML_FILES)
 endforeach()
 message(STATUS "QML cross-file cache deps wired for ${_decenza_qml_units_wired} units")
 
-# The gate. `add_custom_command(APPEND)` against an OUTPUT that does not match an
-# existing command is a SILENT no-op, so a drift in Qt's path formula would leave
-# this whole block doing nothing while still printing a healthy-looking count
-# above. Verify by making the build system REFUSE, not by reading a number:
+# The gate. NOTE: `add_custom_command(APPEND)` against an OUTPUT that does not match an
+# existing command is a hard configure-time error, NOT a silent no-op — verified by
+# reproduction with this project's CMake 3.30.5. So path drift fails the build by itself,
+# and this target covers the other half: that the accepted dependencies actually reach the
+# generator and cause a rebuild. Exercise it rather than reading the count above:
 #
 #   cmake --build <builddir> --target qml_dep_wiring_check
 #
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1362,6 +1362,7 @@ set(QML_FILES
     qml/components/UnsavedChangesDialog.qml
     qml/components/DatePickerDialog.qml
     qml/components/SwipeableArea.qml
+    qml/components/DecenzaDialog.qml
     qml/components/CrashReportDialog.qml
     qml/components/McpConfirmDialog.qml
     qml/components/De1CommunicationErrorDialog.qml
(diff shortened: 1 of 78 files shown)


skialpine:
### Code review

Found 5 issues. All are now fixed on the branch; listed for the record since several are instructive.

1. `StyledSwitch` silently lost Material's `padding: 8`, shrinking every switch in the app (bug: `material/Switch.qml:19` sets it, and this file's implicit sizes are written in terms of it — `implicitHeight` went `max(28, 8+24+8)=40` to `max(28,24)=28`, `implicitWidth` lost 16px). The pre-existing comment two lines above says that width exists to keep the touch target larger than the drawn switch, so this undid a deliberate accessibility fix.

https://github.com/Kulitorum/Decenza/blob/06784fa461bcb103a39cd9056d2632d82afce245/qml/components/StyledSwitch.qml#L4-L10

2. Three dialogs that set `title:` render no title at all (bug: Material's `Dialog.qml:58-65` renders `title` only through its `header: Label`, and `T.Dialog` supplies no header). Affects `SteamPage.qml:562`, `ConversationOverlay.qml:1226`, `ProfileImportPage.qml:393`. The base's own comment asserted "the four files that show a header or footer declare their own" — a claim derived from the 27 file roots and stated as if it covered all 131 sites.

https://github.com/Kulitorum/Decenza/blob/06784fa461bcb103a39cd9056d2632d82afce245/qml/components/DecenzaDialog.qml#L20-L28

3. `qml_dep_wiring_check` could never pass, and had never been run green. `QML_FILES` is 221 entries = 214 `.qml` + 7 `.js`; `EXPECTED_UNITS` counts all 221 while the regex matched only `_qml.cpp`, so `hit_count` topped out at 214 and the `LESS` test always fired. A gate written to make the build refuse, that refuses unconditionally.

https://github.com/Kulitorum/Decenza/blob/06784fa461bcb103a39cd9056d2632d82afce245/cmake/CheckQmlDepWiring.cmake#L56-L61

4. That gate's stated rationale was factually wrong. `add_custom_command(OUTPUT ... APPEND ...)` against a non-matching OUTPUT is a **hard configure-time error**, not the "silent no-op" claimed in `CheckQmlDepWiring.cmake`, `CMakeLists.txt` and `BUILD_PERFORMANCE.md` — reproduced with this project's own CMake 3.30.5. Path drift fails the build loudly by itself; the check's real value is proving the accepted dependencies reach the generator.

https://github.com/Kulitorum/Decenza/blob/06784fa461bcb103a39cd9056d2632d82afce245/CMakeLists.txt#L1858-L1864

5. Test-helper timing was wrong in the direction that matters, and its bound sat on the threshold it guards. With `SETTLING_WINDOW_SIZE=6` and `SETTLING_CLEAN_CAPTURE_MS=250` at 50ms cadence, the capture gate fires around sample 11 (~500ms), not "around sample 7, ~350 ms" as the comment claimed — and the 20-iteration cap is exactly 1000ms, i.e. `SETTLING_STABLE_MS`, the ceiling it exists to stay under. Also, `QVERIFY2` inside a non-slot helper returns from the helper, so a failure there was followed by assertions against known-bad state.

https://github.com/Kulitorum/Decenza/blob/06784fa461bcb103a39cd9056d2632d82afce245/tests/tst_settling.cpp#L47-L51

<details>

- [Worth a look] `DecenzaDialog`'s "no file in the tree uses `standardButtons`" is false — `SettingsConnectionsTab.qml:1338` sets it on a deliberately-unmigrated `ComboBox.popup: Dialog`. No functional break; claim now scoped. (score: 60)
- [Worth a look] "the only Dialog-specific API is BrewDialog's `onRejected`" is false — `SettingsScreensaverTab.qml:117` is a second user, on a migrated dialog. `T.Dialog` has the signal, so no break. (score: 60)
- [OK to skip] `runDetachedDbThread`'s counter is not decremented if the body throws or `start()` fails, latching `isDbWorkIdle()` false. No production caller waits on it today. (score: 55)

</details>

Verified correct and not re-litigated: every Qt source citation in the new comments (`qquickcontrol.cpp:1749-1757`, `Qt6QmlMacros.cmake:3841-3847`, `qquickpopup.cpp:1274-1279`, `backgroundDimColor` = `0x99303030`, the Material inset/padding values); the "all 104 inline blocks override `background`" count, by brace-depth parsing; the atomic memory ordering in `runDetachedDbThread`; and the absence of any `as <Type>` / `property <Type>` cast for the newly re-rooted types, which was #1715's defect class.

PR #1720, opened 2026-07-30
Stop an equipment edit that only ADDS information from forking the grinder (#1713)

### LINKED ISSUE #1713: [Bug] Grind settings only showing a single digit (no decimal point)
### What happened?

Decenza started showing only a single digit for grind settings in the latest build, when it used to show a decimal before (for example: only 1, 2, 3 instead of 1.1, 1.2, 1.3, 1.4, etc.).
My grinder is an Eureka Single Dose, and single digit is way too course to track.

### Steps to reproduce

open decenza, pick any recipe and try to change the "grind" setting.

### Decenza version

2.0.1

### Platform

Android

### Device / OS version

Decent tablet P80X Android Pie

### Scale (if applicable)

_No response_

[debug-2.zip](https://github.com/user-attachments/files/30552475/debug-2.zip)

_No response_

### Screenshots or video

_No response_

### Anything else?

_No response_

skialpine:
Fixes [#1713](https://github.com/Kulitorum/Decenza/issues/1713).

## The bug

Equipment identity is copy-on-write: editing any identity component of a package that has shots forks a new package and retires the old one. That is right for a *swap* and wrong for *recording what was always there*. The reporter typed the burr make/model into a Eureka he had used for years and got a brand-new grinder.

The history really is gone, not just relabelled — the lookups are keyed on exactly what the fork changed:

- `buildGrinderCalibrationBlock` matches shots on grinder model **and burrs**, so nothing pulled before the edit matches. The reporter's log shows it down to a single pair: `pairs= 1 ... validUGS=[ 1.5 , 5 ]`.
- Dial-in session grouping keys on `COALESCE(equipment_id, 0)`, so it splits at the fork too.

## What changed

**Enrichment no longer forks.** `supersedeOrEditStatic` gains a tier between "unchanged" and "fork": when every differing component was EMPTY and now has a value, the edit applies in place — same id, still in inventory, every shot still attached — whatever the shot count. Replacing a value or clearing one still forks, so a genuine burr swap keeps the earlier shots on the gear they were pulled with. Puck prep compares whole canonical strings, so `"wdt"` → `"wdt,rdt"` is a change, not enrichment.

**`equipment_merge` MCP tool** (level `settings`) folds one package into another by explicit id: shots, bags **and recipes** repointed, supersession pointers carried over, target revived, source deleted — atomically. Accepts a retired source or target, which is what makes undoing a fork possible. Re-selects the target when the merged-away package was active. MCP-only by decision; no app or web UI.

**Migration 35 heals the users this already happened to.** Each lineage pair (`older.superseded_by = newer.id`) differing *only* by empty-vs-set burrs is folded, newer surviving. Left alone: a burr swap between two named sets, a cleared field, a basket/puck-prep difference, and two lookalike packages with no lineage between them. Runs inside the migration transaction with the version bump gated on it, so a failure retries next launch rather than half-merging. Chains collapse one link per pass, bounded at 10.

**Model matching folds case and whitespace.** Every model-scoped history lookup (6 in `shothistorystorage_queries.cpp`, plus model *and* burrs in the AI calibration block) now folds the way `findPackageByGrinderIdentityStatic` — which *writes* those rows — always has. An exact compare made a model differing only in case or padding read back as a grinder with no history, and an empty result is indistinguishable from a new grinder.

**Pre-existing gap fixed in passing:** `requestDeletePackage`'s reference pre-check never counted `recipes.equipment_id`, so a hard delete could strand a recipe.

## Accepted trade-off

A real burr swap whose *old* burrs were never named is indistinguishable from an enrichment edit, and is treated as one — by the new rule and by the heal. With the old burrs unrecorded there is no difference in the data to preserve, and the alternative is the reported bug. Stated in the change's design.md.

## Tests

Full suite via Qt Creator: **110 passed, 0 failed, 0 warnings.**

- `tst_equipment`: enrichment in place; burr swap still forks; clearing forks; adding to an existing puck prep forks; whole-tuple enrichment stays in place.
- `tst_equipment::mergePackagesById`: fork → merge; shots and bag land on the survivor, survivor revived with no supersession pointer, source gone, third-package lineage follows, refusals are clean no-ops.
- `tst_equipment::healEnrichmentForks`: the fork is reunited; a named-to-named swap and two lineage-less lookalikes survive; a second run folds nothing; a chain heals only its enrichment link.
- `tst_mcptools_write::equipmentMergeMovesHistoryAndDeletesSource`: response shape, on-disk verification, self-merge refused.
- `tst_dialing_blocks::calibrationBlock_grinderIdentityMatchIsCaseAndSpaceInsensitive`: folded lookup finds the history, a different grinder still does not.
- Schema-version expectations bumped 34 → 35 (`tst_dbmigration` ×18, `tst_coffeebags` ×3).

OpenSpec change `fix-equipment-enrichment-fork` archived in the last commit.


skialpine:
### Code review

Found 7 issues:

1. A merge revives a target that is itself superseded, resurrecting a retired package into the inventory. Chain case migration 35 will hit on real databases: A retired into B, then B retired into C; the heal merges A into B and leaves B with `in_inventory = 1` AND `superseded_by = C`. `loadInventoryStatic` filters only on `in_inventory = 1`, so B reappears as a live duplicate carrying the same derived name as C. `healEnrichmentForks` stages exactly this shape but asserts only the count, the remap and the shot pointer.

https://github.com/Kulitorum/Decenza/blob/5c1c78f458c99fcb298397ad2969e25dec6ba4b8/src/history/equipmentstorage.cpp#L1200-L1213

2. `equipment_update`'s description now says an unused package is "always" edited in place. The merge branch is checked before the edit-in-place branch, so an unused package whose new identity matches an existing package returns a different id. The word is new in this PR and the description is read by an LLM deciding how to act.

https://github.com/Kulitorum/Decenza/blob/5c1c78f458c99fcb298397ad2969e25dec6ba4b8/src/mcp/mcptools_write.cpp#L2517-L2525

3. `equipment_merge` never fills in `view.shotCount`, so `packageToJson` reports the survivor as having zero shots — right after moving shots onto it. `equipment_update` had this exact bug fixed in #1596 and carries a comment explaining why; the new tool sits beside that fix and omits it.

https://github.com/Kulitorum/Decenza/blob/5c1c78f458c99fcb298397ad2969e25dec6ba4b8/src/mcp/mcptools_write.cpp#L2763-L2770

4. The enrichment predicate treats a deliberately grinder-less package (basket-only tea setup) as "grinder not yet recorded". Adding a grinder to a used tea package is judged enrichment and applied in place, so shots pulled with no grinder at all start reporting one. `EquipmentItem::isValid()` distinguishes "no row" from "row with empty fields"; `filledIn` does not consult it.

https://github.com/Kulitorum/Decenza/blob/5c1c78f458c99fcb298397ad2969e25dec6ba4b8/src/history/equipmentstorage.cpp#L1378-L1390

5. `equipment_merge` deletes a package without emitting `packagesChanged()`, so an open Equipment page keeps showing it. Tapping it writes the deleted id onto the active bag through `setActiveEquipmentId` before the package lookup fails — a dangling `equipment_id`, which is the orphaning this PR exists to fix.

https://github.com/Kulitorum/Decenza/blob/5c1c78f458c99fcb298397ad2969e25dec6ba4b8/src/mcp/mcptools_write.cpp#L2786-L2795

6. Migration 35 runs the heal even when `m_db.transaction()` returned false, so every merge autocommits individually while the `if (txn) m_db.rollback()` guard silently does nothing — a mid-heal failure leaves a half-applied merge. The added comment claims the writes are "staged for this transaction to roll back", which is only true when the BEGIN succeeded. Separately, `dbutils.h` says to use `DbWriteTxn` for any transaction that reads before it writes, and the comment's justification ("DbWriteTxn refuses to nest") does not apply — the nested `mergePackagesUnlockedStatic` opens no transaction.

https://github.com/Kulitorum/Decenza/blob/5c1c78f458c99fcb298397ad2969e25dec6ba4b8/src/history/shothistorystorage.cpp#L1828-L1840

7. Two added comments call `findPackageByGrinderIdentityStatic` "the identity matcher that WRITES these rows". It is a read-only SELECT; the writers are `updateGrinderItemStatic` / `createPackageWithGrinderStatic`. The same comments claim parity with it, but it only LOWERs the stored side and trims the search value in C++ — the new queries trim both sides, so they are strictly more robust than the function they cite. CLAUDE.md: "An un-sourced claim in a comment gets believed and then licenses wrong code."

https://github.com/Kulitorum/Decenza/blob/5c1c78f458c99fcb298397ad2969e25dec6ba4b8/src/ai/dialing_blocks.cpp#L919-L932

<details>

- [Worth a look] The `LOWER(TRIM(IFNULL(...)))` fold predicate is hand-rolled at 12+ sites across three files rather than centralized — CLAUDE.md's centralization rule names formats and policies, not SQL fragments, so it is a judgement call (score: 50)
- [OK to skip] SQLite `LOWER()` folds ASCII only, and `findPackageByNameStatic` rejects SQL LOWER for exactly that reason — but the write-side identity matcher also uses SQL LOWER, so read and write still agree, and a non-ASCII model fails identically on both sides (score: 50)
- [Worth a look] `validateMergeStatic` returns `"samePackage"` for a non-positive id too, collapsing "invalid id" into a reason that says the ids name one package; only reachable by direct callers since the MCP tool pre-validates (score: 50)
- [OK to skip] The `q.finish()` comment cites the `DbWriteTxn` unfinished-SELECT precondition, but the scan is already stepped to exhaustion and this path never calls `DbWriteTxn::begin()`; the call is still right, the stated mechanism is not (score: 50)

</details>

skialpine:
### Code review (second pass, scoped to 848e2df7)

Scoped to the review-fix commit and the spec commit, with the 50 scripted log conversions as the priority target. 49 of the 50 conversions were faithful — argument count, order, tier and tag all preserved. Every finding below is in hand-written code or wording. All seven are fixed in 8ed797a6.

1. Migration 35 took the write lock while a SELECT was still open. The `schema_version` read ends in `LIMIT 1` and is stepped once, so it never reaches `SQLITE_DONE` and Qt never resets it (`qsql_sqlite.cpp:326-332`) — the connection sits in an implicit read transaction, which is what `DbWriteTxn`'s PRECONDITION forbids, and SQLite skips the busy handler while a connection is in one. Introduced by switching this migration to `DbWriteTxn`; masked today only because no other storage has opened the file yet at that point in startup.

https://github.com/Kulitorum/Decenza/blob/8ed797a67e4a4e4dbdcd2fc31b1de9df9b1cb5e8/src/history/shothistorystorage.cpp#L1841-L1860

2. The fork log line chained eight `.arg()` calls over six free-text fields. Chained `arg` fills the lowest-numbered placeholder remaining anywhere in the string, including inside text a previous `arg` substituted, so a burrs value containing a literal `%7` would capture the model — corrupting the before/after comparison the line exists to record.

https://github.com/Kulitorum/Decenza/blob/8ed797a67e4a4e4dbdcd2fc31b1de9df9b1cb5e8/src/history/equipmentstorage.cpp#L1676-L1694

3. Four INFO lines announced an outcome before the transaction committed. `done()` rolls back and returns -1 on a failed commit, so the log asserted a fork or merge the next line contradicted (LOGGING.md: "Don't report a transition that did not happen") — which the same change already got right for migration 35.

https://github.com/Kulitorum/Decenza/blob/8ed797a67e4a4e4dbdcd2fc31b1de9df9b1cb5e8/src/history/equipmentstorage.cpp#L1630-L1646

4. The equipment import failure was a bare `qWarning` in a file this change had already touched for equipment logging, so a `[Equipment]` filter could not find it.

https://github.com/Kulitorum/Decenza/blob/8ed797a67e4a4e4dbdcd2fc31b1de9df9b1cb5e8/src/history/shothistorystorage.cpp#L3729-L3735

5. The `q.finish()` comment carried a second wrong mechanism after the first was corrected. Qt resets the statement itself on `SQLITE_DONE`, and SQLite does not refuse the write (checked against libsqlite3), so the call is defensive only. Both disproven stories are now recorded next to it.

https://github.com/Kulitorum/Decenza/blob/8ed797a67e4a4e4dbdcd2fc31b1de9df9b1cb5e8/src/history/equipmentstorage.cpp#L1383-L1396

6. `equipment_merge`'s description promised the survivor is returned to inventory, which stopped being unconditional when the revive was made conditional. An assistant repairing a three-way split would read that as a failed merge and retry.

https://github.com/Kulitorum/Decenza/blob/8ed797a67e4a4e4dbdcd2fc31b1de9df9b1cb5e8/src/mcp/mcptools_write.cpp#L2733-L2740

7. The logging header's example lines did not match the format strings they illustrate, so grepping a submitted log for a shown phrase would miss.

https://github.com/Kulitorum/Decenza/blob/8ed797a67e4a4e4dbdcd2fc31b1de9df9b1cb5e8/src/history/equipmentlogging.h#L12-L20

Checked and clean: the revive CASE expressions across all four target states, `gainingAGrinder` against `updateGrinderItemStatic`'s `wantNone`, heal-loop termination, `fillShotCount` captures and lifetime, `notifyPackagesChangedExternally` thread affinity, and the archived spec text against the code and tests.

Full suite after the fixes: 110 passed, 0 failed. Log-marker gate green.

Issue #639, opened 2026-04-04
Improve multi-shot conversation context


skialpine:
## Summary

The multi-shot conversation context (`AIConversation::multiShotSystemPrompt()`) is only 3 sentences appended to the system prompt. It tells the AI to "track progress" and "keep advice to ONE specific change per shot" but provides no guidance on:

- How many shots constitute a "trend" (2? 3?)
- What to do when the user ignores advice and makes a different change
- What to do when multiple shots show the same persistent problem (escalate?)
- When to suggest the dial-in is complete (e.g., score 85+ two shots in a row)
- How to handle bean changes mid-conversation

## Approach

Expand the multi-shot context with specific rules. Needs real exported conversation data to validate — collect 5-10 multi-shot conversations first, identify where the AI gives poor multi-turn advice, then add targeted rules.

## Dependencies

- Collect test conversations (Phase 0 item 9 in `docs/AI_ADVISOR.md`)

Related: #635 prompt review finding #40

PR #1722, opened 2026-07-30
Add GPT-5.6 models, guard prose grinder recommendations

skialpine:
Adds `gpt-5.6-terra` and `gpt-5.6-luna` to the OpenAI catalog with Terra as the default, keeps both 5.4 entries as options (four total), and fixes a silent adherence-tracking corruption found while testing.

## Why Terra leads

Cheaper than `gpt-5.4` on both axes ($2.00/$12 vs $2.50/$15) and a generation newer. Prices verified against [OpenAI's own pricing page](https://developers.openai.com/api/docs/pricing) — third-party pricing pages were checked and found wrong (one listed Terra at $2.50/$15).

Chosen by replaying the app's **real assembled prompts** (captured via `ai_advisor_invoke` `dryRun`) byte-for-byte against all four models, using the exact request shape `OpenAIProvider::analyze()` builds. Six scenarios, $1.03 total spend.

| | blowout flagged | `nextShot` blocks @none | numeric grind | price /1M |
|---|---|---|---|---|
| `gpt-5.6-terra` | yes | 4/4 | 6/8 | $2.00/$12 |
| `gpt-5.6-luna` | yes | 4/4 | 8/8 | $0.20/$1.20 |
| `gpt-5.4` | **no** | 4/4 | 8/8 | $2.50/$15 |
| `gpt-5.4-mini` | **no** | 4/4 | 7/8 | $0.75/$4.50 |

On a shot whose recent history contained a 64.3 g / 8.5 s blowout, both 5.6 models flagged it and both 5.4 models missed it — **the split was generational, not tier**. `gpt-5.4-mini` additionally reproduced its documented failure mode, inventing a grind trend that did not exist.

**Luna measured at least as well as Terra at a tenth the price.** It is the opt-in rather than the default only because six scenarios of single runs is too thin to promote the smallest tier against prior evidence that small tiers are weak — which is exactly the belief this run undermined. Flagged in the docs as an open question worth settling with a wider run, not as a settled preference.

## Two comments corrected against measurement

- **`reasoning_effort: "none"` is kept, but not for the reason the code gave.** The comment claimed it stops hidden reasoning tokens from overrunning the output cap and truncating the trailing `nextShot` block. A 4-model × 4-scenario × 2-effort matrix emitted the block **16/16 at `"none"` and lost 5 of 16 at `"low"`** — but **no run hit `finish_reason: "length"`**; reasoning ran 208–1245 tokens against a 4096 cap. The models finish cleanly and simply omit the block while reasoning. Right setting, wrong mechanism, stated as settled fact.
- The `web_search` `"low"` floor is a **5.4-generation** constraint, not a universal `web_search` requirement — 5.6 accepts `"none"`. It stays at `"low"` because that is valid across the whole catalog.

## Bug fix: prose in `grinderSetting` was reported as ignored advice

The same replay caught Terra emitting `"a touch coarser than 9"` into `structuredNext.grinderSetting` (and mini `"slightly coarser than 9"`). Prose matches no real setting and parses as no number, so `grinderMatches()` failed both arms and `computeAdherence()` returned **`"ignored"`** — telling the *next* turn's prompt the user disregarded advice they may have followed exactly. Silent, and it corrupts the feature #639 shipped.

`isJudgeableGrinderRecommendation()` now treats prose as unscoreable and warns, while still allowing genuinely non-numeric settings like `3F` through the existing exact-match arm. Two tests cover both halves.

## Harness checked in

`tools/ai_model_eval/` holds the replay script, scenario manifest and findings log so the next model generation doesn't start over. Captured prompts and responses are **gitignored** — they embed personal shot history and this repo is public.

It documents two traps that silently void a run: an emission test over a shot with no taste feedback measures the taste gate rather than emission, and printing per-model cost breaks the blind (which is how the blind broke on this run — disclosed rather than papered over).

## Testing

Full suite via Qt Creator: **110 passed, 0 failed, 0 skipped**, no test warnings. Build clean apart from a pre-existing `__eh_frame section too large` linker warning unrelated to these changes.

## Not done

The wiki manual isn't updated — the model picker is user-visible, so it wants an entry. Happy to follow up.


skialpine:
### Code review

Found 11 issues:

1. The prose guard converts a false "ignored" into a false "followed" — the more damaging verdict. The `else` branch warns but never sets `anyRecommendation`, so when `grinderSetting` is the only recommendation (the common case) `computeAdherence()` falls through to the ranges-only path and returns `"followed"`, under a comment that says "no parameter changes requested" — untrue for this input. Per the adherence teaching in `shotsummarizer.cpp`, `"ignored"` means "STAY THE COURSE" while `"followed"` means "the experiment ran and failed". The fix makes the model believe an experiment ran when the user may have changed nothing.

https://github.com/Kulitorum/Decenza/blob/b97d1b159875807beee567f4a91efc130c41f603/src/ai/dialing_blocks.cpp#L637-L653

2. `replay.py` does not mirror the app's parser, so the emission numbers it produced are not what the app would accept. `FENCE.search()` takes the **first** ```` ```json ```` block; `aimanager.cpp` takes the **last** fence pair and rejects it unless the closer is followed only by whitespace. `clean_tail` is computed but never gates the verdict. That verdict matrix is the source of the "16/16 at none, lost 5 of 16 at low" claim now written as fact into `aiprovider.cpp` and `AI_ADVISOR.md`.

https://github.com/Kulitorum/Decenza/blob/b97d1b159875807beee567f4a91efc130c41f603/tools/ai_model_eval/replay.py#L143-L163

3. The wiki manual is not updated for a user-visible change (CLAUDE.md says "When adding or changing a user-visible feature, update the wiki manual as part of that work... a shipped feature with no manual entry is incomplete"). The OpenAI picker goes from 2 entries to 4 and the default changes. No OpenSpec change backs this PR, so there is no `tasks.md` to defer to. `Manual.md` around L1189-1193 still lists the old two models (and already ordered `GPT-5.4 mini` ahead of `GPT-5.4` as if mini were the default, which was backwards before this PR too).

https://github.com/Kulitorum/Decenza/blob/b97d1b159875807beee567f4a91efc130c41f603/src/ai/aiprovider.cpp#L214-L220

4. An errored call is indistinguishable from a real omission in the summary table, and unknown models report zero spend. A 429 or bad model id is skipped without recording a result, so the matrix prints `n` — reading as a model capability failure in the very table the catalog decision rests on. `spend()` silently returns `0.0` for a model absent from `PRICES`, i.e. exactly when evaluating a newly-released model, which is the harness's purpose.

https://github.com/Kulitorum/Decenza/blob/b97d1b159875807beee567f4a91efc130c41f603/tools/ai_model_eval/replay.py#L164-L169

5. Bumping the translator's fallback to a newer reasoning-model generation exposes that "kept in step" only syncs the model **id**, not the request shape. `tst_aiproviders` asserts the id matches `availableModels().first().id` and nothing more. `TranslationManager::sendNextAutoTranslateBatch()` hand-rolls its own bodies: the OpenAI branch sends no `reasoning_effort`, and the Anthropic branch never calls `disableAnthropicThinking()` — which per the #1691 fix lets adaptive thinking consume the whole budget on `claude-sonnet-5` and return no text. `translationModelFor()` reads the same setting the advisor UI writes. Pre-existing, but this PR touches the coupling.

https://github.com/Kulitorum/Decenza/blob/b97d1b159875807beee567f4a91efc130c41f603/src/core/translationmanager.cpp#L3471-L3476

6. A numeric `grinderSetting` is scored as full adherence. `QJsonValue::toString()` returns an empty `QString` for a non-string type, so a model emitting `"grinderSetting": 4.75` unquoted yields `recommended == ""` — which the new guard passes as "no grind change", `grinderMatches()` accepts on its `isEmpty()` early-out, and the turn scores `"followed"` as *matched*, with no warning. Same class of model misbehaviour this PR set out to harden against, on the line it touched.

https://github.com/Kulitorum/Decenza/blob/b97d1b159875807beee567f4a91efc130c41f603/src/ai/dialing_blocks.cpp#L637-L641

7. The "any whitespace means prose" heuristic misclassifies a real, already-supported notation. `grinderaliases.h` documents Compound notation accepting **both** `"1+4"` and `"1 + 4"` (regex `\s*\+\s*`), used by 17+ grinder entries (Eureka Mignon and 1Zpresso families). `"1 + 4"` is a setting, not prose, and now goes unscored — then falls through to the false `"followed"` of issue 1.

https://github.com/Kulitorum/Decenza/blob/b97d1b159875807beee567f4a91efc130c41f603/src/ai/dialing_blocks.cpp#L579-L591

8. No trimming, so one stray space disables scoring. `"4.75 "` or `" 9"` is rejected as prose. The comment's own premise ("a real setting is a single token") argues for trimming first rather than rejecting a padded token; `grinderMatches()` also compares untrimmed against the DB value.

https://github.com/Kulitorum/Decenza/blob/b97d1b159875807beee567f4a91efc130c41f603/src/ai/dialing_blocks.cpp#L583-L590

9. The new prose test's setup is inert and locks in the bug. It seeds prior grind `9.0` and follow-up `8.75` with the comment "The user DID act on 'a touch coarser than 9'", but that data never reaches `grinderMatches()` — the guard short-circuits first, and `"followed"` arrives via `!anyRecommendation`. Changing the follow-up to `9.0` (user did nothing) would still pass. The sibling `recentAdvice_nonNumericSettingStillScores` has the same blind spot.

https://github.com/Kulitorum/Decenza/blob/b97d1b159875807beee567f4a91efc130c41f603/tests/tst_dialing_blocks.cpp#L1396-L1400

10. Any non-HTTP failure aborts the run and destroys the blind key. Only `urllib.error.HTTPError` is caught, so a socket timeout, a missing `choices`, or a null `content` raises out of `run()`. `key.json` is written only after every scenario completes, so an abort leaves `A`/`B`/`C`-labelled responses with no mapping — the paid run becomes unjudgeable. `audit_block()` also raises `TypeError` on a numeric `grinderSetting` (issue 6's input).

https://github.com/Kulitorum/Decenza/blob/b97d1b159875807beee567f4a91efc130c41f603/tools/ai_model_eval/replay.py#L241-L245

11. The guard's comment overstates its scope. It asserts unguarded prose means `computeAdherence()` "reports **ignored**", but that verdict requires `matched == 0` across all four fields — with a matching `rpm` or `doseG` alongside, the same input yields `"partial"`. Mirrored in `AI_ADVISOR.md`.

https://github.com/Kulitorum/Decenza/blob/b97d1b159875807beee567f4a91efc130c41f603/src/ai/dialing_blocks.cpp#L566-L580

---

<details>

- [Worth a look] `README.md` says "gpt-5.6-sol was never tested" while `aiprovider.cpp` says acceptance was "verified live, all three tiers" — both true but about different things (Sol was probed for parameter acceptance, not advice quality); one-line wording fix (score: 50)
- [OK to skip] The measured figures "16/16 at none / 5 of 16 at low" are duplicated across `aiprovider.cpp`, `AI_ADVISOR.md` and the harness README — CLAUDE.md's centralization rule targets executable behaviour rather than documentation prose, so scope is ambiguous (score: 50)

</details>

skialpine:
### Review fixes pushed (7faa1e36)

All 11 findings at 75+ addressed, plus both below-threshold items. Full suite green: **110 passed, 0 failed, 0 skipped**, no test warnings.

**1, 6, 7, 8, 11 — the guard, redesigned rather than patched.** `classifyGrinderRecommendation()` returns a tri-state and unscoreable now yields a distinct **`"unclear"`** verdict instead of falling through to the ranges-only `"followed"`. It has its own instruction in the system prompt ("treat like ignored; give a concrete value this time"). The same change type-checks `isString()` before reading (so an unquoted `4.75` can't read as empty and score as matched), trims first, and replaces the whitespace heuristic with a new `GrinderAliases::looksLikeSetting()` that shares its regexes with `parseGrinderSetting()` — so `"1 + 4"` scores normally for all 19 compound grinders.

**9 — tests.** They were inert, exactly as reported. Rewritten around a shared `adherenceFor()` driver with each case asserting **both** verdicts, so they cannot pass by never scoring. Added compound, padded, and numeric-JSON cases. Verified the new functions are in the built binary rather than assuming the run covered them.

**2 — harness parser.** Now a port of `AIManager::parseStructuredNext` (last fence pair, clean tail, case-insensitive tag). The 32 saved responses were re-audited offline under the corrected rule: **counts unchanged** — 16/16 at `"none"`, 5 of 16 lost at `"low"`, every rejection a genuinely absent block, not a malformed one. The claim in the code comment is now earned rather than assumed.

**4, 10 — harness reporting.** Errors print `E`, distinct from a real omission `n`; unknown model prices warn instead of silently reporting $0; the blind key persists after every call, and non-HTTP failures degrade to a recorded per-call error instead of aborting the run.

**5 — the pre-existing Sonnet 5 bug.** New header-only `src/ai/airequestshape.h` holds both request-shape rules; `aiprovider.cpp` and `translationmanager.cpp` both call it. Header-only with no AI-stack dependency on purpose — `decenza_testlib` compiles `translationmanager.cpp` but not the providers, which is precisely how the drift happened.

**Below-threshold, both fixed:** the "all three tiers" wording (Sol's *parameter acceptance* was verified; its advice quality was not), and the miss-vs-invent imprecision. The duplicated figures are now centralized, with the code comment pointing at the findings log.

**Three previously-assumed INVARIANTs are now verified live** rather than asserted:

| Check | Result |
|---|---|
| `claude-sonnet-4-6` / `claude-sonnet-5` accept `thinking: {type:"disabled"}` | PASS — and both return a `text` block (a thinking-only reply is the #1691 symptom) |
| `gemini-2.5-flash` `thinkingBudget: 0`, `gemini-3.5-flash` `thinkingLevel: "minimal"` | PASS — `thoughtsTokenCount: 0` on both, so effective, not merely accepted |
| `temperature: 0.3` on `gpt-5.6-terra` (translator body) | PASS |

The INVARIANT comments are kept but scoped to *future* additions, since that is where the risk lives — Gemini's legal `thinkingLevel` values vary by model (`gemini-3-pro-preview` takes only `low`/`high`), so `"minimal"` is not a safe default for the next 3.x entry.

**3 — wiki manual** updated for the four-model picker, including a pre-existing error that listed mini as the OpenAI default. Held locally pending the usual push timing.

skialpine:
### Code review

Second pass, against 7faa1e36. Found 11 issues. Notably, three of the first four are the *same bug class the previous round fixed*, one field or one call site over.

1. The `rpm` axis fails OPEN, producing the false `"followed"` the `"unclear"` verdict was introduced to prevent. `QJsonValue::toInt()` returns `0` for a JSON string (`"rpm": "1400"`), for `null`, and for a model writing `"rpm": 0` to mean unchanged — and `rpmMatches()` opens with `if (recommended <= 0) return true;`, i.e. counts as MATCHED. With `rpm` the only field present the verdict is `"followed"`, which the prompt reads as "the experiment ran". `anyUnscoreable` never sees it. `doseG` and `profileTitle` degrade toward `"ignored"` (fail closed), so `rpm` is the only inflating axis.

https://github.com/Kulitorum/Decenza/blob/7faa1e363a11b1f470d34411c9b414e7c62b0bd0/src/ai/dialing_blocks.cpp#L653-L658

2. The widened gate and the comparator disagree about what a setting is. `looksLikeSetting()` deliberately admits `"1 + 4"` and `"24 1400rpm"`, but `grinderMatches()` compares by byte-exact equality or `QString::toDouble()`, which rejects trailing garbage — so for those forms only a byte-identical string can score `"followed"`. A recommendation of `"23.5"` against a recorded `"23.5 1400rpm"` (the user did exactly what was asked) scores `"ignored"`. The new tests pass only because they seed both sides identically.

https://github.com/Kulitorum/Decenza/blob/7faa1e363a11b1f470d34411c9b414e7c62b0bd0/src/ai/dialing_blocks.cpp#L629-L640

3. `looksLikeSetting()`'s single-token fast path returns `true` without validating against either regex, so a bare `"coarser"` / `"finer"` / `"unchanged"` is accepted as a setting, classified `Scoreable`, fails `toDouble()`, and scores `"ignored"` — the false-non-adherence bug this PR exists to fix. The prose tests are all multi-word, so the branch is untested.

https://github.com/Kulitorum/Decenza/blob/7faa1e363a11b1f470d34411c9b414e7c62b0bd0/src/core/grinderaliases.h#L437-L441

4. `analyzeConversation()` still hand-rolls the rule the PR's own new header exists to centralize (CLAUDE.md: "Centralize anything produced at more than one site… When you touch code that already has copies, collapse them in that pass"). `analyze()` and the translator were converted; this one — the dial-in conversation path — was not. The Anthropic side of the same file was migrated at all four call sites. Its comment also now dangles: `analyze()` no longer holds the "5.4 generation dropped minimal" explanation, which moved to the header.

https://github.com/Kulitorum/Decenza/blob/7faa1e363a11b1f470d34411c9b414e7c62b0bd0/src/ai/aiprovider.cpp#L475-L477

5. The new fourth `adherence` value contradicts the live spec, which is unchanged and carries no delta in this PR. `openspec/specs/advisor-user-prompt/spec.md:130` declares an exhaustive three-value enum, and `:213` requires the prompt "SHALL describe the three `adherence` values" — the shipped prompt now describes four. `docs/CLAUDE_MD/AI_ADVISOR.md` was updated, so the two sources of truth now disagree.

https://github.com/Kulitorum/Decenza/blob/7faa1e363a11b1f470d34411c9b414e7c62b0bd0/openspec/specs/advisor-user-prompt/spec.md#L129-L131

6. The harness parser port is missing the app's `doc.isObject()` check (`aimanager.cpp`), so a trailing block parsing as an array or scalar scores as accepted where the app rejects it — and `audit_block()`'s `obj.get(...)` then raises `AttributeError` outside `call()`'s try/except, aborting the run. That is the same "abort kills the paid run" class fixed one commit ago, in the function whose entire purpose is parser parity.

https://github.com/Kulitorum/Decenza/blob/7faa1e363a11b1f470d34411c9b414e7c62b0bd0/tools/ai_model_eval/replay.py#L99-L102

7. "19 catalog grinders" is wrong — the count is **17** (12 Eureka + 5 1Zpresso). `grep -c "SettingNotation::Compound"` returns 19 because two matches are code references inside `parseGrinderSetting()`/`formatGrinderSetting()`, not catalog rows. The figure is repeated in five places (`grinderaliases.h`, `dialing_blocks.cpp`, `AI_ADVISOR.md`, `tst_dialing_blocks.cpp`, `replay.py`) in a PR whose stated purpose is correcting comments that asserted unverified things as fact.

https://github.com/Kulitorum/Decenza/blob/7faa1e363a11b1f470d34411c9b414e7c62b0bd0/src/core/grinderaliases.h#L422-L425

8. Two documentation claims the code does not back. The harness README still points at `isJudgeableGrinderRecommendation()`, renamed to `classifyGrinderRecommendation()` in the same commit that edited this README. And `airequestshape.h` ("Re-run tools/ai_model_eval/ probes when the catalog changes") plus `aiprovider.cpp` ("Probe a new entry before adding it; tools/ai_model_eval/ has the shape") send readers to a harness that is OpenAI-only — `replay.py` hardcodes `api.openai.com` with no provider branching and prices only OpenAI models. The Anthropic/Gemini probes actually run for this PR were ad-hoc and never committed.

https://github.com/Kulitorum/Decenza/blob/7faa1e363a11b1f470d34411c9b414e7c62b0bd0/tools/ai_model_eval/README.md#L124-L126

9. `max_tokens = 4096` in the translator is a literal duplicate of `AIProvider::MAX_OUTPUT_TOKENS`, sitting between the two lines this PR added to that very function for the identical "hand-rolled body missed the shared rule" reason. Pre-existing, but squarely in the pass.

https://github.com/Kulitorum/Decenza/blob/7faa1e363a11b1f470d34411c9b414e7c62b0bd0/src/core/translationmanager.cpp#L2643-L2645

10. Two claims in this PR's own "Review fixes pushed" comment are inaccurate: the wiki is described as "held locally" but `1da12b7` is already on `origin/master`, and "the duplicated figures are now centralized" — they were removed from `aiprovider.cpp` but still appear verbatim in **both** `AI_ADVISOR.md` and the harness README, i.e. de-duplicated from code, not centralized.

https://github.com/Kulitorum/Decenza/blob/7faa1e363a11b1f470d34411c9b414e7c62b0bd0/docs/CLAUDE_MD/AI_ADVISOR.md#L77-L79

11. Four comment-accuracy defects: `looksLikeSetting()`'s "see the notation rules below" now points backwards (the rules are above it); the long `parseGrinderSetting()` doc comment is separated from its function by `namespace detail` and reads as documentation for the namespace; the Qt citation gives a filename with no line number where CLAUDE.md requires "file-and-line in any comment that asserts what Qt does"; and `AI_ADVISOR.md` puts a paraphrase in quotation marks that differs from the actual prompt text.

https://github.com/Kulitorum/Decenza/blob/7faa1e363a11b1f470d34411c9b414e7c62b0bd0/src/ai/dialing_blocks.cpp#L590-L592

skialpine:
### Second-review fixes pushed (e6cc535a)

All 11 findings addressed. Suite green (verified the five new tests are in the binary that ran, not just that a run passed).

**The three that were the same bug class, one field over** — fixed as a class rather than individually:

- **rpm failed open.** Every axis now goes through a classify step folded through one shared lambda, so `rpm`/`doseG`/`profileTitle` cannot diverge from `grinderSetting`. A fifth field added without one should look obviously wrong.
- **Gate widened without the comparator.** `grinderMatches()` now compares through new `GrinderAliases::compoundKey()` / `leadingDialNumber()`, sharing the notation regexes with `parseGrinderSetting()`. `"1 + 4"` vs `"1+4"` and `"23.5"` vs `"23.5 1400rpm"` both score `followed`.
- **Single-token fast path.** `looksLikeSetting()` no longer trusts any whitespace-free string; every accepted form must begin with a dial number, so a bare `"coarser"` is unscoreable rather than `"ignored"`.

**Centralization finished:** `analyzeConversation()` calls the helper (it was the only remaining OpenAI copy, and the path the helper's rationale is written about); `max_tokens` is now `AIRequestShape::kMaxOutputTokens`.

**Spec:** `openspec/changes/add-unclear-adherence-verdict` carries the delta — the fourth value, the uniform classification, and the notation-equivalence rule. Archive queued as the final commit.

**Claims made true rather than reworded:** the Anthropic/Gemini/OpenAI probe is now committed as `tools/ai_model_eval/probe_request_shape.py`, and the README leads with a table stating that `replay.py` is OpenAI-only. `replay.py` gained the app's `doc.isObject()` check.

**Corrections:** "19 catalog grinders" was wrong — it is **17**; the figure came from a `grep -c` that swept up two code references alongside the catalog rows, repeated into five files. Replaced with "every Eureka Mignon and 1Zpresso entry", which cannot drift. Stale `isJudgeableGrinderRecommendation()` pointer renamed. Qt citation has a line number. The `parseGrinderSetting` doc comment sits with its function again. The paraphrase-in-quotation-marks is gone.

**Finding 10 was about my own previous comment, so to correct it here rather than leave it standing:** the wiki was *not* "held locally" — it was pushed at your request (`1da12b7`, live on the Manual page). And "the duplicated figures are now centralized" overstated it: they were removed from `aiprovider.cpp` but still appear in two docs. The code comment now points at the findings log instead of restating counts.

**Verified against the running app over MCP**, not only the test suite:

| Check | Result |
|---|---|
| `"unclear"` teaching in the live assembled system prompt | present |
| `recentAdvice` scores real turns | `"ignored"` on a 7.75-vs-8 no-movement case; no spurious `"unclear"` |
| OpenAI catalog in the running app | 4 entries, `gpt-5.6-terra` first |
| Default resolution | `aiModel` → `gpt-5.6-terra`, display "GPT-5.6 Terra" |

The rpm classification is not exercised by real data (no stored turn recommended an rpm), so that path rests on the two new unit tests — stating that rather than implying live coverage.

PR #1725, opened 2026-07-31
Keep the grind step correct at the moment it is read

skialpine:
## Why

The grind picker's step for a Niche Zero with 28 distinct observed settings was stuck at `1.0` instead of the `0.25` the history plainly supports — so the wheel stepped in whole numbers and a recorded `7.75` reformatted to `8` on render.

The data was never wrong. `dialing_get_context` reported `stepSize: 0.25` from the same rows: the AI path derived it **live**, while the widget and the web endpoint went through the **async distinct-value cache**. Only the cached path failed. `requestDistinctCache()` cleared every key but refilled only its six bare columns, so the composite key `grinder_setting:<model>` was dropped by every shot save; recovery depended on a consumer re-asking, and when that re-fetch was discarded mid-flight nothing ever asked again.

So the bug was not a badly-refreshed cache. **The bug was that there was a cache** — two call sites deriving the same number two ways, and the cached one had a failure mode the live one structurally cannot have.

## What changed

**The distinct-value cache is deleted.** `m_distinctCache`, `requestDistinctCache()`, `requestDistinctValueAsync()`, the pending/dirty/refreshing flags and `distinctCacheReady()` are gone. Every `getDistinct*()` getter runs through one `queryDistinctList()` against the live database, and the nine `invalidateDistinctCache()` call sites are removed.

It was invalidated **more often than it was read**: every shot save, delete and metadata edit (a rating, a note, one taste slider) wiped it and kicked a six-query refresh, to serve dialogs that might never open.

The grind step follows the same route — `grindStepForGrinder()` / `grindRpmStepForGrinder()` call the helpers the AI path already used, so the widget and the AI payload are the same number by construction.

`historyDataChanged()` replaces the old signal: emitted on **writes only**, never on the async fills the cache signal also carried. QML derives the step inside its existing snapshot functions rather than holding it in a resident binding, so a write costs nothing until rows are next built.

## Cost

Measured with a fresh connection per run, real DB (1,124 shots / 18.5 MB) and a 16× copy:

| query | real (median / worst) | 16× (median) |
|---|---|---|
| grind-step derivation | 3.3 ms / 87 ms | 37 ms |
| a single distinct column | 0.36–1.9 ms | 1.6–17 ms |
| all six bare columns (one old refresh) | 4.66 ms | 55 ms |

## Not in this change

An earlier draft kept a resident map of every grinder's steps behind a covering index and a new schema version 36, plus a supersession guard and a failed-key set — ~615 lines. **All of it was reverted.** There is no schema change, no migration, and no new index. The index bought 37 ms → 1.2 ms on a background thread nothing waits on, and the machinery keeping the map honest is the same class that produced the original bug.

`CLAUDE.md` gains the guardrails that should have caught it: the main-thread DB I/O rule rewritten around user impact and machine operation, "complexity has to come with a measurable win, stated in units the USER feels", and a rule against engineering around migrations.

Net **−200 lines** of production code. 110/110 tests pass.

## Note on issue numbers

The cache bug has **no issue number** — it was found from a device log. #1713 is the equipment-identity fork, a different mechanism that produces the same visible symptom (step falls back to 1.0, decimals lost). Earlier revisions of this branch cited #1724 and then #1713 for the cache mechanism; both were wrong and are corrected.


skialpine:
### Code review

Found 7 issues:

1. `coffeeSuggestions()` still runs a live `SELECT DISTINCT` on every keystroke of the Roaster field. `fRoaster` is rewritten per character by `roasterInput.onTextEdited`, and `suggestions:` is a plain binding over it, so the one-entry memo keys on a string that changes every character and never hits. CLAUDE.md says to thread or hoist a read "when it sits on a repeating path — a binding that re-evaluates on every change", and `shothistorystorage.h` adds "Do not call one from inside a binding that depends on the text the user is typing". The sibling fix in `PostShotReviewPage` (`_baristaHistory`) is the correct shape; this one was only half-applied, while the header comment claims "PostShotReviewPage and ChangeBeansDialog both had to be fixed for exactly that".

https://github.com/Kulitorum/Decenza/blob/a8a04b1b1d9860e4509869202eca7d8fde7e3869/qml/components/ChangeBeansDialog.qml#L309-L317

2. CLAUDE.md asserts test coverage that does not exist: "(`tests/tst_dbmigration.cpp` covers the re-injection this causes)". All 12 `DELETE FROM schema_version` in that file are immediately paired with an `INSERT`; no test seeds a genuinely empty `schema_version` and re-initializes to check `crossedSchemaVersion()`.

https://github.com/Kulitorum/Decenza/blob/a8a04b1b1d9860e4509869202eca7d8fde7e3869/CLAUDE.md#L94-L97

3. `grindRpmStepForGrinder()` hard-codes `sampleCount = 0` on its success path, so a *successful* derivation logs `= 100, derived from 0 distinct numeric setting(s)` — self-contradictory. `grinderWideRpmStep()` returns only a `double`, so the caller has no count to pass. Separately, the empty-model early return is labelled `TooThin` ("too thin to derive") when the real reason is a documented precondition — no grinder supplied. Both defeat the purpose of the outcome enum, which this PR added specifically so a reader can tell these apart.

https://github.com/Kulitorum/Decenza/blob/a8a04b1b1d9860e4509869202eca7d8fde7e3869/src/history/shothistorystorage_queries.cpp#L1654-L1672

4. The PR description is stale and describes a reverted design: a covering index behind "migration 36", a `deriveGrinderSteps()` function, a resident in-memory map, and schema assertions moving 35 to 36. None exist at `a8a04b1b` — schema tops out at 35, `deriveGrinderSteps` is absent, and the derivation is a live per-call query. A reviewer reading the description would approve something materially different from the code.

https://github.com/Kulitorum/Decenza/pull/1725

5. QML binding loop. `_beanTypesFor()` reads `_typesMemoKey`, establishing a dependency, then writes it — from inside a function reachable from the declarative binding `suggestions: root.coffeeSuggestions()`. The write notifies while the binding is still evaluating, which is the condition Qt reports as `Binding loop detected for property`. Expect one per keystroke. The value assigned is still correct, so this is log noise rather than wrong data — but it pollutes exactly the subsystem this PR is trying to make diagnosable.

https://github.com/Kulitorum/Decenza/blob/a8a04b1b1d9860e4509869202eca7d8fde7e3869/qml/components/ChangeBeansDialog.qml#L309-L317

6. `historyDataChanged()` triggers a burst of synchronous main-thread queries. `grindStep` and `rpmStep` are eager `readonly property` bindings over `historyVersion`, and each evaluation costs the query this PR measures at 3.3 ms median / 87 ms worst on a real 18.5 MB database. Two to three `GrindRowSource` instances are typically live (the resident bar widget plus a pushed page or open dialog), so a taste-slider autosave or a shot save costs several queries inline. The post-save emit lands in the `saveShot` callback as the machine leaves Pouring, while BLE telemetry is still on the main thread — the tighter-budget case CLAUDE.md calls out. Deriving inside `grindRowsFor()`/`rpmRowsFor()` (already snapshot-based) and keeping `historyVersion` only as a rebuild trigger would avoid holding the value in a resident eager binding.

https://github.com/Kulitorum/Decenza/blob/a8a04b1b1d9860e4509869202eca7d8fde7e3869/qml/components/GrindRowSource.qml#L48-L78

7. `#1713` is misattributed throughout. The issue thread describes an equipment-identity fork — "adding that info to the grinder did create a new grinder" — which is what `equipmentstorage.cpp` and `equipmentlogging.h` correctly cite it for. The cache-invalidation mechanism described here (composite key dropped, re-fetch discarded mid-flight) is a different bug that appears to have no issue number. Both produce the same symptom (step falls back to 1.0, decimals lost), which is likely how they got conflated. An earlier commit in this PR corrected five citations of `#1724` to `#1713`; the number is still wrong for these call sites.

https://github.com/Kulitorum/Decenza/blob/a8a04b1b1d9860e4509869202eca7d8fde7e3869/src/history/shothistorystorage_queries.cpp#L41-L45

---

<details>

- [Worth a look] `isDbWorkIdle()`'s comment says "the ten read queries"; there are 9 `runDetachedDbThread()` call sites after this PR removes two. One-word fix in a comment the PR already edits (score: 60)
- [Worth a look] CLAUDE.md cites `shothistorystorage.cpp:197` and `:242` for `runMigrations()` and `m_ready = true`; they are now `:198` and `:243`. The lines shifted within this same PR, and the comment explicitly invites the reader to check them (score: 60)

</details>

Verified clean: `historyDataChanged()` emit coverage across all write paths, `queryDistinctList` bind order, no dedup-key collision between the settings and rpm report families, no dangling references to the removed cache API, sort ordering preserved from the cached path, and the deleted QML mechanisms (`_cacheConn`, `_autoTextPendingHistory`, the wheel-snapshot rationale) are genuinely obsoleted rather than orphaned.

skialpine:
All 7 review findings fixed, plus the 2 below-threshold items.

| # | fix |
|---|---|
| 1, 5 | Bean-type list refreshes on roaster **commit** (blur / suggestion pick), not per keystroke — removes the query from the binding, which also removes the read-then-write binding loop |
| 2 | CLAUDE.md no longer claims test coverage that did not exist. Writing that test showed the premise was wrong too: wiping `schema_version` on a migrated DB does not over-report crossings — `createTables()` re-seeds to 1, the chain re-runs against existing tables, several migrations fail and `initialize()` returns false, stranding at version 14. The sentence now says that, having been checked |
| 3 | `grinderWideRpms()` returns its values, so a successful RPM derivation reports a real sample count instead of a hard-coded 0; added a `NoGrinder` outcome so an empty model is not logged as "too thin to derive" |
| 4 | PR description rewritten to describe the actual design, with an explicit "Not in this change" section for the reverted index/migration |
| 6 | `grindStep`/`rpmStep` are functions called inside `grindRowsFor()`/`rpmRowsFor()` — already explicit snapshots — instead of resident eager bindings. A write no longer costs a query per live `GrindRowSource` |
| 7 | The cache bug is described rather than cited: it has no issue number. #1713 is the equipment-identity fork, a different mechanism with the same visible symptom. CLAUDE.md notes this so the wrong link is not re-derived |
| F | `isDbWorkIdle()`: "ten" → "nine" detached reads |
| G | CLAUDE.md's `:197`/`:242` citations → `:198`/`:243` (my own later edits had shifted them) |

Two further changes from maintainer review after the fixes:

- The grind-step tests moved from `tst_dbmigration.cpp` (which implies schema work; there is none) to `tst_dialing_blocks`, which already covers `queryGrinderContext`'s `stepSize` and copies a prebuilt schema template per test. Deliberately **not** a new target — 106 already exist and each is a compile+link on every build.
- `grindStepAgreesWithDialingContext` deleted: after this branch both sides call the same function, so it asserted `f(x) == f(x)` and could not fail.

`TESTING.md`'s "Adding New Tests" led with *create a file, add a target* — the default that produced the misplacement. It now leads with adding to an existing target, and adds the rule that a test must be shown to fail before it is kept, with the three pass-forever shapes that have shipped here.

Grind picker, wheel re-open, bean dialog and post-save step verified manually — QML has no harness here. 110/110 automated. OpenSpec change archived, two specs synced.

Issue #1726, opened 2026-07-31
[Bug] Grind settings only showing a single digit (no decimal point)


mcastaldelli:
### What happened?

Creating a new issue to replace #1713, as the fix didn't work, and I don't have permission to re-open the original issue.

Grinder: Eureka Single Dose with Lebrew Sweet burrs/

This latest build introduces inconsistencies in the grind scroll wheel. It works as expected inside the recipe (in the edit recipe button), but it shows incorrectly in the recipe one line editor in the home page (i.e., editing without opening the recipe). In the latter case it shows only integers if there's no non-integer setting in history, and then decimals moving by a full unit if the history contains such value. For example: introducing a single setting 1.2 makes the list 1.2; 2.2; 3.2; etc. If the history has a setting 1.5, it shows 1.5; 2.5; 3.5; etc.

The screenshots show the difference:

1) Correct grind wheel inside "edit recipe"

<img width="1964" height="2618" alt="Image" src="https://github.com/user-attachments/assets/fb8e4126-3732-4d39-b6f3-c62fc65f31f4" />

2) Incorrect grind wheel showing only integer in the main page one liner:

<img width="1964" height="2618" alt="Image" src="https://github.com/user-attachments/assets/e01332c7-6730-4889-b848-9b9cbaf21b38" />

3) Incorrect grind wheel with .2 decimal in history:

<img width="1964" height="2618" alt="Image" src="https://github.com/user-attachments/assets/0e011876-db5b-49e9-a61c-e6d776e34622" />

4) Incorrect grind wheel with .5 decimal in history:

<img width="1964" height="2618" alt="Image" src="https://github.com/user-attachments/assets/780d5c1d-636b-4c81-a413-6e8502a5e06b" />

### Steps to reproduce

click on "grind setting" in the recipe one liner on the home screen.

### Decenza version

2.0.1

### Platform

Android

### Device / OS version

Decent tablet P80X Android Pie

### Debug log / shot debug log

[debug-3.zip](https://github.com/user-attachments/files/30601579/debug-3.zip)

skialpine:
Thanks for the updated log. I think I fixed it this time, and if not I added a bunch of logging to fix it from the next log.  New release should be available now.

mcastaldelli:
Thanks, Jeff! It does seem like it's fixed now ;-)

skialpine:
Happy to hear it!

PR #1727, opened 2026-07-31
Use the active grinder, not every grinder, when none is injected

skialpine:
Second grind-step defect, found while reading the [#1726](https://github.com/Kulitorum/Decenza/issues/1726) reporter’s log. **Not that issue’s reported cause** — that was the composite-key distinct cache, fixed in #1725 and already on `main`.

## The bug

`GrindRowSource.grindStep()` passed its injected grinder model straight to `grindStepForGrinder()`. An empty model there means *pool every grinder’s history*.

That is right for exactly one caller — the ShotServer `/beans` form, where a new bag has no equipment chosen yet. It is wrong for the picker, which reaches it whenever the host has no grinder to name:

- a bag or recipe with no equipment package
- a shot recorded before it had one

Pooling mixes every grinder’s dial resolution into one estimate. A Niche stepping `0.25` and an EK43 stepping `1` produce a step belonging to neither, and nothing on screen says the wheel is not about your grinder.

**A user with one grinder cannot tell the two apart, because the pool *is* their grinder.** That is why nothing has flagged it — including in the #1726 log, where the post-shot review reached this branch with an empty model and the answer looked correct.

## The fix

Resolve to the active grinder before querying. Pooling stays reachable only when there is no active grinder either — genuinely no grinder anywhere, where every grinder is the only pool there is.

Only the **source of history** falls back. Notation, click-indexing and `rpmCapable` stay on the injected identity; reading the active grinder for those is the bug the context injection exists to prevent, and the header comment now says which of the two this is.

## Not changed, deliberately

- `rpmStep()` — `grinderRpmCapable("")` is false, so the empty case never reaches its query. Noted at the call site so nobody adds a redundant fallback.
- `_observedFallback` / `_medianObservedAnchor` — already guard `model.length > 0` and return nothing rather than pooling.
- `ShotServer::sendGrindCandidates` — the one deliberate consumer of the pooling branch; the C++ comment now names it as such.

## Testing

Full suite: **110 passed, 0 failed, 0 skipped**, no tests with warnings. Built clean (the one build warning is an unrelated `__eh_frame` linker note).

The QML half is manual, per project convention.

One caveat on the record: an intermediate snapshot during the first run showed `tst_CoffeeBags::settingsDyeDoseLadder` failing on a warning. I re-ran before capturing the raw log, which destroyed the evidence — the completed run and two further runs are green, and nothing here touches `CoffeeBagStorage` or the dose ladder, but that is not the same as ruling it out. No issue filed because there is nothing to attach to it.

### DIFF
--- a/qml/components/GrindRowSource.qml
+++ b/qml/components/GrindRowSource.qml
@@ -11,10 +11,15 @@ import Decenza
 // shot's grinder in post-shot review, the recipe's package in the wizard, the
 // bag's equipment in the beans dialog, the active grinder on the brew bar) and
 // every derived behaviour — step size, notation, observed-history fallback,
-// RPM capability — resolves against it. Nothing here reads the active grinder.
+// RPM capability — resolves against it.
 // (The pre-split _observedFallback read Settings.dye.dyeGrinderModel — the
 // active grinder — which was harmless while the brew bar was the only host and
 // is exactly the bug this injection exists to prevent.)
+//
+// The ONE place that still reads the active grinder is grindStep(), and only
+// when nothing was injected at all — see its comment. That is a source of
+// history to measure from, not a semantic: notation, click-indexing and RPM
+// capability never leave the injected identity.
 //
 // Per-candidate stepping is catalog-first via SettingsDye.stepGrinderSetting
 // (numeric AND Compound "a+b" notation), then plain-numeric / number-in-text /
@@ -38,8 +43,7 @@ QtObject {
 
     // Per-grinder grind step, derived from the user's own shot history for the
     // INJECTED grinder by the same noise-filtered estimator the AI dialing
-    // context uses. Falls back to 1.0 when history is too thin. With no grinder,
-    // grindStepForGrinder("") derives from full history.
+    // context uses. Falls back to 1.0 when history is too thin.
     //
     // FUNCTIONS, not properties, and deliberately so. Each call runs a live query
     // measured at 3.3 ms median / 87 ms worst on a real 18.5 MB database. As eager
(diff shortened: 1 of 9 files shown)


skialpine:
### Code review

Found 5 issues, all fixed in e36264f9.

1. `grinderRpmCapable("")` is **true**, not false, so the comment claiming `rpmStep()` needs no empty-model fallback was wrong — and it credited the wrong guard. `deriveRpmCapable` returns true for any grinder absent from the registry, and an empty brand+model matches nothing. What actually prevents pooling is `grindRpmStepForGrinder`'s own early return; a reader trusting the comment could have deleted it as redundant. (CLAUDE.md: "an un-sourced claim in a comment gets believed and then licenses wrong code")

https://github.com/Kulitorum/Decenza/blob/dace16a1d520894691708bf86a103cd18002db66/qml/components/GrindRowSource.qml#L98-L104

2. `rpmStep()` had the same defect the PR fixes for `grindStep()` — with nothing injected it returned a blind 50 instead of the active grinder's observed RPM step. Now resolves identically.

3. Scoping can find **less** than pooling, not merely something different: the pooled query carries no equipment join, so it counts shots recorded before equipment existed. A user whose history is mostly pre-equipment gets nothing from the scoped query and fell to the hard-coded `1.0` — regressing the single-grinder case this change is otherwise a no-op for. `grindStep()` now pools as a second attempt.

4. `ShotServer::handleGrindCandidatesApi` feeds three web forms (shots, recipes, bags), not the `/beans` form alone as the comment claimed. Each mirrors an in-app form that now resolves to the active grinder, so the web ones had to as well. (CLAUDE.md: "Keep the two surfaces in sync")

https://github.com/Kulitorum/Decenza/blob/dace16a1d520894691708bf86a103cd18002db66/src/network/shotserver.cpp#L2819-L2829

5. Comment corrections: `sendGrindCandidates` does not exist (`handleGrindCandidatesApi`); the header listed step size among the injected-identity behaviours then carved it out five lines later; the summary above `grindStep()` still named the injected grinder as the only source; the `#1726` attribution now reads as an inference from the attached log rather than a triaged diagnosis, since that issue is still open. Also refreshed the cost comment's caller list, which described the QML side as a binding — true before #1725, not after.

Verified as false positives: `Settings.dye.dyeGrinderModel` is the correct access path, the `1.0` early return is behaviour-identical to the old code, `root.grinderModel.length` cannot throw on a QML `string` property, and `_observedFallback` / `_medianObservedAnchor` were already safe. One agent claimed `rpmStep()` pools every grinder's RPM history — it does not; the C++ guard blocks it.

Not yet built or tested — Qt Creator was busy. Still needs a local suite run before merge.

skialpine:
### Code review (round 2)

Targeted re-review of everything after the first pass. Found 6 issues, all fixed in 9cfcc036.

1. Census counted distinct **strings** where `grinderWideNumericSettings` dedupes **doubles**, so `"3"`, `"3.0"` and `"03"` were one sample to the step derivation and three to the census. Not hypothetical — `GrindRowSource` writes settings with `toFixed()` at a width derived from the current step, so one dial position lands in history at several decimal widths as that step changes. Both log lines say "distinct numeric setting(s)" precisely so they can be compared; counting different things would have manufactured the false diagnosis the census exists to prevent.

2. The key folding was incomplete. The C++ fold put both spellings of a grinder in one row, but the settings query's `DISTINCT` ran on the **unfolded** model, so a pair of case variants incremented the tally twice against a folded shot count — the exact discrepancy the folding was added to avoid.

3. Posting the census from `initialize()` arms `~SerialDbWorker`'s "those writes are being discarded" warning on a task that writes nothing, and races `QTest::failOnWarning()` in any test that constructs a storage and lets it fall out of scope without draining — surfacing under the worker's name in whichever test happened to be running. Now `#ifndef DECENZA_TESTING`, as `importLegacyBeanPresets()` already is directly above it.

https://github.com/Kulitorum/Decenza/blob/dace16a1d520894691708bf86a103cd18002db66/src/core/dbutils.h#L243-L248

4. The third census query had no failure branch — the one whose whole purpose is surfacing grinders that own no shots would have dropped them with no trace. Warns and continues now.

5. `"Its callers are X and Y"` on `grinderWideNumericSettings` read as exhaustive and omitted `queryGrinderContext`, which reaches the query for every non-empty model on the AI dialing path. That sentence is the frequency argument for not threading or caching it, so undercounting call paths is the error that matters there. (This was itself a *replacement* for a stale claim — same failure, one commit later.)

6. `COUNT(*)` over a `LEFT JOIN` counts join rows, and nothing constrains `equipment_items` to one grinder row per package. `COUNT(DISTINCT s.id)`.

Two comment corrections: the header justified `runOnDbThread` over `runDetachedDbThread` on `isDbWorkIdle()` coverage, but that counts detached threads too — the real reason is the FIFO worker is joined in the destructor. And `equipmentlogging.h` catalogs every `[Equipment]` source verbatim by its own rule; `Census` was added without one.

Verified as false positives / correct as written: the lambda captures only `dbPath` by value and the worker is joined before members die; `Row`/`QMap`/`std::as_const` are sound; DEBUG tier and `_STDERR` variant are right per [LOGGING.md](https://github.com/Kulitorum/Decenza/blob/dace16a1d520894691708bf86a103cd18002db66/docs/CLAUDE_MD/LOGGING.md); `MARKER_ONLY_GLOBS` is the correct list given 23 unrelated bare `qWarning` sites in that file; the QML second-attempt cannot recurse or double-query. No test added, deliberately — the only observable effect is log text, no sibling `[Equipment]` line has one, and it would assert wording rather than behaviour.

Still unbuilt — Qt Creator was busy. Needs a local suite run before merge.

PR #1728, opened 2026-07-31
Compile each production source into the test build once, not nine times

skialpine:
## Why

`tests/` is **40% of a clean build** (1,956 s of 4,846 s cpu), and 46 object files in it were pure duplicate work — 17 production sources listed in more than one test target's source list, so compiled once per target. Two of the five slowest edges in the *whole* build were the same `profilemanager.cpp` object built for different targets.

The suite also doubled in July 2026 (60 → 107 files), and the documented cost model pointed at the wrong thing: `TESTING.md` said a test costs "a compile *and a link*". Link is **3.0%**.

## What changed

**Twelve narrow static libraries**, each linked by exactly the targets that need it. Narrow, *not* `decenza_testlib` — that one is linked by every test target, so putting a source there trades a bounded compile fan-out for a relink fan-out across all 106. Measured by touching `profilemanager.cpp` and diffing `.ninja_log`:

| | Today | Via `decenza_testlib` | Via a narrow library |
|---|---|---|---|
| Touch `profilemanager.cpp` | **77.0 s** | ~67 s | **12.2 s** |

`resources.qrc` also stops being compiled three times, via an **OBJECT** library — `STATIC` would silently fail, since nothing references a symbol in a generated resource `.cpp` and the linker has no reason to pull it from an archive.

| | Before | After |
|---|---|---|
| Duplicate object files | 46 (215 s) | **0** |
| qrc objects under `tests/` | 14 / 54 s | 12 / 28 s |

## On the clean-build saving — a range, not a number

`tests/` fell 19.8%, but the untouched app arm fell 14.9%, and that "control" is contaminated: removing 46 parallel compiles reduces contention for everything else. Control-adjusted gives ≈96 s; direct accounting of removed work gives 241 s. **The saving is between 96 s and 241 s and wall-clock cannot separate them.**

The counts are certain; the timings are not. `baseline.md` records this, along with three measurement methods that each produced a plausible and wrong answer first.

## What keeps it at zero

`scripts/check_test_source_duplication.py`, wired into `text-invariants.yml` (build-free, so it fits that job's rule). Fails on a source in two targets, or a source a target already links a library for.

Validated against the pre-change tree: **17 sources / 46 compiles, matching the independent `.ninja_log` audit exactly.** It also cross-checks its own parsed target count against a naive grep and refuses on mismatch — under-parsing fails *open*, and did once during development.

## Docs

`TESTING.md` and `CLAUDE.md` get the corrected cost model plus the two rules the audit showed were missing: state what defect shape a new test catches that the suite does not, and assert one invariant in one place.

## Not done, deliberately

- **No retirement policy, no outcome-history collection.** An earlier draft specified a ledger of which tests have ever failed; it was cut because it had the same shape as the problem — monotonic growth, never retired, a cost on every run forever, no mechanism to re-measure its own value. `design.md` records the reasoning so it is not re-proposed from scratch.
- **The `<MQTTAsync.h>` header fix was measured, attempted, and reverted.** 170 s of real cost, but `Q_PROPERTY(MqttClient*)` needs the complete type, and the opaque-pointer escape would break 8 QML bindings at runtime rather than at compile time. `maincontroller.h` now says so.

## Verification

- **110/110 tests pass.** Full suite run after each restructuring step.
- `tst_coffeebags` is **genuinely flaky** — 2 of 5 full runs, always `settingsDyeKeepFieldsStillResolvesTheDoseRung`, always passing in isolation. Pre-existing and unrelated (no test source was touched). Mechanism found and documented in `TESTING.md`: it fails via `failOnWarning`, because `settings_dye.cpp:56` warns when `bagReady` arrives with an empty map, racing the test's raw-connection seed.


skialpine:
### Code review

Found 3 issues. All fixed in fe8e27ab; 110/110 tests still pass.

1. `tst_difluidr1` would fail to link on iOS — `decenza_refractometerlib` was attached by editing the existing OpenSSL line, which sits inside the `else()` of an `if(IOS)` block. `RefractometerDevice` is needed on every platform, so desktop stayed green while an iOS build would break. `tst_difluidr2` two lines earlier links it unconditionally. Invisible to the new check script, which has no awareness of CMake conditionals.

https://github.com/Kulitorum/Decenza/blob/18870462b9cc64d6a82b925fd0376b0624919e75/tests/CMakeLists.txt#L1092-L1099

2. Relink fan-out stated as 107; it is 106. `decenza_testlib` is linked by the 106 `add_decenza_test` targets — `tst_shotcorpus` is a bare `add_executable` that never links it. The file's own line 137 already said 106.

https://github.com/Kulitorum/Decenza/blob/18870462b9cc64d6a82b925fd0376b0624919e75/tests/CMakeLists.txt#L832-L838

3. Same 107/106 error in `TESTING.md`, contradicting a correct "fan-out of 106" seven lines later — and a `baseline.md` reference pointing at the pre-archive path, which no longer exists.

https://github.com/Kulitorum/Decenza/blob/18870462b9cc64d6a82b925fd0376b0624919e75/docs/CLAUDE_MD/TESTING.md#L434-L436

Worth noting: issue 1 is exactly the platform-conditional class this repo's tooling misses by construction — no local build, no test run, and no PR check exercises `if(IOS)`.

PR #1729, opened 2026-08-01
Heal every pre-enrichment fork, not just the burr ones

skialpine:
Two rules forty lines apart in `equipmentstorage.cpp` answer the same question — *did this edit only ADD information?* — and only one answers it correctly.

## The defect

The **live rule** (`:1547-1553`) tests six components and its comment is explicit that basket and puck prep belong there:

> there is no espresso pulled without a basket, so an absent basket is a basket nobody wrote down, exactly like absent burrs

The **heal** that retro-applies that rule to databases which forked before it shipped restated it instead of sharing it, and got it wrong: it tested the BURRS going empty→named, and additionally required basket brand, basket model and puck prep to be **EQUAL**.

"Equal" is not a stricter form of "empty→named" — for a component that was absent it is the **inverse**. So a fork caused by recording a basket was skipped while the migration logged `merged 0 package(s)`, which reads as nothing to fix.

## Evidence

The maintainer's Android device, verified directly:

| pkg | superseded_by | grinder | burrs | basket | puck prep | shots |
|---|---|---|---|---|---|---|
| 1 | → 2 | Niche/Zero | 63mm Mazzer Kony conical | — | — | **961** |
| 2 | | Niche/Zero | *byte-identical* | Decent/18g Ridged | puckScreen,shaker | **59** |

Forked 2026-06-21, five weeks before the enrichment rule landed (`50182dbe`, #1720). Migration 35 found nothing — correctly by its own predicate, wrongly by the rule it exists to enforce.

## The fix

`isEnrichmentOf()` is now **one function**, called by both the live edit rule and the heal. They are the same question at two moments: an edit deciding whether to fork, and a migration deciding whether a past fork should not have happened. A second copy is how this recurred.

The heal's SQL is now a plain lineage scan; the decision happens in C++ on loaded packages, which also inherits the Unicode-correct folding SQLite's ASCII-only `LOWER()` cannot do.

Two packages differing in **nothing** are excluded explicitly. `isEnrichmentOf` is vacuously true for them, the live rule never sees that case (it returns early on `unchanged`), but a database can hold one since device transfer imports every superseded package as a new row. A byte-identical pair is a duplicate, not a fork.

Migration 36 runs the widened heal. 35 stays in place and unchanged — its version is stamped wherever it ran, and most devices are below it and will run both in one launch, which is harmless because the widened predicate is a superset. 36 resolves the active-equipment id against 35's result rather than QSettings, since QSettings is not written until `MainController` adopts the value after `initialize()`; a chain of 35 folding 1→2 and 36 folding 2→3 would otherwise leave the app pointing at a package 36 just deleted.

The heal log line now names which components were filled in. `merged 0` was undiagnosable precisely because it never said which component was being looked at.

## Verification

**Live run on a real 1,124-shot database** — seven "Zero" packages folded to two:

```
[Equipment][Migration] 36: re-healing enrichment forks against the full rule
[Equipment][Merge] staged fold of package 2 into 4 - 1006 shots, 0 bags, 0 recipes
[Equipment][Heal] package 2 is an enrichment fork of 4 (same gear; basket brand, basket model, puck prep recorded on the newer one only) - reuniting them
  ... 5 → 6 (15 shots), 10 → 4, 11 → 6, 12 → 6
[Equipment][Migration] 36 complete - merged 5 package(s) that an edit recording gear had split off
```

End state: 1029 + 80 = 1109 shots, no shot stranded. The two survivors differ only by `rdt` — a real distinction — and **each keeps its own dial memory** (8.75 vs 8.25, two of the user's own 0.25 steps apart). The census is unmoved at 1109 shots / 28 settings, which is the check that matters: it groups by model, so merging packages must not change it.

This matched a SQL prediction made before the run, package for package, on both a dev and the maintainer's database.

**Tests:** full suite 110/110. Both new tests break-tested — restoring burr-only makes `healFillsAnyComponentNotJustBurrs` report `healed=0` against `1`; disabling migration 36's guard makes `v35ToV36HealsABasketFork` report version 35 against 36.

## Considered and rejected

A `grinders` table making the grinder a shared entity. Two independent review passes measured today's model-string scope against a `grinder_id` scope on both real databases and got identical results — same 1020 shots, same 28 settings, same 0.25 step. It would additionally have split burr-enrichment forks that currently share a history, collapsed six distinct remembered dial positions into one, and deleted the heal this PR fixes.

Separately worth a one-liner, not done here: `grinderModelMatchSql` matches on model and ignores brand. Exactly one genuine cross-maker collision exists in the 196-entry registry (`m4` → Macap, Versalab).


skialpine:
### Code review

Reviewed with five agents (CLAUDE.md compliance, deep bug scan, git-history context, prior PR feedback, comment accuracy). Twelve findings scored 50+; all twelve are fixed in `db23c4e7` and `776a7ad3`. Nothing outstanding.

The substantive ones:

1. **Migration 35's behaviour changed while its own log line did not.** It still said "merged N package(s) that a burr edit had split off" after the predicate was widened past burrs — so a user's log would have named the wrong cause. Reworded to "that an edit recording gear had split off".

2. **`norm()` used `toLower()` while citing a `Qt::CaseInsensitive` precedent.** Different Unicode tables (LowerCase vs CaseFold), so the heal and the identity lookup it was justified by could disagree on non-ASCII gear names. Now `toCaseFolded()`.

3. **`after.hasGrinder = true` was hard-coded and factually wrong** in `identityOfStatic`. Harmless only because `isEnrichmentOf` never read it — a latent trap for the next caller. Now computed.

4. **Byte-identical pairs were being folded as forks.** A lineage pair with nothing filled in is a duplicate, not an enrichment fork; added the `if (filled.isEmpty()) continue;` gate and a test that fails without it.

5. **Three mutually-inconsistent restatements of the predicate in comments**, none of which matched the five equality checks actually in the code. Collapsed to one statement next to the implementation.

Verified against real data twice, not just the harness: on this machine's 1109-shot database the widened heal folded 5 pre-enrichment forks (7 "Zero" packages down to 2) with the shot census unmoved and survivors keeping distinct dial memory. A second run with staged fixtures produced exactly the 3 expected folds and correctly refused the byte-identical, changed-basket and grinder-less cases.

PR #1730, opened 2026-08-01
Precompile the Qt headers every translation unit re-parses

skialpine:
811 of the tree's 816 C++ translation units now share one of six precompiled headers. **No test was deleted, merged or modified.**

## Measured

Five clean rebuilds, back to back in one window so they are mutually comparable (macOS, Ninja, Debug+ASan, 11 cores):

| Config | Wall | vs off |
|---|---|---|
| PCH off | 229.5 s | — |
| 18 headers | 201.3 / 187.9 s (mean 194.6) | 15.2% |
| **24 headers** | **157.0 s** | **31.6%** |

**~30%, conservatively 27%.** The 24-header arm is a single run and the two 18-header arms spread 7%, so the honest interval is 27–35%.

**Do not compare against numbers from another session.** An earlier sitting measured the *same* 18-header config at 169.5 s against a 226.9 s control; today it reads 194.6 s. The machine drifts by more than the effect, which is why the off-control was re-measured in the same window instead of reused.

## Where The Time Was

- **226 of the app target's 410 TUs are generated by qmlcachegen**, all emitting the same ~20 includes. That is 32% of a clean build — larger than every hand-written `src/*.cpp` combined.
- The first header list was built by reading **one** generated file and stopping at the QtQml lines. It missed `<type_traits>` and `qtimezone.h`, present in **216 of those 226 files** — the identical 95.6% share as every QtQml header that did make the list. Those two plus four more at 45.1% took the saving from 15.2% to 31.6%. **Half the win was six lines.**
- A whole 60-line test file compiles in 1.45 s; `#include <QtTest>` alone costs 1.42 s. The body is 3%.

## Structure

Three PCHs serve 119 targets via `REUSE_FROM` — `decenza_test_pch` for all 106 tests, `decenza_testlib` for 13 narrow libraries — plus private PCHs for `Decenza`, `saw_parity`, `profile_sync` and `shot_eval`, whose flag sets differ.

Reuse is what makes it pay small: building a PCH costs ~2 s, so a 2.7 s library loses money on a private one and breaks even instantly on a shared one. `saw_replay`, `tst_shotcorpus` and the canaries are excluded on that arithmetic. `decenza_testresources` is excluded for a different reason — its 10.8 s is a qrc byte array, not header parsing.

## Two Things Worth Reviewing Closely

**Objective-C++ opts out via `SKIP_PRECOMPILE_HEADERS`.** A `$<COMPILE_LANGUAGE:CXX>` genex does **not** do this — it empties the ObjC++ PCH while CMake still passes `-include-pch` to `.mm` lines, and clang refuses: `Objective-C was disabled in precompiled file ... but is currently enabled`. I shipped that bug first and the local macOS build caught it. The skip list is derived from `SOURCES` rather than hand-written, because the `.mm` files sit in `if(APPLE)`/`if(IOS)` blocks — a hand-list compiles clean on Linux and Windows and breaks only the two Apple workflows.

**OFF on Windows** (`DECENZA_ENABLE_PCH`). Windows CI wraps the compiler in sccache, and a compiler cache needs explicit `sloppiness` configuration to keep hitting once a PCH exists. Get it wrong and CI gets *slower*. Untested, and untestable from macOS. Enabling it there is a separate change measured on a Windows runner with a warm cache.

⚠️ **The same question applies to macOS/iOS**, which wrap the compiler in ccache via `CMAKE_XCODE_ATTRIBUTE_CC`. Watch the CI build times on this PR — if the Apple legs get slower rather than faster, ccache sloppiness is the cause and the fix is a follow-up.

## Verified

- 110/110 tests pass, zero warnings, on the exact shipping config
- Xcode generator checked separately with `-DCMAKE_OSX_ARCHITECTURES="x86_64;arm64"`, since a PCH is per-architecture — one `.gch` per slice, correct fat binary
- Test builds dispatched for all six platforms (see below)

## What This Replaces

The original question was whether tests should be retired for build cost. An audit of all 2,515 tests found **zero tautologies**, 93 mergeable into `_data()` tables, and 239 under three lines. Acting on every one saves **under a second** — 90% of test-TU compile time is fixed per *file*, only 10% scales with lines. The cost was never the number of tests.


skialpine:
### Platform test builds

Dispatched on this branch, uploads disabled:

| Platform | Generator | PCH | Run |
|---|---|---|---|
| Linux | Ninja | on | [30675671994](https://github.com/Kulitorum/Decenza/actions/runs/30675671994) |
| Linux arm64 | Ninja | on | [30675672961](https://github.com/Kulitorum/Decenza/actions/runs/30675672961) |
| Android | Ninja | on | [30675673850](https://github.com/Kulitorum/Decenza/actions/runs/30675673850) |
| Windows | Ninja | **off** | [30675674845](https://github.com/Kulitorum/Decenza/actions/runs/30675674845) |
| macOS | Xcode | on | [30675675710](https://github.com/Kulitorum/Decenza/actions/runs/30675675710) |
| iOS | Xcode | on | [30675676649](https://github.com/Kulitorum/Decenza/actions/runs/30675676649) |

What each leg is actually proving, since only two of these were reachable locally:

- **iOS** is the one that matters most. It compiles `ios/SafariViewHelper.mm` and `ios/MachineStatusWidgetBridge.mm`, which exist on no other platform, plus the three `APPLE` `.mm` files. This is the exact surface where the `SKIP_PRECOMPILE_HEADERS` fix has to hold, and where a hand-written skip list would have failed.
- **macOS** confirms the Xcode generator with `x86_64;arm64` — a PCH is per-architecture. Verified in a standalone probe, not yet in this project's real configuration.
- **Android** compiles the `Q_OS_ANDROID` branches and the `QJniObject` includes that are dead code on desktop. `QJniObject` was deliberately kept out of the PCH for that reason.
- **Windows** is the control: `DECENZA_ENABLE_PCH` defaults OFF there, so this leg should be unchanged. If it breaks, the guard is wrong.
- **Linux / arm64** are the plain Ninja legs.

Also worth watching on the Apple legs: build *duration*, not just pass/fail. macOS and iOS wrap the compiler in ccache via `CMAKE_XCODE_ATTRIBUTE_CC`, and ccache needs explicit `sloppiness` settings to keep hitting once a PCH is in play. A green-but-slower Apple build means that follow-up is needed.

skialpine:
### Correction: the compiler-cache question applies to all six platforms, not two

The PR description says the sccache concern is Windows-specific and adds macOS/iOS as an aside. That is wrong. Every workflow wraps the compiler in a cache, and **none of them configures `sloppiness`**:

| Workflow | Cache | `sloppiness` |
|---|---|---|
| linux-release | ccache | not set |
| linux-arm64-release | ccache | not set |
| android-release | ccache | not set |
| windows-release | sccache | not set |
| macos-release | ccache | not set |
| ios-release | ccache | not set |

**The 31.6% in the description is a cold-build number measured on a machine with no compiler cache.** CI steady state is warm-cache, which is a different regime, and this change could be net-negative there.

ccache documents that caching a compilation which uses a precompiled header requires explicit sloppiness (`pch_defines`, `time_macros`); without it the documented behaviour is to skip caching those compilations. If that holds here, every PCH-using TU would miss the cache on every CI run — costing more than the PCH saves, because a warm-cache build is far faster than a cold one. **I have not verified this** (no ccache on the machine that ran the measurements), so treat it as documented-but-unconfirmed.

Windows is unaffected either way: `DECENZA_ENABLE_PCH` is OFF there.

**The runs currently in flight do not answer this.** They restore a cache populated from `main`, which has no PCH, so they are effectively cold builds. They test **correctness on six platforms**, which is what they were dispatched for. Measuring the cache interaction needs a *second* run on this branch, against a cache the first run populated.

Proposed order:

1. Let the in-flight runs finish — correctness gate, especially iOS for the `.mm` opt-out.
2. Re-dispatch the four ccache legs that actually build with PCH (linux, linux-arm64, android, macOS, iOS) for a branch-warm second run.
3. Compare those durations against recent `main` runs of the same workflows.
4. Only if the cache is being missed, add `sloppiness=pch_defines,time_macros` to the existing `Configure ccache for CI` step — and note that is not free, since it relaxes define and `__TIME__`/`__DATE__` checking.

If step 3 shows CI getting slower and step 4 does not recover it, the right outcome is to keep the PCH for local development and disable it in CI, rather than ship a change that speeds up one machine and slows six runners.

skialpine:
### Code review

Reviewed this branch. Found 8 issues, all in code this PR adds; all fixed in 4c6b2076.

**1. One real build defect.** `decenza_test_pch` was declared unconditionally while only its `target_precompile_headers()` call was guarded. That target exists solely to build a PCH, so with `DECENZA_ENABLE_PCH` OFF — the default on Windows and iOS — it compiled and linked a static library against `decenza_testlib` and six Qt modules to produce an artefact nothing reuses. Confirmed by configuring with the switch OFF: it drops from 1138 references in `build.ninja` to 0, while all 284 test compile edges remain.

**2. Four wrong counts**, each re-verified against the tree:

| Claim | Actual |
|---|---|
| "107 test targets" | 106 |
| `decenza_testlib` "~60 production sources" | 43 |
| `saw_parity` "21 production sources" | 20 |
| `saw_replay` "2 TUs" | 1 |

The 107 is the notable one: it is the same error [#1728's review](https://github.com/Kulitorum/Decenza/pull/1728#issuecomment-5147482887) already corrected in this exact file, and the file states 106 in two other places — so the new comment made the file contradict itself.

**3. Two claims that were wrong, not just miscounted.**

`"none of them defines anything of its own"` — `decenza_shotlib` has its own `target_compile_definitions(... PUBLIC DECENZA_TESTING)`. The libraries are still `REUSE_FROM`-compatible, but because that define duplicates one the donor already declares, not because none exists. The rule is now stated as "defines nothing the donor does not already define".

`"Verified to work under every generator ... and Xcode for macOS/iOS"` — written before the iOS leg ran, and left in place after it failed, seventy lines below the comment recording that failure. A passing local probe is not a passing platform.

**4. One arithmetic inconsistency.** `saw_parity`'s PCH was justified as recovering "35-40 s", from applying the 28% figure to its 139 s total. That contradicts the per-TU model stated earlier in the same file, under which 22 TUs buys ~11 s. Both cannot be right; the per-TU one is the measured one.

**Checked and NOT changed:** the qmlcache figures (226 generated files, 216/226 = 95.6%, 102/226 = 45.1%) were queried as unverifiable, but re-counting against the current build confirms them.

110/110 tests pass with zero warnings; PCH on and off both configure.

PR #1732, opened 2026-08-01
Wait for the database workers before quitting

skialpine:
`~SerialDbWorker` warns that it is discarding queued tasks. Nothing in production ever waited for it.

Three storages had an `isDbWorkIdle()` whose only callers were tests, `RecipeStorage` had no accessor at all, and `aboutToQuit` drained the BLE queue but never the database. So a dose, note, rating or Visualizer id written in the seconds before quit could be discarded, with the only evidence a warning emitted while the process was on its way out.

Found from the other side in [#1731](https://github.com/Kulitorum/Decenza/pull/1731), where a test destroyed a `CoffeeBagStorage` with two writes still queued:

```
"CoffeeBagStorageWorker" destroyed with 2 DB task(s) still queued
  — those writes are being discarded.        (src/core/dbutils.h:245)
```

That was test hygiene. This is the same mechanism against real user data, without a witness.

## The change

- `RecipeStorage::isDbWorkIdle()` — the missing fourth accessor.
- `ShotHistoryStorage::isDbWriteWorkIdle()` — writes only, excluding the detached read/backup/import threads `isDbWorkIdle()` also counts.
- `MainController::drainDbWork(int timeoutMs = 750)` — waits for all four storages.
- `aboutToQuit` calls it, then waits for any export threads those writes spawned.
- `factoryResetAndQuit()` now posts the quit instead of calling it synchronously.

## What review changed, including two bugs in the fix itself

Three agents, seven substantive findings. The two that mattered most were introduced by this PR, not found in spite of it.

**The nested event loop ran beneath a live QML handler.** `factoryResetAndQuit()` called `QCoreApplication::quit()` directly from an `onClicked` handler, and that reaches `aboutToQuit` *synchronously* — so the handler stayed on the stack through all of shutdown, including this change's `QEventLoop::exec()`. That is the `qFatal` shape that aborted shipped iOS 2.0.0 (#1692). Pre-existing, but the drain made it far likelier to fire: on that path `Settings::factoryReset()` has just deleted the database files while all four workers are still armed, so the loop is guaranteed to spin. Now posted with `Qt::QueuedConnection`, which is what `QQmlApplicationEngine` already does for `Qt.quit()` — the ordinary Quit button was always safe and this path was the outlier.

**The drain abandoned export threads.** `ShotHistoryExporter` subscribes to `shotSaved` and `shotMetadataUpdated`, is declared after `MainController` in `main()` so is destroyed *first*, and every export thread body starts with `if (*destroyed) return;`. Completing a write during shutdown therefore spawned an export that was deterministically dropped — leaving a current DB row beside a stale exported JSON, silently, with `QSaveFile` guaranteeing no torn file to notice. Consistently-stale was a *better* failure than invisibly-disagreeing. Export threads are counted now and waited on.

Also fixed:

- The drain waited on detached read/backup/import threads, which `SerialDbWorker` never tracks and `~SerialDbWorker` therefore cannot discard — so a quit-during-backup spent the entire budget and then warned about writes that were never at risk.
- It pumped events with the reconnect ladders still armed; a DE1 or scale tick landing inside the loop starts a BLE scan microseconds before the disconnect below. Moved beneath the neutralization, and run with `ExcludeUserInputEvents` so a second tap on Quit cannot exit the drain loop (`qcoreapplication.cpp:1520-1529` exits *every* loop) and discard the very write it is saving.
- `scaleReconnectTimer` was never stopped in `aboutToQuit`, and not even captured. Pre-existing; fixed in passing.

## What this does not cover

**Only `aboutToQuit` triggers it.** An Android low-memory kill or force-stop, an iOS SIGKILL (the *normal* iOS termination — `qioseventdispatcher.mm:434` says so outright), a fatal signal reaching the crash handler's re-raise, or an ASan abort never emits it. Those lose queued writes with no warning at all, since `~SerialDbWorker` never runs either.

Android is the primary platform and is usually backgrounded rather than quit, so this covers the deliberate in-app quit and little else there. `Qt::ApplicationSuspended` is the hook that would cover the rest; it is deliberately out of scope, because a wait on the backgrounding path risks an ANR and needs measurement on a device.

The 750 ms budget is also not sized against `DbWriteTxn`'s own worst case (a contended write can wait the full 5 s `busy_timeout`). That is a deliberate trade against Android's shutdown deadline, and the warning no longer reads as though a timeout is anomalous.

## Verification

Build clean, suite 110/110, zero warnings, both text-invariant gates green.

`tst_RecipeStorage::isDbWorkIdleTracksQueuedWork` covers the accessor the drain depends on — new, previously uncovered, and silently load-bearing, since one answering `true` would make the drain skip recipes with no symptom. Its teeth are the `!isDbWorkIdle()` assertion immediately after posting, which is deterministic rather than a race: the counter's decrement rides a ticket whose last holder is the result callback, delivered to the owner thread by queued connection (`dbutils.h:347`), so it cannot run until the test returns to the event loop. Confirmed by breaking it — stubbing the accessor to `return true` fails both that assertion and `~SerialDbWorker`'s discard warning.

`drainDbWork()` itself is not unit-tested: no test instantiates `MainController` (`TESTING.md:262`), and reaching the timeout branch would need fault injection, which CLAUDE.md treats as a stop sign rather than a testing problem.

**Still needs a human**: the quit path itself. Enter a rating or note, quit immediately, restart, confirm it survived.

### DIFF
--- a/src/controllers/maincontroller.cpp
+++ b/src/controllers/maincontroller.cpp
@@ -64,6 +64,8 @@
 #ifdef Q_OS_ANDROID
 #include <QJniObject>
 #endif
+#include <QEventLoop>
+#include <QTimer>
 #include <QQmlEngine>
 #include <QJSEngine>
 
@@ -2468,6 +2470,42 @@ void MainController::applyHotWaterSettings() {
 
 void MainController::applyFlushSettings() {
     sendMachineSettings(QStringLiteral("applyFlushSettings"));
+}
+
+bool MainController::drainDbWork(int timeoutMs) {
+    const auto allIdle = [this]() {
+        return (!m_shotHistory      || m_shotHistory->isDbWorkIdle())
+            && (!m_bagStorage       || m_bagStorage->isDbWorkIdle())
+            && (!m_equipmentStorage || m_equipmentStorage->isDbWorkIdle())
+            && (!m_recipeStorage    || m_recipeStorage->isDbWorkIdle());
+    };
+
+    if (allIdle())
+        return true;
+
+    // Polled, not signalled: the workers count outstanding tasks in an atomic and
+    // emit nothing when it reaches zero, so there is no edge to connect to. This
+    // is the periodic-task case the timer rule allows, not a timer standing in
+    // for a condition — the condition is checked directly on every tick.
+    QEventLoop loop;
+    QTimer poll;
+    poll.setInterval(20);
+    QObject::connect(&poll, &QTimer::timeout, &loop, [&]() {
+        if (allIdle())
+            loop.quit();
+    });
+    QTimer::singleShot(timeoutMs, &loop, [&loop]() { loop.quit(); });
+    poll.start();
+    loop.exec();
+
+    const bool drained = allIdle();
+    if (drained)
+        qDebug() << "Storage writes drained before exit.";
+    else
+        qWarning() << "Storage writes did not drain within" << timeoutMs
+                   << "ms — queued database work is about to be discarded. Expect a"
+                      "\"destroyed with N DB task(s) still queued\" warning next.";
+    return drained;
 }
 
 void MainController::applyAllSettings() {
--- a/src/controllers/maincontroller.cpp
+++ b/src/controllers/maincontroller.cpp
@@ -2472,16 +2472,21 @@ void MainController::applyFlushSettings() {
     sendMachineSettings(QStringLiteral("applyFlushSettings"));
 }
 
-bool MainController::drainDbWork(int timeoutMs) {
+void MainController::drainDbWork(int timeoutMs) {
+    // isDbWriteWorkIdle() on the shot history, not isDbWorkIdle(): the latter also
+    // counts detached read/backup/import threads, which SerialDbWorker never
+    // tracks and ~SerialDbWorker therefore cannot discard. Waiting on those spent
+    // the whole budget on a quit-during-backup for nothing. The other three have
+    // no detached threads, so their isDbWorkIdle() is already write-only.
     const auto allIdle = [this]() {
-        return (!m_shotHistory      || m_shotHistory->isDbWorkIdle())
+        return (!m_shotHistory      || m_shotHistory->isDbWriteWorkIdle())
             && (!m_bagStorage       || m_bagStorage->isDbWorkIdle())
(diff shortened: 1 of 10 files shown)


skialpine:
### Code review

Three agents (CLAUDE.md compliance, deep bug scan, silent-failure hunt). Seven substantive findings, all fixed in `2ff8f2f`. The two most serious were introduced by this PR rather than found in spite of it.

1. **Nested event loop beneath a live QML signal handler.** `factoryResetAndQuit()` reached `aboutToQuit` synchronously from `onClicked`, so this change's `QEventLoop::exec()` ran with that handler on the stack — the `qFatal` shape that aborted shipped iOS 2.0.0 (#1692). Now posted queued.

2. **Export threads abandoned.** Completing a write during shutdown spawned an export thread that was deterministically dropped, leaving the DB row current and the exported JSON stale — silently, and harder to debug than the loss it replaced. Counted and waited on.

3. **Drain waited on reads and backups.** `isDbWorkIdle()` counts detached read/backup/import threads `SerialDbWorker` never tracks, so a quit-during-backup burned the full budget then warned about writes that were never at risk. Split into `isDbWriteWorkIdle()`.

4. **Pumped events with the reconnect ladders armed**, and a second Quit tap could exit the drain loop and discard the write it was saving. Moved below the neutralization; `ExcludeUserInputEvents`.

5. **`scaleReconnectTimer` never stopped** in `aboutToQuit` — not even captured. Pre-existing, fixed in passing.

6. **Header overstated coverage.** An OS kill, crash or ASan abort never reaches `aboutToQuit`, so on Android this covers the deliberate quit and little else. Now says so, with `ApplicationSuspended` named as the out-of-scope follow-up.

7. **Timeout warning pointed at a line that cannot be in a submitted log** (`~SerialDbWorker`'s discard fires after `AsyncLogger::uninstall()`), and asserted loss that often does not happen. Rewritten; it now names which storage is stuck.

Verified clean by the agents: the `QTimer::singleShot` cannot fire into the destroyed loop, `RecipeStorage::isDbWorkIdle()`'s null-worker `true` is correct in every reachable state, the new test's `!isDbWorkIdle()` assertion is deterministic rather than racy, and the polling loop is the periodic case the timer rule permits rather than a timer standing in for a condition.

PR #855, opened 2026-04-25
Settings split Tier 2: Brew + Dye + Network sub-objects (3 of 4)

skialpine:
## Summary

Continues the Settings sub-object split started in #852 (Tier 1). This PR
lands **3 of the 4 deferred Tier 2 domains** as discussed in
\`openspec/changes/split-settings-tier-2/proposal.md\`. SettingsApp is
deferred to a follow-up PR.

| Metric | Before (Tier 1) | After (this PR) | Goal |
|---|---|---|---|
| \`settings.h\` | 904 lines | **466 lines** | ≤ 200 (after App) |
| \`settings.cpp\` | 3,882 lines | **2,043 lines** | — |
| Sub-objects | 7 | **10** | 11 |

3 commits, one per domain, organized for review:

1. **Brew** — espresso/steam/hot-water/flush + presets + brew/temperature
   overrides + ignoreVolumeWithScale (~700 lines stripped). Bundles all
   shared infrastructure changes (settings.h/cpp forward decls, main.cpp
   qmlRegister calls, consumer migrations in maincontroller, profilemanager,
   settingsserializer, mcptools_*, tests). The Steam pitcher-preset
   \`Connections\` block on SteamPage is retargeted to \`Settings.brew\`.
2. **Dye** — 16 dye* metadata fields + bean-preset CRUD + GrinderAliases
   helpers + recomputeBeansModified chain. The cross-domain rating ↔
   enjoyment connect is rewired to \`m_dye->setDyeEspressoEnjoyment\`.
   Three new \`tst_settings.cpp\` tests prove the cross-domain wiring
   and beansModified NOTIFY chain. SettingsDye holds a non-owning
   SettingsVisualizer* for the espressoEnjoyment default fallback.
3. **Network** — shot-server, web security, auto-favorites, saved
   searches, shot history sort, exportShotsToFile, layoutConfiguration
   + 13 layout invokables, Discuss-Shot URLs.

## Drive-by fixes

Two unrelated UI fixes rolled in for testing convenience (called out in
their respective commit messages):

- **Niche Duo grinder burr suggestions** (commit 1, in BrewDialog.qml):
  the burr-suggestion list was merging stale shot-history entries with
  stock burrs, so a Niche Duo (83mm flat) would offer 63mm conical from
  old Niche Zero shots. Skip the history merge when \`isBurrSwappable\` is
  false.
- **Cup-fill view briefly rendering full** (commit 3, in EspressoPage.qml):
[...]

PR #1736, opened 2026-08-01
Hold the cup fill after a shot ends until the page changes

skialpine:
## Problem

When an espresso finishes, the cup fill view blanks instantly. The coffee disappears while the espresso page is still on screen for the post-shot stop-overlay window — and while the weight text above it still reads the final dose, so for those seconds the view shows "36.2g / target 36g" over an empty cup.

## Cause

`hasExtraction` in `cupGeometry()` listed only `Preinfusion`, `Pouring` and `Ending`. When the DE1 leaves `State::Espresso`, `MachineState` maps it to `Phase::Ready`, the gate goes false, and `fillRatio` collapses to 0.

The redraw that shows the empty cup is *not* the animation timer — that stops at the same phase boundary, deliberately, since 07ab75fc. It is `onCurrentWeightChanged`, still firing at the scale's ~10 Hz, so the blank lands within ~100 ms and then persists.

The comment above that gate said "during/after extraction", but the "after" half was never implemented. The gate came from #855, defending a *pre*-flow case: a residual scale reading (notably a hot-water pour's frozen weight) drawing a full cup during `EspressoPreheating`. That defence had to survive.

A second, independent blank path: `if (root.currentWeight <= 0) return` in the liquid canvas, plus matching guards in the effects canvas. Lifting the cup or portafilter empties it even with the phase gate fixed.

## Fix

Latch `extractionSeen` when the view observes a flow phase; hold the final weight at the transition out of the cycle; route every drawn read through a derived `displayWeight`. One value closes the phase gate and both canvas guards together.

The hold needs no expiry — `StackView.replace()` destroys the view, so "until the screen changes" is the instance's lifetime. No timer, no tuning parameter. The only clearing rule is a *new* cycle on a surviving instance.

**Deliberately local.** `MachineState::scaleWeight()` already has this freeze for hot water, and it would be the tidier home — but that one earned source placement because a *separate* component (the completion overlay) needed it. Here one view does, and `scaleWeight`'s other readers (MQTT, widget snapshot, MCP surfaces, SAW learning) want the truth. Accepted consequence: `CupFillView` is now the one place displayed weight may diverge from `scaleWeight`, and only while holding.

Two things that differ from the written plan, both from implementation:

- **The latch includes `Ending`.** Switching from the chart to the cup view *during* `Ending` would otherwise never latch — `onPhaseChanged` doesn't fire on construction and the next transition is out of the cycle. `Component.onCompleted` seeds the same case.
- **Phase predicates are functions, not just bindings.** The handler needs the predicate for both the new and the previous phase, and reading a binding on `phaseChanged` from inside a handler on that same signal depends on evaluation order QML does not guarantee.

Bonus: repaints now key off `displayWeight`, so during the hold the scale's stream no longer repaints a frame that cannot change.

## Preserved

Pre-flow behaviour from #855: with nothing latched, `Ready`/`EspressoPreheating` still render an empty cup whatever the scale reports. `animTimer` stays stopped through the hold, so it is a static final frame rather than three seconds of decorative repainting.

## Verification

Build clean, full suite green locally (**110 passed, 0 failed, 0 skipped**) — no CI job builds or tests a PR.

Manual, cup fill view selected:

- Normal shot — cup and weight hold until the shot review appears ✓
- Cup lifted during the hold — fill, crema and text unchanged ✓
- Abort during preheat — cup stays empty ✓ (correct: no coffee was extracted)
- Cup resting on the scale, espresso page opened pre-shot — empty ✓

Stopping a shot early turned out **not** to exercise this: that path navigates to the shot review immediately, so the view is destroyed before any hold window exists. The latch is unconditional and costs nothing there, so no special case was added. The mechanism behind that immediate navigation was not investigated — out of scope. Recorded in `design.md` so the hold isn't later "fixed" for appearing inert after an early stop.

## Scope

One file, `qml/components/CupFillView.qml`. No C++, no new QML file (so no `CMakeLists.txt` edit), no schema change. No automated test — QML is verified manually in this project. No wiki manual entry: this restores intended behaviour rather than adding a feature.

OpenSpec change `hold-cup-fill-after-shot` rides along; archive + spec-sync will land as the final commit on this PR.


skialpine:
## Review pass — reworked the capture

Four review agents (code, comments, tests, silent failures). Two critical findings were real and shared one root cause: **sampling the weight once at the phase boundary**. Replaced with peak-tracking since first flow, which fixes both.

**1. Post-stop drip was lost.** Verified: `finalWeight` comes from the *post-settling* weight (`maincontroller.cpp:3580`), settling runs up to ~10 s, and `MAX_PLAUSIBLE_POST_STOP_DRIP_G` is 5.0. A boundary sample froze the cup several grams under what the shot record, the review page and the page's own live readout all report — and that readout sits right beside the cup, so the user would watch one number climb to 36.0 while the big one stayed at 34.0.

**2. Lifting the cup at `Ending` froze a negative weight.** The scale reads negative when the cup comes off. A boundary sample taken at that instant held that value for the rest of the view's life: empty cup, negative number, and putting the cup back no longer fixed it. The freeze converted a one-frame glitch into a permanent wrong result — strictly worse than the bug being fixed.

Peak-tracking also means nothing is captured at the boundary, so the binding-vs-handler ordering question the first version reasoned about is simply gone.

## A pre-existing bug that undermined the premise

The `hasExtraction` gate **never reached the drawing code**. It zeroes `fillRatio`, but the liquid canvas then adds the `+0.12` crema boost and bailed out on weight alone — so a resting cup on the scale drew a **12%-full cup labelled with its own weight**. That has been broken since the gate landed in `304769ea`, and this PR claimed to preserve the guarantee it was supposed to provide. Fixed here rather than deferred.

## Other fixes from the review

- `isEspressoCycle()` now spells out the four phases like every C++ site (`machinestate.cpp` `isInEspresso`, `profilemanager.cpp`, `mcptools_control.cpp`) instead of testing an ordinal range. A phase inserted between `EspressoPreheating` and `Ending` would have silently changed its meaning with nothing failing.
- `onCurrentFlowChanged` no longer repaints while holding. The DE1 keeps streaming samples at idle, which was reinstating exactly the repaints the `displayWeight` keying was meant to avoid — the performance claim was self-cancelling.

## Corrections to my own artifacts

- `design.md` named the widget snapshot and SAW learning as `scaleWeight` readers. **Neither reads it** — `machinestatussnapshot.cpp` has no weight field, and `ShotTimingController` takes weight via `onWeightSample()` straight from the weight processor. Replaced with the verified list (ShotServer, MQTT, four MCP surfaces).
- "The only clearing rule is a new cycle" ignored the chart↔cup toggle, which destroys and recreates the view.
- The latch phase list omitted `Ending`, which the code does latch on.
- `#855` now cites commit `304769ea` — the gate landed incidentally inside that settings-split PR, so the bare PR number sends a reader to unrelated domain-object work.

## Considered and deliberately not changed

Recorded in `design.md` with reasons rather than dropped silently: `Refill` mid-shot freezing a partial pour (that *is* what was extracted, and the modal dialog explains it); the chart toggle losing the hold (the user changed what they were looking at); the live info bar (with peak-tracking the two now agree except when the cup is lifted); and the missing accessible name on the cup's weight text (a gap in the original component).

One adjacent pre-existing issue surfaced and left alone: when a shot save fails, `main.qml:4221` warns to the console and returns without navigating, stranding the user on the espresso page with no feedback. Out of scope here.

## Re-verified

Build clean, suite green: **110 passed, 0 failed, 0 skipped**.

Note for the record: the first build attempt after the rework silently ran against the *other* clone (`Decenza-Desktop`) after Qt Creator's active project drifted mid-session — a 950 ms "success" that compiled none of this. Re-run against the right checkout before trusting the numbers above; those are from the re-run.

PR #1737, opened 2026-08-01
Give the WiFi scale reconnect the same mDNS deadline discovery uses

skialpine:
## What

`DecentScaleWifi::attemptHostname()` called `MdnsResolver::resolveHostname(host)` with no timeout, taking the **2000 ms default**. The discovery path passes **5000 ms**. Same device, same responder, two different deadlines — and the reconnect path had the short one.

One line of behaviour change; the rest is a named constant and the reasoning behind it.

```diff
- const QString ip = MdnsResolver::resolveHostname(host);
+ const QString ip = MdnsResolver::resolveHostname(host, kMdnsResolveTimeoutMs);
```

## Why

The HDS responder regularly takes 2-4 s to reply (likely the ESP32 waking from WiFi power-save). That isn't a guess — it's written down at the discovery call site in `blemanager.cpp` as the reason that path uses 5000 ms. The reconnect path silently took the default and gave up 2 s in.

Cost, from a tablet log (2.0.1 build 3519, SM-X210, 7h52m session):

- **82 consecutive reconnect misses over 7.5 hours**, every one ending at ~2003 ms with `records= 0` — nothing received at all.
- The scale was awake and on mains power throughout (`battery_percent: 100, charging: true` when it finally connected).
- The user spent the whole session on FlowScale instead of the real scale.
- A user-initiated scan three minutes after one of those misses resolved the same hostname in **362 ms**.

The resolver itself was fine: `homeassistant-chv.local` resolved through the *same* `resolveHostname()` function in 3 ms, 300 ms before the first `hds.local` miss. So this is not a multicast lock, not permissions, and not the network.

## Honest limits

**Necessary, possibly not sufficient.** The one observed success resolved in 362 ms — comfortably inside the old 2 s window — while a DNS-SD browse was running concurrently. So "the responder needs browse traffic to wake" survives as an alternative explanation that this change does not address. The header comment records that, and says what to do about it: browse from the reconnect path, where the auto-connect wiring already exists in `BLEManager`'s `resultFound` handler.

If the next beta still shows misses — now ending at ~5003 ms with `records= 0` — that outcome falsifies the timeout theory cleanly and promotes the browse fix.

**Verification is deferred.** The edited line is inside `#ifdef Q_OS_ANDROID`, so the macOS build does not compile it. Local suite is a regression check only; field verification arrives with the next beta.

## Not in this PR

Two further defects found in the same investigation, deliberately left out to keep this one falsifiable on its own:

1. Reconnect never re-resolves after a *transport* error — a cached-IP `Host unreachable` short-circuits before the mDNS fallback, so a scale that moved DHCP lease never comes back without a manual scan. Reproduced on macOS.
2. Reconnect never browses, though `#1689` built the browse and the auto-connect handler that consumes it.

## Test plan

- [ ] Full local suite (macOS) green
- [ ] Android behaviour observed in the next beta — look for `[MdnsResolver] done host= hds.local` with `elapsed=` in the 2000-5000 ms band and `aRecords= 1`

### DIFF
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -2333,14 +2333,16 @@ void BLEManager::scanForDevices() {
     // to its hostname, and running the fallback only when the browse came back
     // empty would hide the old scale whenever a new one is present.
     //
-    // The 5 s A-record timeout is unchanged: the HDS responder regularly takes
-    // 2-4 s to reply (likely the ESP32 waking from power-save). The browse runs
-    // longer, alongside the ~15 s BLE scan, because a DNS-SD browse's first
-    // callback is a dump of the resolver's cache — stale instances included —
-    // and the resolver's own pruning of those arrives seconds later.
+    // The A-record timeout is kHdsResolveTimeoutMs, shared with the reconnect
+    // path; the responder latency that sets it is documented on the constant.
+    // The browse runs longer, alongside the ~15 s BLE scan, because a DNS-SD
+    // browse's first callback is a dump of the resolver's cache — stale
+    // instances included — and the resolver's own pruning of those arrives
+    // seconds later.
     ensureWifiDiscovery();
     m_wifiDiscovery->browse(15000);
-    m_wifiDiscovery->probe(WifiScaleDiscovery::defaultFallbackHostnames(), 5000);
+    m_wifiDiscovery->probe(WifiScaleDiscovery::defaultFallbackHostnames(),
+                           WifiScaleDiscovery::kHdsResolveTimeoutMs);
 
     // USB is otherwise a free-running background poll that the scan button never
     // touched, so a scale plugged in just before a scan appeared only when the
@@ -2374,7 +2376,8 @@ void BLEManager::browseWifiScales(int timeoutMs) {
                        .arg(MdnsResolver::activeBrowseBackendName())
                        .arg(timeoutMs));
     m_wifiDiscovery->browse(timeoutMs);
-    m_wifiDiscovery->probe(WifiScaleDiscovery::defaultFallbackHostnames(), 5000);
+    m_wifiDiscovery->probe(WifiScaleDiscovery::defaultFallbackHostnames(),
+                           WifiScaleDiscovery::kHdsResolveTimeoutMs);
     emit scanningChanged();
 }
 
--- a/src/ble/scales/decentscalewifi.cpp
+++ b/src/ble/scales/decentscalewifi.cpp
@@ -45,6 +45,9 @@ template struct AccessBypass<WsPrivateSocketTag, &QWebSocketPrivate::m_pSocket>;
 } // namespace
 
 #include "../../network/mdnsresolver.h"
+// For kHdsResolveTimeoutMs — the A-record deadline is a property of the HDS
+// responder, shared with the discovery path, so both read it from one place.
+#include "../../network/wifiscalediscovery.h"
 
 #define WIFI_LOG(msg)  SCALE_LOG("DecentScaleWifi", msg)
 #define WIFI_INFO(msg) SCALE_INFO("DecentScaleWifi", msg)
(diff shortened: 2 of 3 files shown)


skialpine:
### Code review

Found 2 issues:

1. The comment claims a timing margin that no longer exists. It says the 5000 ms resolve "stays well inside BLEManager's 20 s scale-connection timer" — but the worst-case chain is now **exactly 20 s**, equal to `m_scaleConnectionTimer`'s interval (`blemanager.cpp:82`, `setInterval(20000)`), leaving zero buffer. Derived: resolve fails (5 s) → `dialCachedIpAfterResolveFailure` → cached-IP recognition timeout (5 s) → re-resolve (5 s) → recognition (5 s). Before this PR the same chain was 2+5+2+5 = 14 s, a real 6 s margin. The tie appears benign today (`recognitionFailed` is only wired for manual entries, so either order ends the attempt safely), so this is not a functional bug — but CLAUDE.md says an un-sourced claim in a comment "gets believed and then licenses wrong code," and the next person adding a step here will read "well inside" and not re-derive it. Suggest replacing the sentence with the derived worst case plus an explicit note that this constant cannot be raised again without shortening that chain or lengthening the timer.

https://github.com/Kulitorum/Decenza/blob/8a8e6af22016c9f4a44197a926ec74dc98486c5e/src/ble/scales/decentscalewifi.h#L248-L253

2. The PR creates a third uncoupled copy of `5000` while claiming the opposite (CLAUDE.md says "Centralize anything produced at more than one site — never hand-roll it per call... A repeated format, prefix, tag, wording, or policy is a drift opportunity: each copy is free to change alone, silently, and nothing fails when one does"). The new comment states "5000 ms is chosen to MATCH the discovery path... one number, one justification, both paths" — but the discovery path's value is a bare literal at two untouched sites, `blemanager.cpp:2343` and `:2377`. After this PR the value lives at three independent sites with nothing linking them. That is the same drift class this PR exists to fix: someone edits the `blemanager.cpp` literals for a documented reason, the header copy goes stale, the two paths disagree again, and nothing catches it. Suggest hoisting one shared constant that all three sites reference.

https://github.com/Kulitorum/Decenza/blob/8a8e6af22016c9f4a44197a926ec74dc98486c5e/src/ble/blemanager.cpp#L2342-L2344

<details>

- [Worth a look] `resolveHostname` is called without the `cancel` token that the discovery path passes (`wifiscalediscovery.cpp:88-135`), so raising 2 s → 5 s widens the window where a detached worker keeps resolving after the object is gone. Not a hang — this is a `QThread::create` thread, not a pool task, so `~QCoreApplication` does not wait on it; resource tidiness and path inconsistency only. Same pattern already exists in `mqttclient.cpp:310`. (score: 50)
- [OK to skip] `guard.data()` is read on the worker thread and passed as the `invokeMethod` receiver; `~DecentScaleWifi`/`~ScaleDevice` bodies run before `~QObject` clears the `QPointer`, so a sub-microsecond window exists where the worker posts to a partially-destroyed object. Pre-existing; this PR only widens the coincidence window 2.5×. Fix would be the `qApp`-receiver shape from `wifiscalediscovery.cpp:105-107`. (score: 50)

</details>

Verified and found accurate: every other factual claim in the new comment — the 2000 ms `MdnsResolver` default (`mdnsresolver.h:99`), the 5000 ms at the discovery call site and its 2-4 s responder rationale (`blemanager.cpp:2337-2343`), `QThread::create` vs `QThreadPool`, and the `resultFound` auto-connect wiring (`blemanager.cpp:2542-2580`). History also confirms the 2000 ms was never a deliberate choice here — it arrived as the function default in `a06c24f6`, and #1603 fixed the non-Android leg only, so this does not undo a prior fix. No concern from #1603, #1624 or #1689 is re-introduced.

skialpine:
Local macOS suite passes. Rebased onto `6d4ed67d` (#1736); clean, no overlap.

Both review findings are fixed:

- The "well inside the 20 s timer" claim is replaced with the derived worst case (5+5+5+5 = exactly 20 s, equal to `m_scaleConnectionTimer`), plus an explicit note that the constant cannot be raised again without shortening that chain or lengthening that timer.
- The three uncoupled copies of `5000` are now one: `WifiScaleDiscovery::kHdsResolveTimeoutMs`, read by both `blemanager.cpp` probe sites and the reconnect resolve. The driver's private duplicate is gone, so `decentscalewifi.h` drops out of the diff entirely.

**New evidence, from a running macOS build.** The premise behind this change — that the HDS responder answers in 2-4 s, not inside 2 s — now has a second, independent observation. macOS resolves through `QHostInfo`, not the Android mDNS path, and a hostname fallback there took **3.68 s**:

```
[ 8.822] Resolving hdstest.local via QHostInfo...
[12.504] Resolved hdstest.local to 192.168.10.241 via QHostInfo
```

Different OS, different resolver, same band. That moves the latency from "something we inferred from one Android log" to a property of the responder itself, which is the claim the constant actually needs. Recorded on the constant.

Also visible in that run, and relevant to the alternative hypothesis this PR does not close: the re-resolve returned **the same IP** the cached-IP attempt had just failed to validate, and the second dial to that identical address succeeded in 343 ms. The cache was not stale — the first dial appears to have woken the scale. Same "needs a stimulus" shape as the browse-traffic theory noted in the description.

Not exercised by this run: the two refactored probe call sites in `blemanager.cpp` (no user-initiated scan occurred), so those are compile-verified only.

Android verification still arrives with the next beta.

PR #1739, opened 2026-08-01
Browse for a saved WiFi scale when the direct reconnect has failed

skialpine:
## Why

A saved WiFi scale cannot recover on its own once its address stops resolving. The reconnect path only ever asks for the hostname directly, and this responder does not reliably answer that — **82 consecutive reconnects received zero mDNS records over 7.5 h** while the scale was awake, mains-powered and serving WebSocket traffic on its IP. The user spent that session on the estimated flow scale. Opening Connections and tapping Scan recovered it immediately, because the scan **browses** for `_decentscale._tcp` instead of asking for a name.

**#1737 is falsified.** It raised the A-query deadline 2 s → 5 s on the theory that a responder answering in 2-4 s was being cut off. The next Android session showed the misses continuing, now ending at ~5002 ms with `records= 0`. The deadline was never the problem. That PR's comment claiming otherwise is corrected here; the constant stays, since agreeing with the discovery path is right on its own terms.

**Not Android-specific.** macOS reproduces it through a different resolver entirely:

```
reconnect:  QHostInfo resolution failed for hdstest.local: Host not found
user scan:  DNS-SD found Half Decent Scale (hdstest) at 192.168.10.241   <- 1.32 s
control:    hds.local resolved by A-query in 3.07 s, same scan
```

So it is a property of the responder, not of one platform's resolver — which is why the browse runs on every platform rather than being gated to Android.

## What

The reconnect ladder browses when a direct attempt has failed, and connects on an exact saved-address match. The auto-connect half already existed — its comment anticipated a browse firing "per user action or reconnect tick", and only the call was missing.

Isolation from user scans is **structural, not conditional**: a separate `WifiScaleDiscovery` instance, because `browse()` only cancels within an instance and `isScanning()` references only the scan's. The reconnect handler also never touches `m_wifiResults` or the device rows.

## Two bugs found by running it, not by reasoning

Worth recording, because the test suite was green through both:

1. The browse was first called from `tryDirectConnectToScale()` — which stops being called once the WiFi→BLE fallback connects a scale. A log showed two timeouts 20 s apart with no attempt between them. It now fires from `onScaleConnectionTimeout()`, at the failure itself.
2. Moving it there exposed the second: stopping the browse on scale-connect **aborted** it, because the fallback connects ~1 s after the timeout and a browse needs ~1.3 s. The browse now runs to its own deadline.

## Verified live (macOS)

```
[22.719] Scale connection timeout — not found
[22.721] Reconnect: browsing for wifi:hdstest.local (a direct attempt already failed)
[22.722] DNS-SD browse for _decentscale._tcp.local (timeout 5000 ms)
[22.722] WiFi scale hdstest.local unreachable — trying Bluetooth
[23.118] Reconnect browse found WiFi scale: hds.local -> 192.168.10.145 (saved=wifi:hdstest.local)
```

- Trigger fires at the timeout, on a path that actually runs
- 5 s deadline applied, not the scan's 15 s
- Browse survives the fallback firing 1 ms later
- Ladder re-ticks afterwards (`Auto-reconnect attempt 1`)
- **Anti-substitution held twice** against a real second scale on the LAN — found `hds.local`, refused to connect

**Browse-driven recovery itself is still evidenced only by the Android logs.** macOS could not be made to keep the direct path broken (the cached IP stayed valid), so the case this exists for was never forced locally. Stated plainly rather than implied by the passing checks.

## Tests

110 passed, 0 failed, 0 skipped locally. Five new slots in the existing `tst_wifiscalediscovery` (no new test file). They are **pure-predicate** tests over the gating and anti-substitution rules — `BLEManager` cannot be constructed in the suite, and reaching the real branches would need a fault-injection harness. Note both bugs above were invisible to exactly this kind of test.

## Spec

`openspec/changes/browse-for-wifi-scale-reconnect/` — modifies `wifi-scale-discovery`: narrows "no browse outside a user-initiated scan" (the requirement already carved out a saved WiFi primary) and adds a requirement for the reconnect browse.


skialpine:
### Code review

Found 4 issues:

1. The reconnect browse result is discarded in exactly the case this PR exists for, so with a BLE scale in range the behaviour is identical to pre-PR (bug due to `main.cpp`'s `scaleDiscovered` handler early-returning on `if (physicalScale && physicalScale->isConnected()) return;`)

`maybeAutoConnectBrowsedScale()` recovers by emitting `scaleDiscovered(...)`. Field sequence: the 20 s timeout starts the browse **and** `beginWifiFallbackToBleScan()`; the BLE fallback connects ~1 s later; the browse resolves ~0.3 s after that; the emit is then a no-op. It compounds — that connect clears `m_wifiDirectAttemptFailed` and stops the ladder, so no further browse runs, and `wifiPreferTimer` cannot rescue it because it probes only the cached IP (the stale one, broken by hypothesis). The app's working switch-back, `switchToWifiPrimary()`, emits `disconnectScaleRequested()` **before** `scaleDiscovered()` precisely to pass that guard; this path omits it. The comment claiming "the worst case is switching from the BLE fallback back to the WiFi scale the user actually saved — which is the point" asserts behaviour the code does not implement.

https://github.com/Kulitorum/Decenza/blob/05c0fed10cd921d7ab5852dbfb33c5a6f9fd4ca6/src/ble/blemanager.cpp#L2607-L2626

2. Unsourced timing figures in a comment (CLAUDE.md says "An un-sourced claim in a comment gets believed and then licenses wrong code")

"it arrives about a second after the timeout ... and a browse needs longer than that (1.3 s measured)". Neither number is measured for this scenario — 1.32 s came from a macOS *user scan*, and in the actual reconnect log the browse returned 0.4 s after starting. The numbers also invert the argument: if the browse finishes in ~0.4 s and the fallback lands ~1 s later, the browse completes before any cancel, so the real justification is the anti-substitution match, not a race.

https://github.com/Kulitorum/Decenza/blob/05c0fed10cd921d7ab5852dbfb33c5a6f9fd4ca6/src/ble/blemanager.cpp#L1690-L1707

3. `m_wifiDirectAttemptFailed` is not reset where its sibling failure flags are, so it survives a saved-scale change (CLAUDE.md says "Centralize anything produced at more than one site ... each copy is free to change alone, silently")

`clearSavedScale()` and `connectToSavedScale()` both reset `m_scaleConnectionFailed` and `m_flowScaleFallbackEmitted` but not the new flag. Scale A times out (flag true), user switches to Scale B, and Scale B's *first* attempt browses in parallel — contradicting this PR's own design decision, "Browse only after the direct attempt has failed, not alongside it."

https://github.com/Kulitorum/Decenza/blob/05c0fed10cd921d7ab5852dbfb33c5a6f9fd4ca6/src/ble/blemanager.cpp#L2040-L2052

4. A pre-existing comment in a touched file is factually wrong about current behaviour (CLAUDE.md says "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'")

The `m_manualEntryDiscovery` comment says the shared instance auto-connects "(and no user-initiated scan is in progress)". There is no such condition — `maybeAutoConnectBrowsedScale()` matches on the saved primary only, and its own comment says auto-connect fires "including during user-initiated scans".

https://github.com/Kulitorum/Decenza/blob/05c0fed10cd921d7ab5852dbfb33c5a6f9fd4ca6/src/ble/blemanager.cpp#L884-L894

<details>

- [Worth a look] `kReconnectBrowseTimeoutMs = 5000` duplicates `WifiScaleDiscovery::kHdsResolveTimeoutMs = 5000`, coupled only by a prose comment — the same drift #1737's review flagged and that PR fixed by consolidation, one commit earlier. One-line fix: `= WifiScaleDiscovery::kHdsResolveTimeoutMs`. (score: 75, grouped here as it is a maintainability rather than functional defect)
- [OK to skip] "Same reasoning that split `m_manualEntryDiscovery` out" overclaims — that instance was split to suppress auto-connect, not for cancellation/indicator reasons. "Same pattern" would be accurate. (score: 50)

</details>

Verified clear: the `maybeAutoConnectBrowsedScale` extraction dropped no guard (there was no `m_userInitiatedScaleScan` condition to lose); no dangling lifetime path; no browse stacking; `isScanning()` correctly excludes the reconnect instance; USB and simulator paths close the gate; the rewritten #1737 comment accurately describes what that PR changed. History supports the change — #1689 deliberately avoided browsing here to prevent silent substitution, and that invariant is preserved by the exact saved-address match.

PR #1741, opened 2026-08-01
Do not switch back to a WiFi primary that is already connected

skialpine:
## The bug

Pressing **Scan for devices** while the saved WiFi scale was connected disconnected it and redialled it — about 8 seconds with no weight readings, every time.

`maybeAutoConnectBrowsedScale()` decided the connected scale was a *backup* from `m_scaleDevice->isConnected()` alone. That test is just as true when the connected scale IS the saved primary, and this function is shared by three callers: the user scan's DNS-SD browse handler, the user scan's A-record probe handler, and the reconnect browse's. So a user scan that re-found the healthy primary asked to switch back to it, and `switchToWifiPrimary()` dropped the scale and dialled the same hostname at the same IP.

From the tablet log, session starting 2026-08-01 11:17:55:

```
[  4.154] WebSocket connected — peer=192.168.10.145:80
[ 22.273] Starting device scan...
[ 22.636] Reconnect browse found the saved primary hds.local at 192.168.10.145
          while a backup scale is connected — requesting switch-back
[ 22.683] (again — once for the DNS-SD hit, once for the mDNS hit)
[ 30.897] WebSocket disconnected (expected)
[ 32.494] Auto-reconnect attempt 1 → connected to the same scale
```

A scan during a shot would have taken the scale away mid-pour.

## The fix

Gate the switch-back on `m_wifiDirectAttemptFailed`, which already carries exactly the needed meaning — set only by the connection timeout for a saved `wifi:` primary, cleared by any connect that is not the WiFi→BLE fallback, cleared again when the saved address changes. So while a scale is connected, `true` can only mean the connect was the fallback (a backup) and `false` means the primary itself is connected. No new state, no existing state repurposed.

Rejected alternatives are in the change's `design.md`: comparing the connected device's hostname needs an accessor `ScaleDevice` does not have plus a downcast, for no more correctness; suppressing the duplicate request treats the symptom, since one needless disconnect is the bug.

Also drops "Reconnect" from that log line — wrong for the two scan-side callers — and corrects the comment in `onScaleConnectedChanged()` that asserted an already-connected scale always routes through `switchToWifiPrimary()`.

## Verification

- `switchBackOnlyWhenTheConnectedScaleIsABackup()` covers all four `(connected, directAttemptFailed)` combinations. Added as a slot in an existing file, not a new one.
- **Confirmed it can fail:** restoring the old semantics (`return directAttemptFailed || scaleConnected`) turned it red — 109 passed / 1 failed. Restored, and the full suite is 110 passed, 0 failed, 0 skipped, no warnings.
- Spec delta under `openspec/changes/no-switch-back-to-connected-primary/`, validated; log-marker gate passes.

No wiki manual entry: this removes an unintended disconnect rather than adding or altering a documented feature.

### DIFF
--- a/openspec/changes/no-switch-back-to-connected-primary/.openspec.yaml
+++ b/openspec/changes/no-switch-back-to-connected-primary/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-08-01
--- a/openspec/changes/no-switch-back-to-connected-primary/design.md
+++ b/openspec/changes/no-switch-back-to-connected-primary/design.md
@@ -1,0 +1,75 @@
+# Design
+
+## Context
+
+`maybeAutoConnectBrowsedScale()` is the single auto-connect path shared by three
+callers: the user scan's DNS-SD browse handler, the user scan's A-record probe
+handler, and the reconnect browse's handler. Sharing it is deliberate — an
+earlier duplicate of the saved-scale match drifted from the row dedupe and let a
+browse hit rewrite a row out from under the matcher.
+
+Its already-connected branch exists because `main.cpp`'s single-scale invariant
+silently drops a `scaleDiscovered` emission while a scale is connected. Routing
+through `wifiPrimaryReachable` → `switchToWifiPrimary()` is what gets past that,
+because the switch-back drops the current scale *before* emitting.
+
+The branch's condition was `m_scaleDevice && m_scaleDevice->isConnected()`,
+which does not distinguish a backup from the primary itself.
+
+## Goals / Non-Goals
+
+- **Goal:** request a switch-back only when there is something to switch back
+  *from*.
+- **Goal:** keep one implementation shared by all three callers.
+- **Non-Goal:** teach `BLEManager` the connected scale's identity. `ScaleDevice`
+  exposes no address or hostname accessor, and adding one for this would mean a
+  downcast to `DecentScaleWifi` or a new tracked member — more machinery than
+  the question needs.
+
+## Decisions
+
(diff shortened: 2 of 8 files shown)


skialpine:
Closing: the premise does not survive checking.

The "~8 s outage on every Scan" this PR was built on never happened. The disconnect I attributed to the switch-back is the DE1 going to sleep — `[30.676] Idle → Sleep`, `[30.767] DE1 sleep + WiFi scale - closing WS for the sleep interval`, `[30.922] Scale disconnect was deliberate (DE1-sleep)` — the same mechanism visible again later in the same session. I joined the switch-back request at t=22.6 to the disconnect at t=30.9 by adjacency without reading the lines between them.

`switchToWifiPrimary()` never ran at all: it logs unconditionally at INFO and that line is absent from the session. `main.cpp` already blocks this case, with a stricter test than the one added here:

```cpp
if (physicalScale->type() == QStringLiteral("decent-wifi"))
    return false;                                  // already on the WiFi primary
```

And the gate itself does not hold. `m_wifiDirectAttemptFailed` is stale-true on the browse feature's own success path: the browse dials the primary without clearing `m_wifiFallbackToBleActive`, so the primary's connect is misread as a fallback connect and the flag is never cleared — meaning the predicate would read true with the primary connected, exactly where this PR claims to prevent it.

What is real, and is being fixed separately: `m_wifiFallbackToBleActive` left set by the browse's primary dial, `m_browsedPrimaryIp` set on a request `main.cpp` rejects and then never cleared, and the log line that says "while a backup scale is connected" when no backup is connected.

PR #1742, opened 2026-08-01
Clear the WiFi reconnect flag when the primary is what connected

skialpine:
Replaces #1741, which was closed — its premise did not survive checking. See that PR's closing comment.

## The bug

`m_wifiDirectAttemptFailed` gates two things: whether the reconnect browse runs (`shouldBrowseOnReconnect`) and whether a browse hit asks for a switch-back. It is meant to be cleared the moment a connect proves the primary is back.

It is not cleared on the browse's own success path — the path the browse feature exists to produce:

1. The direct dial to the saved WiFi primary times out. `onScaleConnectionTimeout()` sets the flag, starts the reconnect browse, then `beginWifiFallbackToBleScan()` sets `m_wifiFallbackToBleActive` and starts a BLE scan.
2. The browse resolves the primary. Nothing is connected yet, so `maybeAutoConnectBrowsedScale()` dials it via `scaleDiscovered` — **without** clearing `m_wifiFallbackToBleActive`, because the BLE scan it is racing is still running.
3. The primary connects. `onScaleConnectedChanged()` reads `wasWifiFallbackConnect == true` and declines to clear the flag, even though the scale that connected is the primary itself.

Nothing clears it afterwards while that scale stays connected: there is no further connect, and `setSavedScaleAddress()` is not called because the address did not change. So a successful recovery leaves the machinery believing the direct attempt is still failing, for the rest of the session.

That is already forbidden by a requirement the spec states — *"Browse is not run when the direct attempt succeeds"* — reached by a route it did not anticipate.

## The fix

Ask what connected, rather than whether a fallback was in flight. `connectedScaleIsWifiPrimary()` uses the same `type() == "decent-wifi"` test [`main.cpp:2465`](https://github.com/Kulitorum/Decenza/blob/19679ee2/src/main.cpp#L2464-L2467) already applies before honouring `wifiPrimaryReachable`, so the two layers cannot drift apart on what "on the primary" means.

Deliberately **not** fixed by clearing `m_wifiFallbackToBleActive` at the browse's dial site: that flag also gates whether an in-flight fallback scan may adopt a discovered BLE scale and whether a second fallback is permitted, so clearing it there would change which scale the ladder lands on in order to fix a bookkeeping error.

## Two smaller defects on the same path

**A log line that describes an event that isn't happening.** Re-finding the primary while it is already connected logged `"found the saved primary … while a backup scale is connected — requesting switch-back"` — naming a backup that does not exist and a switch that does not occur, since `main.cpp` declines the request at `onWifiBackupAndIdle()`'s `decent-wifi` test. This subsystem is diagnosed from user-submitted logs, where that line reads as the app dropping the user's scale. Now a no-op.

**A browsed address outliving its request.** `m_browsedPrimaryIp` is set before a request `main.cpp` may decline, and a declined request consumes nothing, so a stale browsed address could outrank an IP a later probe had just verified. Cleared at the start of each probe, where the competing evidence is produced.

## Verification

- `primaryConnectClearsTheFlagEvenDuringAFallback()` — four-row truth table for the clearing rule, including the row that was wrong. New slot in an existing file.
- **Confirmed it can fail:** restoring the old rule (`return !wasWifiFallbackConnect`) turned it red — 109 passed / 1 failed. Restored: 110 passed, 0 failed, 0 skipped, no warnings.
- Its limit is stated in the test itself: it asserts the *rule*. Whether `connectedScaleIsWifiPrimary()` reports true on the browse path depends on live objects and is not reachable without linking `blemanager.cpp` into a test target — the expensive direction per TESTING.md. That half rides on the field log, as the surrounding browse machinery already does.
- Spec delta under `openspec/changes/clear-wifi-reconnect-flag-on-primary-connect/`, validated. Log-marker and test-duplication gates pass.

The corrected path is only reachable in practice on Android, so this rides to the next beta rather than a local sideload.


skialpine:
### Code review

No functional issues found. The deep bug scan cleared all four scrutiny points (flag clearing across every reachable ordering, the early return's skipped work, the `m_browsedPrimaryIp` staleness trade, and teardown windows). Every finding was a claim that overstated what the code does — most introduced by this branch — and all are fixed in the follow-up commit.

1. `connectedScaleIsWifiPrimary()` claimed the shared test meant "the two layers cannot disagree", but nothing coupled them — `main.cpp` used a bare `"decent-wifi"` literal, so changing the enum mapping would silently desync them. Both sides now read `ScaleTypeIds`, and the other ten hand-written copies went with them (CLAUDE.md: "Centralize anything produced at more than one site... When you touch code that already has copies, collapse them in that pass").

https://github.com/Kulitorum/Decenza/blob/2629c82c1e2d3a2b3b4d5e6f7a8b9c0d1e2f3a4b/src/ble/blemanager.h#L919-L926

2. The same comment called that test's limit "unreachable in practice". It is reachable twice, because the saved address is written later than the connection is reported: a manual "Add WiFi Scale" defers `setSavedScaleAddress()` to `recognizedAsHds` while `connected` is signalled at WebSocket open, and tapping a discovered WiFi scale over a dead `DecentScaleWifi` takes the type-unchanged re-wire branch, which returns without persisting. Both are inert at both call sites — that is the real argument, and it is now the one stated.

3. `probeWifiPrimaryReachable()`'s new comment listed "already on the primary" among `main.cpp`'s decline reasons, but this PR's own early return fires before `m_browsedPrimaryIp` is assigned, so that decline can no longer strand an address. A new comment describing behaviour its own commit removed is the failure that closed #1741 (CLAUDE.md: "a comment nobody re-derives is how a bad decision keeps getting re-approved").

4. The rewritten fallback comment kept `"lands about a second after the timeout"` — unsourced for this path, and the sibling copy of that exact defect was deleted a few hundred lines up in this same branch. Hedge restored too: the early return excludes a connected WiFi scale, not every other way a BLE scale could be connected.

5. `onScaleConnectedChanged()`'s "letting it finish is safe" comment still described the already-connected case as two-way; this branch made it three.

6. `m_wifiDirectAttemptFailed` still documented as "cleared the moment a scale connects" (wrong before this branch, wrong differently after); `m_browsedPrimaryIp` still as "consumed on use" when it is not guaranteed consumed, which is why it is cleared at two sites. The switch-back block comment still named only `main.cpp`'s idle poll as its driver and said "call only after a reachable probe", which the browse path does not do.

Plus predicate counts, one handler described as two, and an orphaned `kReconnectBrowseTimeoutMs` comment above an unrelated member.

Verification after the fixes: build clean, 110 passed / 0 failed / 0 skipped, no warnings. Log-marker and spec gates pass.

<details>

- [OK to skip] The tap-a-discovered-WiFi-scale path can connect a scale it never persists as primary (`main.cpp`'s type-unchanged re-wire branch returns without `setSavedScaleAddress`), so the user's choice is lost on next launch — pre-existing, outside this PR's blast radius, and it is a `main.cpp` fix rather than a `blemanager` one (score: 60)

</details>

PR #1747, opened 2026-08-02
Let users set a flow calibration they already know

skialpine:
Auto flow calibration learns the per-profile multiplier from shots, and the only MCP surface for it was `clear_flow_calibration`. A user who already knows the right number for a profile had no way to say so.

## New tools

| Tool | Category | What it does |
|---|---|---|
| `set_flow_calibration` | settings | Writes the per-profile multiplier by hand. Range **[0.5, 2.7]**, **refused rather than clamped**, with the range in the error text. Reaches a connected machine immediately — `perProfileFlowCalibrationChanged` is already wired to `applyFlowCalibration`, so no explicit apply call. |
| `get_flow_calibration` | read | Stored per-profile value, global fallback, which of the two is in effect (`effectiveSource`), the auto-cal switch, batch progress (`pendingAutoCalShots` of `autoCalBatchSize`), and one plain-language `state` sentence. |

## The trap this had to surface

With auto calibration **off**, `effectiveFlowCalibration()` uses the **global** multiplier and ignores every per-profile one. A per-profile write is therefore stored but inert. Rather than report a success the user cannot observe, the tool returns a `warning` naming the global value and both ways forward. Pinning an exact number still means auto off **plus** a global set — the tool description says so.

## Behaviour change in the store

`setProfileFlowCalibration()` now clears the profile's pending batch ideals. Those were computed against the old C, so folding them into the next batch median at a new C is wrong. The auto-cal path already cleared them a line earlier (no-op there); the manual and settings-import writers did not.

## Drift fixes in passing

- `[0.5, 2.7]` and the batch size `5` were each hand-typed at two sites. Now `SettingsCalibration::kProfileFlowCalMin/Max` and `kFlowCalBatchSize`.
- **`clear_flow_calibration` never declared `confirmed` in its schema** while being gated on it — a client validating arguments against the declared schema has no declared way to answer the gate. Audited the rest of the file: `machine_start_*` use the in-app dialog and `apply_theme`/`mqtt_*` are ungated, so this was the only gap.
- `MCP_SERVER.md`'s confirmation table was missing seven gated tools.

## Testing

`110 passed, 0 failed` locally. Three new slots in `tst_autoflowcal.cpp` (no new test file): batch cleared on set, out-of-bounds refused **without** eating the batch, per-profile value inert while auto cal is off.

Live-tested against a running build — both gates at confirmation level 2, bounds refused at 0.4 and 3.5, the auto-off `warning` path, and read-back after set and clear. Machine restored to its original state.

**Not covered:** the BLE/MMR write was never exercised (no DE1 connected during the live test), and the new test slots have not been watched to fail.


skialpine:
### Code review

Found 3 issues:

1. The three flow-calibration tools accept any string as `profileFilename` and never check the profile exists, so a wrong key writes a permanently dead calibration and reports success. `profiles_delete`, `profiles_rename` and `profiles_set_active` all guard with `ProfileManager::profileExists()`; these do not. `set_flow_calibration {profileFilename: "filter_3.json"}` returns `success: true` with a message saying it is "in effect", while `applyFlowCalibration()` keeps reading the real key and the machine never changes. The orphan entry then persists across restarts and survives settings export/import. `resolveFlowCalProfile()` is already the single choke point, which makes the fix one guard.

https://github.com/Kulitorum/Decenza/blob/d864e580116c0d43afaee0a50ff82221ae24d38b/src/mcp/mcptools_control.cpp#L24-L32

2. `profileFilename` is documented without the "(without .json extension)" qualifier that every profile-taking tool in `mcptools_profiles.cpp` uses (L186, L625, L712, L769). `baseProfileName()` returns the extension-less form, and the calibration map is keyed by the raw string with no normalization — so a caller passing `filter_3.json` creates a second key that the learner (writing `filter_3`) never touches. Combined with issue 1, this is the most likely way a wrong key actually gets written.

https://github.com/Kulitorum/Decenza/blob/d864e580116c0d43afaee0a50ff82221ae24d38b/src/mcp/mcptools_control.cpp#L402-L404

3. `maincontroller.cpp` still hand-types `kCalibrationMin = 0.5` rather than referencing the new `SettingsCalibration::kProfileFlowCalMin`, which is the exact duplication the new constant's own comment warns about: "a second hand-typed copy of the numbers is exactly how the two would drift apart from the one enforced below." This PR already edits the same file to replace the hardcoded batch size with `kFlowCalBatchSize`. (CLAUDE.md: "Centralize anything produced at more than one site" and "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'.") `kCalibrationMax` is correctly left alone, being firmware-dependent.

https://github.com/Kulitorum/Decenza/blob/d864e580116c0d43afaee0a50ff82221ae24d38b/src/controllers/maincontroller.cpp#L2699-L2710

Checked and clean: confirmation gating is server-side only with no handler-side `confirmed` check (the #1219 class), the new `clearFlowCalPendingIdeals()` call is a genuine no-op on the auto-cal path and cannot re-enter, refused writes preserve both the stored value and the batch, `effectiveSource`/`effectiveMultiplier` reproduce `effectiveFlowCalibration()`'s rule exactly, and all constant substitutions preserve values and comparison operators.

PR #1748, opened 2026-08-02
Pin libtailscale decenza-v1.94.1-5

skialpine:
Bumps `TSNET_TAG` and the three per-platform SHA-256 hashes in [cmake/tsnet.cmake](cmake/tsnet.cmake) to pick up [skialpine/libtailscale#2](https://github.com/skialpine/libtailscale/pull/2).

## macOS only — the fork crash

tsnet forked `lsof -n -a -u<uid> -c IPNExtension -F` on **every LocalAPI request**. `TsnetGetAuthURL`/`TsnetGetBackendState` are polled during connect, so it fired repeatedly.

It never did anything for us. That lookup finds the Mac App Store Tailscale GUI's random LocalAPI port and token; tsnet serves its own LocalAPI over an in-process `memnet` listener with no `RequiredPassword`, so the token is never checked, and with no GUI installed `lsof` matches nothing anyway.

And `fork()` in our ASan-instrumented debug builds crashed the child pre-exec:

```
EXC_BREAKPOINT (SIGKILL)
BUG IN CLIENT OF LIBPLATFORM: os_unfair_lock is corrupt, or owner thread exited without unlocking
crashed on child side of fork pre-exec
```

Same failure mode as the `dscl` exec that `ts_omit_ssh` fixed.

Verified by measurement, not by reading: a recording `lsof` shim first on `PATH`, over a full `tailscale_up()` cycle against the in-process test control server — **6 forks before, 0 after**. That measurement is also what caught a first attempt at the fix that removed exactly zero forks.

`safesocket_darwin.go` is the only file that installs this lookup, so iOS (`GOOS=ios`) and Android never had it.

## All platforms — Funnel errors stop being silent

`TsnetEnableFunnelToLocalhostPlaintextHttp1` discarded `SetServeConfig`'s error and instead checked `sc.AllowFunnel[hp]` — its own map literal, set to `true` a few lines earlier and never mutated by `SetServeConfig`. The branch was dead and the function returned success unconditionally, so a Funnel refused by tailnet policy looked like a working one.

[`McpTunnelTsnet`](src/mcp/mcptunnel_tsnet.cpp#L185) already handles non-zero correctly — warns and retries on the next poll — so the visible effect is an `enable funnel failed:` line that was previously invisible. No regression risk on Android/iOS.

## Hashes

Taken from the release `manifest.json` and independently verified against the downloaded artifacts:

| artifact | sha256 |
|---|---|
| macos | `9e6b42c8…afc3582` |
| ios | `706cbbd2…bb0b739` |
| android | `80331572…5cbdb3e0` |

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -65,13 +65,21 @@ Read [`docs/SHOT_REVIEW.md`](https://github.com/Kulitorum/Decenza/blob/main/docs
 
 The `cmake`/`ctest` invocations in `docs/CLAUDE_MD/TESTING.md` and `docs/CLAUDE_MD/PLATFORM_BUILD.md` are **reference for humans and CI**, not instructions for an assistant. Run their equivalent through the MCP.
 
-**No CI job builds or tests a pull request — run the full suite locally before opening one.** That is the gate, via `mcp__qtcreator__run_tests` (scope `all`). Nothing on GitHub will catch a compile error or a failing test before merge.
+**No CI job builds or tests a pull request *automatically* — run the full suite locally before opening one.** That is the default gate, via `mcp__qtcreator__run_tests` (scope `all`). Nothing on GitHub fires on its own to catch a compile error or a failing test before merge.
+
+**But CI can be pointed at a branch on demand, and for the test suite that is `linux-release.yml`.** It configures `-DBUILD_TESTS=ON` and runs `ctest --output-on-failure --repeat until-pass:3`, then reports any test that only passed on retry. It is `workflow_dispatch`, so it runs only when dispatched:
+
+```bash
+gh workflow run linux-release.yml --ref <branch> -f upload_to_release=false
+```
+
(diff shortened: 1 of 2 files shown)


skialpine:
Added `51cefdf`: a guard so a stale build directory can't silently link the wrong libtailscale.

`TSNET_TAG` is a `CACHE STRING`, so bumping it in `cmake/tsnet.cmake` does nothing to an existing build dir — CMake keeps the cached value. Verifying this very PR locally initially "passed" while the cache still said `decenza-v1.94.1-4`, i.e. against the build *without* the fix. The same staleness applied to all four earlier tag bumps; nothing ever reported it.

Re-reading the file doesn't catch it either, because the three SHA-256s are plain variables while the tag is cached. A mismatched pair fails as `download of libtailscale-macos.zip failed` — which reads like a network problem, not a wrong pin.

The guard compares the cached tag against a non-cached `_TSNET_EXPECTED_TAG` and refuses with the exact fix:

```
CMake Error at cmake/tsnet.cmake:34 (message):
  tsnet: this source tree pins decenza-v1.94.1-5, but this build directory
  has TSNET_TAG=decenza-v1.94.1-4 cached.

  The hashes in cmake/tsnet.cmake belong to decenza-v1.94.1-5, so building
  against decenza-v1.94.1-4 would link the wrong libtailscale.

  Fix the build directory:

      cmake -DTSNET_TAG=decenza-v1.94.1-5 <builddir>
```

Verified both directions: a cache at `-4` against this tree refuses as above; a matching cache configures clean. The real project build dir reconfigures fine.

**CI is unaffected** — checked, not assumed: no workflow caches `CMakeCache.txt` or a build directory. The only `actions/cache` uses are OpenSSL (`macos-release.yml:135`) and Gradle (`android-release.yml:129`), so CI always configures fresh and takes the file's value. This only ever fires locally.

PR #1749, opened 2026-08-02
Collect stable-key caches on superseded tags too

skialpine:
The "superseded release tags" step in [prune-caches.yml](.github/workflows/prune-caches.yml) filtered on a timestamped ccache/sccache key, so **stable-key entries on a shipped tag were collectable by nothing**: `qt-*`, `gradle-*`, `openssl-*`.

They're ref-scoped like every other entry, so no future tag can restore them. The timestamp filter existed to identify compiler caches and had been read as a safety boundary. The deleted-branch step doesn't reach them either — it deletes on a liveness signal, and a tag is always live.

Measured on v2.0.1: **2.4 GB across six entries** stranded per release, against a 10 GB cap.

| entry | size |
|---|---|
| `qt-macos-v2…` | 629 MB |
| `qt-windows-v3…` | 498 MB |
| `qt-android-v1…` | 475 MB |
| `qt-linux-arm64-v1…` | 397 MB |
| `gradle-android…` | 373 MB |
| `openssl-3.6.3-universal-macos` | 24 MB |

The filter is now "tag-scoped and superseded", whatever the key shape. The newest tag stays fully protected, stable keys included — those are what keep a re-tag's Qt install fast, which is the same reasoning that got tag-wide save-suppression reverted in `setup-compiler-cache/action.yml`.

## The part that isn't tidiness

`NEWEST` moves to the **same** widened set as the deletion list. Widening one without the other inverts the protection:

once a pass has collected a superseded tag's compiler caches, a later run can see only stable-key tag entries → `max_by` returns null → `NEWEST` becomes the string `"null"` → every real tag compares unequal → the loop deletes the **live release's** caches.

Reproduced against a fixture before fixing. The naive version:

```
### NEWEST the WRONG way (timestamped-only) ###
NEWEST=''
  WOULD DELETE qt-macos-v2      v2.0.1 (LIVE RELEASE)
  WOULD DELETE gradle-android   v2.0.1 (LIVE RELEASE)

### NEWEST the right way (same set) + guard ###
NEWEST='v2.0.1'
  nothing deleted — live release protected
```

A guard now refuses when `NEWEST` doesn't name a tag, rather than treating everything as superseded.

## Verification

Fixture with v2.0.0 + v2.0.1 + main:

- **deleted:** v2.0.0's `ccache-macos` (1870 MB), `qt-macos` (629 MB), `openssl` (24 MB) — the two stable keys are the new behaviour
- **kept:** all three v2.0.1 entries including stable keys
- **never selected:** both `main` entries

YAML parses; all five `run:` blocks pass `bash -n`.

### DIFF
--- a/.github/workflows/prune-caches.yml
+++ b/.github/workflows/prune-caches.yml
@@ -134,17 +134,31 @@ jobs:
       # would make the next re-tag rebuild from scratch. Ordering is by build
       # time for the same reason — see the note on NEWEST below.
       #
-      # Scoped tightly, since the surrounding steps are conservative about tags
-      # for good reason: only timestamped ccache/sccache keys, only on tag refs,
-      # only for tags that sort older than the newest one present. Stable-key
-      # entries (qt-*, openssl-*, gradle-*) are never touched.
-      - name: Drop compiler caches for superseded release tags
+      # Scoped to tag refs, and to tags older than the newest one present.
+      #
+      # This used to also require a timestamped ccache/sccache key, which left
+      # the STABLE-key entries on a shipped tag — qt-*, gradle-*, openssl-* —
+      # uncollectable by anything. They are scoped to that tag's ref like every
+      # other entry, so no future tag can read them either; the timestamp filter
+      # was about identifying compiler caches, and got read as a safety
+      # boundary. Measured on v2.0.1: 2.4 GB across six stable-key entries
+      # (qt-macos 629 MB, qt-windows 498, qt-android 475, qt-linux-arm64 397,
+      # gradle 373, openssl 24) that no step in this file would ever have
+      # touched. Nor would the deleted-branch step below — it deletes on a
+      # liveness signal, and a tag is always live.
+      #
+      # So the filter is now "tag-scoped and superseded", whatever the key
+      # shape. The newest tag is still fully protected, stable keys included:
+      # those are what make a re-tag's Qt install fast.
+      - name: Drop caches for superseded release tags
         if: ${{ steps.gate.outputs.skip == 'false' }}
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
           set -uo pipefail
-          TS='-[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9:.]+Z$'
+          # No TS regex here any more: this step no longer distinguishes
+          # compiler caches from stable-key ones. The other steps still need
+          # theirs and define it locally.
 
           # Tag-scoped entries surface as refs/heads/refs/tags/X (the odd shape
           # the deleted-branch step below skips) as well as plain refs/tags/X.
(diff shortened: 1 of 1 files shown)


skialpine:
## Tested before merge

The step's `run:` block is extracted verbatim from the YAML on this branch (99 lines) and executed against a stubbed `gh` that serves a fixture for `cache list` and records ids for `cache delete`. So this exercises the shipped script, not a paraphrase of it — my earlier fixture check did paraphrase it, which is why it missed nothing but proved less than it looked.

`gh cache list --json id,key,ref,sizeInBytes,createdAt` field names verified against the real `gh` binary, so the fixtures match production shape.

### Cases

```
live-store-no-tags         rc=0 deleted=[]        want=[]        PASS
two-tags-supersede         rc=0 deleted=[4,5,6]   want=[4,5,6]   PASS
stable-only-live-tag       rc=0 deleted=[]        want=[]        PASS
plain-refs-tags-shape      rc=0 deleted=[9]       want=[9]       PASS
hotfix-after-higher        rc=0 deleted=[11]      want=[11]      PASS
empty-store                rc=0 deleted=[]        want=[]        PASS
```

`live-store-no-tags` is the repo's actual current cache list. `plain-refs-tags-shape` covers `refs/tags/X` without the `refs/heads/` quirk prefix. `hotfix-after-higher` is v1.9.5 built *after* v2.0.0 — `createdAt` correctly keeps v1.9.5 where `sort -V` would have collected it.

Step summary produced for the two-tag case:

```
- deleted `ccache-macos-universal-2026-07-24T10:00:00.000Z` (1870 MB) — tag v2.0.0 superseded by v2.0.1
- deleted `qt-macos-v2-mac-desktop-clang_64-6.11.1` (629 MB) — tag v2.0.0 superseded by v2.0.1
- deleted `openssl-3.6.3-universal-macos` (24 MB) — tag v2.0.0 superseded by v2.0.1
Reclaimed 2523 MB from 3/3 superseded entries.
```

### Negative controls

Six green cases prove nothing on their own, so both directions were falsified.

**A — is the harness sensitive to the change?** Same fixture, old step extracted from `main`:

```
OLD (main)      deleted=[4]        # ccache only
NEW (this PR)   deleted=[4,5,6]    # + qt, openssl
```

**B — is the same-set `NEWEST` + guard load-bearing, or decoration?** Built the naive variant (widen `CACHES`, leave `NEWEST` on timestamped keys, drop the guard) and ran the stable-only fixture, where both entries belong to the *live* release:

```
NAIVE widening  deleted=[2,3]      # deletes the LIVE release's caches
NEW (this PR)   deleted=[]
```

That is the failure the guard exists for, reproduced against the real script.

### What this does not cover

- The gate step, the `workflow_run` wiring, and real `gh cache delete` calls are stubbed out.
- No superseded tag currently holds caches, so the first production run will find nothing to collect until v2.0.2 ships. These fixtures are the evidence until then.

PR #1757, opened 2026-08-03
Add a 1x/2x/3x flow scale and a graph options menu

skialpine:
## Why

Flow is drawn against the same 0–12 axis as pressure, so the part of the curve a user actually dials on sits crushed against the floor of the chart.

Measured over this machine's own history — **1,068 espresso shots with flow curves**, straight out of `shots.db`:

| | mL/s |
|---|---|
| pour sample, median | 1.81 |
| pour sample, p99 | 3.98 |
| per-shot pour peak, p90 | 3.42 |
| per-shot pour peak, p99 | 5.47 |

The median pour therefore occupies **15% of chart height**. What *does* reach the top is the puck-fill spike — peak flow p50 **8.01 mL/s at t≈2.5 s**, over within ~4 s, and not something anyone dials on. The axis is scaled for the one part of the curve that carries no information.

Every other DE1 app has solved this differently: de1app's Insight skin gives flow its own chart with a 0–8 axis, DSx2 pre-multiplies flow by 2 and labels a second axis at half scale, Decaid leaves flow on a fixed 0–11. DSx2's approach is the one that fits a single-chart layout, which is ours.

## What changed

- **Flow scale of 1x / 2x / 3x**, applied to flow, weight flow rate and the dashed flow goal *together* — following de1app, which multiplies `espresso_flow_2x`, `espresso_flow_weight_2x` and `espresso_flow_goal_2x` by the same factor.
- **Default 2x.** It raises the median pour to 45% of chart height while its 6.0 mL/s ceiling still clears the p99 pour peak of 5.47, so no pour detail is clipped at the default. 3x is there for those who want the resolution and accept clipping above 4.0.
- **Flow as a third right-axis mode**, joining weight and temperature, so a multiplied trace stays numerically readable. Mode and multiplier are deliberately **independent**.
- **The left axis title drops to `bar`** whenever a multiplier is active — the only marker of it that survives into a screenshot, which matters when graphs arrive in bug reports.
- **The Advanced Curves button becomes a menu opener** on shot detail, post-shot review and shot comparison; advanced curves moves inside beside the flow scale. The live screen keeps its existing extraction view selector and gains the flow scale there.

## The bit that would have shipped broken

The multiplier is applied to the mapped **range**, never to the data — the live graph shrinks its renderer range, the comparison graph routes flow through its own axis key, the history graph scales at append time from source arrays it leaves untouched.

That is not a style choice. `pressureAxisMax` sizes the history axis by walking the flow arrays: feed it multiplied values and the axis grows by the multiplier, the traces land in **identical pixels**, and the zoom silently does nothing while every test still passes. Keeping the source arrays pristine also makes every readout true by construction — Decaid ships the opposite (its weight trace is plotted at ÷10 and the tooltip prints the plotted number, so a 36 g shot reads "Weight: 3.6 g", `shot_chart.dart:105`,`:359`).

## Included refactor: `SettingsGraph`

The 13 `graph/*` keys had no domain sub-object and were read through `Settings.boolValue()` — a plain `Q_INVOKABLE`, so a property bound to it records **no dependency**. Every graph was reading a one-shot snapshot taken at construction, and three separate compensations had grown on top: `GraphLegend` wrote the setting *and* poked the graph's property, `HistoryShotGraph` hand-maintained a 12-case string switch, `LastShotChartSource` hand-maintained a key list that had already shipped one entry short.

This change would have added two more keys to that scheme. Instead the keys move into a `SettingsGraph` domain with real `Q_PROPERTY`/`NOTIFY`, all three compensations delete, and the omission bug becomes unrepresentable.

Deduplication that fell out of it — all of it predating this feature:

| collapsed | into |
|---|---|
| 59 mirror declarations across 5 files | direct bindings |
| 4 copies of the series list | `GraphSeries` singleton |
| 2 verbatim copies of the right-axis label column | `GraphRightAxisLabels` |
| 4 copies of the Advanced button | `GraphOptionsButton` |
| 3 copies of the checkbox card | `GraphOptionToggleCard` (`ExtractionViewSelector` 386 → 214 lines) |
| duplicated `_niceTimeAxisStep` / `_timeAtPixel` | `GraphUtils.js` |
| `shotReview/advancedMode` mirrored in 4 pages | one property |

**Net −449 lines** in existing production files.

## Settings migration

`graph/showWeightAxis` (bool) → `graph/rightAxisMode` (string enum). Resolution is read-time only and never writes back, so a device that never opens a graph is never migrated and its old value survives a rollback. Every storage key is preserved.

## Verification

- Build clean, **110/110 tests**, no warnings, **qmllint 229/229** (7 new files, tree stays at zero).
- Resolver tests are data-driven over the fresh-install, out-of-range, non-numeric, INI-string and migration cases.
- The `GraphSeries` completeness test was retargeted and **positive-controlled** — removed an entry, watched it go red with the right message, restored.
- **Walked in the running app at tablet size**, shot detail and shot comparison: menu, all three multipliers, the axis title rule, the three-way right-axis cycle, flow labels reading `0.0–7.5` (= autoscaled max 15 ÷ 2), and inspect reporting `2.4 mL/s` / `1.5 g/s` while plotting the scaled values. Fresh-install default confirmed 2x with the keys actually deleted.

That pass caught a regression the compiler, qmllint and the suite all missed: `GraphInspectBar` read the deleted mirror properties off a `var`-typed `graph`, blanking every readout. Fixed here.

## Not done

- Post-shot review and the live espresso screen weren't walked (same components as the two that were; live needs an actual shot).
- Wiki manual page still to write.
- `openspec archive` still to run as the last commit before merge.


skialpine:
### Code review

Found 6 issues. All are fixed in 1aefb033, pushed to this branch.

1. Live graph's flow renderers are unclipped, so at the 2x default the flow trace paints outside the plot area (bug due to `FastLineRenderer` mapping `y` with no clamp — `px[i].y = h - (y - m_minY) * scaleY` in `src/rendering/fastlinerenderer.cpp` — and adding no clip node). At 1x the ceiling was 12 and pump flow never reached it; at 2x it is 6.0 mL/s, which the puck-fill spike clears on 93% of the shots this PR's own statistics are drawn from. The dashed flow goal beside it already clips (`DashedLineSeries.qml`), so goal and trace would visibly disagree exactly where the trace escapes. This is on the live espresso screen, the one surface the manual verification pass never opened.

https://github.com/Kulitorum/Decenza/blob/55f0555cddd8a9641101646a5a56e43fe32bc895/qml/components/ShotGraph.qml#L277-L279

2. `ComparisonDataTable` still hand-typed all eleven series, so the change's own spec requirement is false — `specs/graph-settings-domain/spec.md` says the series set "SHALL be defined in exactly one place" and that adding one propagates "without any of them being edited". Adding a curve to `GraphSeries` would not have reached the comparison table, and nothing guarded the drift: the completeness test in `tst_backgroundpresets.cpp` only diffs `HistoryShotGraph` against `GraphSeries`.

https://github.com/Kulitorum/Decenza/blob/55f0555cddd8a9641101646a5a56e43fe32bc895/qml/components/ComparisonDataTable.qml#L76-L79

3. `writeBool`'s no-op guard compares against `!value` instead of the property's real default, so while a key is unset, writing a property its own default (`Settings.graph.advancedMode = false` on a fresh install) stores the key and emits a change nothing made — waking every graph, the legend, the inspect bar and the last-shot chart's cache key.

https://github.com/Kulitorum/Decenza/blob/55f0555cddd8a9641101646a5a56e43fe32bc895/src/core/settings_graph.cpp#L36-L38

4. A comment claims `tst_historyshotgraph` pins the auto-range invariant. No such test exists anywhere in the repo, and this change's own `tasks.md` says the opposite: 8.2 is unticked because "these need a QML-level harness, which this suite does not have". It guards precisely the failure mode the design exists to prevent — CLAUDE.md says "An un-sourced claim in a comment gets believed and then licenses wrong code".

https://github.com/Kulitorum/Decenza/blob/55f0555cddd8a9641101646a5a56e43fe32bc895/qml/components/HistoryShotGraph.qml#L307-L309

5. Three pages keep an empty `Connections { target: Settings }` shell after its `advancedMode` body was deleted, each still under a comment claiming it "picks up toggle changes made on any other page sharing this setting". Inert code asserting a mechanism that no longer exists. Same in `PostShotReviewPage.qml` and `ShotComparisonPage.qml`.

https://github.com/Kulitorum/Decenza/blob/55f0555cddd8a9641101646a5a56e43fe32bc895/qml/pages/ShotDetailPage.qml#L73-L79

6. `GraphOptionsDialog` reads `parent.width` unguarded while parented to `Overlay.overlay`, which is null until the declaring item is in a window. `SelectionDialog.qml`, `EquipmentInfoDialog.qml` and `ChangeBeansDialog.qml` all guard theirs.

https://github.com/Kulitorum/Decenza/blob/55f0555cddd8a9641101646a5a56e43fe32bc895/qml/components/GraphOptionsDialog.qml#L19-L21

<details>

- [Worth a look] `ComparisonGraph`'s flowAxis comment claimed flow keeps its proportion of plot height under the 12→20 advanced-curve axis expansion; it actually shrinks 40% — one-line comment fix in code already being changed (score: 60)
- [Worth a look] `GraphUtils.paddedAxisEnd` was dead while all three graphs hand-rolled the same padding, which was the helper's stated purpose (score: 55)

</details>

Also checked and clean: no surviving reader of any deleted mirror property; each graph's right-axis labels agree with what that graph plots; no multiplier leaks into any readout; the settings resolvers handle unset/out-of-range/non-numeric/INI-string/wrong-case; `cycleRightAxisMode` advances from the resolved value so the legacy choice survives the first tap; all eleven `DECENZA_GRAPH_BOOL` expansions carry the right key, default and signal; no delegate role-injection breakage from `pragma ComponentBehavior: Bound`; the de1app (`gui.tcl:3426`, `:3451`, `:3551`) and Decaid (`shot_chart.dart:105`, `:359`) citations are byte-exact.

Still open on this branch: the wiki manual page (task 10.1), and `openspec archive` as the last commit before merge.

skialpine:
### Added: fix for an unrelated shutdown crash (abb53ca9)

An intermittent ASan abort on exit was reported against a build of this branch. It is **not caused by the flow scale** — nothing here touches `ShotServer`, `WebDebugLogger` or teardown order — but it is fixed on this branch rather than deferred.

**Mechanism.** `WebDebugLogger` is a `QML_SINGLETON` whose `create()` returned a raw `new`-ed object with no parent, so the QML engine took ownership and destroyed it at engine teardown. `s_instance` was never cleared and the message handler never uninstalled, so any `qDebug()` after that read freed memory:

`main.cpp:4833` → `~MainController` → `deleteChildren` → `~ShotServer` → `ShotServer::stop()` logs → `~QDebug` → `messageHandler` → `s_instance->handleMessage()` → `QMutexLocker` on freed `m_mutex` → `__asan_report_load8`.

**Why intermittent:** it needs a log line to land in the window between the engine deleting the logger and process exit, and whether `ShotServer` logged at all varied with how teardown interleaved. ASan is what makes it fatal — in release it is a silent read of freed memory, which is likely why it went unnoticed.

**Fix:** logging outlives QML by definition, so the singleton is marked `CppOwnership` and the engine no longer deletes it. The window stops existing rather than being narrowed. The class also gains a destructor that clears `s_instance`, since leaving a static pointing at freed memory is the bug itself and the testing constructor lets a test own one by value. The handler is deliberately not uninstalled — `btlogfilter` installed over us and it is not ours to restore, and `messageHandler()` already null-checks `s_instance` and forwards to the previous handler.

Build clean, 110/110 tests, six launch/quit cycles with no crash report.

PR #1758, opened 2026-08-03
fix: keep bottom bar action buttons on screen with a long profile name

skialpine:
## Problem

On Shot Review and Shot Detail, a long profile name pushes the bottom-bar action buttons off the right edge of the screen — reported with Undo + Visualizer + AI Advice visible and a profile named `Collin Espresso/6 bar simple bloom 16g Co…`.

## Cause

A child of a Qt Quick Layout that is not itself a layout and has no explicit `Layout.fillWidth` gets `QLayoutPolicy::Fixed` and cannot shrink below its preferred width; nested layouts get `Preferred` (`qtdeclarative/src/quicklayouts/qquicklayout.cpp:1304-1332`). So the action buttons are Fixed — they are never squeezed. The profile name was capped only at `page.width * 0.3`, which is itself a Fixed contribution, and once title + subtitle + buttons exceed the bar the row's minimum sum can't be met and the whole row is laid out past the right edge.

## Fix

`BottomBar` publishes the width actually left over, and the subtitle is capped by it so it elides instead of overflowing the row:

```qml
readonly property real leftContentMaxWidth: Math.max(Theme.scaled(48),
    width - titleRow.implicitWidth - contentRow.implicitWidth
          - (rightTextLabel.visible ? Math.min(rightTextLabel.implicitWidth, width * 0.4) : 0)
          - Theme.chartMarginSmall - Theme.spacingLarge
          - Theme.spacingMedium * (rightTextLabel.visible ? 4 : 3))
```

Details worth knowing:

- **Floored at `Theme.scaled(48)`, not 0.** `Layout.maximumWidth: 0` is a legal zero-width cell — the text would vanish outright, no ellipsis, and the `Accessible.name` on the collapsed item with it.
- **`titleRow` needs no cap of its own** but `rightTextLabel` does: a nested layout's `implicitWidth` is its engine-preferred size, which already clamps each child to that child's `Layout.maximumWidth` (`qquicklayout.cpp:1279`, `boundSize`). `rightTextLabel` is a plain `Text`, whose raw `implicitWidth` is unclamped.
- **Gap count varies with `rightTextLabel.visible`** — invisible children are dropped from the layout engine entirely (`qquicklayout.cpp:883-890`, `shouldIgnoreItem`), so the row has 4 gaps with `rightText` set and 3 without.
- **Shot Review's upload status labels are now capped.** They are `fillWidth` + `WordWrap`, so the layout can shrink them, but `contentRow` still reports their *uncapped* implicit width as preferred — which starved the budget to its floor whenever an upload error was on screen.

## Also in this PR

- **New `qml/components/BottomBarSubtitle.qml`.** The profile-name/date block was duplicated verbatim across both pages; each is now a six-line usage. It takes the bar's `contentColor` rather than `Theme.textColor` — under a background preset `textColor` derives from the page while `barColor` is a different surface, the divergence `BottomBar.contentColor` exists to absorb — and adds `activeFocusOnTab` alongside `Accessible.focusable`, which neither page had.

## Testing

- Build clean; full suite 110 passed / 0 failed locally via Qt Creator.
- Manual check on Shot Review / Shot Detail with a long profile name. No QML test harness in this repo, so the layout behaviour itself is not covered by an automated test.
- `text-invariants.yml` ran on this PR and passed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1352,6 +1352,7 @@ set(QML_FILES
     qml/components/ProfilePreviewPopup.qml
     qml/components/ProfileInfoButton.qml
     qml/components/BottomBar.qml
+    qml/components/BottomBarSubtitle.qml
     qml/components/ThemedIcon.qml
     qml/components/TranslationErrorToast.qml
     qml/components/AccessibleMouseArea.qml
--- a/qml/components/BottomBar.qml
+++ b/qml/components/BottomBar.qml
@@ -23,16 +23,36 @@ Item {
     readonly property color contentColor: Theme.contentColorOn(barColor, Theme.iconColor)
 
     // How much width the title-side content may take before it starts pushing the
-    // action buttons off the end of the bar. Nothing in the row has a minimum width,
-    // so an oversized leftContent does NOT get shrunk by the layout — the buttons are
-    // squeezed to near-zero cells instead and their centred labels paint outside the
-    // bar, which is what a long profile name did on Shot Review / Shot Detail.
-    // Pages bind their leftContent's Layout.maximumWidth to this so the text elides.
-    // No binding loop: nothing here depends on leftContentRow's own width.
-    readonly property real leftContentMaxWidth: Math.max(0,
+    // action buttons off the end of the bar.
+    //
+    // A child of a Qt Quick Layout that is not itself a layout and has no explicit
+    // Layout.fillWidth gets QLayoutPolicy::Fixed — it cannot shrink below its own
+    // preferred width; only nested layouts get Preferred
+    // (qtdeclarative/src/quicklayouts/qquicklayout.cpp:1304-1330, effectiveSizePolicy_helper).
+    // The action buttons are therefore Fixed, so an oversized leftContent does not
+    // squeeze them: the row's minimum sum simply exceeds the bar and the whole row
+    // is laid out past the right edge, which is what a long profile name did on
+    // Shot Review / Shot Detail. Capping leftContent lowers that minimum sum, so
+    // pages bind their leftContent's Layout.maximumWidth to this and the text elides.
+    //
+    // Floored at a couple of characters rather than at 0: Layout.maximumWidth: 0 is a
+    // legal zero-width cell, so the text would vanish outright — no ellipsis, and the
+    // Accessible.name on the collapsed item goes with it. The subtraction can reach
+    // that point because contentRow.implicitWidth counts a fillWidth/WordWrap label
+    // (e.g. the upload-error text on Shot Review) at its full unwrapped width even
+    // though the layout can shrink it. Overflowing by this floor beats disappearing.
+    //
+    // No binding loop: nothing here depends on leftContentRow's own width. A nested
+    // layout's implicitWidth is its engine-preferred size, which already clamps each
+    // child to that child's Layout.maximumWidth (qquicklayout.cpp:1278, boundSize),
+    // so titleRow's term needs no cap of its own — rightTextLabel is a plain Text, so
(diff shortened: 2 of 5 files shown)


skialpine:
### Code review

Found 4 issues:

1. `leftContentMaxWidth` floors at exactly 0, so the profile name and date render nothing — not an ellipsis — instead of eliding. `Layout.maximumWidth: 0` is a legal zero-width cell, and the `ColumnLayout` carrying `Accessible.name` collapses with it, so screen readers lose the subtitle too. Reached whenever `contentRow.implicitWidth` is large: PostShotReviewPage's upload-error `Text` is `Layout.fillWidth` + `WordWrap` inside the content slot, so its full unwrapped width lands in that term the moment an upload fails — the profile name disappears exactly when the user needs to know which shot failed. Also reachable at larger `Theme.scaleMultiplier` and in verbose locales. The old `page.width * 0.3` could never reach 0.

https://github.com/Kulitorum/Decenza/blob/c331966a4b1b8d1ac3d29dda4994547cc529ba2d/qml/components/BottomBar.qml#L31-L36

2. The comment states the layout mechanism backwards, uncited. It claims "Nothing in the row has a minimum width, so an oversized leftContent does NOT get shrunk by the layout — the buttons are squeezed to near-zero cells instead". Per `qtdeclarative/src/quicklayouts/qquicklayout.cpp:1304-1330` (`effectiveSizePolicy_helper`), a child that is not a layout and has no explicit `Layout.fillWidth` gets `QLayoutPolicy::Fixed` and cannot shrink below its implicit width; nested layouts get `Preferred`. So the buttons are not squeezed — the row's minimum sum exceeds the bar and the whole row overflows past the right edge. CLAUDE.md: "Read it instead of guessing at Qt behaviour, and cite file-and-line in any comment that asserts what Qt does" and "An un-sourced claim in a comment gets believed and then licenses wrong code".

https://github.com/Kulitorum/Decenza/blob/c331966a4b1b8d1ac3d29dda4994547cc529ba2d/qml/components/BottomBar.qml#L24-L32

3. The spacing term is hard-coded to `Theme.spacingMedium * 3`, which is only right while `rightText` is empty. The outer `RowLayout` has 4 visible children with `rightText === ""` (3 gaps) and 5 with it set (4 gaps) — invisible children are dropped from the engine entirely (`qquicklayout.cpp:883-890`). So in exactly the configuration the conditional `rightTextLabel` term was written to serve, the property over-reports free width by one `spacingMedium`. Dormant today (neither `leftContent` user sets `rightText`), but this is shared component API and several pages already set `rightText`.

https://github.com/Kulitorum/Decenza/blob/c331966a4b1b8d1ac3d29dda4994547cc529ba2d/qml/components/BottomBar.qml#L32-L36

4. The duplicated `leftContent` subtitle block was edited in both copies rather than collapsed (CLAUDE.md says "When you touch code that already has copies, collapse them in that pass."). The profile-name/date `ColumnLayout` is near-verbatim identical across the two pages, including the four `Layout.maximumWidth` bindings this PR rewrites — so the off-by-one and floor fixes above each have to be applied twice.

https://github.com/Kulitorum/Decenza/blob/c331966a4b1b8d1ac3d29dda4994547cc529ba2d/qml/pages/ShotDetailPage.qml#L1597-L1623

https://github.com/Kulitorum/Decenza/blob/c331966a4b1b8d1ac3d29dda4994547cc529ba2d/qml/pages/PostShotReviewPage.qml#L2340-L2366

---

<details>

- [Worth a look] `titleRow.implicitWidth` is subtracted uncapped although the title `Text` carries its own `Layout.maximumWidth: root.width * 0.5`, while the `rightText` term in the same expression does mirror its cap — inconsistent, and over-restrictive for a long localized title. Errs safe (never over-reports). (score: 68)

</details>

Issue #1759, opened 2026-08-03
[Bug] USB Charging appears to not work on iPad


ItsGoodCoffee:
### What happened?

Issue is that I have a USB-A to USB-C cable (tested as able to charge my iPad when connected to a standard power source) but when I connect via the USB port on the front of the Decent, and regardless of what charging setting I have selected in settings, I still get a NOT CHARGING banner where the battery icon is on my iPad.

Running the latest stable Decent firmware, latest iPad OS, Decenza 2.0 from the App Store. Will resubmit through GitHub if account creation is the only barrier, thanks

[debug.zip](https://github.com/user-attachments/files/30676918/debug.zip)

### Steps to reproduce

1. Connect iPad to Decent machine with USB-A - USB-C cable
2. Open Decenza
3. Ipad shows NOT CHARGING by battery icon
4. Checked that battery charging was enabled in settings - cycled through all options with same result of NOT CHARGING

### Decenza version

2.0 from iOS App Store

### Platform

iOS

### Device / OS version

iPadOS 26.6

ItsGoodCoffee:
Feedback suggests this is a power supply issue from the Decent and not a Decenza issue so closing the ticket

PR #1760, opened 2026-08-03
fix: stop aboutToQuit from re-entering itself thousands of times on quit

skialpine:
## What

The `aboutToQuit` shutdown handler in `src/main.cpp` can re-enter itself, unbounded. Add a re-entrancy latch at the top of it.

## Why

The handler pumps events three times — the BLE queue drain wait (`waitLoop.exec()`), `mainController.drainDbWork()`, and the shot-export wait. Each of those can deliver a second `exit()`/`quit()`, which re-enters the handler from the top, which starts another nested loop.

Qt's own once-only guard does not stop this. `QCoreApplication::exit()` emits the signal and only **then** sets the flag that suppresses a second emission (`qtbase/src/corelib/kernel/qcoreapplication.cpp:1520-1522`, Qt 6.11.1):

```cpp
if (!d->aboutToQuitEmitted) {
    emit self->aboutToQuit(QCoreApplication::QPrivateSignal());
    d->aboutToQuitEmitted = true;   // after the emit returns
}
```

The emit is synchronous, so for the entire lifetime of the handler the flag is still `false`.

## Evidence

Not hypothetical. The iOS debug log attached to #1759 carries **3064** recursions, ~1 ms apart:

```
[1915.609] DEBUG Sending DE1 to sleep on app exit
[1915.610] DEBUG Waiting for BLE queue to drain before exit...
[1915.611] DEBUG Application exiting - shutting down devices
[1915.612] DEBUG Sending DE1 to sleep on app exit
[1915.612] DEBUG Waiting for BLE queue to drain before exit...
[1915.613] DEBUG Application exiting - shutting down devices
```

Each level re-sends the DE1 sleep command, so the machine was told to sleep roughly three thousand times on a single quit, on a stack thousands of frames deep.

Note the existing comment at `drainDbWork()` already knew that a second `exit()` delivered inside a nested loop exits *every* loop in `data->eventLoops`; what it missed is that the same call also **re-emits `aboutToQuit`**.

## Approach

A plain latch — what Qt's flag would be if it were set before the emit instead of after.

Kept the nested loops rather than removing them: each is bounded and each exists for a stated reason (finish the BLE sleep/charger writes; finish the database writes) already commented at the call sites. Removing them is a much larger change and would give up those guarantees.

If a nested loop still receives a second `exit()`, behaviour is unchanged from today: `quitNow` is set and every loop exits, so the outer handler resumes at the same place it would have on the existing timeout path.

## Testing

- Full suite via Qt Creator: **110 passed, 0 failed, 0 skipped**.
- No new test. `main.cpp`'s `aboutToQuit` handler has no harness, and reaching the recursive branch requires injecting a second `exit()` from inside a nested event loop — the fault-injection stop sign in `TESTING.md`.

## Not in scope

Issue #1759 itself (iPad reports NOT CHARGING) is **not** a Decenza defect and is not addressed here. That log shows the app commanding the port ON correctly every 60 s while the battery fell 65% → 60% with iOS reporting `CHARGING` — the DE1 USB port cannot sustain that iPad. The recursion above was found while reading the same log.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4856,19 +4856,20 @@ int main(int argc, char *argv[])
         // alone therefore made the DB row current while the exported JSON stayed
         // stale, silently. Same bounded shape as the drain above.
         //
-        // Skipped after a re-entrant quit for a different reason than the two above:
-        // processEvents() ignores quitNow, so this loop would still spin — but on
-        // iOS the dispatcher returns false without dispatching anything once
-        // termination is announced (qioseventdispatcher.mm:522-529), so it would
-        // burn the full 750 ms doing no work, inside the OS's termination grace
-        // window, and warn as if it had waited for something.
+        // NOT skipped on a re-entrant quit, unlike the two waits above, and the
+        // asymmetry is deliberate. Those use QEventLoop::exec(), which quitNow
+        // disables outright; this uses processEvents(), which ignores quitNow
+        // entirely (qcoreapplication.cpp:1393-1402). So off iOS this loop still
+        // does real work after a second quit, and skipping it would risk the stale
+        // exported JSON it exists to prevent. On iOS after termination is announced
+        // it spins without dispatching, wasting up to 750 ms of a process the OS is
+        // about to kill — the cheaper mistake of the two, so it is the one taken.
         {
             QElapsedTimer waited;
             waited.start();
-            while (!shutdownReentered && !shotHistoryExporter.isExportWorkIdle()
-                   && waited.elapsed() < 750)
+            while (!shotHistoryExporter.isExportWorkIdle() && waited.elapsed() < 750)
                 QCoreApplication::processEvents(QEventLoop::ExcludeUserInputEvents, 20);
-            if (!shutdownReentered && !shotHistoryExporter.isExportWorkIdle())
+            if (!shotHistoryExporter.isExportWorkIdle())
                 STORAGE_WARN_STDERR("Export", QStringLiteral(
                     "shot export threads still running at exit - the exported JSON for a "
                     "just-edited shot may be stale"));
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4606,6 +4606,40 @@ int main(int argc, char *argv[])
 
     // Cleanup on exit
     QObject::connect(&app, &QCoreApplication::aboutToQuit, [&accessibilityManager, &batteryManager, &de1Device, &de1ReconnectTimer, &physicalScale, &engine, &weightThread, &relayClient, &machineStatusSnapshot, &mainController, &scaleReconnectTimer, &shotHistoryExporter]() {
+        // Qt's own once-only guard on aboutToQuit does not hold for this handler,
+        // because this handler pumps events.
+        //
+        // QCoreApplication::exit() emits aboutToQuit and only THEN sets the flag
+        // that suppresses a second emission (qcoreapplication.cpp:1520-1522 in
+        // 6.11.1):
+        //
+        //     if (!d->aboutToQuitEmitted) {
+        //         emit self->aboutToQuit(QCoreApplication::QPrivateSignal());
+        //         d->aboutToQuitEmitted = true;   // <-- after the emit returns
+        //     }
+        //
+        // The emit is synchronous, so for the entire lifetime of this handler the
+        // flag is still false. Every nested loop below — the BLE drain wait, the
+        // database drain, the export wait — can therefore deliver a second
+        // exit()/quit(), which re-enters this handler from the top, which starts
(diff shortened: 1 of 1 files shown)


skialpine:
## Review pass — three agents, findings applied in 49e5a86d

### Correction to this PR's own description

**"3064 recursions on a single quit" is wrong.** That is a `grep -c` over a file containing seven `SESSION START` markers. Re-derived per session:

| session | count |
|---|---|
| 2026-08-03T09:53:43 (incl. trimmed head fragment) | 937 |
| 2026-08-03T09:17:19 | 534 |
| 2026-08-03T08:50:38 | 534 |
| 2026-08-01T19:00:07 | 534 |
| 2026-08-02T14:20:59 | 525 |

Six independent quit episodes over three days. **Worst single quit is 534**, at ~2 ms per cycle (the ~1 ms figure was the gap between *log lines*, three per cycle). "A stack thousands of frames deep" was likewise ~534. The comment in the code now states both numbers and which is which.

### The re-entry is unconditional on iOS, not a race

`qtbase/src/plugins/platforms/ios/qioseventdispatcher.mm:522-529`:

```cpp
if (applicationAboutToTerminate) {
    // Re-issue exit, and return immediately
    qApp->exit(kApplicationWillTerminateExitCode);
    return false;
}
```

Set at `:445` in `applicationWillTerminate`, **never cleared**. So every `processEvents()` after that re-issues `exit()`. This is why the report is iOS-only, and it forecloses the `ExcludeUserInputEvents` alternative — `flags` is never reached.

### The drains were already dead, and were lying about it

The re-issued `exit()` sets `quitNow`, and `qeventloop.cpp:141-142` is `if (threadData->quitNow) return -1;`. So on that path:

- `waitLoop.exec()` returns instantly → `qWarning` "BLE queue drain timed out after 2000 ms"
- `drainDbWork()`'s `loop.exec()` returns `-1` → `STORAGE_WARN` "did not drain within 750 ms"

Both after waiting **0 ms**. The export wait isn't `quitNow`-sensitive but spins the full 750 ms doing nothing, for the same dispatcher reason, inside iOS's termination grace window.

This predates the latch — it was buried under 534 recursions per quit. It is not a regression introduced here, contrary to one agent's reading; no recursion level ever drained, because the dispatcher refuses to dispatch at every level.

All three are now skipped when re-entered, each saying so. #1759 was diagnosed *from* a submitted log, so warnings describing multi-second waits that never occurred are an active cost.

### Also applied

- **Log the first re-entry once, at WARN.** The guarded return was the only path in this handler that changed shutdown behaviour with no trace, while every neighbouring bounded wait logs. Once, not per re-entry — 534 lines would bury the story. `src/main.cpp` is in `MARKER_ONLY_GLOBS`, so a bare `qWarning` with no leading bracket is the gate-passing form; `scripts/check_log_markers.py` passes.
- **Dropped the false rationale.** "The loops exist to let the writes finish" is true on a normal quit and false on the path this fixes.
- **Fixed the thread-safety citation.** `exit()` being "documented main-thread-only" is a request Qt does not enforce (unlike `exec()`, which warns at `:1444-1447`). Replaced with what actually holds it: `QCoreApplicationPrivate::quit()` posts from other threads (`:2141-2146`). Also noted the second emission site, Windows `WM_ENDSESSION` (`qwindowscontext.cpp:1168-1171`), which emits directly and sets the flag *before* the emit — the opposite order from `exit()`.
- **Recorded that the latch never resets** where Qt's flag is cleared on every `exec()` entry (`:1456`), safe only because `main()` calls `exec()` once.

### Verification

Build clean. Full suite via Qt Creator: **110 passed, 0 failed, 0 skipped**. `scripts/check_log_markers.py`: OK.

### Still open

The root cause of the *first* `exit()` is iOS telling us to terminate, which is legitimate. What is not established is why iOS terminated the app in those six episodes. Out of scope here; the latch is correct containment either way, and the new WARN is what will show it still firing.

skialpine:
Closing. The fix is correct, but the case for it doesn't clear the bar.

What it buys is real — the shutdown handler's second half has never run on iOS (every burst in the #1759 log ends mid-cycle at `Waiting for BLE queue to drain`, process killed while still descending), and 37% of that submitted log is spam from this. But the one item with a user consequence, `ensureChargerOn()`, is already covered by the suspend handler ~40 ms earlier, and iOS is killing the process regardless. Nothing here is felt by someone making coffee, and it carries ~100 lines of comment that can't be verified at runtime from this machine.

Branch `fix/abouttoquit-reentrancy` is deleted; the analysis is in the commits and comments above if this resurfaces. The tell to look for is a burst of `Application exiting - shutting down devices` lines ~2 ms apart in a submitted iOS log.

Issue #514, opened 2026-03-20
SAW never arms when scale sample rate is below gate threshold


skialpine:
## Problem

In the debug log from #513, the SAW (Stop At Weight) system never armed during the entire shot. Every SAW-Worker check logged "Flow too low for SAW check" with `flowShort= 0` because the scale's `shortDt` (~0.33-0.35s) was consistently below the gate threshold (0.4s).

## Evidence from debug log

```
[06:51:36.567] SAW-Worker: Flow too low for SAW check: flowShort= 0 weight= 0.2 target= 29 samples= 1 shortWindow= 500 ms shortDt= "0.000" s gate= "0.400" s
[06:51:36.970] SAW-Worker: Flow became valid: flowShort= "3.33" flowLong= "0.00" weight= 1.5 at "6.8" s
[06:51:42.423] SAW-Worker: Flow too low for SAW check: flowShort= 0 weight= 8.4 target= 29 samples= 9 shortWindow= 500 ms shortDt= "0.338" s gate= "0.400" s
[06:51:47.473] SAW-Worker: Flow too low for SAW check: flowShort= 0 weight= 12.7 target= 29 samples= 9 shortWindow= 500 ms shortDt= "0.349" s gate= "0.400" s
[06:51:52.826] SAW-Worker: Flow too low for SAW check: flowShort= 0 weight= 18.2 target= 29 samples= 11 shortWindow= 500 ms shortDt= "0.338" s gate= "0.400" s
```

Result: `No SAW - emitting shotProcessingReady immediately`. The shot only stopped because `target_volume` was set (36 ml). If `target_volume` had been 0, the shot would have had **no stop mechanism**.

## Scale

Skale scale — sample interval ~2 seconds (lines 17→24→27→32→34→36→39→41), giving a `shortDt` of ~0.33-0.35s which is consistently below the 0.4s gate.

## Impact

Any user with a slow-reporting scale (Skale, or any BLE scale with >1.5s intervals) will never get SAW protection. The shot runs until volume limit or the user manually stops.

## Possible fixes

- Lower the gate threshold to accommodate slower scales (e.g. 0.25s)
- Make the gate adaptive based on observed scale sample rate
- Log a warning when SAW fails to arm for the entire shot

PR #1761, opened 2026-08-04
Refractometer recovery tick + SAW de-jitter interval calibration

skialpine:
Two fixes found by reading a device debug log for 2026-08-03. Unrelated mechanisms, same source.

---

## 1. Refractometer recovery tick reported no-op attempts

Six consecutive `Auto-reconnect attempt N` lines at **INFO**, each immediately followed by `Auto-reconnect skipped: a scan is already in flight`. Not one was a real attempt.

The R2 is pursued only while the post-shot review page is open (the "hunt"), and hunt mode keeps scans running back-to-back — so there is always a scan in flight and `tryDirectConnectToRefractometer()` declines every time. The tick logged the attempt before finding that out.

The attempt counter made it more than cosmetic: incrementing on a no-op walked the backoff ramp out to its 60 s tail while nothing was wrong, so a chain that then genuinely died (`onScanError` clears the scan flag and deliberately does not re-chain) waited a minute for recovery instead of 5 s. The safety net degraded itself while idle.

Now returns early when a scan is in flight, re-arming at the first rung without touching the ramp — a watchdog polling for a dead chain rather than a second scanner. Silent on the healthy path; at a 5 s cadence a line there would dominate every hunt window.

Unchanged: page entry still scans immediately via `setRefractometerHunt(true)` (3 ms after open, in every window of the log), the connections page never sets hunt so the tick is stopped while pairing, and real recovery still walks 5 s → 30 s → 60 s.

---

## 2. SAW went blind mid-pour on the WiFi scale

Six `Flow too low for stop-at-weight check` lines with `flowShort=0.00` while the cup was filling at over 2 g/s, each carrying a `shortDt` just under the gate (0.299–0.312 s against 0.325 s).

**Root cause: the de-jitter interval was calibrated only from non-batched gaps.** On a transport delivering several frames per read, every gap reaching that calibration is an *inter-burst* gap — a 10 Hz feed arriving as 5-frame bursts learned ~490 ms instead of 100 ms. Synthetic timestamps advanced ~5× faster than real time, ran past the 2-interval lead cap, and every later frame in a burst was pinned to the same clamped value (the monotonicity guard tested `<`, not `<=`). One burst stamped `[0, 496, 992, 992, 992]`. `computeLSLR` saw a span far under 65% of its window and returned `0.0`; where it did fit, the stretched spacing under-read the slope by 40%.

Neither failure is cosmetic. SAW's threshold is `target - expectedDrip(flow)`, so under-read flow under-predicts drip, raises the threshold and stops **late** — the two shots in that log stopped 0.22 g and 0.36 g past their own thresholds, the second finishing 36.52 g against a 36.00 g target.

### Fixed in the timestamps, not the gate

That distinction is deliberate. [#514](https://github.com/Kulitorum/Decenza/pull/514) already lowered this gate once (80% → 65%, *"so SAW works on 5Hz scales"*) — same defect shape, patched by moving the threshold, and a bursty 10 Hz feed then ate the headroom again. Lowering `qBound`'s 500 ms floor would work too and is also wrong: a shorter window makes `flowRateShort` noisier, and that value feeds drip prediction on every shot.

- **Calibrate from arrival rate** over a 1 s wall-clock window, counting every arrival. Ten arrivals in a second is 100 ms/sample whether they came evenly or all at once. The first measurement replaces the bootstrap outright — EMA-ing toward it took ~10 s, most of a pour.
- **Allow synthetic to lead wall-clock by the LSLR buffer length**, not two intervals. A burst is a whole cadence period of *sample* time delivered at one instant of *arrival* time, so the lead is the model working. Clamping is now strictly monotonic and can never emit a duplicate.
- **Advance a full cadence step, not +1 ms**, when a non-batched sample lands behind synthetic. The first frame of every burst hit that path, injecting a ~200 g/s spike into the fit once per burst.

Simulated against this file's own arithmetic: a true 2.00 g/s pour reads **1.22 g/s** before, **2.25–2.59 g/s** with the first two changes, and **exactly 2.00 g/s** with all three.

### Diagnostic

The throttled log line gains `shortN` and `interval`. `samples` counts the whole 1 s buffer, so a short `shortDt` beside a healthy `samples` reads the same whether timestamps collapsed or the feed went quiet. `shortN` separates a de-jitter fault (ours) from a delivery gap (the transport's); `interval` names the input that was wrong here. Reading the original log without these cost a round trip.

---

## Testing

Full suite via Qt Creator: **110 passed, 0 failed, 0 skipped**, no warnings.

One new test slot in the existing `tests/tst_weightprocessor.cpp` — no new TU. It asserts flow **accuracy**, not merely non-zero, and that matters: the first version checked only "1–3.5 g/s" and **passed against the unfixed code**, because this burst shape yields 1.22 g/s rather than 0. Verified red before green — with the fix reverted, `tst_weightprocessor` fails.

No test for the refractometer tick: it is a lambda inside `main()` with no seam to drive it.

### DIFF
--- a/src/machine/weightprocessor.cpp
+++ b/src/machine/weightprocessor.cpp
@@ -164,7 +164,6 @@ void WeightProcessor::processWeight(double weight)
     // the calibrated scale interval, so this adapts to any scale rate (10Hz, 5Hz, 2Hz).
     constexpr qint64 kBatchThresholdMs = 20;    // Below this, events are batched
     constexpr qint64 kReconnectGapMs = 2000;    // Above this, ignore as reconnect
-    constexpr double kEmaAlpha = 0.3;           // Smoothing factor for interval estimate
     constexpr qint64 kRateWindowMs = 1000;      // Arrival-rate calibration window
 
     qint64 sinceLast = m_lastWallClockMs > 0 ? (wallClock - m_lastWallClockMs) : -1;
@@ -186,9 +185,10 @@ void WeightProcessor::processWeight(double weight)
     // 10 arrivals in 1000 ms is 100 ms/sample whether they came evenly or all at
     // once. The old EMA is kept below as the bootstrap for the first window, since
     // this one cannot report until kRateWindowMs of wall time has passed.
-    if (m_rateWindowStartMs == 0 || (sinceLast > kReconnectGapMs)) {
+    if (m_rateWindowStartMs == 0 || sinceLast > kReconnectGapMs) {
         m_rateWindowStartMs = wallClock;
         m_rateWindowCount = 0;
+        m_rateRecentCount = 0;  // measurements across a reconnect describe a different stream
     }
     ++m_rateWindowCount;
     const qint64 rateSpanMs = wallClock - m_rateWindowStartMs;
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Found 2 issues, both now fixed in [`85b0eb97`](https://github.com/Kulitorum/Decenza/commit/85b0eb97). Reporting them because each reproduced the exact defect this PR exists to remove, and neither was caught by the tests as written.

1. Ring-buffer desync on the reconnect path: `m_rateRecentCount` was cleared without `m_rateRecentNext`, but the count doubles as the index bound for the minimum loop, which is only valid while writes restart at zero. After a >2 s gap the fresh measurement was written at a stale index and never read, while the pre-reconnect value it had just been told to discard was — for two windows (~2 s), arriving exactly when a feed comes back. Both fields now move through `resetRateCalibration()`, mirroring `resetStallTracking()`'s discipline, so the half-reset is unwritable rather than remembered.

https://github.com/Kulitorum/Decenza/blob/85b0eb97998cb777b522ea356fd63da3b71561fd/src/machine/weightprocessor.cpp#L188-L199

2. The bootstrap seed was biased by construction on the transport it was meant to help. Seeding the interval from the first non-batched gap picks up the *inter-burst* gap on a bursty feed (~492 ms for a 100 ms cadence), and the lead allowance cannot contain the result because it is computed from the same wrong interval and inflates in lockstep. Simulated: synthetic ran 1960 ms ahead inside one 8 ms burst, and when the rate window corrected the interval the allowance collapsed, the cap bit, and four consecutive samples landed 1 ms apart reading 0.00 g/s. Removing the seed leaves the first second merely uncalibrated — bounded and self-clearing. Post-warm-up bad samples went 9 to 0.

https://github.com/Kulitorum/Decenza/blob/85b0eb97998cb777b522ea356fd63da3b71561fd/src/machine/weightprocessor.cpp#L346-L372

Also corrected: the test tolerance was 20% while the comment justifying it claimed 10% (CLAUDE.md: comments are "claims with expiry dates"); a new test covers the reconnect path, which nothing reached; and six comments that overclaimed or had gone stale, including `main.cpp` asserting `isScanningForScales()` implies the hunt chain is scanning — [`blemanager.cpp`](https://github.com/Kulitorum/Decenza/blob/85b0eb97998cb777b522ea356fd63da3b71561fd/src/ble/blemanager.cpp#L1479-L1485) warns about that exact flag. The guard is still correct, for a different reason, now stated.

Two notes on the new test, both of which were wrong on the first attempt and are worth knowing before editing it: the post-gap feed must be **bursty** (an evenly-paced feed takes wall-clock on every sample and never consults the estimate, so it passes with the ring desynced), and it must be **short** (the desync costs two windows, so at 8+ bursts the buggy ring converges too and the test goes quiet). Verified red at 1.43 g/s against the committed defect, green after.

Full suite: 110 passed, 0 failed.

<details>

- [OK to skip] The "batched but uncalibrated" branch can still emit duplicate timestamps for two arrivals in the same wall-clock millisecond before first calibration — pre-existing, reachable once per session, and now the bounded warm-up the fix deliberately prefers (score: 55)
- [OK to skip] A mid-session scale swap inherits a stale non-zero interval since it is intentionally kept across shots; recovery depends on rate-window convergence rather than a reset (score: 55)
- [OK to skip] Several comments cite numbers from development-time simulations of intermediate states that no commit preserves; now labelled as simulation rather than coverage (score: 60)

</details>

PR #1764, opened 2026-08-04
De-jitter diagnostics, and a rejected median change recorded

skialpine:
Follow-up to [#1761](https://github.com/Kulitorum/Decenza/pull/1761). **No behaviour change** — the cadence estimate is byte-for-byte what `main` already does.

## Why

Two shots this morning ran on the merged de-jitter fix, and diagnosing them required inferring three things the log could not state:

| I inferred | From | Should have been able to read |
|---|---|---|
| the interval estimate is unstable | shortN ÷ shortDt vs the negotiated 100 ms | the per-window measurements feeding it |
| frames are being lost | shortN=4 where 5-6 were expected | arrivals against expected |
| the feed bursts at all | never observed | burst sizes |

`m_burstFrames` already existed and was never logged. The per-window `measured` value was computed and discarded — only its aggregate was ever visible.

## What this adds

Two lines, both DEBUG, neither able to dominate the view:

- **De-jitter summary**, one per shot at `stopExtraction()`: arrivals and rate, the interval range committed across the shot, percentage of arrivals that batched, deepest burst, and the count of mid-pour samples where the flow estimate came back at exactly zero. That last number was previously unknowable — its log line is throttled to one per 5 s.
- **Rate-window disagreement**, emitted only when a closing window differs from the committed estimate by more than 15%. Silent while the estimator is stable.

Deliberately *not* added: a line per rate window (~30 per shot of "everything is fine") or per burst. That is the cry-wolf shape [#1761](https://github.com/Kulitorum/Decenza/pull/1761) removed from the refractometer tick.

## A change that was tried and rejected

The estimate was switched from minimum-of-three windows to **median**-of-three, on reasoning that sounds right: contamination is one-directional so the minimum rejects it, but ordinary jitter is two-directional, so the minimum sits on the low tail of its own noise.

Measured against a new jittered-bursty fixture at a true 2.00 g/s:

| statistic | reads |
|---|---|
| minimum (current) | 2.25 g/s |
| median | **2.96 g/s** |

The median is twice as wrong. Reverted, with the reasoning written into the code so the next person does not re-derive it.

## The finding that measurement actually produced

Under jittered burst **timing**, this function over-reads flow whatever statistic it commits to. Remove the jitter and the same fixture reads exactly 2.00 g/s. The error is in how a burst is spread against a moving wall-clock, not in which window measurement is chosen — and jittered burst timing is the normal condition on a WiFi feed.

No fix attempted here. The direction of this effect was predicted wrong twice during investigation, so the next step is measurement on device, which is what the diagnostics above exist for.

The defect is now a running `QEXPECT_FAIL` test carrying the measured numbers, rather than a note someone has to rediscover. If it ever XPASSes, the behaviour changed and wants explaining before the marker is deleted.

## Testing

110 passed, 0 failed. The rejected median was verified by measurement in both directions, not by argument.

### DIFF
--- a/src/machine/weightprocessor.cpp
+++ b/src/machine/weightprocessor.cpp
@@ -207,41 +207,58 @@ void WeightProcessor::processWeight(double weight)
             qMax(1, static_cast<int>(rateSpanMs / (m_rateWindowCount - 1)));
         // MINIMUM of the last few windows, not an average and not the latest.
         //
-        // Contamination here is one-directional, which is what makes the minimum
-        // the right statistic rather than merely a robust-looking one: a dropped
-        // frame or a sub-reconnect hiccup removes arrivals from a window and can
-        // only ever push `measured` UP. Nothing pushes it down — no mechanism
-        // delivers more samples than the scale sent. So the smallest recent window
-        // is the one least contaminated, and taking it discards a bad window
-        // outright instead of blending its error in.
+        // Contamination here is one-directional: a dropped frame or a sub-reconnect
+        // hiccup removes arrivals from a window and can only ever push `measured`
+        // UP, since nothing delivers more samples than the scale sent. So the
+        // smallest recent window is the least contaminated, and taking it discards a
+        // bad window outright instead of blending its error in. An EMA was tried
+        // first and is worse: simulated on a 10 Hz feed with one 1.5 s hiccup it
+        // pulled the estimate to 145 ms and ~1 s of pour read 1.38-1.53 g/s against
+        // a true 2.00 g/s.
         //
-        // The alternative tried first was rejecting the contaminated GAP before it
-        // reached the average, by restarting the window on anything more than a few
-        // times the cadence. That cannot work on a bunching transport: the
-        // inter-burst gap (~490 ms against a 100 ms cadence) is itself many times
-        // the cadence, so the window would restart on every burst and never close.
-        // Averaging ACROSS bursts is the whole point; the gap is signal, not noise.
+        // A MEDIAN was tried and REJECTED, which is worth recording because the
+        // argument for it is persuasive and wrong. The reasoning: ordinary jitter is
+        // two-directional, so the minimum of three noisy windows sits on their low
+        // tail, and a median would reject the same one-directional contamination
+        // without that bias. Measured against cadenceEstimateDoesNotTrackTheLowTail
+        // — a jittered bursty 10 Hz feed at a true 2.00 g/s — the minimum reads
+        // 2.25 g/s and the median reads 2.96 g/s. The median is twice as wrong.
         //
-        // Simulated on a 10 Hz feed with one 1.5 s hiccup: an EMA pulled the
-        // estimate to 145 ms and ~1 s of pour read 1.38-1.53 g/s against a true
-        // 2.00; the minimum holds 100 ms and 2.00 g/s throughout.
+        // Neither is right, and that is the actual open problem: under jittered
+        // burst TIMING this function over-reads flow whatever statistic it commits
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 5 issues, all fixed in [`05cfa601`](https://github.com/Kulitorum/Decenza/commit/05cfa6019f93485cf9e2d2f17ab20c2f2ade6424). The first invalidates this PR's headline claim.

1. **The median comparison was measured wrong, and the reported defect is understated.** The PR claimed minimum reads 2.25 g/s and median 2.96 — "twice as wrong". Those were two different *frames* of an oscillation compared against each other (the min run failed at sample 115, the median run at 116). At matching positions across one burst: minimum `2.25 2.92 2.92 2.25 0.04`, median `2.16 2.96 2.96 2.16 0.04` — identical in shape and magnitude. The statistic makes no difference. The actual defect is that flow **oscillates 0.04 to 2.92 g/s within a single burst** on a true 2.00 g/s feed, with the last sample of every burst near zero — stop-at-weight going blind on a regular beat, not occasionally. The test now asserts the spread across a burst instead of one sample.

https://github.com/Kulitorum/Decenza/blob/05cfa6019f93485cf9e2d2f17ab20c2f2ade6424/src/machine/weightprocessor.cpp#L219-L239

2. **The summary fired for steam, flush and hot water.** `stopExtraction()` is wired to `MachineState::shotEnded`, which fires on any flowing→not-flowing edge, and [`isFlowing()`](https://github.com/Kulitorum/Decenza/blob/05cfa6019f93485cf9e2d2f17ab20c2f2ade6424/src/machine/machinestate.cpp#L95-L106) covers steam, `HotWater` and `Flushing`. A flush after a shot logged a second summary carrying that shot's counters over an elapsed time measured from its flow start — indistinguishable from a real one, and double-counting blind samples across a session. Now latched to one per shot.

https://github.com/Kulitorum/Decenza/blob/05cfa6019f93485cf9e2d2f17ab20c2f2ade6424/src/machine/weightprocessor.h#L243-L249

3. **The arrival rate was roughly double.** Counters reset at `startExtraction()` (cycle start, i.e. preheat) while `elapsedMs` starts at `markExtractionStart()` (flow). Numerator covered preheat + pour, denominator only the pour. A 20 s preheat plus 30 s pour on a 10 Hz scale reported ~16.7/s beside a committed interval of 100 ms — which reads as the transport delivering far more samples than the estimate believes, the exact signature being hunted, manufactured by the counter.

https://github.com/Kulitorum/Decenza/blob/05cfa6019f93485cf9e2d2f17ab20c2f2ade6424/src/machine/weightprocessor.cpp#L879-L887

4. **Six counters bypassed the reset discipline this file already has** (CLAUDE.md: "Centralize anything produced at more than one site"). Hand-written assignments including a 0-means-unset sentinel, sitting beside `resetStallTracking()` and `resetRateCalibration()` — both of which exist because a half-reset shipped in [#1761](https://github.com/Kulitorum/Decenza/pull/1761). `resetForRetare()` was already missing them. Now behind `resetShotDiagnostics()`.

5. **The disagreement line was filed under `[SAW]`.** It answers whether readings were timed right, which the file header, the adjacent constant-weight diagnostic, and `LOGGING.md` all state is a `[Scale]` question. Moved to `SCALEFEED_LOG`. Also: the two summary lines printed `m_estimatedIntervalMs` twice one line apart, labelled "de-jitter calibrated" and "committed"; the older line now states its scope (trailing 1 s) and leaves the interval to the summary.

Verified: the cadence estimate is unchanged from `main` after the median round trip — rename only, same seed, same loop bounds.

110 passed.

PR #1767, opened 2026-08-04
fix: raise the web layout editor's action picker above the editor modal (#1765)

### LINKED ISSUE #1765: [Bug] Web App customise Recipe button actions - popup obscures choices
### What happened?

Image

See the initial popup where you choose long or double click obscures the choices

### Steps to reproduce

open web app
go to layout settings, Recipes button, 
choose either double click or long press and see the problem
if you click the back popup, the front one disappears, but the choice you make is not kept.

### Decenza version

2.0.2

### Platform

Windows

### Device / OS version

11

### Scale (if applicable)

_No response_

[4 aug debug.zip](https://github.com/user-attachments/files/30714891/4.aug.debug.zip)

_No response_

### Screenshots or video

_No response_

### Anything else?

_No response_

skialpine:
Fixes #1765.

## The bug

In the web Layout Editor, the action picker opened *behind* the editor that launched it.

`.action-overlay` was at `z-index: 200`; `.editor-panel` — the fixed full-screen modal used by both the Custom widget editor and the built-in widgets' Options editor — is at `z-index: 210`. The picker's list was painted under the editor's backdrop, so the choices could not be read or clicked, and a click on the visible part landed on the editor instead, closing it with no choice made.

Reported against the Recipes button's **Long press** / **Double-click** rows in the Options editor (the gesture overrides added in #1763). The Custom widget editor's Click/Long/DblClk selectors share the same overlay and had the same defect.

## The fix

`.action-overlay` moves to `z-index: 220`, with a comment recording why it must sit above `.editor-panel` so the constraint survives the next edit.

## Verification

Against the running server, with the editor open, `document.elementFromPoint()` at the centre of a picker item:

- `z-index: 220` — hit is the picker item.
- `z-index: 200` — hit is the editor content underneath (bug reproduced).

Full test suite run locally: passing.


skialpine:
Two follow-ups pushed.

**1. Comment corrected after review.** The first version generalised one editor's geometry to both. The 280px `.action-dialog` is taller than the short readout-options card, so it peeked out above and below and a click on the peeking part hit the backdrop — that is the reported symptom. Under the 900px Custom widget card it was covered outright and nothing responded at all. The comment now says which is which, notes that neither `closeEditor()` nor `closeReadoutOptions()` clears the picker (the reporter's "the choice you make is not kept"), and states the requirement rather than a call-site census that goes stale. A back-reference at `.editor-panel` makes the 210/220 coupling visible from both ends.

**2. `WEB_CSS_TOAST` extraction folded in.** Review flagged `.lib-toast` as hand-copied between the Layout Editor and the Theme Editor, and already drifted — five style values including z-index 9999 vs 300, plus a behavioural one: only the Layout Editor's copy cleared its dismiss timer, so on the Theme Editor a second toast inherited the first's countdown and vanished early. Both surfaces now use `WEB_CSS_TOAST` / `WEB_HTML_TOAST` / `WEB_JS_TOAST` from `src/network/webtemplates/toast.h`.

- The rem-based metrics win; the Theme Editor's toast moves ~4px and gains the timer clear (fixing the early dismissal).
- The one genuinely per-page value — where the toast sits in that page's stack — is `var(--z-toast, 9999)`, which the Theme Editor overrides with 300 to stay under its CRT shader overlay.
- CLAUDE.md records the case and the rule that a duplicate spotted in passing gets collapsed in the same change.

Build clean (one pre-existing `__eh_frame` linker warning, unrelated). Full suite: 110 passed, 0 failed.

PR #1772, opened 2026-08-06
fix(scale): make WiFi scale discovery work on Android

skialpine:
Fixes WiFi scale discovery on Android, where a tablet spent five days unable to find a scale two feet away and reachable by IP.

## What was actually wrong

Two independent defects, found in that order.

**1. The responder answers per peer.** Measured 2026-08-06, one LAN, two scales on openscale 3.1.13: the tablet sent 7 A-record queries per attempt, 5 s per attempt, and got **0 records** — every attempt, across a session of several hours — while a Mac resolved the same name in **272 ms, one query**. Opening one TCP connection from the tablet to a scale flipped it to 1 query / 272 ms, and **only for that scale**; the second scale, never contacted from the tablet, stayed silent to it while answering the Mac normally in the same window. The tablet's send path was verified on the wire by a third host on 224.0.0.251:5353.

That refutes what the spec said — "not a property of the responder", "a host-side condition" cleared by a reboot. One resolver stack cannot produce opposite outcomes for two peers at the same moment.

**2. The app held no MulticastLock.** Android's Wi-Fi driver discards multicast frames not addressed to this device without one. The only lock in the tree belonged to ShotServer, whose setting **defaults to false** — while three comments described it as held "for the whole app lifetime". Nothing failed, because a query from an ephemeral port is answered by *unicast*, which the filter never touches. Every mDNS answer the app has ever received on Android came back unicast, which is exactly why it stopped working against a responder that will not unicast to a stranger.

## What changed

- **Queries go out from port 5353** (the openscale maintainers' recommended client-side workaround), falling back to ephemeral only if the bind fails. mjansson then drops the QU bit itself, so answers come back multicast with nothing per-peer in the path.
- **The app takes its own reference-counted MulticastLock** for each lookup and browse. This and the port are two halves of one change — shipping the port alone would have *reproduced* the older "5353 gives records=0 on Android" measurement rather than fixed it, which is what that measurement always was.
- **NsdManager browses alongside our own on Android.** The system daemon owns 5353, so it sees multicast answers and unsolicited announcements — the only path that can find a scale this device has never talked to. Restores a mechanism removed in #1249 for browsing the wrong service type, which openscale v3.0.9 fixed.
- **A cached IP that answers nothing is kept, not evicted.** Silence says nothing about which device owns an address, and dialling that address is what puts this device back on the responder's answering side.
- **macOS defaults to the mjansson browse.** Bonjour is faster there (66-113 ms vs 160-270 ms) and stays compiled and selectable — but macOS is the development platform (~2 installs) and Android is the shipped one (hundreds), and defaulting to Bonjour meant nobody ran the browse three platforms ship. Run `backend=bonjour` before an iOS release; that is the early warning this gives up.
- **Every WiFi scale row shows its hostname.** openscale only embeds the name in its DNS-SD instance label when the scale has been *renamed*, so a default scale's row read "Half Decent Scale (WiFi)" with nothing identifying it — while the same scale, once connected, was named "Half Decent Scale (hds) (WiFi)". One scale, two spellings, the useless one in the picker.
- **Diagnostics:** `devices_wifi` gains `resolver` and `queryPort`; every lookup and browse logs the port it bound to, so `records=0` is finally interpretable. `resolver=system` is refused on Android, where it would disable `.local` resolution for the session.

## Corrections to the record

The ARP mechanism (legacy query → unicast answer → responder must ARP → `ETHARP_TRUST_IP_MAC` off) fits every measurement but is **not proven**: no ARP frames were captured, and there is a counter-example on file — misses continuing against a scale that had served a WebSocket 16 s earlier. Every site asserting it now says so. The "host-side resolver state" rationale is replaced by the two-scale control. The measured window (several hours) and the user-visible outage (five days) are no longer used interchangeably.

## Verification

- **110/110 tests pass**, build clean.
- **On-device A/B on macOS**: both scales resolve on either query port; the 5353 socket sees 247 records where the ephemeral one saw 0, and correctly reports `(none)` for hosts that do not exist. Speed is a wash between the two ports — the port was never a latency fix.
- **Android CI is the outstanding gate.** The JNI in `startNsdBrowse()` and `WifiScaleNsdHelper.java` is the only code macOS never compiles. Three earlier dispatches: one real compile error (fixed), two lost to the GitHub Actions incident.

OpenSpec: `openspec/changes/add-android-nsd-scale-browse/`

Bug report delivered to the openscale team separately; this is what the app can do without waiting for firmware.


skialpine:
### Code review

Found 8 issues. The NSD half needs rework before merge; the Android build is already red on (1).

1. Does not compile on Android — `SCALE_INFO_TAGGED` in lambdas with no `this`. The macro emits `logMessage(...)`, a non-static member; both new sites capture only `self`. This file's own comment at L112 warns about exactly this, and the two `SCALE_WARN_STDERR_TAGGED` calls in the same function do it correctly. Confirmed by [run 31114480361](https://github.com/Kulitorum/Decenza/actions/runs/31114480361): `error: 'this' cannot be implicitly captured in this context`.

https://github.com/Kulitorum/Decenza/blob/4ec5622a1314cd9a05f3a8e929d0206a6d6809e6/src/network/wifiscalediscovery.cpp#L344-L358

2. NSD results are silently discarded or keyed wrongly. `InetAddress.getHostName()` on an address built from an IP literal does a blocking reverse-DNS lookup, and returns either the IP literal (mapped to `""` here, so `upsertByHostname` drops the result entirely) or a router PTR answer like `hds.lan` (a different identity key from the mjansson path's `hds.local` — duplicate row, and no match against a saved `wifi:hds.local` primary). Either way the scale the NSD path found does not reach the list. The adjacent comment claiming "Deduplication is the caller's, by address" is also wrong: the caller dedupes by hostname.

https://github.com/Kulitorum/Decenza/blob/4ec5622a1314cd9a05f3a8e929d0206a6d6809e6/android/src/io/github/kulitorum/decenza_de1/WifiScaleNsdHelper.java#L222-L228

3. Process-wide `sActive` makes browses cancel each other, breaking a documented invariant. `BLEManager` owns three `WifiScaleDiscovery` instances; `wifiscalediscovery.h` L32 states "Instances do not cancel each other, which is the whole reason the reconnect path has its own." `browseBlocking()` calls `cancelBrowse()` on entry and stops whatever listener is registered, with no ownership check — so a user tapping Scan kills the reconnect ladder's browse, and the ladder's own exit then kills the scan's.

https://github.com/Kulitorum/Decenza/blob/4ec5622a1314cd9a05f3a8e929d0206a6d6809e6/android/src/io/github/kulitorum/decenza_de1/WifiScaleNsdHelper.java#L117-L121

4. The NSD worker cannot be cancelled — reintroduces the quit hang `m_browseCancel` exists to prevent. Every other blocking path threads a cancel token because `~QCoreApplication` calls `QThreadPool::waitForDone()` unconditionally. `startNsdBrowse` captures no token, and the Java `cancelBrowse()` is never called from C++. Quitting 1 s into a 15 s scan blocks teardown for the remaining ~14 s.

https://github.com/Kulitorum/Decenza/blob/4ec5622a1314cd9a05f3a8e929d0206a6d6809e6/src/network/wifiscalediscovery.cpp#L296-L300

5. `resolver=system` permanently disables `.local` resolution on Android for the session. The setting is process-global and nothing restores it, so the reconnect ladder and manual-entry dialog inherit it. This is NOT symmetric with the `backend` selector cited as precedent: on Android `resolveBackend()` substitutes mjansson for any request, so pinning the browse backend is harmless, whereas Android's `getaddrinfo` NXDOMAINs `.local` — as the new code's own comment says. One MCP call leaves discovery dead until restart, with nothing on screen to explain it. Restore the prior value when the browse finishes, or refuse `system` on Android.

https://github.com/Kulitorum/Decenza/blob/4ec5622a1314cd9a05f3a8e929d0206a6d6809e6/src/mcp/mcptools_devices.cpp#L82-L89

6. Concurrent `resolveService()` drops the second scale. NsdManager serialises resolutions and fails the second listener with `FAILURE_ALREADY_ACTIVE`; the handler only logs at `Log.d` with no retry, and `onServiceFound` fires once per instance. On a two-scale LAN — the case the class doc describes — scale #2 never appears.

https://github.com/Kulitorum/Decenza/blob/4ec5622a1314cd9a05f3a8e929d0206a6d6809e6/android/src/io/github/kulitorum/decenza_de1/WifiScaleNsdHelper.java#L137-L141

7. A comment this PR invalidates was not updated. `mdnsresolver.cpp` L12-15 states resolveHostname "is still only *called* on Android — WifiScaleDiscovery guards its call site — so compiling it elsewhere changes no behaviour". This PR replaces that guard with a runtime branch and adds an MCP argument whose documented purpose is calling it on desktop. It sits directly above the "Do NOT narrow this to Q_OS_DARWIN" warning, so a stale claim there is load-bearing. (CLAUDE.md: "a comment nobody re-derives is how a bad decision keeps getting re-approved".)

https://github.com/Kulitorum/Decenza/blob/4ec5622a1314cd9a05f3a8e929d0206a6d6809e6/src/network/mdnsresolver.cpp#L10-L17

8. The same measurement is hand-copied into four files and has already drifted. Two copies say the failure lasted "hours", two say "five days", for one dated observation with the same 272 ms control figure. (CLAUDE.md: "Centralize anything produced at more than one site... each copy is free to change alone, silently, and nothing fails when one does.") `WifiScaleNsdHelper.java` L34 and `wifiscalediscovery.cpp` L289 say hours; `decentscalewifi.cpp` L840 and `tst_decentscalewifi.cpp` L937 say five days.

https://github.com/Kulitorum/Decenza/blob/4ec5622a1314cd9a05f3a8e929d0206a6d6809e6/src/ble/scales/decentscalewifi.cpp#L838-L842

Separately, the ARP mechanism is asserted as settled in three comments, but `decentscalewifi.cpp` L336-346 records a case it cannot explain — misses continuing "against a scale that had served a WebSocket on its IP 16 s earlier". The two-scale control does establish the behaviour is per-peer on the responder side; the specific ARP cause does not follow from it and should not be stated as fact next to a counter-example.

<details>

- [Worth a look] NSD emits its whole harvest after the window instead of incrementally, contradicting the documented "rows appear as scales answer" contract (score: 70)
- [Worth a look] `hostnameResolverReportsWhatActuallyRan` restores `Auto` only on the success path; a failed QVERIFY leaves it pinned globally and cascades into later tests (score: 65)
- [Worth a look] NSD result lambdas check `m_browseGeneration` but not `m_browseInFlight`, so `resultFound` can land after `browseFinished` (score: 65)
- [Worth a look] "NSD browse finished — N resolved" counts lines skipped by the malformed-payload guard (score: 60)
- [Worth a look] Indentation destroyed at `wifiscalediscovery.cpp` L165/L173 when the `#else` block became an `else` branch (score: 55)
- [OK to skip] "no handshake means nothing was there to judge" overstates it — a peer that accepts TCP and stalls mid-upgrade is bucketed with silence (score: 55)

</details>

Checked and clear: the `m_wsHandshakeDone` eviction gate (flag cleared per attempt, set only in `onConnected`, silence branch genuinely reachable, peer-answered-non-HDS still evicted by `onError`); the tab/pipe parsing round-trip; `McpSurfaceVersion` 1.1.0 with an unchanged fingerprint (the check digests tool and action names only); the `std::atomic<HostnameResolver>` substitution logic.

skialpine:
### Code review

Found 7 issues:

1. `MulticastLock` uses the **Activity** context while its comment asserts the opposite. `QNativeInterface::QAndroidApplication::context()` returns `g_jActivity` when an activity exists (Qt 6.11.1, `qjnihelpers.cpp:390-399`), and `nullptr` when none does — so on the background-reconnect path the comment names, the lock may silently never be taken. (CLAUDE.md: "Read it instead of guessing at Qt behaviour... An un-sourced claim in a comment gets believed and then licenses wrong code.")

https://github.com/Kulitorum/Decenza/blob/dd84974c17fdf7112d0e957fe5b0b555f7aee2d3/src/network/multicastlock.cpp#L31-L40

2. A stuck `resolveService()` permanently wedges a browse. `resolveInFlight` is cleared only from `onServiceResolved`/`onResolveFailed`; `NsdManager.resolveService()` has no timeout and can invoke neither. Every queued instance is then never reported and the browse logs "0 resolved" — the exact "found nothing vs never ran" confusion this class exists to prevent. `stopBrowse()` also never calls `stopServiceResolution`, so the orphaned listener retains the whole `Browse`.

https://github.com/Kulitorum/Decenza/blob/dd84974c17fdf7112d0e957fe5b0b555f7aee2d3/android/src/io/github/kulitorum/decenza_de1/WifiScaleNsdHelper.java#L224-L269

3. `resultFound` can arrive after `browseFinished`. The mjansson lambda checks `generation` and `m_browseInFlight`; the NSD lambda checks only `generation`. A row can appear after the Scan button reports done, and `isBrowsing()` reads `m_browseInFlight` — false while an NSD browse is still live — so the reconnect ladder can start a second overlapping browse, which `blemanager.h`'s own comment forbids.

https://github.com/Kulitorum/Decenza/blob/dd84974c17fdf7112d0e957fe5b0b555f7aee2d3/src/network/wifiscalediscovery.cpp#L417-L433

4. `stopBrowse()` cannot cancel the NSD worker once mjansson has finished. `m_browseInFlight` is cleared only by the mjansson lambda, so the early return skips setting the cancel token; `browse()` then replaces `m_browseCancel`, making the old token unreachable even from the destructor. `~QCoreApplication`'s unconditional `waitForDone()` then blocks — the quit hang the comment claims to have designed away.

https://github.com/Kulitorum/Decenza/blob/dd84974c17fdf7112d0e957fe5b0b555f7aee2d3/src/network/wifiscalediscovery.cpp#L452-L462

5. `onServiceFound()` enqueues with no dedup. `b.reported` dedups the output, not the queue, and resolves are serialized one at a time, so repeated announcements can grow `b.pending` without bound and waste the single resolve slot re-resolving the same instance.

https://github.com/Kulitorum/Decenza/blob/dd84974c17fdf7112d0e957fe5b0b555f7aee2d3/android/src/io/github/kulitorum/decenza_de1/WifiScaleNsdHelper.java#L145-L154

6. `[MulticastLock]` is hand-written at five call sites in a brand-new file, and is not a registered marker in `logtags.h`. (CLAUDE.md: "A log/error prefix belongs in one helper, not at the call site" and "Centralize anything produced at more than one site — never hand-roll it per call".) The marker gate is silent because the new file is in neither `COVERED_GLOBS` nor `MARKER_ONLY_GLOBS`, so it advertises a `debug_get_log` filter that returns an incomplete answer.

https://github.com/Kulitorum/Decenza/blob/dd84974c17fdf7112d0e957fe5b0b555f7aee2d3/src/network/multicastlock.cpp#L36-L45

7. `mdnsresolver.h` still says `resolveHostname()` is "Needed on Android ONLY" — falsified by this PR's own runtime `HostnameResolver` selector and the `resolver=mjansson` MCP argument, whose stated purpose is running Android's path on a desktop. (CLAUDE.md: fix pre-existing issues in files you touch; this header is heavily modified here.)

https://github.com/Kulitorum/Decenza/blob/dd84974c17fdf7112d0e957fe5b0b555f7aee2d3/src/network/mdnsresolver.h#L14-L21

---

<details>

- [Worth a look] `displayName()` suppresses the hostname when it is a substring of the label, so a scale named `scale`/`half`/`decent` renders as "Half Decent Scale (WiFi)" with no hostname — the exact defect the hunk fixes (score: 72)
- [Worth a look] `g_warned` latches the "could not take the lock" warning off for the session after one failure, so a run that never gets the lock logs one line and then goes silent (score: 62)
- [Worth a look] `blemanager.cpp:2888-2898` still carries the "host-side, NOT a property of the responder" claim this PR refutes elsewhere; that file is not otherwise touched here (score: 50)

</details>

PR #1779, opened 2026-08-07
fix(visualizer): upload this shot's yield, not the previous shot's

skialpine:
## Reported

Two users, one root cause.

**Report 1** — a shot the app correctly shows as 15.0 g → 45.4 g appears on Visualizer as 15.0 g → 60.1 g. 60.1 g was the *previous* shot's yield. The uploaded weight lags one shot behind.

**Report 2** — adding notes to an already-uploaded shot and re-uploading fixes the displayed weight, but the shot's metadata still reads back the previous shot's number.

## Cause

`MainController::onShotEnded()` copied `ShotMetadata::drinkWeight` from the sticky DYE setting `dyeDrinkWeight()`. That setting is only updated to *this* shot's weight inside the save callback — and the Visualizer auto-upload fires from that same callback, eleven lines earlier, carrying metadata built before the shot was saved. `buildShotJson` then preferred `metadata.drinkWeight` over the measured `finalWeight` argument it was also handed, so the stale value won.

Swapping recipes is not the trigger — it only makes the wrong number visible, because consecutive yields then differ. Someone pulling the same yield every time would never notice.

The local database was never affected: `saveShot()` ignores that field and stores the measured `finalWeight`. That is why the app showed the right number and Visualizer did not.

Report 2 is the same bug with its display half repaired. Re-upload is a PATCH built from the stored row, so the displayed weight corrects itself; the original uploaded `.shot` blob keeps the stale number in `app.data.settings.drink_weight`, and that blob is what an AI reading the shot back sees.

## Fix

`ShotMetadata` no longer carries a drink weight at all.

The yield is a per-shot **measurement** and already travels as an explicit argument to `uploadShot()`/`buildShotJson()`. Carrying a second copy inside a struct that is otherwise sticky *setup* state — bean, grinder, bag, recipe — is what let one shot's yield reach Visualizer under another shot's name. `buildShotJson` now reads its `finalWeight` argument directly, and the bug is structurally unavailable rather than corrected at one call site.

`beanWeight` deliberately keeps its metadata-first resolution. The dose setting and the `doseWeight` argument genuinely differ — the caller derives a profile fallback for it — so that ternary does real work.

An intermediate version of this branch fixed only the assignment, keeping `finalWeight > 0 ? finalWeight : dyeDrinkWeight()`. Review caught that the fallback reinstated the stale value whenever `finalWeight` was 0, and — by making the field non-zero — also suppressed the uploader's own flow-integrated fallback. Its comment was wrong twice over besides. That history is in the commits.

## Also in this PR

All downstream of the same investigation:

- **Deleted `MainController::uploadPendingShot()`** — `Q_INVOKABLE`, no caller anywhere, wired to QML twice historically and unwired for good in `4f578544`. It carried its own copy of the metadata block and the same stale read. Three review agents mistook it for the live manual re-upload path and reasoned from that, which is the cost of leaving it. Its members go with it; the shot-end epoch becomes a by-value lambda capture like every other value the save callback uses.
- **Extracted `buildShotMetadataFromSettings()`** — 23 sticky DYE assignments were duplicated across the real and simulated shot paths, with a third copy in the deleted method.
- **The dev simulated-shot gesture no longer calls `setDyeDrinkWeight()`** — it was planting an invented 40 g in real persisted user data.
- **Recorded the miss in `settings_dye.h`** — that comment already stated the test a sticky DYE field must pass ("if anything ever feeds those fields automatically, they become this bug") and listed the tolerated fields without naming `dyeDrinkWeight`, which fails it on every shot. The list read as complete when it was not.

## User-visible effect

Automatic uploads report the shot just pulled. When no weight source produces a number, the upload carries the flow-integrated volume or no weight field, instead of the previous shot's yield stated as fact.

Shots already uploaded keep the stale value inside their original blob. Re-uploading fixes what Visualizer displays but not that embedded copy — no migration, consistent with #1573's precedent for the identical bug shape on `dyeEspressoEnjoyment`.

No wiki manual entry: this restores intended behaviour rather than adding or changing a feature.

## Testing

Build clean, full suite green locally through Qt Creator: 110 passed, 0 failed, 0 warnings.

No new test. The changed logic is an ordering/ownership property of `MainController::onShotEnded()`, which has no test harness — standing one up needs BLE, settings and storage. `buildShotJson` is private with no testing seam, unlike its `public static` siblings. Called out rather than papered over with a test that asserts something easier. The structural fix is what prevents recurrence here: the field a stale value travelled through no longer exists.

Neither report has a GitHub issue, so this closes nothing.

### DIFF
--- a/src/controllers/maincontroller.cpp
+++ b/src/controllers/maincontroller.cpp
@@ -2237,6 +2237,47 @@ QString MainController::currentHotWaterSpecJson() const {
     return compactJson(o);
 }
 
+ShotMetadata MainController::buildShotMetadataFromSettings() const {
+    // Every field here is sticky DYE state that describes the SETUP — the bean,
+    // the grinder, the bag, the recipe — and is identical for a real shot and a
+    // simulated one. Per-shot measurements (beanWeight, drinkWeight) and
+    // per-shot provenance (yieldMode, yieldAnchorValue) are deliberately absent:
+    // each caller supplies them, so the reasoning about where a weight comes
+    // from stays next to the assignment. A third copy of this block lived in
+    // uploadPendingShot() until that dead method was deleted; centralizing is
+    // what stops the next copy drifting (CLAUDE.md, "Centralize anything
+    // produced at more than one site").
+    ShotMetadata metadata;
+    metadata.beanBrand = m_settings->dye()->dyeBeanBrand();
+    metadata.beanType = m_settings->dye()->dyeBeanType();
+    metadata.roastDate = m_settings->dye()->dyeRoastDate();
+    metadata.roastLevel = m_settings->dye()->dyeRoastLevel();
+    metadata.grinderBrand = m_settings->dye()->dyeGrinderBrand();
+    metadata.grinderModel = m_settings->dye()->dyeGrinderModel();
+    metadata.grinderBurrs = m_settings->dye()->dyeGrinderBurrs();
+    metadata.grinderSetting = m_settings->dye()->dyeGrinderSetting();
+    metadata.equipmentId = m_settings->dye()->activeEquipmentId();
+    metadata.rpm = m_settings->dye()->dyeGrinderRpm();
+    metadata.drinkTds = m_settings->dye()->dyeDrinkTds();
+    metadata.drinkEy = m_settings->dye()->dyeDrinkEy();
+    metadata.espressoNotes = m_settings->dye()->dyeShotNotes();
+    metadata.barista = m_settings->dye()->dyeBarista();
+    metadata.beanBaseJson = m_settings->dye()->dyeBeanBaseData();
+    // Coffee bag snapshot: which bag this shot was pulled with, and the
+    // beans' freeze lifecycle at shot time (bean-bag-inventory).
+    metadata.bagId = m_settings->dye()->activeBagId();
+    metadata.frozenDate = m_settings->dye()->activeBagFrozenDate();
+    metadata.defrostDate = m_settings->dye()->activeBagDefrostDate();
+    metadata.storageHint = m_settings->dye()->activeBagStorageHint();
+    metadata.openedDate = m_settings->dye()->activeBagOpenedDate();
+    // Recipe provenance (add-recipes): the recipe active at shot time and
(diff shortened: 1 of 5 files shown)


skialpine:
### Code review

Three rounds of automated review ran against this branch, plus a cleanup pass. Every finding that survived scoring has been fixed in `a425c274`; nothing is left open.

**Round 1** — against the original one-line fix (`582b0e5a`):

1. The `finalWeight > 0 ? finalWeight : dyeDrinkWeight()` fallback reinstated the stale value it was meant to remove, and by making the field non-zero also suppressed the uploader's flow-integrated fallback (score 90).
2. The comment called `dyeDrinkWeight()` a user-entered value — it is overwritten by the app after every save — and described the fallback as the no-scale case, which it is not (score 75).
3. "Every Visualizer upload" overstated the blast radius; only automatic post-shot uploads were affected (score 95).
4. `uploadPendingShot()` unreachable, carrying the same stale read (score 100).

A fifth finding, that a cross-file claim needed a file:line citation, scored 0 — that CLAUDE.md rule is scoped to Qt behaviour. Dropped.

**Round 2** — against the fixes for round 1. Caught a header comment claiming the shot epoch made the upload and the shot record "agree on the time"; `ShotHistoryStorage::saveShot()` stamps its own timestamp, so that guarantee did not exist. Also corrected my claim that `dyeDrinkWeight` is never user-written — [PostShotReviewPage.qml:844](https://github.com/Kulitorum/Decenza/blob/a425c274a06b0e3c4c9d13a1b1d2ba4bf9b1ff5e/qml/pages/PostShotReviewPage.qml#L843-L845) and MCP `settings_set` both write it. The code stayed; the reasoning had to change.

**Round 3** — clean. No false claims, all line citations exact.

**Cleanup pass** produced the extraction of `buildShotMetadataFromSettings()` (23 duplicated assignments across two paths, a third copy in the deleted method), and the altitude finding that drove the final shape of this PR: the fix belonged at the struct, not the assignment.

Worth stating plainly for anyone reading this later: the first attempt at this fix introduced a fresh instance of the bug it was fixing, and the comment written to explain it asserted three things that were false. Both were caught by review, not by the build or the tests — all of which were green throughout.

PR #1783, opened 2026-08-09
feat(steam): one steam-heater policy, two settings, built-in Heater off

skialpine:
## The reported bug

> "the steamer doesn't seem to start heating even though I've picked a recipe with milk"

The steam-heater target was derived in several places from different subsets of its inputs. `ProfileManager::uploadCurrentProfile()` carried a private copy that knew only two of five, and because the upload it triggers is deferred behind `m_uploadInFlight` it landed **after** a recipe activation and re-sent `steam = 0`.

`SteamHeaterPolicy` is now the single derivation. Every path that writes ShotSettings or reports heater state resolves through it, and a source scan in `tst_steamheaterpolicy` fails the build if a second derivation grows back.

## The model

```
PERMISSION  Keep warm when idle (state)
            an explicit steam action, or a steaming recipe's SHOT START (event, revoked at Idle)
VETO        effective pitcher is "Heater off" · the transient flag
            Let the recipe decide + an active recipe that steams nothing
```

| Keep warm when idle | Let the recipe decide | The heater is |
|---|---|---|
| ✓ | — | always warm |
| ✓ | ✓ | warm, unless the active recipe has no pitcher |
| — | ✓ | warm once you start a milk recipe's shot |
| — | — | warm only when you press Steam |

Recipe intent is **three**-state on purpose: "no recipe" and "an espresso is active" pull in opposite directions, and collapsing them turns *Let the recipe decide* into a heater-off switch for anyone who has not activated anything.

Permission arrives at **shot start**, not activation — users park a recipe as the machine's resting state, so a selected latte says nothing about when milk is wanted.

Event permission outranks every veto: a GHC press puts the DE1 into Steam in firmware and the app cannot refuse it, so resolving "off" there would make every readout a lie about a boiler that is actively heating. A toast explains that the last pitcher's settings are in use.

## Also in here

- **Built-in "Heater off" entry** replaces user-created Off presets. Appended **last** so no stored index shifts; selected by sentinel `-1`.
- **Migration** removes the old Off presets and remaps the selection. Without the remap the boiler silently turns on at upgrade for exactly the users who chose to keep it off.
- **A recipe's pitcher is an override**, unwound on deactivation, instead of permanently overwriting the standing selection.
- **The readouts show "Off"** from the resolved state, never from the measured temperature — a boiler switched off minutes ago still reads hot (the simulator reports 25 °C ambient, a plausible-looking number).

## Three bugs found by driving a live simulated DE1

None of these were visible to the test suite:

1. **Nothing pushed ShotSettings when a machine first became live.** The app and the machine disagreed from launch — every setting said warm/160 while the DE1 held 0.
2. **Steam settings written over MCP never reached the machine** (pre-existing for `steamTemperature`/`steamTimeout`/`steamFlow` too). Fixed at the signal, so QML, MCP, the web UI and a settings import all reach the machine by one route rather than each remembering to.
3. **The MCP kept its own copy of pitcher-select and had already drifted** — selecting "Heater off" then a real pitcher left the boiler cold.

## Extracted while here

- `SteamLabels` — one definition of the "Heater off" / "Off" labels, which were about to have a fifth copy. The built-in stores no name (a stored one would be stale after a language change, and `name` is what recipes match on), so each view had to supply it.
- `StatusToast` — three byte-identical toast blocks in `main.qml`, each with its own Timer.

## Upgrade behaviour — for release notes

- `keepSteamHeaterOn` becomes two settings; **Let the recipe decide defaults ON** for existing installs. Anyone with a parked pitcher-less recipe will find the heater off where it used to be warm.
- User-created "Off" presets are removed in favour of the built-in entry, and a selection that pointed at one lands on it.

## Verification

- **111 tests pass**, clean build, QML gate 234/234, log-marker / test-duplication / MCP-budget gates green, `openspec validate` clean.
- The migration's selection remap was **falsified deliberately** — broken, rebuilt, watched go red, restored.
- Driven on a live simulated DE1 over MCP and on screen: startup push, both toggles, the Heater off round-trip, and the readouts showing **Off** while the boiler measured 25 °C.

## Known gaps

- Tests 2.3 / 6.5 / 7.3 and the wiki screenshot (10.3) are still open in `tasks.md`; being addressed on this branch.
- Wiki `Manual.md` / `FAQ.md` edits are staged locally in the wiki repo, not pushed.


skialpine:
### Code review

Found 7 issues:

1. Tapping "Heater off" in the app stores the entry's **display slot** (`steamPitcherCount()`), not the sentinel `-1`. Only the MCP path normalises. The stored value stops meaning "Heater off" the moment a preset is added or deleted, and `removeSteamPitcherPreset`'s clamp then converts it into the last real pitcher — the boiler turns on with no user action. This is the two-index-space hazard the design claims appending-last avoided; `isHeaterOffPitcher()` answering to both representations is what created it. Fix at the single write seam.

https://github.com/Kulitorum/Decenza/blob/42e583fa9820d8702f661c70c72cb4f4c5601bb1/src/core/settings_brew.cpp#L538-L543

2. `letRecipeDecide` is never consulted when a recipe overrides the pitcher, contradicting this change's own spec ("WHEN Let the recipe decide is off THEN activating any recipe leaves the effective pitcher and the heater state unchanged"). A user with Keep warm on and Let the recipe decide **off** who activates a recipe carrying the Heater off marker gets a cold boiler.

https://github.com/Kulitorum/Decenza/blob/42e583fa9820d8702f661c70c72cb4f4c5601bb1/src/controllers/maincontroller.cpp#L1940-L1944

3. Shot start warms a recipe explicitly set to "Heater off". The intent provider derives `WantsSteam` from `hasMilk` alone and never reads the `heaterOff` marker; event permission then short-circuits every veto. The migration makes this reachable — it writes `heaterOff: true` while leaving `hasMilk` untouched — so a recipe whose entire purpose is "don't heat the boiler" lights it when its shot starts.

https://github.com/Kulitorum/Decenza/blob/42e583fa9820d8702f661c70c72cb4f4c5601bb1/src/controllers/maincontroller.cpp#L134-L140

4. `steamHeaterOn` has no notify path for the active recipe, which is a `resolve()` input. Activating a pitcher-less recipe commands the DE1 to 0 correctly but leaves every steam readout showing a temperature until some unrelated setting changes; deactivation has the mirror bug. Emit on `activeRecipeChanged`.

https://github.com/Kulitorum/Decenza/blob/42e583fa9820d8702f661c70c72cb4f4c5601bb1/src/controllers/maincontroller.cpp#L160-L166

5. The migration's removed-name list is consumed before the async rewrite can fail. `takeMigratedHeaterOffNames()` deletes the key, then `requestHeaterOffPitcherRewrite()` is fire-and-forget; a failed SELECT or UPDATE loses the names permanently. The header claims the opposite: "Reading CLEARS them, so the rewrite runs once and a partial run is not silently forgotten."

https://github.com/Kulitorum/Decenza/blob/42e583fa9820d8702f661c70c72cb4f4c5601bb1/src/controllers/maincontroller.cpp#L391-L395

6. The settings-backup import drops `disabled` presets and then applies the stored `selectedPitcher` after only a range check — the exact renumbering hazard the constructor migration exists to prevent, reintroduced on the import path with no remap. A backup of `[Off, Small, Large]` with selection 0 restores as "Small", heater on.

https://github.com/Kulitorum/Decenza/blob/42e583fa9820d8702f661c70c72cb4f4c5601bb1/src/core/settingsserializer.cpp#L494-L500

7. Two comments assert facts that are not true (CLAUDE.md: "a comment nobody re-derives is how a bad decision keeps getting re-approved"). `heaterOffPitcherEntry()` writes an `off` key justified as "what the add-dialog wrote" — no writer has ever produced that key on a pitcher preset (`addSteamPitcherPresetDisabled` wrote only `disabled`), and nothing reads it. Separately, `SteamHeaterPolicy` claims the active recipe "is mutated at six sites"; the actual count is 4 direct sites plus a shared helper.

https://github.com/Kulitorum/Decenza/blob/42e583fa9820d8702f661c70c72cb4f4c5601bb1/src/core/settings_brew.cpp#L493-L498

<details>

- [Worth a look] The standing pitcher is stored as a raw index and no preset mutator adjusts it, so deleting a pitcher while a recipe override is active can unwind to the wrong pitcher (score: 70)
- [Worth a look] Stale comments naming the deleted `sendSteamTemperature` in `SteamPage.qml` and `machinestate.cpp`, and a "keep in sync" comment left directly above the call that removed the twin in `IdlePage.qml` (score: 70)
- [Worth a look] `currentSteamSnapshot()` in the wizard never emits `heaterOff`, so promoting current state with the built-in selected loses the choice (score: 60)
- [OK to skip] The built-in pill has no QML guard on double-click/long-press rename; Save/Delete are refused in C++ with a `qWarning` (score: 55)

</details>

PR #1785, opened 2026-08-10
fix(firmware): wait 10s post-erase everywhere, gate on notifications, detect spliced downloads

skialpine:
## The correlation

Two users failed a v1333 → v1352 flash at the identical point — `Verification failed at block 0.72.0` (FirstError `00 48 00`, address `0x4800`). **Both were on iOS.** A successful reflash captured on an Android tablet against a DE1+/PCB 1.3 supplied the timings below.

The post-erase wait was **10 s on Android and 1 s everywhere else**. Both failures ran the 1 s arm; the success ran the 10 s arm.

That constant was a misreading of de1app, which branches on `$::has_bluetooth`, not on platform (`de1_comms.tcl:941-947`):

```tcl
if {$::has_bluetooth} {
    userdata_append "Erase firmware do: ..." [list de1_comm write "FWMapRequest" $data] 1
    after 10000 write_firmware_now
} else {
    after 1000 write_firmware_now
}
```

The `else` is the no-BLE dry run — a few lines above it, that same branch fakes the connection handles because no machine is attached. **Every real flash de1app performs waits 10 s.** We read `has_bluetooth` as "is Android" and handed iOS, macOS, Windows and Linux the no-machine timing, then documented it as a workaround for an "Android-specific BLE race" that nothing in de1app supports.

1 s is also just too short. The captured flash reported erase-complete at **+1.31 s**, so on iOS the chunk pump started while the bootloader was still erasing.

## Changes

**1 — Post-erase wait is 10 s everywhere, and is now a fallback rather than the trigger.** The chunk pump starts when the erase-complete notification arrives (+1.31 s measured, against a 10 s wait), matching reaprime/decaid. The timer stays because the notification can genuinely fail to arrive — the A009 CCCD subscription is fragile enough that verify has to re-subscribe. `fwToErase=1` ("still erasing") does not trigger the transition.

The completion check deliberately **ignores `FirstError`**: the DE1 echoes back whatever we put in that field on the erase request. We send `0,0,0` and get `0,0,0`. decaid's `_isEraseComplete` asserts `FF FF FF` because `FF FF FF` is what it happens to send — it is testing its own outbound bytes. Only `fwToErase`/`fwToMap` carry machine state there. Verify is different: we send `FF FF FF` and success comes back as `FF FF FD`.

**2 — Verify accepts a verdict only when `WindowIncrement == 0 && fwToMap == 1`.** `WindowIncrement` was decoded and then dropped on the way to `fwMapResponse`, so we accepted the first notification of any shape and reported its `FirstError` as a corrupt-block address. If the bootloader emits progress mid-scan, that turns a healthy flash into a permanent-looking failure that survives Retry — the next attempt reports the same cursor. The signal now carries the field; non-terminal notifications are logged and ignored, with the 60 s verify timeout as the backstop.

**3 — Header validation and splice detection.** `validateFile()` gained reaprime/decaid's structural checks (`ByteCount` non-zero, `0 < CpuBytes <= ByteCount`, reserved word zero, `CheckSum`/`DCSum`/`HeaderChecksum` non-zero, IV not all-zero) as `Validation::MalformedHeader`.

BoardMarker is identical in every DE1 image ever published and sits at offset 4, so BoardMarker plus a size floor accepts a file **spliced from two revisions** — which `downloadIfNeeded()`'s `Range:` resume could produce by appending a new revision's tail onto an old revision's body. Right total length, valid-looking header, corruption found only by the DE1 at the end of a ~18-minute flash.

The splice is caught by `versionMatchesMeta()`: the sidecar records the Version the server reported for the ETag this channel is serving, so a cached or downloaded file whose header disagrees is discarded rather than flashed. A resume is now attempted only when the on-disk bytes are `Truncated` **and** their Version matches — a genuine interrupted download of the same revision.

**4 — The DE1 is put to sleep before erase.** An awake machine runs heaters, PID and refill for the ~18 minutes a flash takes, on the same MCU doing the programming. decaid does this; de1app has the line commented out. It must run before `setFirmwareFlashInProgress(true)`, since `goToSleep()` carries its own guard check and is dropped once the flash is engaged.

## Tests

63 → 75. The synthetic blob builders now populate the checksum words and the IV, because a bare-zero header is correctly rejected.

## What this does not settle

An erase still running when writes begin should corrupt from address 0, so it predicts `0.0.0`, not `0.72.0`. **The correlation with the 1 s wait is exact; the mechanism producing that specific address is not established.** Treat the offset as unexplained rather than solved.

Also unchanged: I deliberately did not adopt decaid's one-write-at-a-time upload. Ours is equivalent in ordering, gives honest ACK-driven progress, and the captured flash pushed 28,992 chunks with zero retries.

Local build and test suite have **not** been run yet.

### DIFF
--- a/docs/CLAUDE_MD/FIRMWARE_UPDATE.md
+++ b/docs/CLAUDE_MD/FIRMWARE_UPDATE.md
@@ -21,9 +21,12 @@ Two buttons:
 
 **Expect ~15–20 minutes on Android** — not 45 seconds. Android BLE write-with-response is serialised at ~30 ms per ACK; 28,992 chunks at that rate is ~16 minutes of upload alone. On desktop platforms BLE is faster (macOS/Linux can hit closer to 2–3 minutes), but the Android tablet is the common case.
 
-- **Phase 1 — Erase**: `FWMapRequest(erase=1, map=1)` sent on characteristic `A009`. DE1 erases the inactive flash bank. Modern firmware sends a single erase-complete notification (`fwToErase=0`). We then wait a fixed 10 s (Android) / 1 s (elsewhere) per de1app's convention before starting the chunk pump.
+- **Phase 0 — Sleep**: the DE1 is put to sleep before the erase command. An awake machine runs its heaters, PID and refill logic for the ~18 minutes a flash takes, on the same MCU that is erasing and programming flash. reaprime/decaid does this; de1app has the line and leaves it commented out. It has to happen *before* `setFirmwareFlashInProgress(true)` — `goToSleep()` carries its own guard check and is dropped once the flash is engaged.
+- **Phase 1 — Erase**: `FWMapRequest(erase=1, map=1)` sent on characteristic `A009`. DE1 erases the inactive flash bank. Modern firmware sends a single erase-complete notification (`fwToErase=0`). The chunk pump starts as soon as that notification lands, with a 10 s fallback on every platform if it never does. Measured at +1.31 s on a DE1+/PCB 1.3, so the notification is the normal path and the timer is the exception. Matches reaprime/decaid; de1app gates on nothing and always waits the fixed 10 s.
 - **Phase 2 — Upload**: 28,992 sequential 20-byte packets streamed to characteristic `A006`. Each packet is `[len=16][addr24 BE][16-byte payload]`. Length byte is 16 (`0x10`) — same format as MMR writes which use length 4. Address is 24-bit **big-endian**. ACK-driven progress so the bar tracks bytes on the wire, not bytes in the BLE queue.
 - **Phase 3 — Verify**: `FWMapRequest(erase=0, map=1, FirstError=FF,FF,FF)` on `A009`. DE1 verifies the inactive bank and replies with a 3-byte `FirstError`. `{0xFF, 0xFF, 0xFD}` = success; anything else = the byte offset of the first corrupt block.
+
+  **Only a notification with `WindowIncrement == 0` and `fwToMap == 1` is a verdict.** Anything else is the bootloader talking mid-scan and its `FirstError` is a cursor, so reading it as a corrupt-block address turns normal progress into a permanent-looking failure that survives Retry. We keep waiting; the 60 s verify timeout is the backstop. reaprime/decaid filters on exactly these two fields.
 - On success, the DE1 remaps the inactive bank as active. It does **not** auto-reboot in practice — on every captured flash (upgrade and downgrade) the machine keeps running the old firmware until the user flips its power switch. Decenza prompts for a power-cycle immediately after verify (see `FirmwareUpdater::AwaitingReboot`) and its auto-reconnect logic re-establishes BLE once the DE1 comes back up on the new firmware.
 
 ### BLE subscription timing (matters)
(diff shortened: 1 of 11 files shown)


skialpine:
### Code review

Found 7 issues. All are fixed in 798095c6; listing them because two were worse than the bug this branch set out to fix.

1. Verify predicate was incomplete — `FirstError == FF FF FF` accepted as a verdict, reporting "Verification failed at block 255.255.255" on a flash that was still verifying. `FF FF FF` is the bootloader's "no error found" value *and* what we write into the request, which the DE1 echoes. decaid's `_isTerminalVerificationResponse` tests four conditions; the comment claimed we matched it while testing two.

https://github.com/Kulitorum/Decenza/blob/08e5d9d761203808fc3eade47af6dcd56947ab88/src/controllers/firmwareupdater.cpp#L830-L845

2. Erase completion could fire on a stale notification. A terminal verify notification is byte-identical to an erase-complete one, and the retry path delivers exactly that: verify times out at 60 s with the DE1 still scanning, user taps Retry, the late response lands in the new Erasing window and streams the upload into a bank still being erased.

https://github.com/Kulitorum/Decenza/blob/08e5d9d761203808fc3eade47af6dcd56947ab88/src/controllers/firmwareupdater.cpp#L798-L812

3. A legitimate download newer than the sidecar was rejected as spliced, unrecoverably. `startUpdate()` reaches `downloadIfNeeded()` with no fresh HEAD, so `m_meta.version` is from the last availability check — 30 s after launch, then weekly. Against the nightly channel a good download is routinely newer. It accused the server of splicing, deleted the file, and left Retry looping on the identical rejection.

https://github.com/Kulitorum/Decenza/blob/08e5d9d761203808fc3eade47af6dcd56947ab88/src/core/firmwareassetcache.cpp#L435-L450

4. The splice defence was bypassed when a server answers a Range request with 200 and the whole body. Nothing checked for 206, and `NoLessSafeRedirectPolicy` makes a header-dropping redirect reachable. The resulting oversized file has the partial's header, so it passes the version check by construction.

https://github.com/Kulitorum/Decenza/blob/08e5d9d761203808fc3eade47af6dcd56947ab88/src/core/firmwareassetcache.cpp#L364-L372

5. Resume could survive a failed delete — the offset was re-derived from the file's size after an unchecked `QFile::remove()`, turning a rejected file back into a resume target (CLAUDE.md: "the compiler only enforces this for *annotated* APIs... so check unannotated results yourself").

https://github.com/Kulitorum/Decenza/blob/08e5d9d761203808fc3eade47af6dcd56947ab88/src/core/firmwareassetcache.cpp#L345-L352

6. `m_meta.version == 0` made `versionMatchesMeta()` vacuously true, so a cache file with a missing or malformed sidecar was both accepted and resumable — the exact state the guard exists to reject.

https://github.com/Kulitorum/Decenza/blob/08e5d9d761203808fc3eade47af6dcd56947ab88/src/core/firmwareassetcache.cpp#L96-L101

7. `goToSleep()` returned void, so the pre-erase sleep was silently skipped when a profile upload was in flight — the deferred retry is dropped by the flash guard engaged immediately after, and the machine flashed fully awake with one DEBUG line as the only trace.

https://github.com/Kulitorum/Decenza/blob/08e5d9d761203808fc3eade47af6dcd56947ab88/src/controllers/firmwareupdater.cpp#L586-L592

Also corrected: three de1app citations were wrong (`has_bluetooth` if/else is 940-950 not 941-947; pre-verify subscribe is 991 not 962; commented-out erase-phase subscribe is 903 not 876), five comments described behaviour the same PR had changed, and the test-count figures in `FIRMWARE_UPDATE.md` were internally inconsistent — removed rather than re-guessed.

Test failure fixed: `simulateFullUpload()` waited on a total write count assuming one non-chunk write precedes the chunks. Sleeping the machine added a second, so it acked one chunk short and the updater sat in Uploading. It now counts chunk writes specifically.

Not changed, deliberately: the git history confirms the "hang in Erasing forever" rationale for the fixed timer was written in the original feature commit before any hardware testing, not after an incident, and the 10 s fallback plus the 30 s erase timeout both remain — so the worst case is identical to the old behaviour, not a hang.

PR #1789, opened 2026-08-11
Brew bar: slide the center column up so the preset band stays visible (#1581 follow-up)

loganross2001:
### The regression (#1581)
When you tap an action preset on the idle screen, the center column expands
downward. Once it reaches the bottom-anchored lower-mid band, #1581 made the
band **fade out to `opacity: 0`** to avoid the overlap. So selecting a preset
silently hides whatever the user had placed in that band.

The band can't simply be pushed down — it sits on `bottomBar.top` (modulo the
user's zone Y-offset), so "down" runs it straight into the bottom action bar.

### The fix
Slide the **column** up instead of hiding the **band** — which is exactly what
the picker popups already do via `bottomPanelClearance`. The band stays visible.

- `_bandOverlap` — the raw, un-offset overlap of the expanded column into the
  band (0 when they don't touch). Read against the band's *static* top so the
  slide can never feed back into its own input (no binding loop).
- `presetBandClearance = min(_bandOverlap, _maxPanelClearance)` — how far the
  column slides, bounded so it never rides up under the status bar.
- The `centerContent` transform slides by `max(bottomPanelClearance,
  presetBandClearance)`, so a picker popup opened *over* an active preset still
  gets its full clearance.
- `carouselOverlapsBand` (the fade) is kept only as a **fallback** for the cases
  the slide can't clear, and its condition accounts for **both** transforms — the
  column's slide *and* the band's own `-bottomPanelClearance` slide — so the
  residual overlap is `_bandOverlap - max(0, presetBandClearance -
  bottomPanelClearance) + topPanelClearance`. It re-engages on a very short
  viewport (the `_maxPanelClearance` cap bites), an upper-half picker (whose
  `topPanelClearance` pushes the column back down), or a lower-half picker open
  over an active preset (band and column ride up together, so the slide can't
  clear the overlap). It's gated on a real active-preset overlap, so a picker
  alone never fades the band.

### Scope
One file (`qml/pages/IdlePage.qml`), no new API, reuses existing clearance
machinery. Behavior is unchanged except that the band no longer disappears when
a preset is active.

### DIFF
--- a/qml/pages/IdlePage.qml
+++ b/qml/pages/IdlePage.qml
@@ -172,16 +172,42 @@ T.Page {
         idlePage.topPanelClearance = 0
     }
 
-    // Center-zone inline carousel: when the expanded center column would reach
-    // the bottom-anchored lower-mid band, the band is HIDDEN (faded out) rather
-    // than shoved down — the band sits on bottomBar.top (modulo the user's zone
-    // Y-offset), so "down" runs it into the bottom action bar. One-way: reads the column's
-    // un-offset bottom against the band's static top.
-    readonly property bool carouselOverlapsBand: {
+    // Center-zone inline carousel: when the expanded center column reaches the
+    // bottom-anchored lower-mid band, slide the COLUMN up so the band stays
+    // visible, rather than hiding the band. The band sits on bottomBar.top
+    // (modulo the user's zone Y-offset) and can't be shoved down without running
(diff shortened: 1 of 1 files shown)


skialpine:
I do not like this change and am removing it. it causes the buttons you are using in the center of the screen to jump around to stop buttons that you are not using from being moved. That's bad UI. It also slows down the usability of the center buttons as they take time to settle in place now instead of being instant.

PR #1791, opened 2026-08-11
Brew bar: Temp Quick-Select layout widget (#1790)

### LINKED ISSUE #1790: Brew bar: Temp Quick-Select pill (the temperature twin of ratio/grind quick-select)
### Summary

The brew bar has quick-select pills for **ratio** (#1455) and **grind** (#1456), but not for **brew temperature**. A Temp Quick-Select pill would complete that trio: tap the current brew temperature to nudge it ± a configurable step, without opening the full temperature editor.

### Motivation

Temperature is one of the three things a user adjusts most between shots (alongside grind and ratio). Today changing it mid-session means opening BrewDialog; a brew-bar pill makes it a one-tap nudge, consistent with how grind and ratio already work.

### Proposed behavior

- A `temperatureQuickSelect` layout widget (same base/pill as the ratio/grind siblings).
- Shows the effective brew temperature (override if armed, else the profile target).
- Tapping opens a small value picker of the current temp ± 5 steps, clamped to the same 70–100 °C range BrewDialog's override editor uses; picking a value writes `temperatureOverride`.
- A global step preference (default 0.5 °C).
- When the profile temperature is outside the override range (tea/calibration profiles), the pill is a plain read-out rather than a button over an empty picker.

### PR

Implemented in the linked PR.

loganross2001:
Closes #1790

# Temp Quick-Select layout widget

Adds a **Temp Quick-Select** brew-bar pill — the temperature twin of the existing
`ratioQuickSelect` and `grindQuickSelect` layout widgets.

## What it does

Shows the effective brew temperature as a pill. The effective value is the brew
temperature override when one is armed (`Settings.brew.temperatureOverride`), otherwise
the active profile's target temperature — the same source `TemperatureItem` and Brew
Settings read. When that temperature is within the quick-select range, tapping the pill
opens a value picker listing the current temperature ± 5 steps (up to 11 entries,
current centred, clamped to the 70–100 °C brew range — the same window `BrewDialog`'s
override editor uses — and de-duplicated). Tapping a value writes it to
`Settings.brew.temperatureOverride`.

When the effective temperature is **outside** that range (e.g. a tea or calibration
profile), the pill is a non-interactive read-out rather than a button over an empty
picker: a single `canQuickSelect` property (does the ladder actually contain the current
temperature) gates the tap, the pill opacity, and the accessibility role/name together,
so the guards can't drift.

Values are stepped and stored in Celsius (the internal unit); the pill and the picker
display them in the user's unit via `Theme.formatTemperature`, the same helper
`TemperatureItem` uses. The pill highlights (via `Theme.highlightColor`) when the
session's temperature is a real override of the active baseline
(`MainController.temperatureIsRealOverride`), exactly like the ratio pill's override
highlight.

## The step setting

The picker step is a new global preference `Settings.brew.temperatureQuickSelectStep`
(double, default 0.5 °C), declared in `SettingsBrew` mirroring `doseCupTareWeight` in
every spot: `Q_PROPERTY` + getter/setter/NOTIFY in `settings_brew.h`, QSettings-backed
impl in `settings_brew.cpp` (key `espresso/temperatureQuickSelectStep`), and both the
export and import blocks of `settingsserializer.cpp`. The setter clamps the step to a
positive range (0.1–5.0 °C) so 5 steps stay inside a usable window and a non-positive
step can't collapse every picker row onto the current value.

A global preference is the right fit because neither sibling pill has a matching
per-instance "step": the ratio pill uses three fixed global presets
(`ratioPreset1/2/3`) and the grind pill derives its step from shot history. There is no
per-instance step option, matching the siblings (neither declares readout option keys).

## The picker

Ratio and grind each ship their own picker file (`RatioPresetDialog`,
`GrindPickerDialog`), so this adds `TempPickerDialog.qml` — a plain value-list picker
modelled on the grind picker's value-picker role (not the ratio preset editor). It takes
the item's computed `rows` and emits the tapped value in Celsius.

## Registration (mirrors ratio/grind)

- `qml/components/layout/items/TempQuickSelectItem.qml` — new widget (base
  `LayoutWidgetItem`, the same upstream base both siblings use)
- `qml/components/TempPickerDialog.qml` — new picker
- `CMakeLists.txt` — both new QML files added to the `qt_add_qml_module` file list
- `qml/components/layout/LayoutItemDelegate.qml` — `case "temperatureQuickSelect"` arm
- `src/core/settings_network.cpp` — `widgetCatalogTable()` row (id
  `temperatureQuickSelect`, palette "Temp Quick-Select", chip "Temp")

The two brew-bar **default-preset** definitions (`ZoneOptionsPopup.populateBrewBar`,
`shotserver_layout.cpp` `brewBar`) were left untouched — those seed a curated 5-widget
default arrangement, and `grindQuickSelect` isn't in them either, so adding the widget
to the catalog (where users pick it) is the correct parity, not editing the default
layout.

No settings control for the step was added: the sibling steps are edited inline
(ratio) or derived (grind), there is no central brew-settings tab hosting a comparable
global step, and the property has a sane default and is persisted via the settings API.
(This UI control was marked optional.)

## Conventions

All visible strings go through `TranslationManager.translate` / `Tr`; styling is all
`Theme.*` (no hardcoded colours/sizes); accessibility roles/names on interactive
elements; `pragma ComponentBehavior: Bound` on the picker with `required property` on
the delegate role.

## Verification

- qmllint (with the built module import root): both new files clean, zero warnings —
  identical to the upstream sibling under the same invocation.
- `cmake --build build/pr --target Decenza`: succeeded, app binary produced, zero
  `error:` / `FAILED:` lines in the build log.
- `tst_settings`: added one slot, `temperatureQuickSelectStepRoundTripAndClampsImport`
  — round-trips the step through export→import (catches a typo on either hand-mirrored
  serializer key) and asserts both the upper (99→5.0) and lower (0.001→0.1) clamps.
  Full suite: 170 passed, 0 failed.

## Follow-up

This is a new user-visible brew-bar widget, so it wants a short Manual (wiki) entry —
happy to add one if you'd like it in this PR or as a follow-up.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1351,6 +1351,7 @@ set(QML_FILES
     qml/components/SteamGraph.qml
     qml/components/StatusBar.qml
     qml/components/RatioPresetDialog.qml
+    qml/components/TempPickerDialog.qml
     qml/components/GrindField.qml
     qml/components/GrindPickerDialog.qml
     qml/components/GrindRowSource.qml
@@ -1477,6 +1478,7 @@ set(QML_FILES
     qml/components/layout/items/MilkWeightItem.qml
     qml/components/layout/items/RatioQuickSelectItem.qml
     qml/components/layout/items/GrindQuickSelectItem.qml
+    qml/components/layout/items/TempQuickSelectItem.qml
     qml/components/layout/items/ShotPlanItem.qml
     qml/components/layout/items/SpacerItem.qml
     qml/components/layout/LayoutActions.qml
--- a/qml/components/TempPickerDialog.qml
+++ b/qml/components/TempPickerDialog.qml
@@ -1,0 +1,184 @@
+// The value-row delegate declares its injected model role required, so Bound
+// cannot break role injection here and this file's ids resolve statically.
+pragma ComponentBehavior: Bound
+
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+import Decenza
+
+// Value picker opened from the temperatureQuickSelect layout widget. Lists the
+// brew temperatures the widget generated (current ±5 steps, clamped to the brew
+// range) and applies the tapped value by emitting its Celsius value. A plain
+// VALUE picker — not a preset editor; the value-picker sibling of
+// GrindPickerDialog (a plain tap-a-row list rather than Tumbler wheels, no shared
+// code).
+//
+// The list (`rows`) is computed by the item; each row is
+// { value: <Celsius double>, label: <display string>, isCurrent: bool },
+// ordered cool -> hot (ascending Celsius). The emitted value is CELSIUS (the
+// internal/stored unit); labels are already in the user's display unit.
+//
+// Roots at DecenzaDialog (the shared dialog base every dialog in the app uses),
+// not QtQuick.Controls Dialog: the app base keeps AOT compilation and the app's
+// dialog theming (dim, enter/exit transitions). See DecenzaDialog.qml.
+DecenzaDialog {
+    id: root
+    parent: Overlay.overlay
+    anchors.centerIn: parent
+    width: Math.min(Theme.scaled(360), parent ? parent.width * 0.95 : Theme.scaled(360))
+    height: Math.min(contentCol.implicitHeight + Theme.scaled(40),
+                     parent ? parent.height * 0.92 : Theme.scaled(640))
+    modal: true
+    closePolicy: Dialog.CloseOnEscape | Dialog.CloseOnPressOutside
+    padding: 0
+
+    // [{ value: double (Celsius), label: string, isCurrent: bool }] — cool -> hot.
+    property var rows: []
+
+    // Emits the picked temperature in CELSIUS (the internal/stored unit).
+    signal valuePicked(double value)
+
+    background: Rectangle {
+        color: Theme.dialogBackgroundColor
+        radius: Theme.cardRadius
+        border.width: 1
+        border.color: Theme.borderColor
+    }
+
+    contentItem: Flickable {
+        contentWidth: width
+        contentHeight: contentCol.implicitHeight
+        clip: true
+        boundsBehavior: Flickable.StopAtBounds
+        ScrollBar.vertical: ScrollBar { policy: ScrollBar.AsNeeded }
+
+        ColumnLayout {
+            id: contentCol
+            width: parent.width
+            spacing: Theme.spacingSmall
+
+            // --- Header ---
+            ColumnLayout {
+                Layout.fillWidth: true
+                Layout.topMargin: Theme.spacingLarge
+                Layout.leftMargin: Theme.spacingLarge
+                Layout.rightMargin: Theme.spacingLarge
+                spacing: Theme.scaled(2)
+                Text {
+                    text: TranslationManager.translate("temp.picker.title", "Brew Temperature")
+                    color: Theme.textColor
+                    font.pixelSize: Theme.scaled(24); font.bold: true
+                }
+                Text {
+                    text: TranslationManager.translate("temp.picker.subtitle", "Tap a value to set the brew temperature")
+                    color: Theme.textSecondaryColor
+                    font: Theme.labelFont
+                }
+            }
+
+            // --- Cooler end label (top: rows ascend, so lowest = coolest is first). ---
+            Text {
+                visible: root.rows.length > 2
+                Layout.fillWidth: true
+                Layout.leftMargin: Theme.spacingLarge
+                Layout.rightMargin: Theme.spacingLarge
+                horizontalAlignment: Text.AlignHCenter
+                text: TranslationManager.translate("temp.picker.cooler", "Cooler").toUpperCase()
+                color: Theme.textSecondaryColor
+                font: Theme.captionFont
+            }
+
+            // --- Value rows ---
+            Repeater {
+                model: root.rows
+                delegate: Rectangle {
+                    id: rowRect
+                    required property var modelData
+                    readonly property bool isCurrent: modelData.isCurrent === true
+                    Layout.fillWidth: true
+                    Layout.leftMargin: Theme.spacingLarge
+                    Layout.rightMargin: Theme.spacingLarge
+                    Layout.preferredHeight: Theme.scaled(48)
+                    radius: Theme.buttonRadius
+                    color: rowMa.pressed ? Qt.darker(Theme.backgroundColor, 1.1) : Theme.backgroundColor
+                    border.width: isCurrent ? 2 : 1
+                    border.color: isCurrent ? Theme.primaryColor : Theme.borderColor
+
+                    Accessible.role: Accessible.Button
+                    Accessible.name: String(modelData.label)
+                                     + (isCurrent ? ", " + TranslationManager.translate("temp.picker.current", "current") : "")
+                    Accessible.focusable: true
+                    Accessible.onPressAction: rowMa.clicked(null)
+
+                    RowLayout {
+                        anchors.fill: parent
+                        anchors.leftMargin: Theme.spacingMedium
+                        anchors.rightMargin: Theme.spacingMedium
+                        spacing: Theme.spacingSmall
+                        Text {
+                            text: String(rowRect.modelData.label)
+                            color: rowRect.isCurrent ? Theme.primaryColor : Theme.textColor
+                            font.pixelSize: Theme.scaled(20)
+                            font.bold: rowRect.isCurrent
+                            Accessible.ignored: true   // the delegate Rectangle carries Accessible.name
+                        }
+                        Item { Layout.fillWidth: true }
+                        Text {
+                            visible: rowRect.isCurrent
+                            text: TranslationManager.translate("temp.picker.current", "current").toUpperCase()
+                            color: Theme.primaryColor
+                            font: Theme.captionFont
+                            Accessible.ignored: true   // the delegate Rectangle carries Accessible.name
+                        }
+                    }
+
+                    MouseArea {
+                        id: rowMa
+                        anchors.fill: parent
+                        onClicked: {
+                            root.valuePicked(rowRect.modelData.value)
+                            root.close()
+                        }
+                    }
+                }
+            }
+
+            // --- Warmer end label (bottom: highest = hottest). ---
+            Text {
+                visible: root.rows.length > 2
+                Layout.fillWidth: true
+                Layout.leftMargin: Theme.spacingLarge
+                Layout.rightMargin: Theme.spacingLarge
+                horizontalAlignment: Text.AlignHCenter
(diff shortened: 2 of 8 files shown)


loganross2001:
Per CLAUDE.md (user-visible feature → Manual entry), here's a short **wiki Manual** draft you can drop in on merge (kept to the "what it is / where to find it / the one non-obvious rule" shape):

---

### Temp Quick-Select

A brew-bar pill showing the current brew temperature. Tap it to pick a nearby temperature (the current value ± 5 steps); your choice becomes the brew-temperature override, the same one the temperature editor sets. Add it from the layout editor's widget palette, like the Ratio and Grind quick-select pills.

The picker only covers the 70–100 °C espresso range — for a profile set outside that (a tea or calibration profile) the pill is a read-only display rather than a button.

---

Happy to open a wiki PR for this myself if you'd prefer.

skialpine:
### Code review

Found 8 issues:

1. **The picked temperature never reaches the machine.** `applyValueC` writes only the setting; the override is applied to the DE1 exclusively inside `ProfileManager::uploadCurrentProfile()`, which shifts the frames (`framesShiftedToTemperature`) and pushes the group temp via `setShotSettings`. Every existing writer pairs the write with an upload — `activateBrewWithOverrides` ([profilemanager.cpp:616-634](https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/src/controllers/profilemanager.cpp#L615-L635)), restore-from-shot (maincontroller.cpp:1086), recipe activation (maincontroller.cpp:2179). The only connection to `temperatureOverrideChanged` is maincontroller.cpp:789 → `brewBaselineChanged`, which is UI-only. So the pill, `TemperatureItem` and the Shot Plan all show 94.5 °C while the machine still holds the 92 °C frames, and the shot record stores a temperature that was never brewed.

https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/qml/components/layout/items/TempQuickSelectItem.qml#L81-L88

2. **Two widgets cannot both be called "Temp" in the layout editor.** The new catalog row reuses the chip fallback `"Temp"`, already held by the pre-existing `temperature` readout at line 954 — both `inPalette: true`. That chip is what the layout editor's zone list, the library card and the web editor all render (`Settings.network.widgetChipNames()`, [LayoutEditorZone.qml:238-268](https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/qml/pages/settings/LayoutEditorZone.qml#L237-L269)), and it also feeds `"Edit %1 options"` at LayoutEditorZone.qml:324 — so a screen reader gets two identical buttons too. The one other collision in the table (`connectionStatus`/`machineStatus` → "Machine") is a commented legacy alias with `inPalette: false`. `widgetCatalogInvariants` asserts type uniqueness but never chip-fallback uniqueness, so nothing catches it. Give the new row a distinct chip, e.g. `"Temp Quick"`, matching how `Ratio`/`Grind` stay distinguishable.

https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/src/core/settings_network.cpp#L953-L965

3. **`TempPickerDialog` roots at `QtQuick.Controls` `Dialog` — the only file in the tree that does.** `DecenzaDialog.qml:5` is "the base every dialog and popup in the app roots at", and [BUILD_PERFORMANCE.md](https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/docs/CLAUDE_MD/BUILD_PERFORMANCE.md#L266-L272) records why: a Controls `Dialog` resolves to the style's composite, whose base chain `qmlcachegen` cannot walk, so every `root.<prop>` in the file loses AOT — the reason #1717 converted 131 sites. This file has 6 such references. `RatioPresetDialog.qml`, the file it mirrors, is a `DecenzaDialog`. Side effects: the modal dim comes from the style rather than `Theme.dialogDimColor`, and the background uses `Theme.surfaceColor` where the sibling uses the dialog token.

https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/qml/components/TempPickerDialog.qml#L20-L22

4. **Child `Text` inside `Accessible.name`-bearing parents missing `Accessible.ignored: true`**, in three places (`ACCESSIBILITY.md:232`: "TalkBack announces button name AND text content, doubling the announcement"; `:236` extends it to delegates that summarise content in `Accessible.name`). The row label and the "CURRENT" tag in the picker delegate, the Close button's `Text`, and the pill's `tempValue`. `ACCESSIBILITY.md:260` also requires fixing this in files you touch — the sibling `RatioQuickSelectItem` has the same gap, while `ClockItem`/`BeansItem`/`BatteryLevelItem` all apply the rule.

https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/qml/components/TempPickerDialog.qml#L115-L125

5. **The `rows` comment asserts Qt behaviour that is wrong, and the file contradicts it.** `Theme.formatTemperature` is a plain QML JS function whose `tempIsFahrenheit()` reads `Settings.app.temperatureUnit` in JS, and [Theme.qml:407-411](https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/qml/Theme.qml#L406-L412) states plainly that "that read is what makes bindings re-evaluate on a unit toggle (property reads inside C++ **methods** are invisible)". The `void(Settings.app.temperatureUnit)` sites (BrewDialog.qml:955, ShotPlanText.qml:177) exist because they call the C++ `ProfileManager.temperatureDisplay()`. `TemperatureItem.qml:60,113` calls `formatTemperature` in bindings with no workaround — and so does `valueText` eleven lines above this comment, so if the premise held the pill would be stale while the picker was not. CLAUDE.md: "cite file-and-line in any comment that asserts what Qt does. An un-sourced claim in a comment gets believed and then licenses wrong code."

https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/qml/components/layout/items/TempQuickSelectItem.qml#L57-L62

6. **The pill skips the "equals baseline ⇒ clear" rule and can never disarm.** `setTemperatureOverride` arms `m_hasTemperatureOverride` unconditionally (settings_brew.cpp:1246-1254); the canonical writer deliberately calls `clearTemperatureOverride()` within 0.1 °C of the profile temperature ([profilemanager.cpp:616-619](https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/src/controllers/profilemanager.cpp#L615-L620)). Tapping the row marked *current* therefore arms an override at the baseline, persisted to disk, and the ladder holds only temperatures — so the only exits are Brew Settings' Clear or a profile load. `temperatureIsRealOverride`'s 0.1 tolerance keeps the highlight off, which makes the armed state latent rather than visible.

https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/qml/components/layout/items/TempQuickSelectItem.qml#L81-L88

7. **`temperatureQuickSelectStep` is unreachable — nothing writes it.** Sole reader is `tempStepC`; there is no QML control, no `settings_set` coverage, no web surface, so the step is permanently 0.5 °C and the clamp plus half the new test cover values only a hand-edited file or a migration import can produce. `doseCupTareWeight`, the property this mirrors, has an editor at BrewDialog.qml:1077. The archived change `2026-07-17-grind-widget-observed-step` **deleted** the sibling global `grindQuickSelectStep` rather than leave it stranded. CLAUDE.md: "Write the benefit as 'X, which the user experiences as Y' … if Y is empty, so is the case for it."

https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/src/core/settings_brew.h#L30-L34

8. **No wiki manual entry or task for a new user-visible widget** (root [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/909a1d794e04dd0fb3c14fadc7661e7a294289a9/CLAUDE.md#L13-L15): "update the wiki manual as part of that work — add a task for it in the change's `tasks.md` … Don't leave it as an afterthought"). The PR defers it to an optional follow-up; a draft already sits in this PR's comments. The wiki being a separate repo explains its absence from the diff, not the absence of a tracked task.

<details>

- [Worth a look] No `openspec/specs/layout-brew-widgets/spec.md` requirement block for the new widget, though the spec carries one for each of the six existing brew-bar widgets including both quick-select siblings, and #1762/#1763 shipped their spec changes in-PR. CLAUDE.md never states this as a hard requirement and specs can be synced separately. (score: 65)

</details>

skialpine:
One more comment, The app has moved to all adjusted temps being 0 based where 0 is the temp in the profile or recipe and you adjust from there. There is a ladder for how the right temp is picked for the shot. Make sure that this widget is just adjusting the same thing as the brew settings page. Also I think it is a little confusing that you are using temp here when the rest of the app is using an adjust from the baseline model but probably can live with it in an optional widget. Maybe you could have the widget have an option for which way it displays temp 0 based adjustment and relative temp.

loganross2001:
Thanks for the thorough review — all 8 addressed (force-pushed):

1. **Reaches the machine now.** Added `Q_INVOKABLE ProfileManager::applyTemperatureOverride(double)` — the temperature slice of `activateBrewWithOverrides` (set-or-clear vs `m_currentProfile.espressoTemperature()`, then `uploadCurrentProfile()`). The pill calls that instead of writing the setting raw.
6. **Disarms at baseline** (folded into 1): tapping the row equal to the profile temp hits `qAbs(baseline−baseline) ≤ 0.1` → `clearTemperatureOverride()`, so it can't pin a redundant override.
2. **Chip** fallback `Temp` → `Temp Quick` (key was already `chipTempQuick`).
3. **Picker** re-rooted `QtQuick.Controls Dialog` → `DecenzaDialog`, `Theme.surfaceColor` → `Theme.dialogBackgroundColor`, mirroring `RatioPresetDialog` (AOT + dialog theming restored).
4. **A11y** `Accessible.ignored: true` on the picker label + CURRENT tag, the Close text, the pill value — and the sibling `RatioQuickSelectItem` per ACCESSIBILITY.md:260.
5. **Comment rot / crutch** removed — `Theme.formatTemperature` reads the unit in JS so the bindings are already reactive (Theme.qml:409-411); dropped the `translationVersion`-style read and the wrong claim.
7. **Stranded `temperatureQuickSelectStep` removed** end-to-end (property/impl/serializer/test); step is now a fixed 0.5 °C.
8. Manual draft is in the comment above; wiki is a separate repo so it's not in the diff.

On the 0-based vs absolute display: agreed it's the odd one out. Left it absolute for this pass (you noted it's livable in an optional widget); happy to add a per-instance "absolute / adjust-from-baseline" display option as a follow-up if you'd want it in.

Build 781/781, qmllint 0 on the widget/picker, tst_settings 169/0.

PR #1808, opened 2026-08-11
fix(saw): quarantine a corrupt SAW blob instead of destroying it

skialpine:
## The bug

`loadPerProfileSawHistoryMap` / `loadPerProfileSawBatchMap` reset a bad JSON blob to `"{}"` and logged *"per-profile history lost"* — wording that presents the loss as unavoidable while overwriting the only copy of the bytes. Truncation from a partly flushed write is precisely the case where those bytes are partly salvageable.

**Second half, and why this is the same severity class as the seed paths fixed in #1807:** the pre-basket basket seed would then run against the emptied store, find nothing to copy, and close its one-way `saw/basketKeyMigrated` flag — turning a recoverable corruption into a permanent one.

## What changed

- The raw blob is copied to `<key>.corrupt` with an ISO timestamp before the reset. The **first** capture is kept: a later parse failure is reading the `"{}"` we wrote, so the earliest is nearest real data.
- A session latch defers the seed when the store was reset this way, so it cannot close over an emptied store.
- `resetSawLearning()` clears the latch — it discards the buckets deliberately, so there is nothing left to protect, and a stuck latch would block the seed for the rest of the session over data the user threw away.
- The two loaders were near-identical copies, so the quarantine went into one shared `loadSawMap()` rather than being added to one and not the other.

## Test

`aCorruptHistoryBlobIsQuarantinedAndDoesNotCloseTheSeed` seeds a truncated blob, asserts the raw bytes survive under the quarantine key with a timestamp, and asserts the seed leaves its flag open. Fails on both halves without this change.

Writing it also surfaced the missing latch reset: the latch leaked between test slots because `Settings` is shared, which is the same gap a user would hit after a deliberate reset.

## Note on process

This was found in the #1807 review and I deferred it as "pre-existing". That was the wrong call — CLAUDE.md is explicit that pre-existing issues in files you touch get fixed rather than excused, and the deferral was raised as such.

Full suite: **111 passed, 0 failed**.


skialpine:
### Code review

Found 4 issues. All four are fixed in the follow-up commit on this branch; listing them because each was a real hole in the guard this PR exists to provide.

1. The corruption latch was session-scoped while the reset it guards is persisted, so the seed closed its one-way flag on the *next* launch. Restoring bytes can only happen between runs; by then the store parses clean, the latch is false, `preBasketCount` is 0, and every other guard passes over an empty store. Gate is now the persisted quarantine key.

https://github.com/Kulitorum/Decenza/blob/398625b4c04e2e690d883d6ecd67e4d7f06a324e/src/core/settings_calibration.cpp#L1279-L1292

2. The gate sat *before* the batch map was read, so batch-only corruption quarantined the blob and then closed the flag in the same call. `loadPerProfileSawBatchMap()` runs later in the seed than the history load. Both maps are now read before the gate.

https://github.com/Kulitorum/Decenza/blob/398625b4c04e2e690d883d6ecd67e4d7f06a324e/src/core/settings_calibration.cpp#L1325-L1329

3. Keep-first was the wrong rule and its stated reason was false: `{}` is valid JSON, so a later parse failure is never re-reading what we wrote — it is fresh data that got corrupted. Keep-first would retain a stale blob and overwrite the valuable one, which is the destruction this change exists to prevent. Newest capture wins now.

https://github.com/Kulitorum/Decenza/blob/398625b4c04e2e690d883d6ecd67e4d7f06a324e/src/core/settings_calibration.cpp#L1832-L1836

4. The quarantine had no reader: `sawLearningExport()` omitted it, so a device transfer taken after a corruption migrated the emptied store and left the only copy behind — and since the quarantine is also the gate, the new device would close the seed the transfer was meant to preserve learning across. Now carried by export/import, and dropped by `resetSawLearning()` (which also makes its "wipes EVERY saw/* key" comment true again).

https://github.com/Kulitorum/Decenza/blob/398625b4c04e2e690d883d6ecd67e4d7f06a324e/src/core/settings_calibration.cpp#L647-L652

<details>

- [OK to skip] `isEmpty()` check misread literally-empty corrupt bytes as an unused quarantine slot — moot, keep-first is gone (score: 30)
- [OK to skip] `sawLearningImport()` did not clear the session latch — moot, the latch is gone (score: 50)

</details>

Verification: build clean, full suite 111 passed / 0 failed. Three new tests, each failing without its corresponding fix.

skialpine:
### Live verification on a real store

Ran this against the actual QSettings store (`com.decentespresso.Decenza`, 11755-byte history / 4041-byte batch), backed up and restored byte-identical afterwards (`sha256 ab2d4d876b993177`, no leftover keys).

**1. Quarantine keeps the bytes.** Truncated the history blob to 5877 bytes (mid-token, so a guaranteed parse error) and launched:

```
WARN [SAW][Learning] Corrupt saw/perProfileHistory JSON: missing name separator
     — 5877 bytes quarantined at saw/perProfileHistory.corrupt, store reset
WARN [SAW][Learning] Basket seed deferred: a corrupt SAW blob is quarantined and no
     pre-basket bucket is present to copy — ...
```

Quarantined bytes are byte-identical to what was there (`sha256 d9dd1d7723326923`), and one appended `}]}` recovers **7 of 9 buckets / 34 medians**. On the old code those were gone.

**2. The gate survives a relaunch — the reason this PR exists.** Relaunched with nothing changed. Store now parses clean (`{}`), so the old session latch would have been false and the flag would have closed over an empty store. It deferred again, `saw/basketKeyMigrated` still absent, `corruptAt` unchanged (no spurious re-quarantine).

**3. The gate releases when the bytes come back.** Restored the full blob, left the quarantine in place: corruption gate silent, seed proceeded.

### One further defect found by running it

Step 3 surfaced a second problem in the same guard, now fixed in 4b11853d. "Created nothing" was reported as a key-derivation failure in every case — but an already-per-basket store also creates nothing, and that state is reachable in production: `sawLearningImport()` clears the flag unconditionally ([settings_calibration.cpp#L671-L679](https://github.com/Kulitorum/Decenza/blob/4aecc317d2b2cd54e6da49bd3ba2a5c60c2b18e5/src/core/settings_calibration.cpp#L671-L679)) and a donor device is normally already seeded, so **every launch after a device transfer** warned `treating as a key-derivation failure` and never closed the flag.

Separated by whether any pre-basket bucket matched the window: `matched == 0` is the real derivation failure and still warns; `matched > 0` with every target present closes the seed. Verified on the real store — cleared the flag on a seeded store and launched:

```
INFO [SAW][Learning] Basket seed (complete): 18 profile(s) in window,
     created 0 history and 0 pending bucket(s), 0 pre-basket bucket(s) had no basket in the window
```

No WARN, flag closed, history unchanged at 11755 bytes.

Suite 111/111, 0 warnings.

PR #1809, opened 2026-08-13
fix(inventory): make the whole bag/equipment card tappable (#1798)

### LINKED ISSUE #1798: [Bug] tap tap in the box
### What happened?

Hey Jeff,
small thing.. the tap to select the specific recipe only works if you tap in line with a line of text, or on the text, but not in an empty space within the box. see video  

### Steps to reproduce

https://github.com/user-attachments/assets/2802bc08-39a0-4db5-9ee8-1943a71770a1

### Decenza version

2.02

### Platform

Android

### Device / OS version

12

### Scale (if applicable)

_No response_

_No response_

_No response_

### Screenshots or video

_No response_

### Anything else?

_No response_

skialpine:
Fixes #1798.

## The bug

On the Bag Inventory page, the tap-to-select `AccessibleMouseArea` wrapped only the info row (thumbnail + text lines), not the card. Taps in the action row's whitespace or the card's bottom padding hit nothing, so selection appeared to work "only if you tap in line with a line of text".

## The fix

Move the tap area to the card root — `anchors.fill: parent`, `z: -1` — which is the pattern `RecipesPage`'s `RecipeCard` already uses. The whole card selects; the buttons and thumbnail render above the area and still win their own taps. The `Item` wrapper that existed only to host the old area is removed, along with two ids nothing referenced.

`EquipmentCard` mirrors `BagCard` and carried the identical dead-space bug — fixed in the same pass rather than left to be reported separately.

## Testing

Builds clean (0 errors). QML behaviour is runtime-only — needs a tap on device: card whitespace selects the bag, and Find in Bean Base / Bag finished / Edit / Thaw / Mark Opened / delete still do their own thing rather than selecting.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1382,6 +1382,7 @@ set(QML_FILES
     qml/components/BeanSummary.qml
     qml/components/LinkBeanBaseDialog.qml
     qml/components/ChangeBeansDialog.qml
+    qml/components/CardTapArea.qml
     qml/components/BagCard.qml
     qml/components/SwitchEquipmentDialog.qml
     qml/components/EquipmentCard.qml
--- a/qml/components/BagCard.qml
+++ b/qml/components/BagCard.qml
@@ -247,6 +247,21 @@ Rectangle {
         }
     }
 
+    // Tap anywhere on the card to select the bag (#1798: the tap area used to
+    // wrap only the info row, so the action row's whitespace and the card's
+    // padding were dead space). z: -1 keeps every button and the thumbnail
+    // above it, so they still win their own taps — the RecipeCard pattern.
+    AccessibleMouseArea {
+        anchors.fill: parent
+        z: -1
+        accessibleName: card.accessibleSummary
+        accessibleItem: card
+        onAccessibleClicked: {
+            if (card.bag && card.bag.id !== undefined)
+                Settings.dye.activeBagId = card.bag.id
+        }
+    }
+
     Timer {
         id: deleteRefusedTimer
         interval: 4000  // UI auto-dismiss (allowed timer use)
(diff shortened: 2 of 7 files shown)


skialpine:
### Code review

Found 4 issues:

1. The comment claims the thumbnail "wins its own taps", but `BeanThumbnail` has no `MouseArea` or `TapHandler` at all — a tap on the photo falls through to the card-wide area and selects the bag. CLAUDE.md says an un-sourced claim in a comment "gets believed and then licenses wrong code".

https://github.com/Kulitorum/Decenza/blob/61f8ebf959ed0fdd80b71444c41336aafb4c9c27/qml/components/BagCard.qml#L250-L255

2. `EquipmentCard`'s header says the card is "informational + edit/remove (no global \"selected\" state)", which the code contradicts — `selected` reads `Settings.dye.activeEquipmentId` and the tap handler calls `Settings.dye.switchToEquipment()`. Pre-existing, but this PR adds a comment 70 lines below saying the tap area "switches to this package" and widens the target that exercises it (CLAUDE.md: "Fix pre-existing violations in any file you touch").

https://github.com/Kulitorum/Decenza/blob/61f8ebf959ed0fdd80b71444c41336aafb4c9c27/qml/components/EquipmentCard.qml#L6-L11

3. The two mirrored files point opposite directions for the same structure: `BagCard` says the tap area is "below" when it is declared at L254, above this line; `EquipmentCard` says "above" and is right.

https://github.com/Kulitorum/Decenza/blob/61f8ebf959ed0fdd80b71444c41336aafb4c9c27/qml/components/BagCard.qml#L289-L292

4. This is the third and fourth hand-copy of the same `anchors.fill: parent` + `z: -1` tap-area glue (the others are `RecipesPage.qml:474-483` and `RecipeWizardPage.qml:2347-2349`). CLAUDE.md: "When you touch code that already has copies, collapse them in that pass" — a shared component taking `accessibleName`, `accessibleItem` and an optional `enabled` would stop a future card silently omitting `z: -1` and swallowing its own buttons.

https://github.com/Kulitorum/Decenza/blob/61f8ebf959ed0fdd80b71444c41336aafb4c9c27/qml/components/BagCard.qml#L249-L264

Checked and cleared: `z: -1` hit-testing (the root `Rectangle` accepts no buttons, so it is never a pointer target; buttons at default z consume the press first), flick-stealing inside the host `Flickable` (`AccessibleMouseArea` acts on release and has an `onCanceled` handler, so a drag cannot select), the deleted wrapper `Item`'s `implicitHeight` (the card's own `implicitWidth`/host-assigned `width` make it irrelevant), and screen-reader discovery of the action buttons.

PR #1814, opened 2026-08-14
fix(shot-analysis): confirm a frame exit whose threshold crossed between BLE samples (#1813)

### LINKED ISSUE #1813: [Bug] First Step Skipped?  Firmware v1352
### What happened?

Updated firmware to v1352.  Running Decenza v2.0.2   D-Flow shot reported "First step skipped".  Screen shot below, and visualizer link.  

### Steps to reproduce

D-Flow

### Decenza version

2.0.2 Build 3531 iOS

### Platform

iOS

### Device / OS version

iPad mini (A17 Pro)

### Scale (if applicable)

_No response_

[debug.zip](https://github.com/user-attachments/files/31081017/debug.zip)

[shot338.json](https://github.com/user-attachments/files/31081024/shot338.json)

https://visualizer.coffee/shots/630cf262-9f31-4e1a-b6b1-ccdfb2ed0037

### Screenshots or video

Image
Image

### Anything else?

Recently update firmware to v1352.  Also saw issue previously with firmware v1333.

skialpine:
Fixes #1813.

## The report

A D-Flow shot on firmware v1352 was badged **"First step skipped"**. The shot was fine — frame 0 executed exactly as the profile asked.

Frame 0 ("Filling") is configured `exit_type pressure_over`, `exit_pressure_over 2.10`, `seconds 25.00`. From the attached `shot338.json`:

```
[12:34:28.838] EXTRACTION STARTED at frame 0
[12:34:30.726] MainController: Frame 0 exit reason unconfirmed - exitType: "pressure_over" pressure: 2.0481 flow: 7.71045 recorded as "pressure_unconfirmed"
```

| t (s) | P (bar) |
|-------|---------|
| 1.465 | 1.673 |
| 1.675 | 1.869 |
| **1.888** | **2.048**  ← transition |
| 2.071 | 2.228 |

Pressure was rising ~0.84 bar/s. The 2.10 bar threshold was reached at roughly **1.95 s** — about 60 ms after the last sample the app could see. The shot then ran normally to 19.8 g in 36.9 s.

## Root cause

`MainController` compared the transition sample against the threshold with **no tolerance** (`m_lastPressure >= prevFrame.exitPressureOver`), so it recorded `pressure_unconfirmed`. `ShotAnalysis::detectSkipFirstFrame` deliberately trusts only confirmed `pressure`/`flow`/`weight`, so it fell through to its duration branch: `min(2.0, 0.5 × 25.0)` = 2.0 s cutoff, actual 1.888 s → flag.

The DE1 evaluates its exit condition on its own internal cadence, an order of magnitude faster than the ~10 Hz BLE stream. A fast-rising fill frame crossing its threshold between two samples is the normal case, not an ambiguity — so that comparison could never confirm exactly the frames that exit fastest.

## The fix

Frame-exit inference confirms when the threshold falls **within one sample-interval's worth of the observed change toward it**. The tolerance is the measured per-sample delta, never a constant:

- flat → no tolerance
- moving away from the threshold → no tolerance
- no preceding sample in the shot → no tolerance

That last one matters: a genuinely skipped frame has no rise to extrapolate from, so it still lands in the unconfirmed branch, which is what the skip detector exists to catch.

The inference moves out of `MainController::onShotSample`'s frame-change block into a header-only `FrameExit::inferReason` (`src/machine/frameexitreason.h`) so it is unit-testable. Vocabulary and precedence (weight → sensor → time → unconfirmed) are unchanged.

## Logging

Three lines, all naming the firmware build — this is reported against specific builds (#1813 cites v1333 and v1352) and a submitted log otherwise leaves the reader guessing which one produced it:

- **WARN** when the machine opens extraction at a non-zero frame — frame 0 never reported at all. The genuine firmware skip, observed live.
- **WARN** when frame 0 ends unconfirmed under the detector's own cutoff, i.e. the badge is about to fire. The cutoff is mirrored from `detectSkipFirstFrame`'s short-first-step branch so the line predicts the badge rather than approximating it. Carries thresholds, both samples, and configured seconds.
- The existing unconfirmed line now carries the preceding sample, and confirmed-by-extrapolation gets its own sibling line — so the log says which path a marker took and with what numbers.

## Scope

Existing shots keep their persisted reason strings, so their badges do **not** change on recompute. Only new shots benefit. No BLE/protocol change, no schema change, no migration.

## Tests

New `frameExitReasonInference` slot in `tests/tst_shotanalysis.cpp` (existing file — no new test target): #1813's exact numbers, one-sample-short, flat, falling, no-preceding-sample, both `_under` arms including the zero-sensor guard, the flow arms, and precedence (weight / time expiry / no exit condition). Plus the marker that shot now produces asserted through `detectSkipFirstFrame` as not flagging.

Verified by refusal, not by a green run: with the extrapolation disabled in both `reachedOver`/`reachedUnder`, `tst_shotanalysis` **fails**; restored, it passes.

Full suite: **111 passed, 0 failed, 0 skipped**, no warnings.

## Not changed

Worth recording from the same shot, deliberately left alone:

- Grind Arm 1 measured mean flow 0.724 vs a 1.000 ml/s goal, **delta −0.276** against a 0.4 threshold — real signal, correctly under the bar. The last ~13 s ran against the 9 bar limiter at roughly half the commanded flow.
- The choked-puck arm was structurally blind to it: it reads pressure-mode phases only, frames 0/1 never exceed 3.48 bar (`CHOKED_PRESSURE_MIN_BAR` is 4.0), and the entire 9 bar stretch sits inside a *flow*-mode frame. Zero pressurized samples, no data.

That gap is a separate design question, not part of this fix.

## OpenSpec

`openspec/changes/fix-frame-exit-confirmation-tolerance/` — the `frame-transition-reason` requirement previously said a confirmed value is recorded only when the reading "satisfied the threshold at the transition sample", which this changes. Archived as the last commit before merge.


skialpine:
### Code review

Found 2 issues:

1. The skip-detection cutoff formula is hand-copied rather than shared. The new predictive log line computes `min(2.0, 0.5 * seconds)` itself, and the identical policy already lives in `ShotAnalysis::detectSkipFirstFrame`. The line's own comment says it is "mirrored from its short-first-step branch so this line predicts the badge rather than approximating it" — which is only true while the two copies stay in lockstep, and nothing fails when one moves. (CLAUDE.md says "Centralize anything produced at more than one site — never hand-roll it per call, never copy a helper's body. A repeated format, prefix, tag, wording, or policy is a drift opportunity: each copy is free to change alone, silently, and nothing fails when one does.") Extract it, e.g. a `skipFirstFrameCutoffSec(double configuredSeconds)` helper both sites call.

https://github.com/Kulitorum/Decenza/blob/d89146928145bd031b1c63c49d9adb1e25a4c6ad/src/controllers/maincontroller.cpp#L4680-L4684

https://github.com/Kulitorum/Decenza/blob/d89146928145bd031b1c63c49d9adb1e25a4c6ad/src/ai/shotanalysis.cpp#L720-L724

2. Two comments describing when a confirmed reason is recorded are now false, and were not updated. `detectSkipFirstFrame` still asserts MainController records `"pressure"`/`"flow"`/`"weight"` "solely when the sensor exit was verified against the transition sample" — this PR's whole point is that a confirmed value is now also produced when only the one-sample extrapolation reaches the threshold (#1813's own numbers: transition sample 2.048 bar against a 2.10 bar threshold yields `"pressure"`). The same stale phrasing sits on `HistoryPhaseMarker::transitionReason` ("threshold not seen in the BLE sample at transition"). `docs/SHOT_REVIEW.md` was updated correctly in this PR, so the doc and the code comments now disagree — and the wrong half is the one a reader of the detector will find first. (CLAUDE.md says "When you change how something is read or called, re-read the comments justifying its optimisations: they are claims with expiry dates, and a comment nobody re-derives is how a bad decision keeps getting re-approved.")

https://github.com/Kulitorum/Decenza/blob/d89146928145bd031b1c63c49d9adb1e25a4c6ad/src/ai/shotanalysis.cpp#L692-L697

https://github.com/Kulitorum/Decenza/blob/d89146928145bd031b1c63c49d9adb1e25a4c6ad/src/history/shothistory_types.h#L35-L41

---

<details>

- [Worth a look] The extrapolation uses the raw per-sample delta with no bound on how much real time separates the two samples, so a stalled BLE notification inflates the tolerance — the same function already gates a different consumer of consecutive samples on `deltaTime > 0 && deltaTime < 1.0` ([maincontroller.cpp#L4483-L4499](https://github.com/Kulitorum/Decenza/blob/d89146928145bd031b1c63c49d9adb1e25a4c6ad/src/controllers/maincontroller.cpp#L4483-L4499)). Needs a large gap AND a sensor near its threshold AND still rising to flip a verdict, and `FrameExit::Inputs` carries no timestamp to express it, so no test covers it. (score: 65)

</details>

Also checked and cleared: `reachedOver`/`reachedUnder` sign and boundary handling; preservation of the original precedence and the `> 0` guards on the `_under` arms; the capture point of `m_prevPressure`/`m_prevFlow`/`m_prevValid` relative to the early returns in `onShotSample`; whether the tolerance could swallow a genuine skip (it cannot — no preceding sample means no tolerance); consumer impact of `pressure_unconfirmed` to `pressure` on the grind limiter-tail trim and the display surfaces; and `check_log_markers.py` compliance for the new log lines.

One reported issue was refuted on inspection and is not listed: the frame-0 prediction comparing `frameElapsed` against the detector's absolute `phase.time`. `ShotTimingController` re-anchors `m_displayTimeBase` at extraction start, and `m_frameStartTime` is set at that same instant, so for `prevFrameIndex == 0` the two values coincide.

PR #1818, opened 2026-08-15
feat(profiles): show what a copy changed from the built-in it matches

skialpine:
## What

The profile knowledge dialog (✨) now opens with **"Your changes from ‹profile›"** — one row per dial-in value where the user's profile departs from the bundled profile the knowledge was authored against.

The prose in that dialog describes the **original**: its temperature, its yield, its pressure targets. Read against a re-tuned copy, there was no way to tell which of those numbers still describe what is about to be brewed. The app already knew; it just didn't say.

## Three rules decide whether the block appears

**Shape equality is the gate, not resolution origin.** A title-resolved profile gets the block too, whenever it is the same shape as the bundled profile its entry was authored against. That is the larger population — a user editing a built-in's temperature in place and keeping the name never reaches the shape step, yet their change is exactly as worth showing. A title match of a *different* shape gets nothing: a title says nothing about frame structure, and diffing there would render `frame 4 does not exist` noise while falsely presenting the profile as a modified copy.

**The base is a bundled FILE, never a KB id.** One entry can be authored against several bundled profiles — `gentle-flat-long-preinfusion-family` has four — so "distance to an entry" is not defined. `ProfileShapeIndex` buckets now retain the resource path beside the id.

**Fewest differing dial-in fields wins.** Nearness is deliberately *not* how far apart the values are. A magnitude comparison needs a weighting between bar, mL/s, degrees and grams that nothing in the domain supplies, and that weighting would silently decide the outcome. Counting needs no weighting — and the winner's difference list *is* the list that selected it, so the selection and the thing selected cannot disagree.

**A tie is answered only when the tied candidates would say the same thing** — same KB id *and* equivalent difference lists — and is then named for the entry rather than for an arbitrary member of it. Same entry but different numbers still shows nothing: the six `tea_portafilter_*` profiles all tie at one row while each states a different brew temperature, so no single "before" column is true. Anything else abstains; it exists to stop a coin flip being presented to a user as a fact.

## One traversal, two audiences

`Profile::fieldDeltas()` now feeds both this block and `frameDiffReport()` — the developer text behind `profile_sync` and the TCL import parity gate. Each row declares which audience it belongs to. Three deliberate divergences, each recorded at its definition:

| | developer report | dial-in block |
|---|---|---|
| inactive setpoint axis | yes — a mismatch is an import defect | no — the machine never applies it |
| profile-level yield / limits | no — would widen what the parity gate rejects | yes |
| frame rename | no — not a portability defect | yes — a real signal |

A change repeated identically on every frame collapses to one row: raising a three-frame profile's temperature is one edit, and listing it three times is both longer and less true.

An empty list *with* a base selected is not the same as no base — it means an unchanged copy, which the surface states, because that tells the reader the knowledge applies without qualification.

## Per-surface source

Shot pages compare against the profile the shot was **pulled** with, through a new `openForShot(title, profileJson)`. Editing the catalog profile afterwards must not rewrite what an old shot appears to have been brewed with.

## Also in this change

- Extracts `ProfileManager::loadProfileByFilename()` from `getProfileByFilename`, which had the four-step lookup inlined. `loadProfile()` keeps its own copy **on purpose** — its first step consumes `m_profileJsonCache` with `take()` and it tracks an `Origin` the readers here have no use for — and the declaration now says so, so the next reader does not "fix" it.
- Documentation in `docs/PROFILE_KNOWLEDGE_BASE.md`; a four-sentence wiki manual entry is staged and will be pushed with the merge.

## Two spec corrections found while implementing

Both folded into the change rather than silently diverged from:

1. The field list was missing frame volume and exit weight, and had no collapse rule — a temperature edit would have rendered three identical rows.
2. The nearness rule started as a magnitude vote and became fewest-differences. Same answer on both real collisions, no unit weighting, one computation instead of two.

## Gates

Build clean · qmllint 239/239 · **111 tests passed, 0 failed, 0 warnings**.

Two defects were caught by those gates and fixed in the **code**, not the test:

- The new Repeater delegate reached outer ids unqualified. Fixed with `pragma ComponentBehavior: Bound`, which is safe only because the delegate already declared `modelData` required — the pairing `QML_GOTCHAS.md` warns about.
- `profileDialInDiffForJson` emitted a `qWarning` on unparseable shot JSON, contradicting its own comment claiming it stayed quiet (`Profile::loadFromJsonString` warns). It now parses directly and returns "no base"; a shot with bad profile JSON is already reported where the row is read, and re-reporting on every dialog open tells the user nothing they can act on.

One thing worth a reviewer's eye: the call-site comment justifying the inline read states plainly that it is **unmeasured**, and gives the bound instead. An earlier draft carried invented millisecond figures; they are gone rather than dressed up.

## Test coverage

Selection tests use the real colliding bucket, not synthetic pairs — a re-tuned `Preinfuse then 45ml` copy must pick that file over `Hybrid pour over`, and an equidistant fixture must produce no base, with the equidistance asserted as a precondition so a future retune of either bundled profile fails loudly instead of quietly skipping the tie path.

`frameDiffReport()`'s rendered text is now pinned by a test. `tst_tclimport` only ever asserted it came back *empty*, which catches a field appearing or vanishing but says nothing about how a difference renders — a format change would have passed every gate while breaking `profile_sync`'s output.


skialpine:
### Code review

Note: I authored this PR, so treat this as a second pass rather than an independent one. Findings below were each verified against the code or the shipped profile data, not taken on an agent's word.

Found 6 issues:

1. A same-entry tie relabels the base but keeps the **first** candidate's deltas, so the block shows values from a profile the user never brewed. `tiedIds` records only kb ids; `bestDeltas`/`bestPath` still belong to whichever tied candidate came first in resource-path order. The `tea_portafilter` bucket is six files under one `tea` entry, so any tea copy at a temperature other than the six bundled ones ties all six, and the "before" column renders black tea's 105 °C. The same gap defeats the self-check one line above: chinese green and white tea are byte-identical apart from `title`, so opening the dialog on the bundled **white tea** scores chinese green first, `bestTitle != p.title()`, and the block prints "Unchanged copy of Tea" on the profile that *is* the documentation — contradicting the contract in `profileshapeindex.h`. `dialInBase_aTieInsideOneEntryNamesTheEntry` asserts `baseKbId` and `baseTitle` and never the row values, so it passes.

https://github.com/Kulitorum/Decenza/blob/2d80002db60f7112f204dcdaca434f246b4f4a88/src/ai/profileshapeindex.cpp#L310-L333

2. `maxFlowOrPressureRange` is tagged `kCount`, whose tolerance is 0.5 "for integer-valued rows" — but the limiter P/I range is continuous, and shipped profiles carry 0.20, 0.90, 1.00, 1.50, 2.50, 3.00 and 3.50. Rows are filtered at emission, so a difference in (0.1, 0.5] never reaches `frameDiffReport()`'s own 0.1 re-filter. `main` compared this field at `> 0.1`. That falsifies the PR's byte-identical claim, silently loosens the TCL import parity gate (`tst_tclimport` and `profile_sync` both gate on the report being empty), and breaks the pre-existing invariant "Empty exactly when `functionallyEqual()` is true" — `functionallyEqual` still uses 0.1 for the same field. Found independently by two reviewers.

https://github.com/Kulitorum/Decenza/blob/2d80002db60f7112f204dcdaca434f246b4f4a88/src/profile/profile.cpp#L1905-L1910

3. Pre-existing, from #1817, but this PR edits the QML that consumes it: `ShotProjection::profileKbDerivedFrom` is only assigned inside `convertShotRecord`'s `!record.cachedAnalysis.has_value()` branch, while `loadShotRecordStatic` sets `cachedAnalysis` unconditionally before returning. Every production path goes through that function, so the field is always empty and the "Based on X" label on the shot pages never renders.

https://github.com/Kulitorum/Decenza/blob/2d80002db60f7112f204dcdaca434f246b4f4a88/src/history/shothistorystorage_serialize.cpp#L168-L172

4. `fieldDeltas()`'s doc comment still says "Doubles compare at a 0.1 tolerance, the same one frameDiffReport() has always used." Both halves are false — `toleranceFor()` returns 0.005/0.05/0.5 per unit, and the whole point of the change was that it is *not* the tolerance `frameDiffReport` uses. It contradicts `ProfileFieldDelta::tolerance` a few lines above. (CLAUDE.md: "a comment nobody re-derives is how a bad decision keeps getting re-approved".)

https://github.com/Kulitorum/Decenza/blob/2d80002db60f7112f204dcdaca434f246b4f4a88/src/profile/profile.h#L511-L516

5. The tie rule is still documented as always-abstain in two places, after the code changed to proceed on a same-entry tie. Header: "Both then converge on fewest-differences, abstaining unless one candidate is strictly nearest." Doc: "Fewest differing dial-in fields wins; a tie shows nothing."

https://github.com/Kulitorum/Decenza/blob/2d80002db60f7112f204dcdaca434f246b4f4a88/src/ai/profileshapeindex.h#L147-L152

https://github.com/Kulitorum/Decenza/blob/2d80002db60f7112f204dcdaca434f246b4f4a88/docs/PROFILE_KNOWLEDGE_BASE.md#L176-L180

6. `design.md` states the dialog-open cost was "Measured and recorded at the call site before merge", while the call site says "NOT MEASURED, and deliberately not given a number here". Two shipped artifacts in one PR asserting the opposite, and the stale one is what gets archived into spec history. Either take the reading CLAUDE.md's inline-read rule asks for, or update the design doc to record the bounded-work argument that was actually made.

https://github.com/Kulitorum/Decenza/blob/2d80002db60f7112f204dcdaca434f246b4f4a88/openspec/changes/summarize-profile-changes-from-builtin/design.md#L118-L122

<details>

- [Worth a look] `tankDesiredWaterTemperature` is tagged `celsius` (0.005) but serializes at 1 decimal (`ProfileJson::TankTemp`), so its tolerance should be 0.05 — the unit token is shared by two fields with different precisions, which the "derived from ProfileJson" claim does not survive (score: 65)
- [Worth a look] `findProfileByTitleForKb` is documented as serving "the three KB surfaces"; this PR makes it four (score: 60)
- [Worth a look] Nothing tests the `espressoTemperatureHealed()` guard on the new brew-temperature row — flipping or deleting the condition would not fail the suite (score: 55)
- [OK to skip] The "48 ms, 85 shapes" index-build measurement predates the added `isValid()` check and second sort; likely still close, and nothing treats it as load-bearing (score: 50)

</details>

Checked and confirmed accurate: the "69 of 100 mixed-pump" figure, the six-file tea bucket, "espresso_temperature is authored not derived" against PR #961, "Direct Setpoint Control fields are never serialized", the `profile_kb_id`-is-not-identity boundary from #1817, the `ComponentBehavior: Bound` / `required property` pairing, and no new test files.

skialpine:
### Code review

Found 6 issues. All six are fixed in [`89072a07`](https://github.com/Kulitorum/Decenza/commit/89072a07acb9448b931edd10bf8d9da0c587ab70); listed here because each was a real defect in the reviewed head.

1. **The "Based on X" line dies on the first autosave.** `clonePersistedShot()` is an explicit whitelist and did not carry `profileKbDerivedFrom` — the field this branch had just brought to life in production. Autosave fires on every commit point, so dragging the enjoyment slider once emptied it for the life of the page. Three neighbours were missing the same way: `recipeId` (five readers, including the gates deciding whether the recipe card or the "no recipe" prompts show), `steamJson`, `hotWaterJson`. Fixed together per [CLAUDE.md](https://github.com/Kulitorum/Decenza/blob/89072a07acb9448b931edd10bf8d9da0c587ab70/CLAUDE.md) ("Fix pre-existing violations in any file you touch").

https://github.com/Kulitorum/Decenza/blob/2d80002d0b8b1ee9a7a2b1a05b9f1e3d0e5d2b6f/qml/pages/PostShotReviewPage.qml#L769-L772

2. **A change could render as two identical numbers.** Rows are emitted at a 0.005 tolerance for `bar`/`mlPerSec` (what ProfileJson writes and what the editor steps at, per `profilejson.h:21-31`) but were displayed at one decimal. A 9.00 → 8.98 edit read `Pour · Pressure 9 bar → 9 bar` — a row asserting a change while showing none. Decimals are now chosen per row.

3. **Hand-copied helper body.** `Theme.cToDisplay(value).toFixed(1) + Theme.tempUnitSuffix()` is verbatim the body of `Theme.formatTemperature()`, used at nine other call sites. CLAUDE.md: "Never copy a macro/helper body to specialize it — alias it."

4. **An unresolvable entry name named an arbitrary sibling.** When `canonicalNameForKbId()` returned empty, the tie branch fell through to the single-winner path and labelled the block `winner->title` — the first tied candidate's own bundled title, by iteration order. Exactly the arbitrary pick the tie conditions exist to prevent. Abstains now.

5. **The spec requirement contradicted itself one line down.** The title still read "SHALL abstain on a tie" above a body describing the same-entry exception; `proposal.md` still carried the old unconditional rule.

https://github.com/Kulitorum/Decenza/blob/2d80002d0b8b1ee9a7a2b1a05b9f1e3d0e5d2b6f/openspec/changes/summarize-profile-changes-from-builtin/specs/profile-dial-in-diff/spec.md#L108-L110

6. **Two comments stated the tea tie count from memory — "seven frames" and "one row" — and both were wrong.** The eighth frame's cooler step carries a different "before" value, so the collapse rule does not fire and neither figure holds. The count is now pinned by the test that depends on it, alongside a precondition asserting the tied candidates disagree on the values; the comments describe it without a figure.

Checked and clean: the tie-selection loop, the per-unit tolerance table against `ProfileJson`, `celsiusTank` not colliding with `celsius` by value, the cached/uncached `ShotRecord` paths, the `dialInDeltas` collapse loop, the QML unit and label coverage, and compliance with #1817's rule that a shape match never reaches `profile_kb_id`.

Issue #1819, opened 2026-08-16
[Bug] Build 3549 not pulling shots on UI


mcastaldelli:
### What happened?

The most recent build does not initiate a shot on the tablet screen (i.e., no line charts) when the pull shot button is pressed on the DE1. The profiles load correctly in the machine (i.e., purple flash), and the scale data updates on the main Decenza screen, but the shot routine does not trigger, and the shot also keeps going until manually stopped. 
See attached logs. I had to downgrade the app to get the machine back to an usable state.

### Steps to reproduce

Try to pull a shot as normal.

### Decenza version

build 3549

### Platform

Android

### Device / OS version

Pie P80X

### Debug log / shot debug log

[debug-4.zip](https://github.com/user-attachments/files/31116366/debug-4.zip)

mcastaldelli:
things got back working on the most recent build. Closing this ticket

PR #1820, opened 2026-08-16
fix(ble): one shared GATT queue across the DE1, scales and refractometers (#1819)

skialpine:
Fixes the root cause of #1819: build 3549 reported the DE1 `CONNECTED` with `STATE_INFO`, `SHOT_SAMPLE` and `WATER_LEVELS` never enabled. No chart, no shot detection, no stop-at-weight — the reporter's shot ran until they stopped it by hand.

## What actually happened

From the attached log, one session, times from app start:

```
[  8.222] Waiting for the DE1 to finish connecting before connecting the scale (15 s cap)
[ 19.069] Found DE1
[ 23.009] DE1 did not connect within 15 s — connecting scale anyway
[ 23.134] Starting characteristic discovery for DE1 service
[ 29.145] Characteristics ready: 18 registered
[ 29.188] Descriptor error (suppressed): DescriptorWriteError
[ 32.265] Notification subscribe timed out (0000a00e)
[ 35.265] Notification subscribe timed out (0000a00d)
[ 38.265] Notification subscribe timed out (0000a011)
[ 38.414] DE1 CONNECTED (BLE)
```

Three CCCD writes rejected while a scale ran connect and discovery on the same GATT stack, each rejection logged at DEBUG and dropped, each then waited out for a full 3 s — and the app went on to report the machine ready. The one `Sleep → Idle` came from the explicit `read(STATE_INFO)` in the subscribe sequence, not from a notification.

**Not a 3549 regression.** The 3546→3549 diff contains no BLE commit. The 15 s scale gate expired because that morning's DE1 scan was ~9 s slower (Found DE1 at 19.07 s vs 10.15 s in the working session). The 3544 downgrade session is confounded — the scale never connected in it at all.

## Why a shared queue

**Qt serializes GATT work per CONTROLLER, not per adapter.** `readWriteQueue` and `pendingJob` are instance fields of `QtBluetoothLE`, one instance per `QLowEnergyController` (`QtBluetoothLE.java:949-950`); the darwin backend has no cross-peripheral queue either. No backend supplies the cross-device guarantee, which is why this is unconditional rather than platform-gated.

de1app has one queue for the DE1 and the scale together (`::de1(cmdstack)`, a single `::de1(wrote)` flag), and defers a scale connect while that queue is backed up (`bluetooth.tcl:2276`). That makes this state unreachable rather than handling the error better. **decaid does not** — it sets `queueType = QueueType.perDevice`, serializing per device and deliberately not across them; an earlier draft of this PR claimed the opposite. Decenza was alone in proceeding on a link whose telemetry never started.

**The asymmetry was the defect.** `BleTransport` knew an operation was outstanding; `QtScaleBleTransport` and `CoreBluetoothScaleBleTransport` had no in-flight concept at all — `writeCharacteristic()` validated the link and called straight through, and the same for discovery, notify-enable and read. The DE1 side could yield and the scale side had nothing to yield with.

## What changed

- `BleGattQueue` — one queue, one operation in flight, every peripheral. Policy (retry budget, delay, discard key) travels with the operation, so the DE1 keeps its tuned constants and everything else defaults to zero retries. Giving scales the DE1's budget would let a dead scale hold the shared slot through a ~32 s retry sequence, starving the machine that budget exists to protect.
- The DE1's subscribe sequencer **collapses** rather than porting: `m_pendingSubscribeQueue`, `m_currentSubscribeUuid`, a 3 s timeout, the `subscribeNext()` recursion and the late-ACK matcher all existed because there was no shared queue to sequence against. Now five submissions, a ready marker and the initial reads.
- `DescriptorWriteError` is handled when it arrives (+45 ms) instead of logged at DEBUG and dropped, and retried within the link's budget.
- `STATE_INFO` and `SHOT_SAMPLE` are **required streams**. If either cannot be enabled the connection fails, with a `de1LinkFault` that `evaluateBleWedge()` can see, rather than reporting an unusable machine as ready.
- `SUBSCRIBE_TIMEOUT_MS` deleted: 3000 against Qt's own `RUNNABLE_TIMEOUT` of 3000, on the same operation.
- One operation clock per transport, covering reads and CCCD writes too. A read that never completed previously held nothing — under a queue it would have held the shared slot forever.
- Both scale transports submit discovery, notify-enable, write and read. One real platform divergence removed: CoreBluetooth's `didDiscoverServices` kicked characteristic discovery for every service from inside the delegate, ahead of the Qt signals, putting discovery on the radio outside any queue.
- Two latent defects fixed on the way: `characteristicRead`/`characteristicChanged` shared one handler, and three early returns in `writeCharacteristic()` reached neither the platform nor a completion.
- A use-after-free in the queue: `dispatchNext()` invoked `m_inFlight->issue()` in place, and a callback that reaches back in destroys the `std::function` whose body is running.

Behaviour changes that are deliberate rather than incidental — no pacing on DE1 operations, `writeUrgent` as position rather than bypass, reads retrying — are each documented at the test that pins them, with the evidence for why.

## Verification

- 113/113 test targets green, no warnings. Both PR-time gates pass.
- The single-in-flight invariant was checked by **removing the guard**, and the first attempt did not go red: the check is duplicated across `scheduleDispatch()` and `dispatchNext()` so neither alone is load-bearing, and a submit arriving while the dispatch timer is still armed never consults it. Both tests were rewritten to submit after the previous operation is in flight; they now go red with the guards removed and green with them restored.
- CI dispatched: [Android](https://github.com/Kulitorum/Decenza/actions/runs/31952396006), [Linux + ctest](https://github.com/Kulitorum/Decenza/actions/runs/31952397137), [iOS](https://github.com/Kulitorum/Decenza/actions/runs/31952421199).

## Ordering that is not a queued operation

A **connect** is deliberately not queued. Queueing it would let a scale reconnect wait behind DE1 work — or worse, stall a DE1 write behind a scale's connect — and stop-at-weight sits on that path. de1app has the answer and it is not a queue entry: it defers a scale connect while the shared queue is backed up (`bluetooth.tcl:2276`, "Too much backpressure, waiting with the connect"). Decenza now does the same, event-driven rather than polled — deferred while any operation is outstanding or queued, released on `BleGattQueue::drained()`.

That replaces the 15 s `m_de1WaitTimer` (group 6), which was a timer standing in for an event, covered one of three orderings, and in the #1819 capture fired at 23.0 s — four seconds *after* the DE1 connected at 19.1 s — releasing the scale straight into the DE1's discovery. `setServiceDiscoveryActive()` went with it: a bool mirrored through five files to ask one scale driver to pause its heartbeat, which the queue now expresses for every device as a guarantee rather than advice.

**decaid is a counter-example here, not a precedent** — an earlier draft of this PR and of `BLE_PROTOCOL.md` had it backwards. It sets `UniversalBle.queueType = QueueType.perDevice` (`universal_ble_discovery_service.dart:403`), serializing per device and deliberately not across them, so it does not prevent this interleaving at all. What it does do — throw out of `_bleConnect()` on a failed subscribe so the reconnect ladder re-runs against a fresh GATT — is the half worth copying, and is what the required-stream handling above does.

## Review pass

- The queue owns **no `QTimer`**. Dispatch is a posted `QMetaObject::invokeMethod`; both hop the event loop, only one of them says so.
- `paceMsAfter` deleted as dead configuration. The 50 ms constant it came from was armed on an enqueue that found the link idle, so it delayed the *first* write after a pause and never paced consecutive ones. Five doc claims about "50 ms pacing" corrected rather than left to be re-read as fact.
- `QLowEnergyService::OperationError` added to every failure arm — it is what Qt emits for a *synchronously* rejected read/write/descriptor-write (`qlowenergyservice.cpp:581/650/724`), not the per-operation error, so without it those rejections released nothing.
- `drained()` now has an emitter on the two paths that reach idle with no slot transition (`discard()`, and a dispatch finding the queue already empty). Posted, because `forget()` is reached from destructors; collapsed to one emission per transition after a test caught it firing twice.
- `validate()` rejects an operation with no `issue` callback or no requester at submit — either would take the slot and be ended by nothing.
- `writeUrgent` is queue **position**, never a bypass; three sites claiming it "writes synchronously" corrected, with the reason one event-loop turn is available on the iOS suspend path.

**Timers.** Removed: `m_de1WaitTimer`, `SUBSCRIBE_TIMEOUT_MS`, `m_commandTimer`, two dead `<QTimer>` includes. Kept, and not removable: the per-transport operation timeout (the only thing that can end an operation the platform never answers — "there is no event" is the condition), the retry backoff (decides *when*, never *whether*), and the Android connect watchdog (#1303). Separately: ~30 fixed delays remain across the 15 scale drivers and are now candidates, but none is testable here and several are device settle times rather than radio workarounds. Own change, one driver at a time.

## Still open before merge

- **Hardware verification** (tasks 9.1–9.3): a slow DE1 connect with a scale present, a refractometer connecting mid-session, and the profile-upload timing delta. The Apple connect-sequence change especially wants a real device.
- **Stop-at-weight latency under scale contention** (task 9.7). The one guarantee the shared queue weakens; bounded to 3 s, unmeasured. Do not build a priority lane before the measurement.
- **Task 9.8** — whether a DE1 write dispatched against a down link should abandon rather than spend its retry budget.


skialpine:
### Code review

Five agents over `b87d5c3b7aebd84dce6bb8eeaa47da7f4c3fbf68`. Found 6 issues at or above threshold, all fixed in 69a3fc8d.

1. **A required stream that fails at SUBMISSION time did not stop `connected()`** — the #1819 symptom itself. `failRequiredStream()` expresses "do not report connected" as `forget()`, which drops the queued ready marker; that only works once the marker is queued. On the two submission-time paths `subscribeAll()` has not reached `submitReadyMarker()` yet, so `forget()` dropped nothing and the marker was queued a moment later. Worse, the test added for this passed because it submitted the marker first — the reverse of the production order.

https://github.com/Kulitorum/Decenza/blob/b87d5c3b7aebd84dce6bb8eeaa47da7f4c3fbf68/src/ble/bletransport.cpp#L216-L228

2. **Two failure reports inside one retry delay armed two retry timers** on the same generation, so both fired: the payload written twice, and the duplicate's late ACK able to release a later operation on the same characteristic. Reachable when a link drops mid-retry (the service `errorOccurred` arm and `InvalidService` both report).

https://github.com/Kulitorum/Decenza/blob/b87d5c3b7aebd84dce6bb8eeaa47da7f4c3fbf68/src/ble/blegattqueue.cpp#L134-L142

3. **`ensureChargerOn()` at app exit was submitted after the drain wait**, and on a quiet quit nothing turns the event loop between it and `de1Device.disconnect()` — `drainDbWork()` returns immediately when nothing is queued and the export loop runs zero iterations — so `forget()` discarded the write unissued, leaving the DE1 USB port off. (bug due to `writeUrgent` now needing one event-loop turn for its posted dispatch)

https://github.com/Kulitorum/Decenza/blob/b87d5c3b7aebd84dce6bb8eeaa47da7f4c3fbf68/src/main.cpp#L4757-L4762

4. **A repeat `discoverCharacteristics()` stalled the whole radio for 20 s** (Qt transport only). `discoverDetails()` returns silently unless the service is in `RemoteService` state (`qlowenergyservice.cpp:585-586`), so a call for an already-discovered service issues nothing that can complete — harmless before the queue, a process-wide stall after it. The CoreBluetooth transport carries this guard; the Qt one did not.

https://github.com/Kulitorum/Decenza/blob/b87d5c3b7aebd84dce6bb8eeaa47da7f4c3fbf68/src/ble/transport/qtscalebletransport.cpp#L204-L212

5. **CoreBluetooth's outstanding-read key was left stale on two release paths** — the operation clock firing, and `disconnectFromDevice()` calling `releaseGattQueue()` directly, neither of which goes through the four shims that cleared it. A later unsolicited notification could then complete a write still on the wire. (CLAUDE.md says "Centralize anything produced at more than one site")

https://github.com/Kulitorum/Decenza/blob/b87d5c3b7aebd84dce6bb8eeaa47da7f4c3fbf68/src/ble/transport/corebluetooth/corebluetoothscalebletransport.mm#L120-L125

6. **`queueDrained()` was missing from two of the four terminal-success paths** (characteristic discovery, and the CCCD ACK). `main.cpp` waits on it at exit to know the sleep and charger writes went out.

https://github.com/Kulitorum/Decenza/blob/b87d5c3b7aebd84dce6bb8eeaa47da7f4c3fbf68/src/ble/bletransport.cpp#L751-L760

**Not a defect, but the one guarantee this design genuinely weakens:** an urgent DE1 write is now position, not bypass, so it can wait behind *another device's* in-flight operation — impossible before, and stop-at-weight is on that path. Expected cost is negligible (the only in-shot scale write is a 1 Hz heartbeat completing in tens of ms); the tail is a scale that stops answering, which held the shared slot for the full 5 s operation clock. Reads and writes now use `RW_TIMEOUT_MS = 3000` — Qt's own `RUNNABLE_TIMEOUT`, past which Qt discards the reply anyway — which bounds the tail without pretending to remove it. Task 9.7 measures it on hardware before any priority lane gets built.

<details>

- [Worth a look] `de1device.h` still said `setUsbChargerOnUrgent` "bypasses the BLE command queue" — the one site the previous pass missed, contradicting the five around it. Fixed (score: 70)
- [Worth a look] `de1-connection-health/spec.md` required that a failed notification-enable NOT be retried in place, while the code retries it 5x under the ordinary write policy; `design.md` and `proposal.md` both had it right. Fixed (score: 70)
- [OK to skip] A stale `bletransport.cpp:753-758` citation in `qtscalebletransport.cpp`, rotted by this branch's own rewrite. Fixed by naming the arm instead of a line number (score: 60)
- [OK to skip] A DE1 write dispatched against a down link spends its full 5x500ms budget re-asking a fixed question. An `abandon()` path was written and reverted — it invalidates the headless seam ten queue tests hold the slot with, for an unmeasured gain, and `disconnect()`'s `forget()` already bounds the window. Recorded as task 9.8 (score: 55)

</details>

113/113 local, Linux CI 111/111 with no retries, both PR-time gates green.

PR #1824, opened 2026-08-18
ci: suppress the glibc resolver leak in the nightly ASan job (+ opsx refresh)

skialpine:
## What

The nightly ASan job went red on 2026-08-18 with a 28-byte LeakSanitizer finding, while ctest itself reported `100% tests passed, 0 tests failed out of 110`. The leak came from the first attempt of `tst_wifiscalediscovery` and not from the `until-pass` retry; the `Summarise` step fails the job on any sanitizer report regardless of ctest's exit code, which is working as designed.

This adds one LSan suppression, `tests/lsan-suppressions.txt`, and wires it into the nightly.

## Why it is not ours

The allocation is entirely inside glibc. `resolv/res_send.c:324` allocates `sizeof(struct sockaddr_in6)` — 28 bytes, matching exactly — into `_res._u._ext.nsaddrs[]`. Its only free path here is `__res_thread_freeres` (`resolv/res-close.c:130`), which glibc's own comment describes as running *"when a thread is exiting"*.

`QHostInfo` resolves on a `QThreadPool` that reuses and parks its threads (30 s expiry), so a pool thread that sent a DNS query can still be alive at process exit holding its per-thread `_res` state. Under load the leak check races that teardown, which is why it is intermittent rather than constant.

Verified by controlled comparison on Linux under ASan (glibc 2.39, Qt 6.11.1, blackholed resolver, 8-way concurrency, 80 runs each):

| Binary | Resolver leaks |
|---|---|
| bare C, threads exit immediately | 0 / 80 |
| bare Qt `QHostInfo`, no Decenza code | 3 / 80 |
| `tst_wifiscalediscovery`, our code | 3 / 80 |

The leak reproduces with all Decenza code removed, at the same rate and with a byte-identical stack. It is bounded — 28 bytes per pool thread, at most 20 threads — and released when the process ends, so it is not a production concern either.

Three of our own code paths were checked first and came back clean: every `QHostInfo` lookup id is stored and aborted (`wifiscalediscovery.cpp:183`, `:617`); the deliberately unjoined browse worker (`:61-64`) never calls into the libc resolver; and `MdnsResolver` uses raw 5353 sockets with no `getaddrinfo`.

## Suppression effect

192 runs each at 16-way concurrency:

```
WITHOUT suppression: leak-ERROR=2/192  suppression-report=0/192
WITH    suppression: leak-ERROR=0/192  suppression-report=9/192

Suppressions used:
  count      bytes template
      1         28 __res_context_send
```

The two counts differ because the underlying race varies with load; both are small samples of the same intermittent event.

## Notes for review

- Leak suppressions belong in `LSAN_OPTIONS`. Putting `suppressions=` on `ASAN_OPTIONS` silently does nothing for leaks, so that is called out at the call site.
- `print_suppressions=1` keeps a used suppression visible in the log rather than making it a silent subtraction from the night's signal.
- The workflow header said findings are "never to silence for a green night". That is no longer strictly true, so it now records this single exception and points at the suppression file, which states the bar for adding another.
- **What this costs:** a genuine leak of ours reached through `getaddrinfo` would also be hidden. The template matches only frames through glibc's resolver send path, so the window is narrow, but it is real and is stated in the file.
- The suppression was verified against an absolute path inside a container; I could not exercise the `${{ github.workspace }}` expansion locally. If that path is wrong on the runner, LSan reads nothing and the night goes red exactly as before — it fails visibly, not silently. Worth a glance at the next nightly for the `Suppressions used:` block.

## Testing

No app code changes. `tst_wifiscalediscovery` was additionally mutation-tested to confirm the two DNS-dependent assertions still earn their place: breaking the per-probe completion semantics fails `multipleHostnamesFinishOnce`, and forcing `ran=false` fails `nonexistentHostnameEmitsProbeFinishedOnly`.


skialpine:
Review pass done (comment-analyzer, code-reviewer, silent-failure-hunter, pr-test-analyzer). The `leak:__res_context_send` line survived every attempt to break it — no reviewer could construct a path by which a Decenza allocation lands under that frame. The prose around it did not, and `f814a10c` corrects it.

**Four substantive corrections, all mine:**

1. **The free path was stated backwards.** `__res_thread_freeres` frees nothing; `__res_iclose` does (`res-close.c:94`, freeing at `:112`) and `__res_thread_freeres` is one of its callers.
2. **The mechanism was asserted and is wrong.** QHostInfo joins its pool before exit — `QCoreApplication::destroyed` → `waitForDone()` (`qhostinfo.cpp:907-908`) → `reset()` waits on and deletes every thread (`qthreadpool.cpp:280-284`) — and the test is `QTEST_GUILESS_MAIN`, so that runs before LSan's check. Worse, a genuinely live thread's TLS is scanned as a root region, so the stated mechanism describes the condition under which the leak would *not* fire. The file now records why it's wrong, offers the thread-exit-window hypothesis with its own unresolved tension against the bare-C control, and states that the suppression does not rest on the mechanism.
3. **The stated cost was wrong in the reassuring direction.** A leak of ours "reached through getaddrinfo" is allocated *above* `res_send` and carries no `__res_context_send` frame — never at risk. The real cost is quantity: the template has no count ceiling.
4. **"At the same rate" over-read 3/80 vs 3/80** — with three events the 95% interval spans roughly 0.8–10.6%.

**And the safeguard was inert.** `print_suppressions=1` bought nothing: `ctest --output-on-failure` discards a passing test's output, so the block never reached `ctest.log`, and Summarise grepped only for `ERROR:` lines — the suppression was exactly the silent subtraction the comment claimed it wasn't. Summarise now reads `Testing/Temporary/LastTest.log` and reports a match into the step summary. `log_path=` is deliberately not used; it would redirect all sanitizer output and break the existing finding greps.

Also added: a file-exists canary on the suppressions path; disclosure that the measurements came from a since-deleted container with an unroutable nameserver, with the bare-C control inlined so the decisive 0/80 row is checkable; the failing nightly run id and a removal criterion; glibc line numbers tied to ubuntu-24.04 with a re-derivation URL; duplicated evidence collapsed out of the workflow header and TESTING.md.

**Confirmed sound by review, previously unverified:** `${{ github.workspace }}` is correct, the option string parses as two flags, and an unreadable suppressions file makes LSan `Die()` — so it fails red, not silently unsuppressed.

**Deliberately left as follow-ups:** an `lsan_canary` asserting LeakSanitizer still *reports* (the existing canary commits a buffer overflow, not a leak); a `check_lsan_suppressions.py` ratchet so the file's bar is enforced rather than prose; and one discriminating mutant on `tst_wifiscalediscovery` — the two I ran would also fail with `localhost`, so they show the assertions can fail but not that the real DNS traffic earns its keep.

skialpine:
Rebased onto `8f46c49d` (the Qt 6.11.2 upgrade), which falsified two version claims in the suppression file — corrected in `c7a16654`: the file said "CI builds 6.11.1" when CI now builds 6.11.2, and the measurements were taken on 6.11.1 *before* that upgrade, so the rate under the Qt version CI now builds is unmeasured. That caveat is now stated rather than implied.

Also added `04e90919`, an unrelated refresh of the OpenSpec slash commands and skills — regenerated CLI output that had been sitting uncommitted in the working tree and needed to land somewhere. It is called out as unrelated in its own commit message, and the two concerns are cleanly separated by commit if you'd rather split them.

skialpine:
Correction to this PR's description, from the first CI run of the merged change ([run 32170570549](https://github.com/Kulitorum/Decenza/actions/runs/32170570549)).

**What the run confirmed:** `${{ github.workspace }}` expands correctly — the runner shows `LSAN_OPTIONS: suppressions=/home/runner/work/Decenza/Decenza/tests/lsan-suppressions.txt:print_suppressions=1` — the file-exists canary passed, and the asan leg was green at 112/112 with zero `ERROR: LeakSanitizer` lines. No `Suppressions used:` block, so no leak occurred to suppress; this run cannot distinguish "suppression works" from "nothing to suppress", as flagged when it was opened.

**What it corrects:** the description said a wrong path "makes LSan `Die()` — so it fails red, not silently unsuppressed". That is wrong, and I asserted it from review analysis rather than measuring it. LSan reads the suppressions file **lazily**, only when it has a leak to check. Measured on ASan/glibc 2.39:

```
leaking binary,     bad path -> "failed to read suppressions file '<path>'", exit 1
non-leaking binary, bad path -> no output, exit 0
```

So a wrong path is silent on every clean night and speaks up only on the rare night a leak occurs — where it reads as the suppression having failed, not as a bad path.

This makes the canary load-bearing rather than belt-and-braces: it is the only thing that catches the misconfiguration, and the old comment was exactly the reasoning that would have justified deleting it. Corrected in [#1829](https://github.com/Kulitorum/Decenza/pull/1829) (`376f745b`), which already touches this file, rather than as a separate PR.

PR #1838, opened 2026-08-19
fix(saw): debounce the untared-cup popup against a stale pre-tare sample

### LINKED ISSUE #1837: [Bug] audto tare disabled - weird bug
### What happened?

TWo vidoes, see the scale readout starting way up off the screen then down to zero then behaving more or less normally

https://github.com/user-attachments/assets/bbf32de9-c9cb-411d-9928-034c64a45407
https://github.com/user-attachments/assets/b19e7e2d-3b07-4b66-8884-dab0a54ea697

### Steps to reproduce

nothing special.
used turboturbo profile

### Decenza version

2.0.4

### Platform

Android

### Device / OS version

12

### Scale (if applicable)

_No response_

[19aug debug (1).zip](https://github.com/user-attachments/files/31219914/19aug.debug.1.zip)

_No response_

### Screenshots or video

_No response_

### Anything else?

_No response_

skialpine:
## Summary
- Fixes #1837 — "auto-tare disabled" popup shown even though tare is happening.
- Root cause, confirmed against the reporter's debug log: a Hot Water shot leaves 70-140g on the scale, Espresso starts right after and re-tares, but the scale's zeroed BLE notification can land ~100-300ms after extraction already started. The untared-cup sanity check (single-sample, >50g in first 3s) fired on that one stale reading before the real zero arrived.
- Requires the high reading to persist past a 350ms debounce window before showing the popup, mirroring the existing oscillation-recovery debounce a few lines above. The "skip stop-at-weight for this sample" safety behavior is unchanged — only the user-facing popup is debounced.

## Test plan
- [x] Updated `tst_saw.cpp::untaredCupDetected` and `tst_weightprocessor.cpp::untaredCupDetectedEarly` to feed a persisting high reading (debounce now requires it).
- [x] Added `tst_saw.cpp::untaredCupTransientStaleSampleDoesNotFire` — a stale sample that resolves within the debounce window (reproducing the exact #1837 log shape) must not fire the popup.
- [x] Full local suite: 113/113 passed via Qt Creator.

### DIFF
--- a/src/machine/weightprocessor.cpp
+++ b/src/machine/weightprocessor.cpp
@@ -552,12 +552,26 @@ void WeightProcessor::processWeight(double weight)
             SAWW_WARN(QStringLiteral("Sanity check: weight %1 g at %2 s into extraction — "
                                      "skipping stop-at-weight (likely untared cup)")
                           .arg(weight, 0, 'f', 2).arg(extractionTime, 0, 'f', 1));
-            if (!m_untaredCupSignalled) {
+            // Debounce the user-facing popup (#1837): a Hot Water shot that leaves
+            // 70-140g on the scale, followed immediately by Espresso, re-tares the
+            // scale but the zeroed BLE notification can land ~100-300ms after
+            // extraction already started — so the stale Hot Water weight reads as
+            // "untared cup" for one or two arrivals before the real zero arrives.
+            // Require the high reading to persist past that window (observed
+            // worst case ~200ms across three logged incidents) before telling the
+            // user, so a self-resolving stale sample never shows a false popup for
+            // a cup that was, in fact, tared correctly.
+            constexpr qint64 UNTARED_CUP_CONFIRM_MS = 350;
+            if (m_highWeightStreakStartMs == 0) {
+                m_highWeightStreakStartMs = wallClock;
+            } else if (!m_untaredCupSignalled &&
+                       wallClock - m_highWeightStreakStartMs >= UNTARED_CUP_CONFIRM_MS) {
                 m_untaredCupSignalled = true;
                 emit untaredCupDetected();
             }
             return;
         }
+        m_highWeightStreakStartMs = 0;
     }
 
     // Suppress SAW during preinfusion frames (matches de1app: SAW only after
@@ -876,6 +890,7 @@ void WeightProcessor::startExtraction()
     m_lastConstantSampleLogMs = 0;
     m_flowBecameValidLogged = false;
     m_untaredCupSignalled = false;
+    m_highWeightStreakStartMs = 0;
     m_lastWallClockMs = 0;
     m_lastSampleTs = 0;
     m_uncalibratedBatchWarned = false;
@@ -1009,6 +1024,7 @@ void WeightProcessor::resetForRetare()
     m_lastLowFlowLogMs = 0;
     m_lastConstantSampleLogMs = 0;
     m_flowBecameValidLogged = false;
+    m_highWeightStreakStartMs = 0;
     SAWW_LOG(QStringLiteral("Reset for auto-retare"));
 }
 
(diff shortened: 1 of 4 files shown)


skialpine:
### Code review

Found 3 issues:

1. The debounce is a wall-clock timer guard, which CLAUDE.md explicitly bans (root `CLAUDE.md`: "Never use timers as guards/workarounds... Everything else — including debounce — should use event-based flags"). `UNTARED_CUP_CONFIRM_MS` gates the popup on `wallClock - m_highWeightStreakStartMs >= 350`, i.e. real elapsed time. The PR's own comment claims this mirrors "the existing oscillation-recovery debounce pattern a few lines above" — but that pattern (`m_settleCount`) counts 3 consecutive samples, not milliseconds, so it's actually the event-based alternative CLAUDE.md asks for, not a precedent for the timer.

https://github.com/Kulitorum/Decenza/blob/3d175632805537011887a7838aa13658f792e6fb/src/machine/weightprocessor.cpp#L548-L575

2. The debounce can be silently bypassed on a scale that alternates readings (this file already documents that behavior — "a scale alternating a loaded reading with a near-zero one," `weightprocessor.cpp:55`/`103-109`). `m_highWeightStreakStartMs` is only cleared at line 574, which sits below two unconditional early returns: the spike-rejection branch (`weightprocessor.cpp:120-129`) and the `!m_tareComplete` branch (`weightprocessor.cpp:538-546`). Sequence: t=0 stale 139.5g sets the anchor; t=200ms the real tare-confirmed 0g arrives but is >100g from the previous sample so it's spike-rejected and returns without touching the anchor or `m_lastRawWeight`; t=400ms another 139.5g sample is no longer rejected (matches the stale `m_lastRawWeight`), falls into the sanity block, and `400-0 >= 350` fires the popup anyway — the exact false positive this PR is meant to fix. The PR's own new test stops one sample short of this sequence (it ignores the "Spike rejected" warning but never sends a fourth 139.5g sample after it).

https://github.com/Kulitorum/Decenza/blob/3d175632805537011887a7838aa13658f792e6fb/src/machine/weightprocessor.cpp#L548-L575

3. `m_untaredCupSignalled` (the latch limiting the popup to once per extraction) is reset in `startExtraction()` but not in `resetForRetare()`, even though this PR edits `resetForRetare()` directly to reset its own new sibling field `m_highWeightStreakStartMs`. `resetForRetare()` runs mid-preheat on a cup-placed retare (wired to `MachineState::flowBeforeAutoTare`), so if the popup fires once before that retare, a genuinely new untared-cup condition later in the same extraction can never re-trigger it. Pre-existing gap, but CLAUDE.md's "fix pre-existing violations in any file you touch... issues compound over time and each change is an opportunity to fix them" applies directly since the PR is already touching this exact function for the analogous field.

https://github.com/Kulitorum/Decenza/blob/3d175632805537011887a7838aa13658f792e6fb/src/machine/weightprocessor.cpp#L1004-L1028

---

<details>

- [Worth a look] A streak that first crosses 50g late in the 3s window (e.g. 2.8s) can have its confirming sample land past `extractionTime >= 3.0`, falling through to the reset and never firing the popup at all — a silent regression vs. pre-PR (any breach in the first 3s used to fire immediately). Low real-world odds since the PR's own logged incidents show the stale window is ~100-300ms, not seconds, and no test exercises this boundary (score: 50)
- [OK to skip] The new comment above `UNTARED_CUP_CONFIRM_MS` cites "100-300ms" and "~200ms" for what reads as the same delay without reconciling them, leaving the actual safety margin unclear (score: 58)
- [OK to skip] The new `m_highWeightStreakStartMs` header comment only mentions the >50g half of its gating condition and omits the `extractionTime < 3.0` half (score: 60)

</details>

skialpine:
Pushed fixes for all issues from the review above (including the two below-threshold ones worth fixing):

- Switched the debounce from a 350ms timer to a 4-consecutive-sample count (event-based, matching CLAUDE.md and the existing oscillation-recovery pattern).
- Fixed the bypass: the streak now also resets on the spike-rejected and `!m_tareComplete` early-return paths, so a real tare-confirmed zero can't be "forgotten" and a stale reading can't revive the streak.
- A streak already under way can now finish confirming past the 3.0s window instead of being silently swallowed.
- `resetForRetare()` now also resets `m_untaredCupSignalled`, so a mid-preheat retare re-arms detection for a later, genuinely new untared-cup condition in the same extraction.
- Reconciled the debounce comment's numbers and completed the streak field's header comment.

Four new regression tests cover the bypass, the alternating-reading case, the boundary streak, and the retare re-arm. 113/113 local tests pass.

PR #1840, opened 2026-08-20
Drop MCP revisions 2024-11-05 and 2025-03-26, delete batch dispatch

skialpine:
## Why

`supportedProtocolVersions()` advertised four MCP revisions. Two of them have never been requested by any client, and one of those two was the sole reason a whole dispatch path existed.

Measured from the device log — 50 `initialize` handshakes across several app runs, every real client this server has:

| Client | Requests | Count |
|---|---|---|
| `claude-code` v2.1.229–234 | `2025-11-25` | 38 |
| `Anthropic/ClaudeAI` (claude.ai connector) | `2025-11-25` | 9 |
| `codex-mcp-client` v0.147.0-alpha.6.5 | `2025-06-18` | 3 |

Zero for `2025-03-26`. Zero for `2024-11-05`.

The ecosystem agrees: the official conformance suite rejects `2024-11-05` as an unknown `--spec-version`, and `2025-03-26` selects zero scenarios. Two of the four revisions we advertised were unverifiable by the protocol's own tooling — and advertising a revision is a claim to implement it.

**`2025-06-18` stays.** Codex is on it. It is the one drop that would have had a known victim.

## What this actually buys

Removing two strings from a list buys nothing. The win is what those strings kept alive:

**Batch dispatch — `handleJsonRpcBatch` (136 lines), `willDeferResponse` (13), and six tests.** Batching is defined by exactly one revision, `2025-03-26`: absent from `2024-11-05`, removed in `2025-06-18`, absent from `2025-11-25`, absent again in `2026-07-28`. No revision this server serves — or will serve — defines it.

It was also the most defect-prone code in the file *by its own record*: a batched `shots_delete` that deleted the row and then reported it refused, and a deferred handler writing a second complete HTTP response onto an already-answered socket. Both from folding a per-message model into an array response, which nothing else here does.

An array body is now refused explicitly rather than ignored — it parses fine, so it would otherwise fall through to `doc.object()` and be answered "method not found", which tells the sender nothing true.

Also collapsed: `instructions` at `initialize`, and `title` on tools and resources, whose old arms are now unreachable. `structuredContent` / `resource_link` (2025-06-18) and `$schema` dialect / `icons` (2025-11-25) keep their gates — the two surviving revisions are distinguishable by exactly those.

**Net −289 lines across `src/` and `tests/`.**

## The bug this nearly shipped

Dropping `2025-03-26` from the supported set turned the protocol's own compatibility mechanism into a hard 400.

That value is the sentinel the spec tells a client to use when it cannot identify the version — *"if the server does not receive an `MCP-Protocol-Version` header, and has no other way to identify the version… assume `2025-03-26`"* — and clients send it explicitly for the same reason. It means "I do not know", not "I want 2025-03-26". Read as a version request and refused, every client using the documented fallback breaks.

It is now treated exactly as an absent header. Accepting it does **not** make the revision negotiable and does **not** grant its semantics — batching stays gone, since a client cannot negotiate the only revision that defines it.

**Found by the official conformance suite within a minute of the build**, on `server-accepts-multiple-post-streams` — the same scenario that found an earlier header-handling defect. Nothing in this change's own reasoning predicted it, and the local suite was green throughout. That is the argument for running conformance as a gate rather than a report.

## Verification

- **Full suite: 113 passed, 0 failed, 0 warnings.**
- **Conformance (ShotServer route): exit 0** — 9 passed, 22 expected failures, no stale entries. Scenario-by-scenario identical to the pre-drop run.
- Adds `tests/conformance/expected-failures.yaml`, the suite's own baseline mechanism, listing only what cannot pass: its fixture surface (it drives a reference server exposing `test_*` tools and `test://` resources) and capabilities we never declare (`prompts`, `completions`, `logging`). The run becomes green-or-regression rather than a score, and it fails on a **stale** entry, so the file cannot rot.
- Live-verified before and after: `2025-03-26` → negotiates `2025-11-25`; `2025-06-18` → honoured; sentinel header → 200, not downgraded.
- Gates clean: `check_log_markers.py`, `check_mcp_tool_budget.py`, `check_test_source_duplication.py`.

Fixture tools were considered and rejected — two of the fourteen fixture scenarios would exercise existing code; the rest mean building content kinds we never emit, a progress feature we lack, and Sampling/Elicitation, which we declined and 2026-07-28 deprecates. They would also spend the `tools/list` budget every client fetches and real clients truncate.

## Not verified

- **The remote route.** `McpRemoteAccess` is a second entry into the same handler and has been missed before. This change does not touch routing, but nothing here confirms that.
- **The two dropped revisions themselves.** The conformance suite has no scenarios for either, so it cannot confirm the removal — only that nothing else regressed.

## Breaking change

A client requesting `2024-11-05` or `2025-03-26` is answered with `2025-11-25` and, per the protocol, should disconnect if it cannot speak it. No observed client does this. **`mcp-remote` is the wildcard** — the auto-recovery branch exists for it and its negotiated revision is not in the sampled log.

The log is one device over a bounded ring buffer. Best evidence available, not a census — which is why this lands on its own, so a break is attributable and revertible without disturbing anything else.

## Follow-up

`add-mcp-dual-era-2026` needs a rebase after this lands. It carries a duplicate copy of the baseline file, and a `MCP-Protocol-Version Request Header` spec delta that must merge with this one's. Both recorded in that change's task 6.1.


skialpine:
## Post-review verification

Full suite **113 passed, 0 failed, 0 warnings.** Conformance against a running app, with the checked-in baseline:

```
Total: 9 passed, 22 failed
Baseline check passed: all failures are expected.     exit 0
```

Scenario-by-scenario identical to the pre-drop run — no stale entries, nothing newly red.

The three log lines added during review, captured from the built binary rather than from tests:

```
DEBUG [MCP][Server] MCP-Protocol-Version 2025-03-26 treated as absent (not negotiable)
                    — session 7b9e4a2a… stays on 2025-11-25
WARN  [MCP][Server] Client requested unsupported protocol 2025-03-26 — answering 2025-11-25.
                    It will receive fields that revision does not define.
WARN  [MCP][Server] Refused a JSON-RPC batch of 2 (ping, tools/call)
                    — no supported revision defines batching
```

Three consecutive DEBUG lines appear during the conformance run itself — the suite's three concurrent POSTs are exactly the population that line was written for.

## What the review changed

Four agents. Every significant finding was a **factual error in a comment this PR wrote**, not a defect in the code it changed.

- **`instructions` was never version-sensitive.** `schema/2024-11-05/schema.ts` declares `instructions?: string;` with the same docblock as every later revision. The removed gate rested on a premise that was false when first written, and this PR restated it verbatim. Corrected in code and in both places the docs repeat it; the data-driven test that claimed to guard it could not break on the one scenario its comment named, and is now a single renamed test.
- **Two more dead gates.** `emitStructured` / `emitResourceLinks` read `>= 2025-06-18` — the floor. Same threshold as the `title` gates collapsed in this diff, opposite treatment. Four artifacts claimed the surviving revisions were "distinguishable by them"; they never were.
- **The text-block correction was an overcorrection.** The tools spec does give backwards compatibility as the reason for the serialized payload in the text block. Both reasons now stated — the denial read as licence to change what goes in it.
- **Unsupported negotiation now warns.** This PR made `supportedVersions.first()` reachable for the dropped revisions, so such a client is answered with the *newest* revision and receives every field the deleted gates withheld. Spec-correct, but the INFO line looked identical to a success.
- **Tests restored** that the batch deletion silently took: refusal-before-dispatch (the replacement used only side-effect-free methods and could not distinguish "refused" from "refused after running" — the invariant whose absence once shipped a `shots_delete` that deleted and then reported refusal), the single-HTTP-body assertion, and 202-for-a-notification. Plus new coverage for `McpSession`'s default version, whose comment argued a safety property nothing asserted.

## Still not verified

**The remote route.** `McpRemoteAccess` is a second entry into the same handler and has been missed before. This change does not touch routing, but nothing here confirms that.

PR #1852, opened 2026-08-22
Scope the AI advisor's shot selection to the equipment package

skialpine:
## Why

The advisor grouped a shot's history by bean and profile and nothing else, so shots pulled on different equipment landed in one dial-in story.

Concretely: a straight-wall Decent 18g basket dialled at 9.75 and a stepped Graph Coffee 58→46mm basket dialled at 17, same grinder. The advisor saw a single dial jumping 9.75 → 17 with pressure rising rather than falling, had no way to learn the basket had changed, and concluded the burrs "may have crossed a zero-point threshold" — then recommended 10.5, a setting for the other basket. The same reply cited a "70/100 shot" that appears nowhere in its context and reasoned from it.

Two failures compound: the history is filtered on the wrong key, and the equipment is absent from the payload, so the advisor can neither name what it is comparing nor notice it is comparing across a basket change.

## What changed

**Seven independent selections** fed the advisor and none consulted equipment. All seven now match on the shot's equipment package, which forks on grinder brand/model/burrs, basket, *or* puck prep — so one integer answers all three questions:

| Selection | Was keyed on |
|---|---|
| In-app recent-shot history | bean + profile + 21-day window |
| MCP `dialInSessions` | profile KB id |
| `bestRecentShot` (the anchor the model treats as the target) | profile KB id + rating + 90-day window |
| `grinderContext` observed settings / explored range | grinder model + beverage + bean |
| `grinderCalibration` mined pairs + current-batch anchor | grinder model + burrs, spanning every package sharing them |
| `dialing_get_grinder_calibration` | same builder as above |
| Advisor conversation key | bean + profile |

`grinderCalibration` is the costliest: it publishes a **number** the user dials in, and its endpoint medians are pooled by `(batch, profile)` *before* any pair forms — so an unscoped pool corrupts the endpoint, not merely a pair that happens to straddle two baskets. Its anchor was package-blind too.

**Matching uses `COALESCE(equipment_id, 0)` on both sides.** A bare `=` drops every unpackaged shot (SQL `NULL = NULL` is not true); bucket 0 holds them all, so the filter is a no-op for a user who has never created a package.

**`equipmentBucketForShot` returns `std::optional`.** Callers pass `-1` as a "no resolved shot" sentinel; collapsing that to bucket 0 scopes to unpackaged shots and returns an empty history to every user who *has* a package. Returning 0 for a missing row looks like the safe default and is the opposite of one — caught by six pre-existing test failures on the first run, not by review.

**Equipment is named wherever it is filtered on**, because a filter the model cannot see is a silent one: the in-app `### Setup:` header gains basket and puck prep, the MCP session `context` gains `basketBrand`/`basketModel`/`puckPrep`, and an empty history now *says so* and names the equipment set rather than being omitted. An absent block is indistinguishable from "this user has no history", and an unanchored model supplies its own anchor.

**Two system-prompt rules**: grind settings compare only within one equipment set (and an out-of-order setting should raise an equipment question before a grinder-mechanism theory); and the model may cite only shots, scores and taste notes present in its context.

**The conversation key gains the package.** A saved thread replays its stored turns on every request, so filtering future context does nothing about contamination already in the transcript. Changing the key retires every pre-upgrade thread — the first advisor use after upgrading starts clean, with no migration step — and stops a later basket switch from re-accumulating a mixed thread.

## Consequences

- After an equipment change (including a puck-prep edit, which forks the package) the advisor's history starts empty until shots accumulate. Intended; the no-history block keeps it legible rather than silent.
- `grinderCalibration` recomputes from the whole database on every call (no window, no limit), so an established package keeps its calibration intact — only a newly forked one starts thin.
- Upgrading discards in-flight advisor conversations. That is the requested behaviour; nothing is deleted, the old threads age out of the five-slot LRU.

## Dropped during implementation

Reporting an equipment swap in the shot-to-shot changes line was planned, then dropped: keying the conversation on the package means every shot in a thread shares one, so the diff could never fire. A branch with no reachable input is the same defect as a test that cannot fail.

## Testing

Full suite via Qt Creator: **113 passed, 0 failed**. New tests were verified to fail against the pre-change behaviour by neutering the scoping and re-running.

No schema migration — `shots.equipment_id` exists and migration 22 backfilled it. No wiki entry: this changes how the advisor selects its own context and surfaces nothing a user must be told exists.

OpenSpec change: `openspec/changes/filter-advisor-history-by-equipment/`


skialpine:
### Code review

Found 4 issues. All four are documentation or contract drift rather than defects in the feature logic — a full trace of the seven selection points, the conversation-key path, import/migration and threading found the algorithm correct (binding order, named-vs-positional placeholders, migration ordering, no main-thread reads, no use-after-move all check out).

1. `docs/CLAUDE_MD/AI_ADVISOR.md` is edited by this PR but still describes the pre-change bean+profile-only semantics. `grep -cin "equipment"` over the file returns **0** — the PR's entire premise is absent from its own domain doc. Worse, the doc positively asserts behaviour the PR makes false: `queryGrinderContext()` is described as "Grinder-model-wide, so it matches the Grind quick-select widget's `grindStepForGrinder()`", but it is now scoped by equipment bucket. The "Dial-in history" row likewise still reads "Last N shots with same profile family". CLAUDE.md lists this file in the pointer table as the doc to read when working in this domain, so the next person here is handed a description that contradicts the code.

https://github.com/Kulitorum/Decenza/blob/3bc603a70fc13bcc4cf36757bbb7e4012169ef86/docs/CLAUDE_MD/AI_ADVISOR.md#L178-L181

2. The `equipmentBucketSql` comment is factually wrong about the SQL the function emits. It states "COALESCE on **BOTH** sides is load-bearing", but the implementation is `COALESCE(%1, 0) = %2` — one COALESCE, on the column side only. The safety property the comment claims is enforced by the helper is actually caller discipline: all six call sites happen to bind a plain non-null `qint64`. A caller that ever binds a null `QVariant` gets `COALESCE(col,0) = NULL`, unconditionally false, silently excluding every row — precisely the failure mode the comment says is already covered. Either make the claim true (`COALESCE(%1, 0) = COALESCE(%2, 0)`) or state the caller obligation.

https://github.com/Kulitorum/Decenza/blob/3bc603a70fc13bcc4cf36757bbb7e4012169ef86/src/history/shothistorystorage.h#L149-L156

3. A comment instructs maintainers to keep a duplicate in sync with a copy this PR deleted. It says one copy is "deliberately NOT collapsed" because `src/ai/dialing_blocks.cpp:1194` inserts an extra burrs clause inside the same subquery. That copy is gone — `dialing_blocks.cpp` now contains zero occurrences of `equipment_items`, `package_id`, `grinderModelMatchSql` or `LOWER(TRIM`, and line 1194 is now `if (grinderModel.isEmpty()) {`. The header at `dialing_blocks.h:390-393` was updated; this comment was not, so the two now contradict each other and the survivor sends the reader to an unrelated line. (CLAUDE.md: comments justifying an optimisation "are claims with expiry dates".)

https://github.com/Kulitorum/Decenza/blob/3bc603a70fc13bcc4cf36757bbb7e4012169ef86/src/history/shothistorystorage_queries.cpp#L1410-L1416

4. `scope.historyReadable = false` is set when the **grinder-context** connection fails to open — a third read on a third connection. The field's contract, stated at `aimanager.h:43-45`, is "false = a history QUERY failed. A DIFFERENT failure from `!bucketKnown` — they describe two reads on two connections and either can fail alone." `emptyHistoryReason` tests `!historyReadable` first, so when the history query ran fine and legitimately returned zero rows for a new equipment package, the model is told "Shot history could not be read for this request… Do NOT conclude that the user has no history" instead of the correct new-equipment sentence built two branches down. Give the grinder-context failure its own flag, or drop the assignment and keep the existing `qWarning`.

https://github.com/Kulitorum/Decenza/blob/3bc603a70fc13bcc4cf36757bbb7e4012169ef86/src/ai/aimanager.cpp#L1198-L1202

<details>

- [Worth a look] The `### Setup:` gate at `aimanager.cpp:1365` tests grinder brand/model, basket brand/model and bean, but not `setupPuckPrep`, so a package whose only recorded component is puck prep builds a non-empty `describeEquipmentSet()` that is never emitted. One-line fix; narrow, since it also requires no bean brand/type. (score: 60)
- [Worth a look] `describeEquipmentSet()` in `aimanager.cpp` and the "brand + model + with + burrs" phrase built independently at `shotsummarizer.cpp:952` are the same formatting rule in two places. Introducing a canonical equipment-phrase builder was the natural moment to collapse it. (score: 55)
- [OK to skip] The `blockedByIndex` comment in `aiconversation.cpp` says "only a REAL thread blocks the restore", but an indexed key with on-disk state `Empty` also blocks. True today only because `indexStoredConversation()` never indexes an `Empty` key — an unstated cross-file invariant. (score: 55)
- [OK to skip] `shothistorystorage_queries.cpp:1193` hand-writes `COALESCE(s.equipment_id, 0) AS equipment_bucket` rather than reusing the helper. It is a projection, not a filter predicate, so it is a different call shape. (score: 50)

</details>

---

**Structural note, offered separately from the findings above.** All four issues are in the maintenance layer, not the algorithm, and that pattern is worth naming. `shothistorystorage_queries.cpp` currently expresses "which equipment does this shot belong to" five different ways: `equipment_id IN (SELECT package_id …)` (L228, L460, L1419); `COALESCE(equipment_id,0) = :equip` (L711, L1294, L1364, L1526); a correlated scalar subquery on grinder brand+model (L1730-1733); a `LEFT JOIN equipment_items` (L705, L1575, L2055, L2085); and the basket/puck-prep subqueries in `equipmentComponentsSql` (L668-675). Idioms 2 and 3 are not equivalent — the comment at L1726 says so itself, since two packages can share a brand+model — so the same shot can be in scope under one filter and out of scope under another.

`equipmentBucketSql` centralises the *fragment*, not the *decision*: it returns a 40-character string the caller must remember to concatenate with the right alias and separately remember to bind, and nothing forces a new query to call it at all. The comment directly above it records that an earlier attempt at a real helper was removed because "it returned `std::optional` and two call sites two lines apart picked opposite policies for nullopt" — but the fix for two callers picking opposite policies is to make the policy un-pickable, not to return the decision to seven hand-assembled call sites.

Issue 2 above is that gap becoming visible in the documentation. A scope value type that owns its predicate and its bindings together, so no caller writes equipment SQL or chooses whether to filter, would collapse idioms 1-3 and remove the class rather than this instance of it.

skialpine:
### Code review — addendum (5th issue)

A historical-context pass finished after the review above and found one issue that is a behaviour change rather than documentation drift. Correcting my own framing: the review said all findings were in the maintenance layer rather than the algorithm. That held for the first four and does not hold for this one.

5. `buildGrinderCalibrationBlock` narrows grinder-calibration history from grinder identity to exact package, unwinding the behaviour PR #1720 shipped for issue #1713 without saying so.

On `main` the query pools every package that shares the same physical grinder:

```sql
WHERE equipment_id IN (SELECT package_id FROM equipment_items
    WHERE kind = 'grinder' AND LOWER(TRIM(IFNULL(model,''))) = LOWER(TRIM(?))
      AND LOWER(TRIM(COALESCE(json_extract(attrs,'$.burrs'),''))) = LOWER(TRIM(COALESCE(?,''))))
```

This PR replaces it with `AND COALESCE(equipment_id, 0) = ?` bound to `cur.equipmentId`. A user with two packages on one grinder and different baskets now gets calibration history siloed per package — the same fragmentation #1713's `equipment_merge` tool and migration-35 heal were built to prevent, arrived at through a basket change instead of a re-typed model string.

https://github.com/Kulitorum/Decenza/blob/3bc603a70fc13bcc4cf36757bbb7e4012169ef86/src/ai/dialing_blocks.cpp#L1274-L1298

Two things in the PR's favour, so this is a question rather than an assertion of a defect. The narrowing carries its own stated rationale — the endpoint medians are pooled by `(batch, kbId)` before any pair is formed, so pooling across packages corrupts the endpoint rather than merely producing a pair that straddles two baskets, and the block publishes a number the user dials in. That is a coherent argument, and it may well be the right call given the PR's premise that baskets make genuinely different coffee.

What is missing is the reconciliation. The change never references #1713, so a reader cannot tell whether the prior behaviour was considered and overruled or simply not noticed. And the comment it deletes recorded a cross-system invariant that is now untrue: the old folding existed *specifically so this read would agree with* `EquipmentStorage::findPackageByGrinderIdentityStatic`, which still treats two packages sharing a grinder model and burrs as the same gear. Those two subsystems now disagree about what "the same grinder" means, and nothing in the tree records that they do.

Worth an explicit decision from the author before merge: is grinder calibration a property of the grinder or of the package? Either answer is defensible; both subsystems answering differently is not.

<sub>Follow-up to the review above.</sub>

skialpine:
### Re: issue 5 — resolved by measurement, the PR's narrowing is correct

I asked above whether grinder calibration belongs to the grinder or to the package. Checked it against real shot history rather than leaving it open. The narrowing this PR makes is right, and the #1713 tension I raised is not a real conflict.

Controlled comparison from live history — same bean (Sweet Bloom Hometown Blend), same profile (D-Flow / Q), same 18 g dose, same 36 g target, same grinder (Niche Zero), same burrs (63 mm Mazzer Kony conical), same puck prep. Only the basket differs:

| Basket | Shots | Grind range | Median | Typical time |
|---|---|---|---|---|
| Decent 18g Ridged | 33 | 8 – 10 | 10 | 26–33 s |
| Graph Coffee Stepped 58→46mm | 6 | 15 – 17.5 | 16.75 | 34–58 s |

The two distributions do not overlap — five full steps separate the coarsest Decent shot from the finest Graph shot. The Graph basket also runs *longer* at 17 than the Decent basket does at 10, i.e. much coarser grind and still more time, which is the 46 mm exit's restriction showing up directly in the data.

Pooling these by grinder identity yields a median of 10, because 33 of 39 shots are Decent. A Graph-basket user would be handed 10 — about 6.75 steps too fine — from a value that looks well-supported by a large sample. That is exactly the failure the endpoint-corruption comment at `dialing_blocks.cpp:1275-1288` predicts, and the measurement confirms it.

So the two mechanisms are answering different questions, and both are correct:

- `EquipmentStorage::findPackageByGrinderIdentityStatic` (#1713) resolves **accidental** forks — one real basket split across two package rows by a re-typed model string. Those should pool; the fragmentation #1713 fixed was an artifact.
- This PR's calibration scoping separates **genuine** difference — two real baskets whose grind settings are five steps apart. Those must not pool.

The suggested change is therefore documentation, not behaviour. The new comment justifies package scoping on endpoint-corruption grounds but does not say that grinder-identity pooling still governs the accidental-fork case, so the two reads appear to contradict each other on what "the same grinder" means. One sentence naming which question each answers — and a reference to #1713 in the calibration comment — would keep the next reader from re-litigating this.

Issues 1-4 in the review above stand as written.

skialpine:
### Correction to the structural note

My structural note on the review above said `shothistorystorage_queries.cpp` expresses one concept five ways and implied the five should collapse to one predicate. That is wrong, and acting on it would introduce bugs. Correcting it here so the note is not followed as written.

Checking each site's enclosing function, the five idioms are **three legitimate concepts**, not one:

| Concept | Correct scope | Sites |
|---|---|---|
| Advice scope | equipment package | grinder calibration, dial-in history, best-recent-shot, observed grind range |
| Grinder-wide derivation | grinder model, deliberately | `grinderModelMatchSql` (L1419), feeding `grinderWideNumericSettings`/`grinderWideRpms` |
| User display filter | whatever the user selected | `buildFilterQuery` (L228), `requestAutoFavoriteGroupDetails` (L1730) |

Two specifics I had wrong:

- **L228 is inside `buildFilterQuery(const ShotFilter&, …)`** — the user-facing shot-history filter. A user who filters on "grinder model = X" is asking for grinder-wide matching and should get it. Narrowing that to package would be a regression.
- **`grinderModelMatchSql` is grinder-wide on purpose.** `stepSize`/`rpmStepSize` must agree with the Grind quick-select widget's `grindStepForGrinder()`, an invariant documented at `docs/CLAUDE_MD/AI_ADVISOR.md:179`. Collapsing it would break the widget's agreement with the advisor.

My observation that idioms 2 and 3 are not equivalent still stands as a fact, but my framing of it as a defect does not: at the display and grinder-wide sites the divergence is correct by design.

The real problem is narrower and, I think, more useful: three distinct concepts are all expressed as ad-hoc SQL fragments with nothing naming which is which, so every maintainer at every site has to remember which one applies. That is a much better fit for what has actually gone wrong here than "five copies of one thing" was. Three named scope types — so that selecting the wrong scope fails to compile rather than needing to be caught in review — addresses it without touching the two categories that are already correct.

Issues 1-4 in the review, and the issue-5 resolution above, are unaffected.

skialpine:
Superseded by #1857, which solves the same problem — scoping the advisor's shot selection to the equipment package — as a value the reads cannot omit rather than a SQL fragment each one has to remember to apply.

Nothing here was wrong. A full trace of the seven selection points, the conversation-key path, import/migration and threading found no defect in the algorithm. The reason for replacing rather than fixing is that the review of this branch's final state found four issues and all four were in the maintenance layer — a comment pointing at a deleted duplicate, a comment describing SQL the function did not emit, a flag violating the contract above it, a domain doc that never named the feature. That is what nine rounds of correct patching leaves behind when the shape stays the same.

#1857 keeps the measured finding this branch established, including the basket comparison in the addendum above, and carries the invariant test that fails on `main` with `4 of 5 advice-scoped selections pooled`. It comes out 902 lines smaller in src, with the eleven touched files below where `main` sits without the feature at all.

Closing in favour of it.

skialpine:
### Correction to the closing comment above

I closed this saying #1857 "solves the same problem". That is not accurate, and I am correcting it rather than leaving it to stand.

#1857 replaces the **selection** half of this PR — the part that was structurally wrong and that generated the review rounds. It does not carry:

- the equipment-keyed conversation key (`aiconversation.cpp/.h` here, +203 lines; #1857 does not touch those files, so a dial-in thread started on one basket still continues into another)
- `describeEquipmentSet` and the `### Setup:` block that name the gear in the payload
- the explicit no-history block for a package with no history (`bucketKnown` / `historyReadable` / `userHasPackages` → `emptyHistoryReason`)
- the two system-prompt rules
- the MCP payload fields (`equipmentPackageId`, `basketBrand`/`basketModel`/`puckPrep`)
- the import/backup `packageIdMap` remap at three call sites
- the QML changes

The line-count comparison I posted on #1857 is therefore overstated: part of the difference is work that branch has not done, not work it avoided.

This branch (`advisor-equipment-grouping`, `3bc603a7`) is intact, so none of the above is lost. The intended path is to port that half onto #1857's `AdviceScope` foundation rather than rewrite it — but that is a decision for @Kulitorum, and reopening this PR is reasonable if the remaining work is better tracked here.

PR #1853, opened 2026-08-22
fix(restore): stop renumbering shots on restore, and remap the references that must move

skialpine:
## What went wrong

A restore renumbers every shot: `importDatabaseStatic` inserts source rows into the destination's `AUTOINCREMENT` sequence, so a shot that was id 1109 on the source becomes 1052 here. It already remapped `packages`, `bags` and `recipes` foreign keys onto the new ids.

AI conversation turns also store shot ids — in QSettings, at `ai/conversations/<key>/messages[].shotId` — and nothing remapped those. After a restore the stored id either resolves to nothing or, as observed in the field on 2026-08-21, to a **real but unrelated shot**. A metadata write from the advisor then landed on the wrong row and a user-supplied roast date was discarded with no error anywhere.

Tracing that turned up a second, independent defect. Merge mode reads the destination's existing uuids to skip duplicates:

```cpp
QSqlQuery uuidQuery(destDb);
uuidQuery.exec("SELECT uuid, id FROM shots");   // result never checked
while (uuidQuery.next()) …
```

A failed `exec()` leaves an empty set, which is indistinguishable from an empty database, so every source shot is inserted as new. Forensics on the affected device support that path: `sqlite_sequence.shots` equalled the max id exactly, and a `DELETE FROM` does *not* reset that sequence (verified empirically — only a DROP does), so the table was dropped and recreated rather than emptied.

## What changed

**One import result.** `importDatabaseStatic` takes an optional `ImportResult*` carrying `shotIdMap` (source → destination), `destShotsBefore`, and the imported/skipped/failed/remapped/cleared tallies. A skipped duplicate maps to the destination row it matched; a failed row is absent. All three call sites pass one. The summary log now states the pre-import count, so merging into an empty destination no longer logs identically to merging into a populated one.

**Merge integrity guards.** The pre-read result is checked; an independent `SELECT COUNT(*)` runs alongside it; and the import aborts *before the first insert* when a non-empty destination yields an empty pre-read. Both guards are inside the existing transaction — nothing is written. Replace mode is exempt: it clears the destination first, so an empty pre-read is correct there.

**One conversation importer.** The hand-rolled AI-conversation import existed in **three** copies (`databasebackupmanager.cpp`, `datamigrationclient.cpp`, `shotserver_backup.cpp` — the plan assumed two). They are replaced by one `importConversationsStatic()` that takes the id map, rewrites each turn's `shotId`, and drops the key when the source id is absent. With **no** map — a conversations-only import, e.g. a LAN migration with shots deselected — every id is cleared: those ids name a database this device does not have.

The backup endpoint needed reordering: its conversation block ran *before* the shot import, so there was no map yet. The payload is now stashed during the zip-entry loop and imported in the existing main-thread continuation once `success` and the map are known. The no-shots-db path stays synchronous and passes a null map.

**Stale ids stop being acted on.** `AIConversation::loadFromStorage` resolves the distinct turn shot ids against the database once and treats unresolvable ones as absent in memory — no write-back, so nothing is destroyed. `shotIdForTurn` returns 0 and `recentAssistantTurns` skips the turn, matching how a turn that never carried an id is already handled. `AIManager` now learns whether a metadata write landed rather than only warning.

**A refusal reads as a refusal.** All three surfaces report a refused import as a failure naming which counts disagreed and stating that nothing was changed, instead of a completed restore.

## Measured, not assumed

The added resolve pass in `loadFromStorage` was benchmarked against the **real tablet database** (1061 shots / 18.1 MB, pulled over `/api/backup/shots`) and a synthetic 4× copy (4244 shots / 45.1 MB), 200 runs per point: median 0.018–0.045 ms, worst 1.02 ms on a cold cache, flat against 4× the rows. Plan is `SEARCH shots USING INTEGER PRIMARY KEY (rowid=?)`. Numbers and the caveat — taken on the Mac against the tablet's file, not on the tablet's own ARM hardware — are in a comment at the call site. No hoist needed.

## Settings sweep

Only one other settings key stores shot ids: `migration16/pendingVisualizerSync` (`maincontroller.cpp:709`, `:4896`), an array of `{shotId, visualizerId}`. It does **not** cross an import — `SettingsSerializer::exportToJson` is an explicit per-property allowlist, not an `allKeys()` sweep, so the key is never exported or restored and stays device-local. SAW learning stores per-profile/basket keys and lag values, no shot ids. No shot-badge persistence outside `shots.db`. AI conversation turns were the only stored shot-id reference that crosses an import.

## Tests

11 new slots across `tst_shotimportdedupe.cpp` and `tst_aimanager.cpp` — no new test file. They cover the id map (imported / skipped-duplicate / failed row), the merge guards, the remap and the clear-with-no-map path, `recentAssistantTurns` skipping an unresolvable turn, a failed metadata write reporting failure, and a regression case where the source ids exceed everything in the destination — the shape that resolves to a real but unrelated shot.

Each was mutation-checked. Three mutations, all caught with the right message; mutation 2 (keep an unmapped source id instead of dropping it) went red on a turn still carrying source id **1096**, reproducing the reported defect exactly.

**One branch is deliberately untested.** "Non-empty destination with an empty pre-read is refused" cannot be reached without stubbing: the count and the pre-read hit the same table in the same transaction, so they cannot be made to disagree. CLAUDE.md treats needing fault injection to reach a branch as a stop sign, so no faked test was written. The guard stays — the field evidence for that state is real — and is documented as untested in `tasks.md` 6.1 and `design.md`.

## Drive-by

`tst_beanbaseclient::inFlightSupersedeEmitsSingleTerminalSignal` used `QTest::qWait(600)` as a gate for two *events* (the request reaching the wire, the terminal signal arriving) — a timer standing in for an event, which CLAUDE.md bans — and lost the race under parallel load. Now `QTRY_COMPARE_WITH_TIMEOUT` / `QTRY_VERIFY_WITH_TIMEOUT`, with a short explicit settle window kept only for the negative assertion, where proving a second signal never arrives inherently needs one.

## Verification

Full suite via Qt Creator: **113 passed, 0 failed, 0 skipped in 41670 ms**. `check_log_markers.py` and `check_test_source_duplication.py` both clean.

Wiki manual updated (`Kulitorum/Decenza.wiki` `f38d97d`) — two sentences under **Backup and Restore**.

Still outstanding, manual on real hardware: restore a real backup onto a populated database (tasks 7.2) and a LAN migration with only AI conversations selected (7.3). `openspec archive fix-restore-id-remap` (7.5) lands as the last commit before merge.


skialpine:
## Review pass applied (`8d0b4229`)

Five review agents ran against `4fbbec0f`. Every finding was verified against the source before acting. The important ones were mine, and the worst was self-inflicted.

### The repair reproduced the defect it was fixing

`existingShotIds()` returned an empty `QSet` both when the database was not ready **and when its query failed** — the exact shape GUARD 1 refuses eleven hundred lines earlier ("a failed read is NOT evidence of an empty destination"). The caller's response to "this id resolves to nothing" is to **delete** the reference, so one transient `SQLITE_BUSY` would have stripped every advisor-to-shot link on the device. The header even blessed it as "the conservative direction"; deleting on an unanswered question is the aggressive one.

Now `std::optional<QSet<qint64>>`. `nullopt` means *could not answer* and the caller leaves the data alone.

### The drop was not reversible, and the comment said it was

`loadFromStorage` mutates `m_messages`; `saveToStorage` writes `m_messages` verbatim. So the next save — a follow-up turn, or `AIManager::switchConversation` saving the outgoing conversation — persists the drop to QSettings. Two comments claimed the opposite, and that claimed reversibility was what made the defect above look survivable. Corrected; the behaviour is acceptable *because* the drop is now conditional on a real answer.

### This change's own defect, one layer up

`DataMigrationClient::m_shotImport` was never reset while every sibling counter was. A second import — `importOnlyAIConversations()`, or a reconnect to a **different** source device — would remap that run's turns through the previous run's map. One line in `startImport()`.

### Also fixed

- `/api/backup/restore` answered `success: true` on a refused import, and nothing read `shotsRefused` — a user whose whole history was refused saw a green "Restore complete". The page now renders it as an error.
- "Nothing was changed." was **untrue** in merge mode, where profiles, settings and media still restore. Scoped to the shot history.
- `m_pendingMetadataWrites` was a `QSet`, but one reply drives two writes for the same shot id — the first outcome consumed the entry and the second, possibly the failing one, was discarded as "not ours" by the filter added to catch failures. Now a refcount.
- `shotMetadataCaptureFailed` had no consumer at all. Wired to the storage toast with advisor-specific wording, because the generic "please try again" is wrong advice for a shot that does not exist.
- `ImportResult::referencesRemapped`/`referencesCleared` were **dead stores** into a by-value lambda copy, duplicating `ImportTally`'s own fields. Deleted, with the `mutable` that existed only for them.
- Counters disagreed with the map on the replace-mode abort path while a comment asserted they could not. `destShotsBefore` is `std::optional<int>` so replace mode stops logging "destination held 0 before" against a populated database.
- Pre-existing silent failures in the same function, per the fix-what-you-touch rule: a failed samples query read as "this shot has no samples" (shot imports, no graph, nothing says why), a failed phases query used as a schema probe (a locked database read as "old schema"), and an unchecked `lastInsertId()` that could put a mapping to **0** in the map — an entry that exists and names nothing.
- Skipped and malformed conversations are counted rather than dropped silently ("3 imported" when 40 arrived).
- The backup endpoint lost conversations entirely if the server was destroyed mid-restore — moving the import below the `*destroyed` guard cost a write the old in-loop version survived.
- A plain shot-import failure in LAN migration fell through every arm and the dialog printed "Import complete".

### GUARD 2 was unreachable — replaced, not kept

`shots.uuid` is `TEXT UNIQUE NOT NULL` with no migration adding it nullable, so a complete read yields exactly `COUNT(*)` distinct keys: `destShotsBefore > 0 && existingByUuid.isEmpty()` is **mutually exclusive by schema constraint**. It is now an equality check, which covers that case as a subset and additionally catches a **truncated** read — `next()` returning false is how end-of-rows and a mid-scan error look the same, and a mid-scan `SQLITE_BUSY` during a restore can actually occur.

I also corrected this PR's own justification for it. `tasks.md` 6.1 said "the field evidence for that state is real"; `design.md`'s conclusion in the same change is that the table was **dropped and recreated**, so it was genuinely empty and the old guard would not have fired on 2026-08-21 either.

### The mutation record was wrong

6.5 claimed a mutation where "making `loadFromStorage` skip the resolve pass" turned a slot red. That was not reproducible: **no test wired a `ShotHistoryStorage` into an `AIManager`**, so the entire resolve block was unreachable by the suite and could have been deleted without one test failing. That block is the change's only remedy for installs already carrying the damage, and it shipped unexercised.

Fixed rather than reworded. Five new slots, no new file:

| Slot | Catches |
|---|---|
| `existing_shot_ids_answers_exactly_and_refuses_when_it_cannot_look` | bind-order/placeholder regression; the `nullopt` contract |
| `loadFromStorageForgetsTurnIdsThatNoLongerNameAShot` | the repair itself, real storage wired |
| `loadFromStorageLeavesTurnIdsAloneWhenTheDatabaseCannotAnswer` | the leave-alone path, with the user-visible consequence |
| `aRatingWrittenToAMissingShotReportsFailure` | the failure actually reaching a consumer |
| `twoWritesToOneShotBothReportTheirOutcome` | the QSet-vs-refcount bug |

Task 6.4's actual requirement — a **real** `ImportResult.shotIdMap` fed into `importConversationsStatic`, then read back from storage — is now `aRealImportMapCarriesStoredTurnsOntoDestinationIds`. Writing it exposed a fixture trap worth recording: an *empty* destination restarts `AUTOINCREMENT` at 1 and hands out the source's own ids, so every mapping is an identity and an un-remapped turn is indistinguishable from a remapped one. The destination is padded to 60 rows so that cannot happen. The weaker `source_ids_beyond_destination_are_all_remapped` slot was folded into the map slot as one identity assertion.

Mutation-checked again: making `existingShotIds` return a bare empty set turns **two** slots red — one on the type, one showing all four references deleted. Reverted; `grep -c MUTANT` is 0.

### Verification

113 passed, 0 failed, 0 skipped, 0 warnings. `check_log_markers.py`, `check_test_source_duplication.py`, `check_translation_key_conflicts.py`, `check_translated_richtext.py` clean; qmllint gate 239/239.

Still outstanding and manual: tasks 7.2 (restore a real backup onto a populated database) and 7.3 (LAN migration with only AI conversations selected). `openspec archive` lands as the last commit before merge.

skialpine:
## Second review round applied (`d15180a6`)

Five agents again, pointed at both commits and told to verify the round-1 fixes rather than re-report them. Two of those fixes were themselves defective, and one was **worse than the code it replaced**.

### A refusal was destroying the links it exists to protect

All three surfaces treated "the shot import was refused" as "no shots came with these conversations", and imported with every `shotId` cleared. The guards fire on a **transient** condition — a mid-scan `SQLITE_BUSY` — so the user's natural next move is to run the restore again. By then the cleared conversations have consumed their keys: the retry hits `existingKeys.contains(key)`, skips them as duplicates, and the linkage is gone permanently. The user does everything right and still loses exactly what this PR exists to preserve.

A refusal now imports nothing and leaves the archive intact for the retry. `integrityFailure` was already there to tell the two cases apart; I had collapsed them with a comment saying "three ways to get there and all want the same answer". They don't.

### Two holes in the repair itself

**`saveToStorage` does not write `m_messages` verbatim** — the claim the whole "the drop is permanent" argument rested on. Its reconcile branch adopts another writer's on-disk copy when that copy is longer, which is the MCP `ai_advisor_invoke` path, and that copy still carries the stale ids. So the repair was silently undone on precisely the installs that have a second writer. Dropped ids are now remembered in `m_forgottenShotIds` and re-applied after reconciliation.

**The repair never ran for the conversation restored at startup.** `AIManager` loads the most recent conversation in its own constructor, before `MainController` wires the storage, so the guard was false and the conversation the user is most likely to continue loaded unrepaired on every launch. Round 1 recorded this as a *performance benefit* — "free today because `m_shotHistory` is still null" — which is how a functional hole stops being re-examined. Now `AIConversation::repairStaleTurnShotIds()`, called again from `setShotHistoryStorage`.

### Round-1 fixes reverted

- **The `*destroyed` rescue** wrote the stashed conversations with a `nullptr` id map while `success` and `shotImport` were both in scope — stripping links the map right beside it could have preserved. Worse than the bare `return` it replaced, because a written conversation consumes its key and the next restore skips it. And it could not run at all: the continuation is delivered queued to `this`, and `~QObject` removes posted events for a destroyed receiver. Deleted; the comment now says what actually happens at teardown.
- **The `newId <= 0` guard** `continue`d *after* a committed INSERT, leaving a shot with no samples and no phases, counted `failed` while present in the database, absent from the map (which means "clear the reference" for a shot that exists), and skipping the rollback every other insert-side failure performs in replace mode. It made the log and the database disagree to guard a state `INTEGER PRIMARY KEY` forbids.
- **`conversationsSkipped` / `conversationsMalformed`** were the same producer-only-field shape round 1 had me delete from `ImportResult`, re-added one layer up. Demoted to locals — and they now also catch a conversation whose `messages` is not an array, which used to import as an empty conversation and be reported as a success.

### Also

- The JS `catch` arm still showed a green "Restore complete". The entire refusal check runs inside that `try`, so a parse failure undid the round-1 fix one path over. The page now gates on the server's own `success` field instead of re-deriving the verdict.
- `hasShotsDb && !m_storage` answered `success: true` with the whole shot history unimported — round 1 fixed only the async branch.
- The migration dialog still printed "Import complete" beside its error. My round-1 commit claimed otherwise; it had only added a second contradicting line.
- The generic "Couldn't save your shot changes — please try again" toast was still **announced in full** to screen-reader users before the advisor message replaced it visually. Retrying cannot help when the shot does not exist, so the wrong advice reached exactly the users who could not see it overwritten.
- The null-map decision is one definition (`ImportResult::idMapOrNull()`), and the producer clears the map on failure so a rolled-back import cannot hand back ids the rollback erased. The three callers were correct only by three *different* local invariants, one of them holding because of a gate 300 lines away in another function.

### Tests

`twoWritesToOneShotBothReportTheirOutcome` now drives both real producers instead of writing the refcount by hand, and asserts the premise (one reply → two writes → same shot id). As written it would have stayed green if `maybePersistBeanCorrectionFromReply` stopped registering its write entirely — and that slot was its only coverage.

Two new slots: the shared importer's **skip-existing-key** branch, which was dead in the suite because every tally slot clears settings first, on code that was just collapsed from three hand-written copies; and the reconcile regression above.

**Mutation-checked, including one that was not caught.** Bean-correction producer stops registering → red at 1-vs-2. `saveToStorage` stops re-applying the drop → red with stale id **987654** restored, the exact defect. The third — producer stops clearing `shotIdMap` on failure — was **not** caught, and that is recorded at the assertion rather than papered over: both guards fire before the first INSERT, so the map is empty on a refusal whatever that line does. The state it guards needs one row to succeed and a later one to fail, which no honest fixture produces. The line stays; it is documented as untested.

### Verification

113 passed, 0 failed, 0 skipped, 0 warnings. All five build-free gates clean; qmllint 239/239.

`design.md` corrected too — it still carried "Nothing is written back" and the old GUARD 2 shape.

Still outstanding and manual: tasks 7.2 (restore a real backup onto a populated database) and 7.3 (LAN migration with only AI conversations selected). `openspec archive` lands as the last commit before merge.

skialpine:
## Live testing on the real database — and a root cause one level down (`3296e282`)

Run against the maintainer's own Mac database: 1135 shots, 18.7 MB.

### What passed

| Check | Result |
|---|---|
| Merge onto a populated DB | `destination held "1135" before; 0 imported, 1135 skipped, 0 failed` |
| Rewritten equality guard, real data | `1135 shot(s); 1135 matched for de-duplication` |
| `destShotsBefore` as `optional` | replace logged `"(not measured, replace mode)"` — the old code printed `0` against a DB holding 1135 |

### The defect, reproduced on demand

A replace restore moved every shot from ids **1–1172 to 1173–2307**. Shot `275d16f4` went 1172 → 2307. Same mechanism as 2026-08-21.

### Then the question this PR had no answer for

> *"why would we want to renumber everything just to renumber it"*

There isn't one. `importDatabaseStatic` simply omitted `id` from its INSERT, so AUTOINCREMENT assigned a fresh one per row — and replace mode clears with `DELETE`, which does not reset `sqlite_sequence`. Restoring a backup into the database it came from renumbered all 1135 shots into a fresh block. Nothing wanted those numbers, and a renumbered shot is precisely what makes an outside reference go stale.

**I built a map to follow the renumbering. The better fix is not to renumber.**

- **Replace** — destination is cleared, so every source id is free: the database comes back with the ids it was backed up with, and the sequence is realigned so the next shot continues the restored history rather than skipping the old high-water mark.
- **Merge** — existing rows untouched; an incoming shot keeps its own id unless that id is occupied.
- **Collisions** go above every id in use in either database.

That last point was wrong in my first attempt, and the new test caught it: letting AUTOINCREMENT place a relocated row meant that with destination ids 1–2 and source ids 1–3, relocating source 1 took id **3** — so source 3, whose own id was free, got renumbered to 5. A relocation must never consume an id a later row is entitled to.

The id map stays and is still needed for genuine collisions and cross-device migration. It is now usually an **identity**, which shrinks the blast radius of any gap in the remap instead of depending on the remap being complete.

### Two more found only by running it

**The `beanbase_id` backfill has been failing on every import.** Five rows hold `beanbase_json = ''`, and `json_extract('', '$.id')` is a hard error that aborts the entire `UPDATE` — so no row was ever backfilled from an import. Marked only by a `qWarning`. Guarded with `json_valid()`. Pre-existing, in a function this change rewrites.

**The restore dialog's merge switch was a trap.** The label was a status line ("Merge with existing data") while the switch beside it was `checked` for **replace** — so an unchecked switch next to the word "Merge" meant a merge was about to run. The maintainer misread his own dialog during this session, on a control that can delete every shot on the device, and reported a "clean restore" that the log shows was a merge. The `accessibleName` was already correct, so this was a sighted-user-only failure. Label is now static, switch ON = merge, default unchanged.

### Spec

Not renumbering is now a stated requirement with scenarios for both modes, and the remap requirement is phrased around the binding staying correct rather than the id changing. `openspec validate` clean.

113 passed, 0 failed, 0 warnings. All five text gates plus qmllint 239/239.

Remaining: re-run 7.2/7.3 against an app rebuilt with id preservation, and the end-to-end conversation remap (plant a turn naming a real shot → Backup Now → restore → confirm it still resolves).

skialpine:
## Live verification complete; archived and ready for review

### 7.3 — LAN migration, AI conversations only, from the tablet

```
importConversationsStatic: 5 conversation(s) imported, 0 already present, 0 malformed;
                           0 shot reference(s) remapped, 36 cleared
                           (no shot import accompanied them — all ids cleared)
```

Verified against the plist rather than the log:

| Conversation | Turns | shotIds |
|---|---|---|
| `37d5711c`, `62c36897`, `b8efc1a7` (local) | 2 each | 1152 / 1155 / 1152 — **kept** |
| `19f4f47c`, `5e0bda29`, `6d4926f1`, `89cd08ec`, `f78e696a` (tablet) | 56 total | **none** |

Both halves of the requirement: an id naming a database this device does not have is cleared, and an existing local reference is left alone. Without the change those 36 tablet ids would have been written verbatim and resolved against this Mac's own 1135 shots — attributing advice to unrelated coffee.

### 7.2 — replace restore on the real database

All 1135 id→uuid pairs came back **byte-identical** to the pre-test snapshot (ids 1–1172, sequence realigned, shot `275d16f4` back at 1172 from 2307), with every sample and phase row still joining (1135/1135, 3757/3757).

### The defect reproduced, unprompted, on real data

A planted fixture was a no-op — QSettings maps `/` to `.` in the macOS plist, so `defaults write "ai/conversations/…"` writes keys the app cannot see. What the startup repair actually fired on was one of the maintainer's **own** conversations holding `shotId: 2288` — an id that existed only inside the window where an earlier test restore had the database renumbered to 1173–2307. The advisor ran during that window, the next restore put the ids back, and the reference was orphaned. `repairStaleTurnShotIds` caught it through exactly the deferred path added for the startup case.

That is the reported defect, end to end, on real data — and the clearest argument for not renumbering in the first place. The link was recovered (2288 → 1152, corroborated by the conversation's context label matching the shot's `D-Flow / Q` profile).

### One more bug, found by reading the log instead of trusting it

The clean-outcome line I added printed `0 turn shot reference(s) all resolve` for a conversation holding two that did — it read the set *after* it had been reduced to the unresolvable ids. Fixed, and a `QTest::ignoreMessage` now pins the wording and the number, so the log line is an assertion rather than decoration. Second count I got wrong in this change; this one can no longer regress silently.

### State

- **74/74 tasks complete.** Archived as `2026-08-22-fix-restore-id-remap`; 8 requirements promoted (one new capability, `restore-merge-integrity`). `openspec validate --specs --strict` — 153 passed, 0 failed.
- **113 passed, 0 failed, 0 warnings.** Five text gates plus qmllint 239/239.
- Seven commits, all pushed.

Ready for your merge decision.

PR #1854, opened 2026-08-23
Stop a moving weight sample from settling on stale stillness

skialpine:
`onWeightSample` measures how long the scale has been still **before** it accounts for the current reading, and the fast path then completes settling **at** that reading. A sample that moved the scale therefore inherited the stillness of the samples before it — a disturbance arriving after a second of stillness logged `stable for 1007 ms` and recorded the disturbed weight as the settled one.

The fix is one line: zero the stillness run on any sample that changed by the 0.1 g threshold.

### What reaches that branch is wider than it looks

The cup-removed guard above it tests **drops only**, and both its clauses are gated on the weight having passed 20 g:

```cpp
bool cupRemoved = (m_weight > 20.0 && weight < m_weight - 20.0) ||
                  (m_settlingPeakWeight > 20.0 && weight < m_settlingPeakWeight - 20.0);
```

So every *increase* arrives at the fast path at any size, and a genuine cup lift arrives on any shot whose yield never passed 20 g — a ristretto, a small stop-at-weight target. That is the case where this matters most.

### Why the test was flaky, and why that was the harness

`feedPlateauUntilCaptured` fed byte-identical samples, which is what armed the fast path: with `delta` pinned at 0 the stillness clock ran across the whole loop, and whether it crossed `SETTLING_STABLE_MS` before the test reached its disruption sample was a race against machine load. Samples now alternate ±0.1 g — above the 0.1 g change threshold so the fast path stays disarmed, below `SETTLING_ABOVE_AVG_MARGIN` so the rolling-average gate is unaffected — and the loop is bounded by sample count rather than wall clock. The `QVERIFY2` that attributed the failure to a loaded machine is gone.

Jittering the plateau makes the existing test unable to reach the fast path, so `movingSampleDoesNotSettleOnStaleStillness` covers the fix directly. It ages `m_lastWeightChangeTime` rather than sleeping, so it is deterministic and off the suite's wall clock. **Verified to fail with the one-line fix reverted.**

### Known, not fixed here

A cup lifted on a sub-20 g shot still settles at ~0 g *and trains SAW on it* — `qAbs(overshoot)` of ~18 is under the 20.0 rejection threshold, and the completion-time cup-removal guard is gated on the same 20 g. That needs the `shot_eval` regression corpus to change corpus-tuned thresholds, so it wants its own change.

Split out of #1852, where it was found via a flaky test and had nothing to do with the AI advisor.

Full suite green locally (113/113).

### DIFF
--- a/src/controllers/shottimingcontroller.cpp
+++ b/src/controllers/shottimingcontroller.cpp
@@ -335,6 +335,24 @@ void ShotTimingController::onWeightSample(double weight, double flowRate, double
         if (delta >= 0.1) {
             m_lastStableWeight = weight;
             m_lastWeightChangeTime = now;
+            // The run of stillness ends AT this sample, so it is 0 ms long — not
+            // the length of the run that preceded it. `stableMs` was sampled before
+            // this update, and the fast path below completes settling at `weight`,
+            // so leaving it stale let a sample that MOVED the scale be recorded as
+            // the settled weight, logging "stable for 1007 ms" for a run that had
+            // ended one sample ago.
+            //
+            // What reaches here is any change of at least 0.1 g the cup-removed
+            // gate above did not claim — and that gate is narrower than it looks.
+            // It tests for DROPS only (`weight < m_weight - 20.0`), and both of its
+            // clauses are additionally gated on the weight having exceeded 20 g. So
+            // every INCREASE arrives here regardless of size (a second cup or a
+            // portafilter set on the drip tray), as does a genuine cup lift on any
+            // shot whose weight never passed 20 g — a ristretto, a small SAW target.
+            // An earlier draft claimed a real cup lift "cannot happen" here; it was
+            // wrong in both of those directions. Found from a suite failure, not
+            // from #1280 — that symptom is attributed to the cup-removed path.
+            stableMs = 0;
         }
 
         // Calculate rolling average
--- a/src/controllers/shottimingcontroller.cpp
+++ b/src/controllers/shottimingcontroller.cpp
@@ -332,6 +332,15 @@ void ShotTimingController::onWeightSample(double weight, double flowRate, double
         // Also track per-sample changes for the old-style fast path
         double delta = qAbs(weight - m_lastStableWeight);
         qint64 stableMs = now - m_lastWeightChangeTime;
+        // How long the run this sample BROKE had lasted, or -1 if it broke none.
+        // `stableMs` is zeroed below on a moving sample, so without this the log
+        // reads `stable: 0 ms` and a submitted log can no longer show how long the
+        // scale had been still before it moved. Only a NEAR-MISS is worth saying:
+        // during a drip most samples move, and each one's broken "run" is just the
+        // sample interval, so reporting every one would add ~100 redundant lines a
+        // shot to the logs this subsystem is diagnosed from. Half the threshold is
+        // the point past which the sample stopped a settle that was under way.
+        qint64 endedStillMs = -1;
         if (delta >= 0.1) {
             m_lastStableWeight = weight;
             m_lastWeightChangeTime = now;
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 4 issues:

1. Bare `1000` ms literal in the branch this PR edits, and its mirror in `onDisplayTimerTick`. `SETTLING_STABLE_MS = 1000` is a named constant in `shottimingcontroller.h` and is used correctly at three other sites; these two fast paths hardcode it and must be kept in sync by hand. A reviewer flagged this exact pattern in this same function in [#689](https://github.com/Kulitorum/Decenza/pull/689#issuecomment-4207475466). (CLAUDE.md says "Centralize anything produced at more than one site" and "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'.")

https://github.com/Kulitorum/Decenza/blob/b09766752ed9c69e191c425cc1d5dcc42c2b04b9/src/controllers/shottimingcontroller.cpp#L370-L372

https://github.com/Kulitorum/Decenza/blob/b09766752ed9c69e191c425cc1d5dcc42c2b04b9/src/controllers/shottimingcontroller.cpp#L508-L510

2. The fast path now has only negative coverage. Every assertion in `tst_settling.cpp` that touches it asserts settling did *not* complete — the new test, `feedPlateauUntilCaptured`'s second `QVERIFY2` (whose jitter exists precisely to keep the fast path from firing), and `weightAboveAvgGuardAllowsSettlementWhenStable`. `tst_settling.cpp` is the only test file driving this code, so nothing anywhere asserts the fast path ever *completes*: make `stableMs = 0` unconditional and the suite stays green while SAW shots would only settle via the avg path or the 10 s timeout. Three lines on the end of the new test (age the clock, feed a byte-identical sample, assert `!isSawSettling()` and weight == 44.0) pin the other half. (CLAUDE.md: "Before keeping a test, break the code it covers and watch it go red.")

https://github.com/Kulitorum/Decenza/blob/b09766752ed9c69e191c425cc1d5dcc42c2b04b9/tests/tst_settling.cpp#L539-L544

3. `cupLiftMidSettlePreservesLastStableAvg_1280` still hand-rolls the fixed `qWait` budget the comment deleted by this PR warned about — 14 samples x `qWait(50)`, including a 7-sample run of byte-identical `42.3` readings that accumulates real stillness. If those overshoot, settling completes on the fast path before the cup-lift artifacts arrive, and the test passes *silently for the wrong reason*: `!isSawSettling()` holds because settling finished normally, and `currentWeight()` is 42.3, inside the asserted band. The PR converted the helper to a condition and left this one behind. (CLAUDE.md: "Never use timers as guards/workarounds"; "Fix pre-existing violations in any file you touch.")

https://github.com/Kulitorum/Decenza/blob/b09766752ed9c69e191c425cc1d5dcc42c2b04b9/tests/tst_settling.cpp#L294-L300

4. The helper comment claims the loop is "bounded by SAMPLE COUNT, not by wall clock". That is true of the fast path only. `m_settlingAvgStableSince` keeps running across the jittered samples, so the rolling-average path can still complete settling at `SETTLING_STABLE_MS`, and the cap of 40 iterations is >= 2000 ms of `qWait` — past that 1000 ms ceiling. The `QVERIFY2` below then misdiagnoses such a stall as "the fast path's stillness clock is accumulating". (CLAUDE.md: an un-sourced claim in a comment "gets believed and then licenses wrong code.")

https://github.com/Kulitorum/Decenza/blob/b09766752ed9c69e191c425cc1d5dcc42c2b04b9/tests/tst_settling.cpp#L48-L53

The core one-line production fix itself checks out: I traced that zeroing on `delta >= 0.1` cannot make settling unreachable (`m_lastWeightChangeTime = now` already ran on that same branch pre-PR, so a dithering scale never accumulated toward 1000 either), the byte-identical fast path is untouched, the reset is correctly ordered above both consumers of `stableMs`, and the new test genuinely fails when the fix is reverted. `git blame` also confirms this restores the pre-`e7168b9` semantics that refactor dropped, rather than reverting a deliberate decision.

<details>

- [Worth a look] The `[SAW] Settling:` line prints `stable: 0 ms` on every moving sample now, since it is emitted after the reset — a post-mortem log can no longer show how long the scale had been still before it moved, and this subsystem is diagnosed from user-submitted logs. Printing the ended run's length under a separate key preserves it. (score: 72)
- [Worth a look] The reworded `QTest::qWait waits AT LEAST its argument` comment asserts Qt behaviour with no file:line citation; CLAUDE.md requires one, and the claim is correct (`qtestsupport_core.cpp:137`), so it is a one-line addition. (score: 62)

</details>

PR #1857, opened 2026-08-23
Scope the advisor to the equipment package: selection, conversation, payload, import

skialpine:
Scopes the AI advisor to the equipment package — the shots it selects, the conversation it writes into, the payload it sends, and what survives an import. Replaces #1852, which solved a narrower version of the same problem and which this closes.

## The measurement behind it

From real shot history: one grinder, one set of burrs, identical puck prep, same bean, same profile, same 18 g dose, same 36 g target. Only the basket differs.

| Basket | Shots | Grind range | Median | Typical time |
|---|---|---|---|---|
| Decent 18g Ridged | 33 | 8 – 10 | 10 | 26–33 s |
| Graph Coffee Stepped 58→46mm | 6 | 15 – 17.5 | 16.75 | 34–58 s |

No overlap; five whole steps of clear air. Pooling yields a median of 10 because 33 of 39 shots are Decent, so a Graph-basket user is handed a setting about 6.75 steps too fine, behind a large and reassuring sample.

A grind number means nothing across a basket change. History from another package is not weaker data, it is misleading data.

## 1. The selection

`AdviceScope` (`src/history/shotscope.h`, 41 lines) is a value, passed as argument 3 to every advice-scoped read. It carries its predicate and its bucket together, so a call site cannot forget to filter, get the alias wrong, mismatch a placeholder style, or bind a null. Those were four separate things to remember at each of five sites.

The bucket is embedded rather than bound — it is a `qint64` from our own schema, so there is nothing to quote. The fixture helper in this change demonstrates why that matters: it bound a default-constructed `QString`, Qt turned it into SQL `NULL`, `x = NULL` was never true, and every fixture shot forked its own package.

**Scoped:** `loadRecentShotsByKbIdStatic` (which `buildDialInSessionsBlock` delegates to), `buildBestRecentShotBlock`, `queryGrinderContext`'s observed settings and RPM axes, and `buildGrinderCalibrationBlock`.

**Deliberately not scoped:** `stepSize` / `rpmStepSize` stay grinder-wide, because the advisor's step must equal the Grind quick-select widget's `grindStepForGrinder()` on the same screen. `adviceScope_stepSizeStaysGrinderWide()` fails if that changes, so the fix cannot over-apply.

### Why a replacement rather than a fix to #1852

#1852 was correct — a full trace of its selection points, conversation-key path, import/migration and threading found no defect in the algorithm. It went through nine rounds of review anyway, and the review of its final state still found four issues, all in the maintenance layer rather than the logic: a comment instructing maintainers to sync a duplicate that had been deleted, a comment describing SQL the function did not emit, a flag violating the contract written directly above it, and a domain doc that never mentioned the feature it documented.

That pattern is the diagnosis. Equipment scoping was expressed as a SQL fragment every query author had to remember to apply. Each review round fixed one site's memory lapse; nothing made the next one impossible, so the rounds could not converge.

### Grinder → equipment

The grinder code predates equipment packages, and packages were bolted on beside it. Calibration still rebuilt a grinder-identity subquery inline and threaded `grinderModel` + `grinderBurrs` through four call sites to feed it. It is now package-keyed, which removes `grinderBurrs` from the signature, the burrs column and local from `AIManager`'s context query, and `grinderModelMatchSql`'s "one copy is deliberately NOT collapsed" paragraph — the copy it pointed at is the one deleted.

This is not a reversal of the identity folding in #1713 / #1720. `findPackageByGrinderIdentityStatic()` answers "are these two rows the same real package" and still merges accidental forks. `AdviceScope` answers "may these shots be compared", which baskets decide. The two only looked contradictory because neither said which question it was answering.

## 2. The conversation key

Scoping the selection alone does not repair the reported case. A saved thread **replays its stored turns on every request**, so a transcript that already discusses another basket keeps informing every future answer no matter how the next payload is built.

The key is therefore derived from the shot including its package (`ConversationKey::derive`). One bean and profile can now hold several threads, one per package; the index snapshots the package label and id so the app, the ShotServer page and `ai_conversations` can tell them apart, through one producer rather than four copies. Pre-upgrade conversations are dropped by the existing one-time wipe — the key change already orphans them, and this stops dead threads holding index slots.

## 3–4. What the payload says

- The identity that reaches the model is the whole package. `ShotIdentity::fields()` is one table naming both ends of each mapping, walked by the hoist, the session-context emit, the per-shot override emit and the fill from a shot. Adding a component is one row. Previously the advisor's read joined the grinder alone, so a basket switch arrived as one setup.
- When nothing matches, the payload carries `noDialInHistory` — the package named, count zero, and an instruction to judge the shot on its own data. An omitted section reads as "no history at all", and an unanchored model supplies its own anchor: the reply that prompted this change cited a "70/100 shot" that appears nowhere in its context.
- "The query matched nothing" and "the query failed" arrived here as the same empty list. They are now distinguished, and the absence block is emitted only for the first — a failure is not evidence that nothing matched.

## 5. System prompt

Two rules: cite only shots present in this context, and compare grind settings only within one equipment package.

## 6. Import and backup

A conversation key holds a package **row id**, and an import renumbers those rows. The archive's key was written through verbatim, producing a thread in the index that no shot on the device could open — or, where the destination held a package with the source's id, one that opened on the wrong basket.

`ImportResult` now publishes `equipmentIdMap` beside `shotIdMap` (the map already existed inside `importDatabaseStatic` to remap shots and bags; it was never handed out), and the importer re-keys through it. Two shapes are refused rather than imported wrong: a key the conversation's own fields do not derive, and a package the map does not contain — bucket 0 would file a thread about one basket under "no basket", which is the mixing the key exists to prevent.

## 8. Auto-favourites

Same defect, different query. The grouping keyed on `equipment_id` in the "+ Grinder" modes and not in the default one, while the stats query mirrored that split — so a card counted one package and averaged another. Verified live: a 41-shot card became 34 + 7, and the 7-shot card's averages are exactly those 7. The details query's package condition is hoisted out of the per-mode branches so a new mode cannot skip it.

## 9. One payload, one format

Both surfaces build their system prompt from `shotAnalysisSystemPrompt` — ~45,000 characters that name the structured blocks 35 times and tell the model to read them by field path. `ai_advisor_invoke` sent that payload; the in-app advisor sent **prose**. One prompt, two formats, and the prose one delivered none of the paths the prompt described.

The in-app path now sends the same object, from the same assembler. That deletes the hand-rendered `### Setup:` header (which is why basket and puck prep reached one surface and not the other), and with it every reader that existed only to undo the wrapping: a question-recovery heuristic that searched for `"Here's my latest shot:"` and guessed whether the trailing text "looked like a question", and a hand-written brace-matching JSON scanner with seven regexes. Nothing had to be kept for old conversations, because the key change wipes them once.

`dialing_get_context` was the last hand-written copy of the block calls, kept for one difference — no cross-profile calibration (#1164). That difference is now an argument, so **one line** in the tree reads `equipmentId` off a live shot, and the MCP tool stops being the only surface with no `noDialInHistory`.

## Bundled: the test harness stopped exhausting the machine's ports

Not part of the feature, folded in to avoid a second PR. `tst_mcpserver_protocol` opened a fresh `QTcpServer` + `QTcpSocket` per request across 88 call sites, `tst_mcpserver_session` across 29, and `tst_mcpremoteaccess` opened 105 real connections to test a rate limiter. A full run left ~42,000 loopback sockets in `TIME_WAIT` and 96% of the ephemeral range held, after which *any* process on the machine failed to connect with `EADDRNOTAVAIL` — including the suite itself, which then hung to the 300 s QtTest watchdog and aborted. That is the "flake" that had aborted the suite four times in one day.

Measured, same test code either side: 895 ports → 14, 29 → 1, 105 → 23.

## What the review changed

Four review agents ran over the branch before it was ready. Their findings are in `ae4cba72`; the ones worth knowing about here:

**The keying was defeated on the app's main entry point.** `PostShotReviewPage`'s `clonePersistedShot()` was a hand-written whitelist of ~60 fields carrying `grinderBrand/Model/Burrs/Setting` and **not** `equipmentId`, basket or puck prep. That clone is what the AI Advice button hands the advisor, so every conversation opened from Post-Shot Review keyed to the unpackaged pool while the same shot opened from Shot History keyed to its real basket. The whitelist existed because QML cannot copy a `Q_GADGET` — `Object.assign` and spread copy own properties, and `Q_PROPERTY`s are accessors on the prototype — but `ShotProjection::toVariantMap()` already did that job completely and simply was not `Q_INVOKABLE`. It is now, and the whitelist is one line.

**Three silent losses**, all "the user is shown the survivors": an import that refused conversations reported only what it imported (worst on the conversations-only migration, where after this change every packaged thread is refused and the button had offered to import 40); and `dialing_get_grinder_calibration` asserted "no qualifying shots" as fact even when the query had failed. Refusal counts now reach all three surfaces with the remedy attached.

**Auto-favourites "Show"** filtered by grinder brand + model + grind setting, returning a strict subset of what the card aggregates. `ShotFilter` gains an exact `equipmentId`. The three grouping labels got new translation keys, since their English changed meaning and a reworded fallback keeps its old translation.

Also: `noDialInHistory` claimed a bean match the query never performed; the system prompt read an absent history block as "no query was asked", which is false when one failed; and `emitRecentShotContext` was a second assembler over `AdvisorContextBlocks` that had already fallen behind, serving a payload with no reader.

## Tests

113 pass. Highlights:

- One **invariant** over all five advice-scoped paths rather than one regression test per bug. Against `main` it reports `4 of 5 advice-scoped selections pooled`, naming each one and the settings it leaked.
- `adviceScope_theAssemblerDerivesItFromTheShotUnderReview` covers the single line that builds a scope from a live shot. It asserts both halves — no foreign package's settings, **and** at least one of the shot's own — because a scope of 0 matches nothing, and a leak check alone would call an empty payload clean.
- Five import tests: re-key onto the destination package, refusal of an unresolvable package, refusal of a pre-equipment archive entry, an unpackaged thread keeping its key, and an export/import round trip.
- A parity test that the two surfaces emit the same identity fields, which fails if either drifts.
- Coverage the review found missing at the edges: that `noDialInHistory` REACHES both payloads (both merge lines could be deleted with the suite green), that the equipment join's basket columns are asserted (deleting them reproduced the original defect and passed), that `equipmentIdMap` comes from the real importer rather than a hand-built map, and the `ok` out-param's false direction.

Every new test was verified by breaking the code it covers and watching it go red. Two worth naming: under a pooling predicate the calibration test reports Londinium at `rgs 5` — the midpoint of Decent's 2 and Graph's 8, not an obviously broken number but a confident and plausible one that is wrong for both baskets. And breaking the live scope construction turns exactly one test red, which is the gap that test was written for.

Two things that verification surfaced. The scope cannot be silently dropped: removing it from the query fails to compile under `-Werror,-Wunused-parameter`. And the calibration assertion was vacuous at first — `rgs` is a formatted string, reading it as a number gave 0 for every profile and passed, until a `checked > 0` guard was added.

Verified live on a real shot, not only in the suite (tasks 9.5 / 9.6): the conversation renders unchanged for the user, and the model's reply cites the tasting feedback, the profile's own temperature and pressure intent, the measured peak, and a specific prior shot at the same grind setting — every one a field path the shared system prompt names and the prose payload never delivered.

## Size

| | src lines |
|---|---|
| `main`, without the feature | 10,889 |
| The same eleven files, **with** the selection work | **10,866** |
| Including the new `shotscope.h` | 10,907 |
| #1852, same feature | 11,809 |

The eleven original files come out below a codebase that does not have the feature. Deletions that paid for it: `requestRecentShotsByKbId` and its signal (no caller in any surface, while three docs described it as live), the redundant grinder subqueries, the prose payload and its readers, and a test covering folding the query no longer performs.

## Docs

`AI_ADVISOR.md` gains a "Scoped to the Equipment Package" section; `MCP_SERVER.md` documents the scoped history, `noDialInHistory`, and the `equipment` field on `ai_conversations`. Several stale claims fixed in passing, including five citations of `getRecentShotsByKbId()` — a symbol removed as callerless — and a design-time "prerequisite" section gating a tool that shipped months ago.

No wiki entry: this changes how the advisor selects its own context, which nothing asks the user to do differently.

## Not included

`convertShotRecord` never copied `equipmentId` into the projection, so every loaded shot read as unpackaged; fixed here, because a scope built from it would otherwise be a silent no-op.

No test covers auto-favourites (section 8.5) — deliberately deferred as a short-lived feature.

Still open, tracked separately: `tst_blegattqueue::anOperationWithARetryBudgetIsReissuedUpToThatBudget` and `tst_scaleprotocol::difluidTiCommandsUseTiService` each failed once and passed on other runs during this work. Both are pre-existing and unrelated; neither is dismissed as noise.


skialpine:
### Correction: this is not a full replacement for #1852

The description above says this replaces #1852. That is wrong, and the comparison table in it is not apples to apples. This PR replaces the **selection** half. Roughly half of #1852's user-visible behaviour is not here:

- **Conversation threads are still not keyed by equipment.** `AIManager::conversationKey` remains `(beanBrand, beanType, profileName)`, so a dial-in thread started on one basket continues into another. #1852 changed this (`aiconversation.cpp/.h`, +203 lines) and this PR does not touch those files at all.
- **The equipment is never named in the payload.** No `describeEquipmentSet`, no `### Setup:` block, no basket brand/model or puck prep. The data is scoped correctly and the model is not told what changed, which is a strange combination: numbers move between conversations with no stated cause.
- **No explicit no-history block.** Fit a new basket and dial-in history is legitimately empty; #1852 said so out loud (`bucketKnown` / `historyReadable` / `userHasPackages` feeding `emptyHistoryReason`). Here the payload is simply empty, which is the silence that invites the model to fill it.
- Missing too: the two system-prompt rules, the MCP payload fields (`equipmentPackageId`, `basketBrand`/`basketModel`/`puckPrep`), the import/backup `packageIdMap` remap at three call sites, and the QML changes.

The size comparison in the description therefore overstates the case. Some of #1852's 920 src lines are work this branch has not done, not work it avoided.

What stands: the structural argument, the measured basket data, the invariant test, and the five selection points that are scoped. What does not: the claim that #1852 can be closed on the strength of this alone.

`advisor-equipment-grouping` still exists at `3bc603a7`, so the remaining half is recoverable and should be ported onto this foundation rather than rewritten.

PR #1863, opened 2026-08-24
Retire the pre-shot zero offset when the shot is saved

skialpine:
## The symptom

After a shot, the scale reads 0 and the app reads 0.2 g. Tapping the weight to tare does not help — the scale zeroes and the app still shows 0.2. Only restarting the app clears it.

Reproduced live: `scale_get_weight` returned 0.1 g with the scale sitting at -0.1 g, and the shot before it logged `[Scale] weight= "-0.8"` (raw) beside ScaleFeed's corrected `-0.6` — an offset of -0.2 g, exactly the discrepancy still on screen minutes later.

## The cause

`WeightProcessor` adopts a per-shot "pre-shot zero" offset at flow start (`markExtractionStart()`) — whatever the scale reads then, which after a tare is typically -0.1 to -0.5 g of residual and preheat drift. Every weight that shot has it subtracted, and `MachineState` mirrors it so the live readout, MQTT, MCP and the widget agree with what SAW stops on and what is saved as `finalWeightG`.

Nothing cleared it after the shot. `startExtraction()` and `resetForRetare()` were the only reset sites; neither `stopExtraction()` nor `endShotCycle()` touches it. So the offset outlived its shot and skewed every surface reading `MachineState::scaleWeight()` — idle readout, steam/milk weight, dose weighing in BrewDialog, MQTT, MCP, the home-screen widget — until the next shot armed or the app restarted. A manual tare could not shift it, because the tare moves the *scale's* zero while the app goes on subtracting the old number.

The invariant was already written down, on `m_preShotZeroOffset` itself: "every reset site must notify, or the surfaces mirroring it keep subtracting last shot's number." There simply was no reset site at the end of the shot.

## The fix

Clear the offset on `ShotTimingController::shotProcessingReady`.

That edge matters. `shotEnded` fires while drip is still settling, and the settle samples run through the same correction — clearing there, or at espresso-cycle exit, would step the graph and the saved `finalWeightG` by the offset in the middle of capture. `shotProcessingReady` is emitted after the SAW settle window, and the clear is queued into the worker thread, so it lands after the direct-connected `onShotEnded()` has already read the shot.

An aborted cycle that never flowed is unaffected: `startExtraction()` sets the offset to 0 and only `markExtractionStart()` (on `shotStarted`) ever adopts a non-zero one.

## Tests

Adds `tst_WeightProcessor::theShotsZeroIsRetiredWhenTheShotIsSaved`. No existing test asserted the offset is ever retired — the four existing pre-shot-zero tests all cover capture, not release. It asserts through `preShotZeroOffsetChanged`, which is the only channel to the mirroring surfaces: an offset dropped without notifying reproduces the bug exactly.

Verified it can fail: with `clearPreShotZeroOffset()` made a no-op the suite goes 112/1; restored, 113/113.

Full suite green locally (113 passed, 0 failed, 0 warnings), build clean.

### DIFF
--- a/src/machine/weightprocessor.cpp
+++ b/src/machine/weightprocessor.cpp
@@ -952,6 +952,11 @@ void WeightProcessor::setPreShotZeroOffset(double offsetG)
     emit preShotZeroOffsetChanged(offsetG);
 }
 
+void WeightProcessor::clearPreShotZeroOffset()
+{
+    setPreShotZeroOffset(0.0);
+}
+
 void WeightProcessor::markExtractionStart()
 {
     if (!m_active || m_extractionStartTime != 0) return;
--- a/src/machine/weightprocessor.h
+++ b/src/machine/weightprocessor.h
@@ -63,6 +63,17 @@ public slots:
     // flow never began, which stopExtraction (gated on shotEnded) misses.
     void endShotCycle();
     void resetForRetare();  // Clear LSLR buffer after auto-tare during preheat
+    // Drop this shot's zero correction once the shot has been saved. The offset is a
+    // PER-SHOT number: it belongs to the pour it was measured at flow start, and every
+    // surface that mirrors it (live readout, MQTT, MCP, widget) keeps subtracting it
+    // from the idle scale until something clears it. Nothing did — startExtraction()
+    // was the only reset, so a shot's offset survived into idle, steam and dose
+    // weighing, and a manual tare could not shift it because the scale zeroed while
+    // the app kept subtracting. Restarting the app was the only cure.
+    // Called on shotProcessingReady, which is AFTER the SAW settle window and the
+    // save: clearing at espresso-cycle exit instead would jump the drip-settle
+    // samples (and so the saved finalWeightG) by the offset mid-capture.
+    void clearPreShotZeroOffset();
 
 #ifdef DECENZA_TESTING
 public:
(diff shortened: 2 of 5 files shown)


skialpine:
Follow-up commit for two findings from reviewing the first one.

**Mid-pour disconnect never cleared the offset.** `shotProcessingReady` is reached only through `shotEnded`. A BLE drop during the pour sets `Phase::Disconnected`, emits `espressoCycleEnded()` and returns before the queued `shotEnded` ever fires, so `endShot()` never runs and the offset a flowing shot had already adopted stayed applied to the idle scale for the rest of the session — the original bug, reached by the same road the `endShotCycle()` wiring beside it already documents.

Now cleared from the `espressoCycleEnded` handler, guarded on `Phase::Disconnected`. The guard is the whole point: that signal is emitted *synchronously* on the normal `Ending -> Idle` exit while `shotEnded` is queued, so an unguarded clear there would precede the settle window and step the drip samples — and the saved `finalWeightG` — by the offset. A disconnect has no such window; the shot is dead and nothing further will be captured.

That guard depends on `m_phase` being assigned before the emit, which nothing asserted. `bleDropMidPourStillEndsTheEspressoCycle` now records the phase as it stands *inside* the emission, which a `QSignalSpy` cannot see. Swapping those two lines would silently stop the guard firing with nothing else failing; verified the assertion goes red when the emit is moved above the assignment.

**Header comment misstated the ordering.** It said `shotProcessingReady` is emitted after the save. It is what *triggers* the save — the clear lands after only because it is queued onto the worker while `onShotEnded` runs directly on the main thread. As written it invited moving the call to a direct same-thread site, which would clear the offset before the shot is read.

Suite green (113/113), build clean.

PR #1870, opened 2026-08-27
Fix Launcher Mode leaving Qt unloaded on Android (#1869)

skialpine:
## Problem

Four field reports over three releases, all the same Java `UnsatisfiedLinkError`, all of them an Android lifecycle callback reaching a `QtNative` native with Qt's libraries not loaded:

| Issue | Version | Qt | Crash site |
|---|---|---|---|
| #1239 | v1.7.5 (May 2026) | 6.11.1 | `QtActivityBase.onResume:193` → `updateApplicationState` |
| #1511 | v1.8.0 (Jul 2026) | 6.11.1 | same |
| #1512 | v1.8.0 (Jul 2026) | 6.11.1 | same |
| #1869 | v2.0.4 (Aug 2026) | 6.11.2 | `QtActivityBase.onNewIntent:372` → `onNewIntent` |

```
No implementation found for void org.qtproject.qt.android.QtNative.onNewIntent(android.content.Intent)
	at org.qtproject.qt.android.QtActivityBase.onNewIntent(QtActivityBase.java:372)
	at android.app.ActivityThread.deliverNewIntents(ActivityThread.java:3969)
	at android.app.ActivityThread.performResumeActivity(ActivityThread.java:4952)
```

## Root cause

Qt 6.11 added a branch to `QtActivityBase.onCreate` that skips loading the Qt libraries **outright** when the activity was launched through an activity-alias:

```java
if (isLaunchedAsAlias()) {
    Log.d(TAG, "Starting an alias-activity, skipping loading of the Qt libraries.");
} else { ...loadQtLibraries()... }
```

`isLaunchedAsAlias()` is a plain name comparison of the intent's component against the activity's own class (`QtActivityBase.java:169-178`).

Decenza's **Launcher Mode** (`app/launcherMode`, `settings_app.cpp:44`) enables the `LauncherAlias` activity-alias at runtime so the app can be the tablet's home screen. Every HOME launch therefore produces an activity with no Qt behind it — no libraries, no registered natives, no `startNativeApplication()` — and the first lifecycle callback into `QtNative` kills the process.

The branch is **new**: absent in `v6.8.0` and `v6.10.0`, present in `v6.11.1` (checked against code.qt.io). Decenza moved to 6.11.1 shortly before #1239, the first of the four reports, and Launcher Mode had been working since the alias got its `android.app.lib_name` metadata in #287. So this is a Qt regression we inherited, not a change of ours.

## Fix

**1. Retarget the alias launch (`retargetAliasLaunch()`, called first in `onCreate`).** Point the intent at the real activity so Qt's comparison matches and the libraries load normally. `QtLoader` reads `android.app.lib_name` from `getComponentName()`, not from the intent (`QtLoader.java:95-96`, `:525`), which `setIntent()` does not affect — and `DecenzaActivity` carries that metadata too, so library resolution is unchanged either way.

**2. A narrow guard for the three callbacks whose loss costs nothing.** `onNewIntent`, `onActivityResult` and `onRequestPermissionsResult` go through one `dispatchToQt` helper that catches `UnsatisfiedLinkError` and logs it. Nothing in Decenza registers a Qt new-intent listener, and a result delivered with no Qt to receive it had no consumer either.

**The lifecycle callbacks are deliberately not guarded**, though they throw the same error. An earlier revision of this PR wrapped them, and review showed that was worse than the crash: with no Qt behind the activity there is no content view, so swallowing keeps a blank window alive that cannot recover — and in Launcher Mode Decenza *is* the home screen, so there is nowhere to escape to while the DE1 runs unattended. A process that dies and is restarted by Android is the better failure. `onDestroy` is worse still: its `QtNative.terminateQtNativeApplication()` runs *before* the `System.exit(0)` that ends it (`QtActivityBase.java:210-223`), so catching there skips the exit and strands the process.

## Verification

Java-only change; no C++ or QML touched, so the local suite is unaffected by it. Compiled `DecenzaActivity.java` against `android-34` with stubbed `QtActivity`/`StorageHelper`/`DeviceShutdownService` — clean.

**Needs a device check that CI cannot give:** enable Launcher Mode, reboot to home, and confirm the app comes up instead of dying — that is the path this PR fixes and the only one that exercises `retargetAliasLaunch()`. An Android CI build should run too, since this file compiles only on that platform.

### DIFF
--- a/android/src/io/github/kulitorum/decenza_de1/DecenzaActivity.java
+++ b/android/src/io/github/kulitorum/decenza_de1/DecenzaActivity.java
@@ -41,39 +41,50 @@ public class DecenzaActivity extends QtActivity {
         }
     }
 
+    // onResume/onPause/onStop go through dispatchToQt for the same reason as
+    // onNewIntent below: each one calls QtNative.setApplicationState(), which
+    // calls the `updateApplicationState` native unguarded (QtNative.java:266,
+    // Qt 6.11.2). That is the crash in issues #1239 (v1.7.5), #1511 and #1512
+    // (both v1.8.0) — same UnsatisfiedLinkError, different callback, and all
+    // three predate the Qt 6.11.2 upgrade.
     @Override
     protected void onResume() {
-        super.onResume();
+        dispatchToQt("onResume", () -> super.onResume());
         Log.d(TAG, "=== LIFECYCLE: onResume (app now in foreground) ===");
     }
 
     @Override
     protected void onPause() {
         Log.d(TAG, "=== LIFECYCLE: onPause (app losing focus) ===");
-        super.onPause();
+        dispatchToQt("onPause", () -> super.onPause());
     }
 
     @Override
     protected void onStop() {
         Log.d(TAG, "=== LIFECYCLE: onStop (app no longer visible) ===");
-        super.onStop();
+        dispatchToQt("onStop", () -> super.onStop());
     }
 
+    // NOT wrapped: QtActivityBase.onDestroy() ends in System.exit() on the
+    // normal path, and swallowing a failure part-way through that teardown
+    // would leave a half-torn-down process alive rather than saving one.
     @Override
     protected void onDestroy() {
         Log.d(TAG, "=== LIFECYCLE: onDestroy (app being destroyed) ===");
         super.onDestroy();
     }
 
-    // Qt's Java lifecycle callbacks that jump straight into JNI.
+    // Guard for the Qt lifecycle callbacks that jump straight into JNI.
     //
     // QtActivityBase.onNewIntent(), onActivityResult() and
     // onRequestPermissionsResult() call a `public static native` method on
     // QtNative with no guard (qtbase/src/android/jar/src/org/qtproject/qt/
-    // android/QtActivityBase.java:369-385, Qt 6.11.2). Those natives are
-    // registered by the QPA plugin's JNI_OnLoad (qtbase/src/plugins/platforms/
-    // android/androidjnimain.cpp:751-761, :907-926), so outside the window in
-    // which Qt is loaded and running the call throws UnsatisfiedLinkError:
+    // android/QtActivityBase.java:369-385, Qt 6.11.2), and onResume/onPause/
+    // onStop reach one the same way through QtNative.setApplicationState()
+    // (QtNative.java:266). Those natives are registered by the QPA plugin's
+    // JNI_OnLoad (qtbase/src/plugins/platforms/android/androidjnimain.cpp:
+    // 751-761, :907-926), so before Qt's libraries are loaded the call throws
+    // UnsatisfiedLinkError:
     //
     //   No implementation found for void org.qtproject.qt.android.QtNative
     //   .onNewIntent(android.content.Intent)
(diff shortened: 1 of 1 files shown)


skialpine:
### Code review

Found 3 issues. All three are fixed in `6c9d6683`, which changes the approach: the crash is now fixed at its source and the guard is narrowed.

1. Wrapping `onResume`/`onPause`/`onStop` converted a crash-and-restart into an unrecoverable blank window. With Qt's libraries unloaded, `startNativeApplication()` never runs, so there is no content view — swallowing the error kept that activity alive instead of letting the process die and be restarted. In Launcher Mode Decenza is the home screen, so there is nowhere to escape to while the DE1 runs unattended.

https://github.com/Kulitorum/Decenza/blob/878f726d0c956a440e81c119319b70517d19f365/android/src/io/github/kulitorum/decenza_de1/DecenzaActivity.java#L50-L66

2. `onDestroy` was left unwrapped on a rationale that does not hold. `QtActivityBase.onDestroy()` calls `QtNative.terminateQtNativeApplication()` — a native from the same `JNI_OnLoad` table — *before* the `System.exit(0)` that ends it. In the state this PR was designed to survive it throws there, skipping the exit and stranding the process. The comment claimed the method "ends in `System.exit()` on the normal path", which was false on the one path the PR created.

https://github.com/Kulitorum/Decenza/blob/878f726d0c956a440e81c119319b70517d19f365/android/src/io/github/kulitorum/decenza_de1/DecenzaActivity.java#L67-L75

3. The comment listed `onResume` among the callbacks Qt guards with `isStarted`. It is not. `QtActivityBase.onResume` calls `setApplicationState()` at `:193`, one line *before* the `isStarted` check at `:194`, and the check only gates the UI work after it — which is precisely why #1239, #1511 and #1512 all crash at `:193`. The count was also wrong (5 guard sites, not "roughly eight"), and `QtNative.setApplicationState`'s bare native call is at `:274`, not `:266`.

https://github.com/Kulitorum/Decenza/blob/878f726d0c956a440e81c119319b70517d19f365/android/src/io/github/kulitorum/decenza_de1/DecenzaActivity.java#L96-L110

**Root cause, found while verifying issue 1.** Qt 6.11 added a branch to `QtActivityBase.onCreate` that skips loading the Qt libraries entirely when the activity is launched through an activity-alias, keyed on a name comparison in `isLaunchedAsAlias()`. Launcher Mode enables exactly such an alias. The branch is absent in `v6.8.0` and `v6.10.0` and present in `v6.11.1` (checked against code.qt.io); Decenza moved to 6.11.1 shortly before #1239, the first of the four reports, and Launcher Mode had worked since the alias got its `android.app.lib_name` in #287. `onCreate` now retargets the intent at the real activity so the comparison matches and Qt loads.

<details>

- [Worth a look] A swallowed error leaves no durable evidence — no `crash.log`, so nothing reaches the auto-filed issue pipeline, unlike the `ble_dead_system` flag-file pattern this file already uses for `DeadObjectException` (score: 60). Much reduced now that the guard covers only three callbacks with no consumer, and the fatal path is fixed rather than caught.
- [OK to skip] `onRequestPermissionsResult` has a live producer (`StorageHelper` calls `requestPermissions`), so a future Java-side consumer would silently lose its result (score: 55). No such consumer exists today.

</details>

Issue #1823, opened 2026-08-17
[Bug] Automated calibration performs poorly


Ubermensch7492:
### What happened?

Using the D-flow profile and my manual calibration value of 1.35, I have an almost exact match between weight-based and volume-based flow (e.g., 1.4g/s and 1.4ml/s).  After five shots, the system automatically updated my flow calibration for this profile to 1.01.  At this calibration, the match between mass and volume flows is quite bad (e.g., 1.7g/s and 1.2ml/s).

### Steps to reproduce

1. Use D-flow profile with a manual setting that matches volume and mass flow rates
2. Run 5 shots to allow autocalibration to calculate a new flow calibaration.
3. Note the lack of fit.

### Decenza version

2.03 currently

### Platform

Android

### Device / OS version

Samsung A7 lite

### Scale (if applicable)

Skale

### Debug log / shot debug log

[debug.zip](https://github.com/user-attachments/files/31159040/debug.zip)

[shot25.json](https://github.com/user-attachments/files/31145700/shot25.json)
[D-Flow___20g-2_50-91 (1).json](https://github.com/user-attachments/files/31145703/D-Flow___20g-2_50-91.1.json)
[shot541.json](https://github.com/user-attachments/files/31145702/shot541.json)
[D-Flow___20g-2_50-91.json](https://github.com/user-attachments/files/31145701/D-Flow___20g-2_50-91.json)

[08:04:03.691] START Shot capture started - 2026-08-16T08:04:03
[08:04:03.692] INFO Profile: D-Flow / 20g-2.50-91
[08:04:03.696] DEBUG [Yield] latched target=50.0g dose=20.0g anchor=absolute:50.00
[08:04:03.702] DEBUG [SAW][Learning] Model: source=perProfile lag=0.806 s profile=d_flow_20g_2_50_91 scale=skale basket=vst-20g-competition historyN=9
[08:04:03.892] DEBUG ShotDataModel: Registered fast renderers (QSGGeometryNode, pre-allocated VBO)
[08:04:03.906] DEBUG [AutoSleep] Reset by phase change: normal=60
[08:04:03.954] WARNING [SAW][Worker] Scale oscillation detected (weight=-66.70 g) — stop-at-weight blocked, awaiting settle
[08:04:03.957] DEBUG [SAW][Worker] Active but tare not complete — skipping stop-at-weight, weight=-66.70 g
[...]

PR #1828, opened 2026-08-18
fix(calibration): use achieved flow, not target, on pressure-capped flow windows

skialpine:
## Summary

- Auto flow calibration's flow-branch formula assumed a flow-controlled frame always reaches its target flow. Frames with a pressure ceiling (D-Flow, D-Flow/Q) often don't, which produces a bad calibration ideal for that window.
- Fixes #1823: reported user's hand-tuned D-Flow calibration (1.35, good mass/volume match) auto-degraded to 1.01 (poor match) after 5 shots. Confirmed against the issue's attached debug log — 5 of 8 real steady-state windows missed target flow (57–67% of target) and clustered low (~0.75–0.90), dragging the batch median down, exactly matching an analysis from 2026-08-02 that was deliberately left unfixed pending real-world evidence.
- Cross-checked against 30 debug logs pulled from other GitHub issues in the last 3 months (unrelated bug reports that happened to include a system log): 20 of 29 accepted flow-calibration windows (69%) had missed target by >10%, across 4 distinct other users, with the same downward-drift pattern visible in at least one (two consecutive batch updates, both down).
- A flow-controlled window whose measured flow **undershoots** the frame's target by >10% is reclassified as pressure-capped and routed through the existing achieved-flow formula and ratio guard, instead of dividing by an unattained target. (Deliberately undershoot-only — see review rounds below.)
- One-time migration clears only *pending* per-profile batches (not stored multipliers, unlike the v2/v3 full resets) — this is a per-window defect the batch median already partially absorbed, not evidence the stored values themselves are wrong.
- OpenSpec change at `openspec/changes/fix-flow-cal-pressure-capped-windows/` (proposal, new `auto-flow-calibration` capability spec, design, tasks — full account of both review rounds in `tasks.md` sections 7-8).

## Review round 1 (`/code-review`, 8-agent multi-angle pass)

8 findings, 1 CONFIRMED + 7 PLAUSIBLE, all addressed:

- **(CONFIRMED, most severe)** The deviation check was originally symmetric — it also reclassified windows that *overshot* target by >10%. A pressure ceiling can only hold flow below its setpoint, never push it above, so overshoot has no pressure-cap explanation, and reclassifying it risked reintroducing the v2 feedback-loop bug (reported flow as denominator on a window that may still be genuinely PID-locked) for that case. Fixed: the check now only fires on undershoot.
[...]

PR #1873, opened 2026-08-27
fix(calibration): auto flow calibration converged on the square root of the correct multiplier

### LINKED ISSUE #1872: Automated flow calibration remains an issue on D-flow (repeat of #1823) [Bug]
### What happened?

I previously reported that the automated flow calibration was making the calibration worse (Issue #1823 ).  The updates made did not resolve the issue.  

### Steps to reproduce

1. Manually set flow calibration to 1.35 and note the good fit between mass and volume flow
2. Allow five shots to accumulate and automatically change the profile to 1.17
3. Note the poor quality of fit between flow measurements

### Decenza version

2.04 (build 3566)

### Platform

Android

### Device / OS version

Samsung A7 

### Scale (if applicable)

Skale

[debug (1).zip](https://github.com/user-attachments/files/31525104/debug.1.zip)

_No response_

### Screenshots or video

_No response_

### Anything else?

_No response_

skialpine:
Closes [#1872](https://github.com/Kulitorum/Decenza/issues/1872). Supersedes the 2.0.4 fix for [#1823](https://github.com/Kulitorum/Decenza/issues/1823), which addressed the same reporter's problem and did not resolve it.

## The defect

The DE1 has no flowmeter — Decent removed it for an open-loop physics model over pump strokes. So reported flow is `pumpModel × C`, and the quantity calibration should converge to is that model's error:

```
e = C · weightFlow / (machineFlow · ρ)      ← C cancels; this is an identity
```

Pressure-controlled windows already used that expression. Flow-controlled windows used `weightFlow / (targetFlow · ρ)` instead, introduced by v3 on the premise that the sensor expression is "proportional to the current factor". That premise holds only if the multiplier scales *reporting* alone. It does not: the DE1 servos its calibrated flow, so a higher multiplier delivers less water and the expression is unchanged. Across three unrelated machines the stored multiplier moved 15–38% while `e` moved 4–15%.

On a window holding target the old expression equals `e / C`. The multiplier is not set to the ideal but blended in at the batch EMA's `alpha = 0.5`, making the update `C := (C + e/C) / 2` — **Babylonian square-root iteration**. Flow-profile machines therefore settled on `√e` instead of `e`.

## Evidence

Four machines, both branches, each landing on what its formula solves for:

| machine | branch | e | √e | converged C |
|---|---|---|---|---|
| #1872 reporter | flow | 1.389 | **1.178** | 1.17 |
| this repo's DE1 | flow | 0.79 | **0.889** | 0.8795 |
| cablecj74 | pressure | **1.369** | 1.170 | 1.3555 |
| mcastaldelli | pressure | **1.303** | 1.141 | 1.30 |

The reporter's log shows the app moving him `1.35 -> 1.22027` — away from the 1.35 he had set by hand and toward √e. He independently measured his own machine at 24% off when C was 1.17, which puts `e` at 1.45.

`evidence/` replays the real algorithm offline over 75 shots from three machines, validated by reproducing this repo's own logged batch update.

## What changes

**One formula for both control modes.** Window classification still selects which ratio guard runs and whether the off-target check applies; it no longer selects a formula.

**Off-target flow windows are skipped** rather than re-routed through the pressure formula, as 2.0.4 did. `e` varies with operating point — 1.08 at 0.72 mL/s down to 0.64 at 1.9 on this repo's DE1 — so a window that missed its target measured a rate the profile never pours at. Re-routing didn't remove that error, it reversed its sign.

**The effective multiplier is recorded per shot** (`shots.flow_calibration`, migration 39), latched at shot start and consumed by the save. Reported flow is a calibrated quantity, so without it a shot's flow curve cannot be interpreted after the fact — which is why diagnosing this needed a debug log beside the shot data.

**Migration v6** clears pending batches only, same shape as v4 and v5. Stored multipliers are left alone: the √e error is small on the machines measured, and every batch now yields the target value directly.

## Corrections to the record

v3's premise is stated as settled fact in `settings.cpp` and `AUTO_FLOW_CALIBRATION.md`; both are annotated rather than deleted, since a comment nobody re-derives is how a wrong diagnosis keeps getting re-approved. The v2 runaway it was blamed on was bad scale data reaching a formula with no ratio guards — the guards fixed it, and changing the formula on top traded a correct expression for one whose fixed point is a square root.

"Flow sensor" was the wrong term throughout; renamed to pump-model language, and `autoFlowCalSensorIdeal` is now `autoFlowCalIdeal`.

An earlier draft claimed a third user's unchanged median showed live well-dialled users were unaffected. That user has auto calibration switched off with C = 1.000, so he is a control, not a live user. Corrected in the proposal.

## Verification

Build green, full suite 116/116, 0 warnings. `check_log_markers.py` and `check_test_source_duplication.py` pass; `openspec validate --strict` passes.

Three new tests pin the properties the change rests on: the formula is a fixed point for a correctly-calibrated machine; it is invariant under the current multiplier (with the superseded expression asserted *not* invariant in the same slot); and the old rule settles on √e under the real EMA while v6 settles on e. A fourth pins that the damping is load-bearing — undamped, `C := e/C` is a period-2 map that oscillates forever, so the square root is a property of the update as a whole rather than the formula alone.

## Known limits

- The identity rests on `reported = pumpModel × C`, which comes from Decent's own calibrator replaying historic shots by the ratio of multipliers, plus four machines agreeing retrospectively. It has not been confirmed by a controlled experiment.
- The skip costs the #1872 reporter 10 of 16 windows and tightens nothing on his machine, where `e` is flat across flow. It is justified by machines where `e` is not.
- `design.md` records the pump-model error across seven machines: machine model dominates and mains voltage does not. n=1 for DE1+, so model and individual machine are not separable from that sample.

Two follow-ups are noted and deliberately not acted on here: the multiplier defaults to 1.0 when a PRO/XL belongs near 1.3, and `e` varies enough with operating point to question whether one scalar per profile is the right shape.

## Review round 2

Five agents over the branch diff. Two independently found the same high-severity defect.

**The consumable latch was defeated by two early returns.** `consume()` sat below `onShotEnded()`'s "never extracted" and maintenance-beverage returns, while `latchForShot()` is armed from `espressoCycleStarted` unconditionally — so a rinse or an aborted cycle latched and never cleared, and the next shot whose cycle-start went unseen stamped that value, possibly from a different profile. Now read and cleared once, above every return.

**The ideal took the shot-end multiplier.** It now takes the poured one, threaded through from the latch. Reported flow carries whatever multiplier was in effect during the pour; the MCP tool writes mid-shot.

**Migration 39's probe conflated "PRAGMA failed" with "column absent"** — which would have frozen `schema_version` at 38 (blocking every later migration) while every shot load and save failed, since both name the column unconditionally.

**One ratio guard for both modes, on the formula's own inputs.** The flow arm bounded the *old* ideal by construction and does not bound v6's; with the off-target check undershoot-only, an overshooting window could pass it and still drive the ideal arbitrarily low.

**The clamp is reported**, and the ceiling warning reads the raw value — it was reading the clamped one, destroying the magnitude it exists to report, and on pre-1337 firmware could never fire at all.

### Deferred items done rather than filed

This repo doesn't track follow-ups, so calling them that is just deferral:

- `[Calibration]` registered as a log marker with a tiered helper. The subsystem had none, so its story was reachable only by guessing a substring. Outcomes moved to INFO — the inverted shape `LOGGING.md` warns about was present, every fault at WARN and every outcome at DEBUG.
- Consecutive rejections counted per profile and surfaced. `flow_calibration` rendered a permanently-rejected profile as "0 of 5 shots collected", which reads as "keep pulling shots" to the one user for whom that never works.
- `flowCalibrationAutoUpdated` has a consumer; it was emitted into nothing.
- `FlowCalibrationPage` replays each shot at the multiplier it was recorded at. `m_shotMultiplier` was hardcoded to 1.0, so the overlay that page exists to judge was drawn at the wrong scale on any machine not at 1.0.

### Corrections to this PR's own claims

- `+2.4% / −2.5%` were residuals *against* √e, misread as the √e-vs-e error — which is ~16%, and was the stated justification for not resetting multipliers.
- The refuted v3 premise was still present-tense fact 160 lines above the block refuting it.
- "30-40%" didn't follow from its own numbers (48%).
- Line citations replaced with symbols after rotting twice.
- An oscillation test that could not fail is gone; the convergence tests now bind to `kAutoFlowCalBatchEmaAlpha` instead of re-typing 0.5.

The derivation now lives in `AUTO_FLOW_CALIBRATION.md` rather than being restated at six sites.

Build green, 116/116, qmllint gate clean.


skialpine:
### Code review

Found 7 issues:

1. Documentation numbers contradict the PR's own evidence: this says "30/30 → 29/30 and the median ideal moves under 2%", while `proposal.md`, `design.md` and the PR description all say 27/30 and that the median does not move at all (0.968 → 0.968). The doc both gets the count wrong and understates its own argument for not resetting stored multipliers.

https://github.com/Kulitorum/Decenza/blob/c0031ccddf9f8d92216bcd18d2e44396478e2d59/docs/CLAUDE_MD/AUTO_FLOW_CALIBRATION.md#L193-L195

2. Task 1.5 asserts `maincontroller.cpp` "is in neither glob set of `scripts/check_log_markers.py`". It is in `MARKER_ONLY_GLOBS`. The gate passes because the new line has no bracketed prefix, but the false claim would license a hand-typed marker in a future edit — exactly the drift CLAUDE.md's logging section exists to prevent ("Which file you are in decides which rules apply").

https://github.com/Kulitorum/Decenza/blob/c0031ccddf9f8d92216bcd18d2e44396478e2d59/openspec/changes/skip-off-target-flow-cal-windows/tasks.md#L6-L8

3. No test exercises `shots.flow_calibration` through the real save/load path. `v38ToV39AddsFlowCalibration` verifies the migration with raw SQL and never calls `loadShotRecordStatic()`, where the column is read by positional index (`query.value(56)`). The two prior migrations that added positionally-read columns both added exactly this guard and say why in their own comments — `v22_shotEquipmentRpmRoundTrip` ("so a future SELECT-list edit that shifts those columns fails loudly") and `v33_tasteAxesReadRoundTrip`. Index 56 is correct today; nothing would catch it becoming wrong.

https://github.com/Kulitorum/Decenza/blob/c0031ccddf9f8d92216bcd18d2e44396478e2d59/tests/tst_dbmigration.cpp#L895-L900

4. The comment justifying the shot-start latch cites "maincontroller.cpp:4078 vs :4261". The actual lines are :4079 (`computeAutoFlowCalibration()`) and :4267 (`saveShot()`); :4261 is an unrelated `stoppedBy = QStringLiteral("manual");`. This citation is the whole justification for latching rather than reading at save, and the PR's own design doc cites the correct pair.

https://github.com/Kulitorum/Decenza/blob/c0031ccddf9f8d92216bcd18d2e44396478e2d59/src/controllers/profilemanager.h#L270-L275

5. The same comment says 0.0 means not recorded "(no settings object at latch time, an imported shot, or a fake/dev shot)". Two of those three cannot apply to this getter: the import path never calls `latchForShot()`, and the dev fake-shot path never reads the getter. Since `m_latchedFlowCalibration` is never cleared by `releaseShotLatch()`, in those scenarios it holds the *previous* shot's multiplier, not 0.

https://github.com/Kulitorum/Decenza/blob/c0031ccddf9f8d92216bcd18d2e44396478e2d59/src/controllers/profilemanager.h#L275-L280

6. `metadata.flowCalibration` is read unconditionally, while the yield fields it sits beside are guarded by `hasShotSnapshot()` ([maincontroller.cpp#L3986](https://github.com/Kulitorum/Decenza/blob/c0031ccddf9f8d92216bcd18d2e44396478e2d59/src/controllers/maincontroller.cpp#L3985-L3987)). Combined with issue 5, the comment's claim "0 when nothing was latched" is wrong: a shot reaching save without its own latch stamps the previous shot's multiplier, which is precisely the "measurement nobody took" the NULL-means-unrecorded contract is built to prevent.

https://github.com/Kulitorum/Decenza/blob/c0031ccddf9f8d92216bcd18d2e44396478e2d59/src/controllers/maincontroller.cpp#L4112-L4115

7. The new "Recorded per shot" subsection is an H3 under `## v4 Migration (Pressure-Capped Flow Window Fix)`, so a per-shot database column reads as part of an unrelated settings migration. It belongs at `##`, beside the v5 section.

https://github.com/Kulitorum/Decenza/blob/c0031ccddf9f8d92216bcd18d2e44396478e2d59/docs/CLAUDE_MD/AUTO_FLOW_CALIBRATION.md#L183-L186

---

<details>

- [Worth a look] Pre-existing comment in `maincontroller.cpp` cites `PostShotReviewPage.qml:844` for the `dyeDrinkWeight` write; the actual line is 809. Out of scope, but this PR's new fake-shot comment leans on that comment for its own justification, so the citation chain is now load-bearing (score: 50)

</details>

Verified as correct and not flagged: positional index 56 (the SELECT contains `json_extract(eg.attrs, '$.burrs')`, whose internal comma makes a naive count give 57); the transfer-copy INSERT's 49 columns / 49 placeholders / 49 executed binds; migration 39's gating, transaction and stamping; and the preservation of #1828's two review-established safeguards (undershoot-only asymmetry, single combined branch).

skialpine:
All seven addressed in 33d0ab7b. 116 tests pass.

**Finding 6 needed a design decision, not a guard.** I checked whether the stale case is actually reachable: `espressoCycleStarted` fires only on the `isInEspresso && !wasInEspresso` transition, and [machinestate.cpp](https://github.com/Kulitorum/Decenza/blob/c0031ccddf9f8d92216bcd18d2e44396478e2d59/src/machine/machinestate.cpp#L754-L757) already carries a safety net for "machine skipped preheating (missed BLE substate notification)". So a missed transition means no latch, and the save would inherit the previous shot's multiplier.

Two obvious fixes are both wrong:
- `hasShotSnapshot()` guard — set on the first latch of the session and never cleared, so it is true for every later shot and guards nothing.
- Clearing in `releaseShotLatch()` — the save runs after settling, i.e. after release, so every shot would record 0.

The latch is now **consumed** by the save, so a value belongs to exactly the shot that latched it; a shot that never latched reads 0 and records NULL. Covered in `tst_profilemanager.cpp`.

The rest: added `v39_flowCalibrationReadRoundTrip` (the guard v22/v33 established for positionally-read columns — nothing previously exercised `query.value(56)` through `loadShotRecordStatic`); corrected the `:4078`/`:4261` citations to `:4079`/`:4267`; dropped the two causes that cannot apply to that getter; fixed the v5 doc numbers to 27/30 with the median unmoved; promoted the per-shot section out from under the v4 Migration heading; corrected the `MARKER_ONLY_GLOBS` claim.

The below-threshold one (`PostShotReviewPage.qml:844` should be `:809`) is left alone — pre-existing and outside these changes.

skialpine:
### Code review

Found 6 issues:

1. Aborted shots feed the new rejection counter. `computeAutoFlowCalibration()` runs before the aborted-shot classifier discards the shot, so a false start or a flush pulled with an espresso profile loaded (duration < 10 s, weight < 5 g) records a rejection for a shot that is then deleted and never appears in Shot History. After five, the user is told "Pulling more shots will not help until that changes" about shots they cannot find — the same wrong-place-to-look the counter exists to prevent.

https://github.com/Kulitorum/Decenza/blob/e03183f1a7890c063964df0e3e94ac899a7bcac2/src/controllers/maincontroller.cpp#L4158-L4162

https://github.com/Kulitorum/Decenza/blob/e03183f1a7890c063964df0e3e94ac899a7bcac2/src/controllers/maincontroller.cpp#L4209-L4213

2. `main.cpp`'s ordering claim is false, and it is the sentence that would license clearing the flow-cal latch in the wrong place. `releaseShotLatch()` hangs off `espressoCycleEnded`; `onShotEnded()` hangs off `shotProcessingReady`, which fires after stop-at-weight settling. In a real shot log the phase change `Espresso -> Idle` was at `18:26:02.157` while the auto-flow-cal block inside `onShotEnded()` ran at `18:26:03.533` — release fires ~1.4 s *before* the save path, not after. Harmless for the yield snapshot, which survives release by design; misleading for a latch that does not. (CLAUDE.md says "an un-sourced claim in a comment gets believed and then licenses wrong code".)

https://github.com/Kulitorum/Decenza/blob/e03183f1a7890c063964df0e3e94ac899a7bcac2/src/main.cpp#L1651-L1656

3. An unrelated OpenSpec change is in this PR. `scope-recipe-grind-recommendations-to-equipment` is about the recipe wizard's grind-history recommendations and has no code implementing it here. It was untracked before this branch and was swept into `08177d0c` by a `git add -A`.

https://github.com/Kulitorum/Decenza/blob/e03183f1a7890c063964df0e3e94ac899a7bcac2/openspec/changes/scope-recipe-grind-recommendations-to-equipment/proposal.md#L1-L6

4. The falsification experiment this change gates itself on has not been run. `design.md` says "Run it before merge"; task 7.10 is unchecked. Every other support for reversing v3's formula is retrospective correlation across machines with different pumps and mains voltages, while `AUTO_FLOW_CALIBRATION.md` states the C-invariance flatly.

https://github.com/Kulitorum/Decenza/blob/e03183f1a7890c063964df0e3e94ac899a7bcac2/openspec/changes/skip-off-target-flow-cal-windows/tasks.md#L50-L52

5. The rejection counter's documented contract overclaims what it counts. The header says "Consecutive shots on this profile that produced NO calibration ideal" and `AUTO_FLOW_CALIBRATION.md` names "a user with no scale connected" as a permanent case the mechanism surfaces — but six no-ideal returns never increment, including that one. It is also logged at `CAL_DETAIL`, and per `docs/CLAUDE_MD/LOGGING.md` the connections views default to `minLevel INFO`, so the user cannot see it either way.

https://github.com/Kulitorum/Decenza/blob/e03183f1a7890c063964df0e3e94ac899a7bcac2/src/core/settings_calibration.h#L107-L112

https://github.com/Kulitorum/Decenza/blob/e03183f1a7890c063964df0e3e94ac899a7bcac2/src/controllers/maincontroller.cpp#L3088-L3092

6. `parseFlowCalRejections()` logs corrupt JSON but does not reset the key, so the next `noteFlowCalRejection()` rewrites it from the empty map and discards every other profile's run — which is the failure its own comment says it prevents. Both siblings (`parseFlowCalBatch`, `allProfileFlowCalibrations`) do reset.

https://github.com/Kulitorum/Decenza/blob/e03183f1a7890c063964df0e3e94ac899a7bcac2/src/core/settings_calibration.cpp#L300-L307

<details>

- [Worth a look] `shothistorystorage.cpp:2152` says "Not `.value_or(false)`" and line 2159 is `after.value_or(false)`; runtime behaviour is safe (unstamped version, retries next launch) but migration 38 handles the same case by returning explicitly (score: 50)

</details>

Verified as correct and not reported: SELECT column count and the index-56 read, all three INSERT arities, the latch take-and-clear ordering against every early return, the ratio-guard arithmetic and its capture range, the marker registration and alias-don't-copy rule, and the undershoot-only asymmetry in `autoFlowCalWindowTargetCheck()` being deliberately left untouched.

PR #1877, opened 2026-08-28
feat(descale): own the steam heater, and confirm before descaling hot

skialpine:
Follow-up to #1876. Prompted by the steam-heater button on the descale page reading as confusing — why ask the user to press it rather than turning the heater off and just showing the temperature.

## The requirement is real, and official

From the Decent Espresso Machine Manual, "Descaling Instruction - With Citric Acid":

> Disable the steam heater on the Steam page before using the descale function. And wait till the steam temperature cools down to 60ºC or lower (It can take an hour).

with the note that disabling it right after waking the machine, before preheating, saves the wait. The page's existing instruction text is a close copy of that paragraph.

Neither reference app helps here. de1app's `descale_prepare` page is four instructions (drip tray, citric acid, blind basket, tank) and never mentions the steam heater; `start_decaling` doesn't touch `steam_disabled`. Decaid has no descale UI at all. Decent's own instruction sends you to a *different screen* to switch the heater off.

## Changes

| Before | After |
|---|---|
| Step 1 told the user to press "Disable" | Heater turned off when the page opens |
| Exit forced the heater on, unconditionally | Full steam state restored when the user leaves via back |
| Start descaled at any steam temperature | Start asks first above 60 °C |
| Bare temperature number | Number plus "Cooling — wait for 60 °C" / "Cool enough to descale" |
| "takes about 6 minutes" | "about 12 minutes" |
| An "Enable" button under "Cool enough to descale" | Removed — the temperature and status stay |

### The hold lives in `MainController`, not in the page

Waiting for 60 °C can take an hour, and the page is not guaranteed to survive it: during the wait the machine phase is Idle, so the shell does not consider an operation active and the auto-sleep countdown runs, replacing the page with the screensaver. A descale started from the GHC also replaces it with a fresh instance. `beginDescaleHeaterHold()` is idempotent so a second instance does not re-capture; `endDescaleHeaterHold()` is called only from the back button, never from destruction.

### The release restores inputs, not an outcome

The hold snapshots every input to `SteamHeaterPolicy::resolve()` that it disturbs, and the release puts all of them back and lets the policy resolve:

- **`steamDisabled`** — the transient veto `turnOffSteamHeater()` sets.
- **Event permission** — also cleared by `turnOffSteamHeater()`, and the one input that cannot be re-derived from settings. Without it, a user who reached the page while a steam grant was live, with every other veto standing, left with the heater off when it had been on.
- **Selected pitcher** — captured though nothing in this flow changes it today. The selected pitcher *is* the steam spec (its own temperature, and whether it is the built-in "Heater off"), so a restore covering the flag but not the pitcher is one edit away from drifting silently.

Restoring inputs rather than a remembered on/off state is what makes both directions correct. Arrive with the heater off for any reason — a "Heater off" pitcher, Keep warm when idle off, a recipe with no steam — and you leave with it off, because those vetoes are independent in `resolve()` and the hold never touched them.

Notably NOT `startSteamHeating()`: that grants event permission which short-circuits every veto and persists with nothing to release it.

### The dialog warns rather than blocks

The 60 °C limit is guidance, and a user who has just replaced the boiler water keeps the ability to proceed. Both start sites route through one `startDescaleChecked()`, so the gate cannot be added to one and forgotten on the other — the same trap the three maintenance states fell into in C++ in #1876.

## The duration correction

"About 6 minutes" comes from the 2018 article the page's instructions were drawn from. Three instrumented runs on firmware 1358 measured 720.1 s, 720.2 s and 720.2 s.

## Review findings, all fixed in-branch

This PR was reviewed before merge and four real bugs came out of it — three introduced by the first version of this change, one pre-existing:

1. The exit restore called `startSteamHeating()`, which grants event permission that short-circuits every veto in `SteamHeaterPolicy::resolve()` and persists with nothing to release it. Same veto-bypass #1783's review flagged at another call site. Fixed in bae4f2f9.
2. Auto-sleep during the cool-down destroyed the page and undid the heater-off, while the page promised the opposite. Fixed in 9b74f71d.
3. A GHC-started descale replaced the page; Qt's `deleteLater` teardown meant the old instance's restore landed after the new one had read the state, turning the heater on mid-descale. Fixed in 9b74f71d.
4. "Run Again" cleared the rinse view before the gate could decline, so choosing Wait dropped the user on the preparation view with the rinse instructions gone and nothing started. Fixed in 9b74f71d.

The event-permission and pitcher halves of the restore came after that review, in 8f6b9f99.

One correction to an earlier version of this description: it claimed the old unconditional `setSteamDisabled(false)` turned the heater on for users with a "Heater off" pitcher or Keep warm when idle disabled. That was wrong — those vetoes are independent in `resolve()`, so clearing `steamDisabled` cannot override them. The real prior bug was narrower: it re-enabled a heater that had been disabled through the flag itself.

## Testing

- Build clean; full suite 116 passed, 0 failed.
- **The QML is unverified.** No QML test harness — entry auto-disable, the back-button restore, the auto-sleep case and the dialog all need exercising on a real machine.

### DIFF
--- a/qml/pages/DescalingPage.qml
+++ b/qml/pages/DescalingPage.qml
@@ -15,9 +15,52 @@ T.Page {
     background: ThemedPageBackground {}
 
 
-    // Restore normal operation when leaving descale page
+    // Decent requires the steam heater OFF and the steam boiler below 60 °C before a
+    // descale ("Disable the steam heater on the Steam page before using the descale
+    // function. And wait till the steam temperature cools down to 60 °C or lower (It can
+    // take an hour)" — Decent Espresso Machine Manual, 2.2 The DE1 Software, 5.
+    // Maintenance & Cleaning). It is a precondition of every descale, not a preference,
+    // so the page satisfies it on open rather than asking the user to.
+    //
+    // Opening the page is also the earliest moment the user has expressed intent, which
+    // matters: the same manual notes that disabling right after waking the machine, before
+    // preheating, is what keeps the wait short.
+    readonly property real steamSafeTempC: 60
+    readonly property real steamTempC: typeof DE1Device.steamTemperature === 'number'
+                                       ? DE1Device.steamTemperature : 0
+    readonly property bool steamTooHotToDescale: steamTempC > steamSafeTempC
+
+    // What the heater was doing before this page took it over, so exit can put it back.
+    // -1 until captured, so a destruction that somehow beats completion restores nothing
+    // rather than guessing.
+    property int steamHeaterWasOn: -1
+
+    // Both start sites go through here so the hot-boiler check cannot be added to one and
+    // forgotten on the other — the same trap the three maintenance states fell into in C++.
+    function startDescaleChecked() {
+        if (descalingPage.steamTooHotToDescale) {
+            hotSteamConfirmDialog.open()
+            return
+        }
+        DE1Device.startDescale()
+    }
+
+    Component.onCompleted: {
+        descalingPage.steamHeaterWasOn = MainController.steamHeaterOn ? 1 : 0
+        if (MainController.steamHeaterOn) {
+            MainController.turnOffSteamHeater()
+        }
+    }
+
     Component.onDestruction: {
-        Settings.brew.setSteamDisabled(false)
+        // Restore what the user had, rather than forcing the heater on. This line used to
+        // be an unconditional Settings.brew.setSteamDisabled(false), which turned the
+        // heater back ON for anyone who arrived with it deliberately off — a "Heater off"
+        // pitcher, or Keep warm when idle disabled.
+        if (descalingPage.steamHeaterWasOn === 1) {
+            MainController.startSteamHeating("descaling-restore")
+        }
+        // The cold-maintenance workaround may have left a 1 °C profile on the machine.
         ProfileManager.uploadCurrentProfile()
     }
 
@@ -400,7 +443,7 @@ T.Page {
                         onClicked: {
                             descalingPage.showRinseInstructions = false
                             descalingPage.wasDescaling = false
-                            DE1Device.startDescale()
+                            descalingPage.startDescaleChecked()
                         }
                     }
 
@@ -462,7 +505,7 @@ T.Page {
                         Tr {
                             Layout.fillWidth: true
                             key: "descaling.warning.steam"
-                            fallback: "\u2022 Disable steam heater and wait until steam temp is below 60\u00B0C (can take 1 hour). Tip: disable it right after waking the machine, before preheating, to save time"
+                            fallback: "\u2022 The steam heater is switched off while this page is open, and restored when you leave. Steam temp must be below 60\u00B0C (can take 1 hour from hot) - opening this page right after waking the machine saves the wait"
                             font: Theme.bodyFont
                             color: Theme.textColor
                             wrapMode: Text.WordWrap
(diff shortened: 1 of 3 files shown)


skialpine:
### Code review

Found 4 issues. All were fixed in-branch during the review; the links below point at the fixed code.

1. Exit restore granted event permission that never gets released (`MainController::startSteamHeating` sets `setEventPermission(true)`, and event permission "short-circuits every veto" in `SteamHeaterPolicy::resolve()`). A user whose heater was on via Keep warm when idle would leave the descale page and thereafter have "Heater off" pitchers silently ignored. Same veto-bypass #1783's review flagged at a different call site. Every other `startSteamHeating()` call site pairs it with `releaseSteamEventPermission()`; a restore has no steam event to release.

https://github.com/Kulitorum/Decenza/blob/9b74f71d1d49aff197e0561c887e9d547620b54b/src/controllers/maincontroller.cpp#L3809-L3822

2. Auto-sleep during the cool-down destroyed the page and undid the heater-off. The wait is documented as up to an hour, and during it the phase is Idle — so `operationActive` is false and `sleepCountdownTimer` runs, replacing the page with the screensaver. The page told the user "leave this page open and it will keep cooling" while the shell was free to revoke exactly that.

https://github.com/Kulitorum/Decenza/blob/9b74f71d1d49aff197e0561c887e9d547620b54b/qml/main.qml#L410-L416

3. A GHC-started descale does `pageStack.replace(null, descalingPage)`, and Qt destroys the outgoing page with `deleteLater`, so the old instance's restore ran after the new instance's `onCompleted` had already read the state — switching the steam heater on in the middle of an active descale, with the new instance unable to correct it. Both this and 2 came from the page owning state that outlives it; the hold now lives in `MainController` with an idempotent begin and a release only on explicit exit.

https://github.com/Kulitorum/Decenza/blob/9b74f71d1d49aff197e0561c887e9d547620b54b/qml/pages/DescalingPage.qml#L51-L67

4. "Run Again" cleared `showRinseInstructions` and `wasDescaling` before the hot-boiler gate could decline, so choosing Wait in the confirmation dropped the user on the preparation view with the rinse instructions and Done button gone and nothing started. Both start paths now clear those flags only when a descale actually begins.

https://github.com/Kulitorum/Decenza/blob/9b74f71d1d49aff197e0561c887e9d547620b54b/qml/pages/DescalingPage.qml#L32-L50

Also corrected: this PR's description and two code comments claimed the old unconditional `setSteamDisabled(false)` turned the heater on for users with a "Heater off" pitcher or Keep warm when idle disabled. That is wrong — those vetoes are independent in `resolve()`, so clearing `steamDisabled` cannot override them.

<details>

- [Worth a look] "The threshold lives once" comment was false — the instructional prose spells 60°C out twice (score: 60). Fixed: the comment now claims only what is true, that everything acting on the threshold reads one property.
- [Worth a look] "12 minutes" had no in-file source, only a commit message (score: 55). Fixed: the measured figures are now recorded next to the string.
- [OK to skip] `descaling.steam.enable` / `.disable` / `.accessible` translation keys are unused after the button removal — no QML references them any more (score: 50).

</details>

One flagged issue was discarded as a false positive: the manual citation's section chain was called structurally impossible, but it is copied verbatim from the source document's own breadcrumb.

Caveat on scope: the QML itself is unverified. There is no QML test harness, so the entry auto-disable, the back-button restore, the auto-sleep case and the dialog all still need exercising on a real machine.

skialpine:
### Code review (round 2)

Re-reviewed after the round-1 fixes. Found 6 issues, all fixed in 9ee8d275.

1. Restoring `SteamHeaterPolicy` event permission violated a documented invariant — it is "transient, revoked on the return to Idle and never restored by a settings re-send", i.e. it means a steam action is under way *now*. The hold spans the hour the boiler needs to cool, so writing a captured `true` back fabricates a grant for an event that has certainly ended, and that grant outranks every veto. No longer captured.

https://github.com/Kulitorum/Decenza/blob/9ee8d2755b4cb6bd25e3d63e0e0e0e5f4e5c5a1f/src/controllers/steamheaterpolicy.h#L88-L94

2. Restoring the selected pitcher could corrupt an unrelated recipe. `setSelectedSteamCup()` emits `selectedSteamPitcherChanged`, which is wired to `stampActiveRecipeSteam()` and rewrites the **active** recipe's persisted `steamJson` — so restoring a stale index an hour later would overwrite the steam spec of whatever recipe the user had activated meanwhile. It was also unnecessary: `turnOffSteamHeater()` never touches the selection.

https://github.com/Kulitorum/Decenza/blob/9ee8d2755b4cb6bd25e3d63e0e0e0e5f4e5c5a1f/src/controllers/maincontroller.cpp#L1626-L1632

3. The "Done" button leaked the hold. It is the natural end of the whole flow — run cycle, read rinse instructions, tap Done — and it exits via `AppShell.idleRequested()`, which never released. The steam heater stayed vetoed off for the rest of the session, and this PR had removed the page's own Enable button, so nothing could explain or undo it.

4. Android hardware back and Escape leaked it too. `main.qml` handles those with a global `Keys.onReleased` on the StackView that pops the page without reaching `BottomBar.onBackClicked` — on the primary platform, the usual way out. Both now route through one `leaveDescaling()`.

https://github.com/Kulitorum/Decenza/blob/9ee8d2755b4cb6bd25e3d63e0e0e0e5f4e5c5a1f/qml/main.qml#L1231-L1240

5. `beginDescaleHeaterHold()` asserted only once. Snapshot-once and assert-every-time are different concerns; collapsing them meant a veto cleared under a live hold was never re-applied — selecting any pitcher clears it deliberately, as does sleep or disconnect. Re-entering the page then claimed "the steam heater has been turned off for you" over a boiler that was heating.

6. Sleep voided the hold outright. The dormant handler clears `steamDisabled`, and the wake path re-asserts the resolved target — so an auto-sleep during the cool-down put the boiler back on heat in exactly the scenario the hold exists for. The clear now skips while a hold is active.

Items 1 and 2 were introduced by 8f6b9f99, which set out to make the restore an exact round-trip. That was the wrong goal: the hold should put back only what it took.

<details>

- [Worth a look] The `deleteLater` claim about StackView teardown was uncited, which CLAUDE.md requires for assertions about Qt behaviour (score: 65). Fixed: now cites `qquickstackelement.cpp:75`.
- [OK to skip] `steamTooHotToDescale` uses `>` where the old colour logic used `>=`, so exactly 60.0 °C now reads as cool enough (score: 50). Deliberate — Decent's wording is "60ºC or lower", so the old strictness was the deviation.

</details>

One below-threshold item was discarded: `descaleHeaterHoldActive()` was flagged as an unused accessor. It has since been removed along with the rest of the round-2 changes.

Still unverified: the QML itself. No harness covers it, so the Done and hardware-back release paths, the auto-sleep case and the dialog all need exercising on a real machine.

Issue #575, opened 2026-03-27
[bug] Stuck in Steam mode


fredphoesh:
See video and log.
I don't think I'd been doing anything unusual, probably just finished a shot... 
https://videoblurb.com/mp/stuckinsteam.mp4

[debug 27-03-22.zip](https://github.com/user-attachments/files/26314062/debug.27-03-22.zip)

PR #1879, opened 2026-08-28
feat(calibration): guided pressure and temperature sensor calibration

skialpine:
## Why

Decent's pressure calibration test profile ends with an instruction Decenza could not satisfy:

> Go to Settings->Machine->Calibrate and enter the held pressure value, and then retest until the two agree.

There was no such field. A user whose DE1 reports 9 bar while the portafilter gauge reads 8.2 had no way to correct the machine from this app. The BLE characteristic that carries the correction — `A012`, `DE1::Characteristic::CALIBRATION` — was declared in `de1characteristics.h` and had never had a reader or a writer.

## What this adds

Two guided wizards, **Pressure Calibration** and **Temperature Calibration**, in a new Sensor Calibration card on Settings → Calibration. They are the only way to change a sensor's calibration — no settings field, no popup, no MCP tool.

**The reason for the wizard shape is a correctness argument, not presentation.** Every failure mode here reduces to the `(reported, measured)` pair being mismatched: a profile target mistaken for a reading, a value remembered from the wrong shot, a stale number in a field. de1app is exposed to all three, because both halves are text fields and it sends `::settings(espresso_pressure)` — a *profile parameter* — as the machine-reported value. That is only correct for a simple Pressure-editor profile. `profiles/A-Flow____default-medium.tcl` is `settings_profile_type settings_2c` and still carries `espresso_pressure 6.0` beside frames whose Pressure Up step holds **10.0 bar**; calibrate from that page with it loaded and you are off by 4 bar.

A new `SensorCalibrationController` removes the class of error rather than guarding it: the machine-reported value can only come from a run the controller watched, live. There is no field to type it into, no default, no fallback to a profile target, and no path that reads a historical shot. If the run never held steady, there is no value and the Apply step does not exist.

## Protocol

The 14-byte big-endian `A012` record (`WriteKey` u32, `CalCommand` u8, `CalTarget` u8, two S32P16 values), packed and parsed in one place beside the UUID. `sendCalibration()` is the only writer; the reads are that call with different arguments.

Replies are demultiplexed by de1app's rule: **`WriteKey == 0` means the reply carries a real stored value; anything else is an echo of our own request.** An echo updates nothing, so a refused write can never read back as accepted. An unanswered read leaves the value *absent*, never zero — zero reads as "no correction" and is the one wrong answer that looks plausible.

`A012` is subscribed **non-required**, so a machine or transport that cannot notify still connects; the wizard then shows values as unavailable and refuses to write.

## Deliberately not included: restore-to-factory

`Command::ResetFactory` (2) is named in the enum and never sent. It has no working reference anywhere:

- de1app's helper documents 2 as the reset command (`de1_comms.tcl:1653`)
- but its three reset buttons passed **3** — read-factory — from the day they were added (2018-02-27, `a2092efc`) until they were commented out ten weeks later (2018-05-15, `69e4277c`)
- both commits have empty messages, so nothing states why
- decaid declares `A012` read-only and implements no calibration at all
- there is no DE1 firmware source in this tree

Shipping it would mean writing an unverified command to a machine in the name of undoing a mistake. The factory value is still **read and displayed** — that path *is* live in de1app (`gui.tcl:2537-2538`) — so a user can see what the machine holds. `tst_de1device_mmrreads` asserts nothing ever sends command 2.

The loop is self-correcting anyway: with a gauge in hand, the fix for a bad calibration is another run.

## Also here

- **Nominal heater voltage** (120 V / 230 V, `MMR 0x803834`) in the existing Heater Calibration popup — behind its destructive-change warning, among expert parameters, off any casually-browsed tab. It is machine state, so "Defaults for cafe" does not touch it. Readback bucketing follows decaid's `De1HeaterVoltage.fromInt`. Unknown renders as unknown with neither option preselected.
- **`SettingsActionRow` extracted.** The Maintenance card's Descaling Wizard and Transport Mode rows were two hand-written copies of the same 45 lines; the calibration rows would have been a third. Both move onto the shared component — no behaviour change there.

## Verification

- Full suite via Qt Creator: **117 passed, 0 failed, 0 warnings**
- `check_log_markers.py`, `check_test_source_duplication.py`, `check_mcp_tool_budget.py` all pass
- qmllint gate passes
- New coverage: 9 slots on the wire format in `tst_binarycodec`, 8 on the device API in `tst_de1device_mmrreads`, a new `tst_sensorcalibration` covering the capture controller (measured/no-hold/aborted, the scale-less early hold the auto-flow finder rejects, both entry guards), and one in `tst_bletransporterror` proving a failed `A012` subscribe does not fail the connect

## Not verified — needs the machine

- **The pressure loop end to end** (gauge portafilter): run the wizard, confirm the correction lands, verify with a second run that the readings agree
- **The Heater Calibration popup still fits without scrolling at 1280×800** — it gained a row, and `heater-calibration-layout` spec-requires it to fit
- **Temperature ships unverified end to end.** Its hardware loop needs a thermocouple basket that was not available; the row, its table entry and its profile wiring are all in place

## Two bugs found while building this

Both the same mistake as the reset command — an invented value instead of a referenced one:

- A `0.15 bar/s` steady-hold threshold, reasoned from "a calibration hold is flatter than a pour", rejected every genuine hold: the DE1's PID moves pressure ~0.1–0.2 bar between samples, which at 5 Hz is already ~0.5–1.0 bar/s. Now uses the auto-flow finder's measured `0.5`.
- Sample timestamps came from wall clock, so event-loop scheduling could read as a fast-moving sensor. Now `ShotSample::timer`, the machine's own clock.

OpenSpec change: `openspec/changes/add-machine-sensor-calibration/`


skialpine:
### Code review

Found 9 issues. The first four are functional; the rest are artifacts and comments left behind by the switch to the vendor method (commit `2c8f65e0`), which deleted the measuring design they still describe.

1. The wizard is destroyed the moment its own calibration shot starts, and its `onDestruction` then reloads and re-uploads the previous profile mid-pour. The phase handler replaces the stack for `EspressoPreheating`/`Preinfusion`/`Pouring`/`Ending` with no exemption for `sensorCalibrationPage`; `replace(null, ...)` destroys it. Both test profiles are frame-based, so `loadProfile` re-uploads. On a GHC machine this is the only path. Note steam and hot water call `saveReturnToPage()` first; espresso does not.

https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/qml/main.qml#L3656-L3664

2. A locale decimal comma silently truncates the gauge reading before it reaches the machine. `parseFloat("88,5")` is `88`. The field sets `Qt.ImhFormattedNumbersOnly` with no validator or locale normalisation, so a European user entering 88,5 against a 90 C hold writes a 2.0 C correction instead of 1.5.

https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/qml/pages/SensorCalibrationPage.qml#L529-L531

3. The calibration read is issued only at page construction. `clearCalibrationCache()` fires on disconnect, so after a BLE drop and reconnect while the page is open, `hasStored` stays false and Apply is disabled permanently — recoverable only by leaving and re-entering, which nothing signposts.

https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/qml/pages/SensorCalibrationPage.qml#L113-L121

4. `profileMissing` is set on a path where the profile *was* replaced. `loadProfile` returns false both when it refuses (nothing changed) and when the profile is not found — the latter loads and uploads the default first. The page treats both the same, skips the restore, and tells the user "Nothing has been changed on your machine."

https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/qml/pages/SensorCalibrationPage.qml#L106-L112

5. `writeCalibration()` has no production caller — only `tst_de1device_mmrreads`. `applyCorrection` calls `sendCalibration` directly. Its own doc names `applyCorrection` as the one place the pair is assembled, which is true of `sendCalibration` and not of this function. The adjacent comment also says "three convenience calls"; there are two.

https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/src/ble/de1device.h#L349-L356

6. The delta spec still requires the deleted design — "The machine-reported value is measured by the wizard, never entered" (L72), "Applying a correction is unavailable until a run has been measured" (L89), "A lost connection or an abandoned run is never treated as success" (L163). The shipped controller never touches `MachineState` or shot samples, and Apply is available as soon as the profile loads. On archive these become the canonical specs.

https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/openspec/changes/add-machine-sensor-calibration/specs/machine-sensor-calibration/spec.md#L88-L100

7. `design.md` documents the removed capture controller (`Idle -> Armed -> Observing -> Measured | NoHold | Aborted`, steady-hold window, median) and an Open Question about tuning its threshold; `tasks.md` 3.2-3.4 are checked done for that machinery. `design.md` also lists "read-factory and restore" as callers of `sendCalibration`, neither of which exists.

https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/openspec/changes/add-machine-sensor-calibration/design.md#L62-L70

8. Four comments describe the deleted design or a dependency that is gone (CLAUDE.md: an un-sourced claim in a comment "gets believed and then licenses wrong code"):

- the page header says `SensorCalibration` "watches the run" and the page is "driven by `MachineState.phase`"; the file has zero `MachineState` references — https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/qml/pages/SensorCalibrationPage.qml#L9-L18
- the singleton registration says the class "needs DE1Device, MachineState and TranslationManager"; the constructor takes two of those. The "14 references across 2 QML files" count is also stale — https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/src/core/contextsingletons_qml.h#L421-L427
- the log-gate entry describes "armed, observing, held or did not hold, aborted"; the only lines are "correction refused" and "applying ... correction" — https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/scripts/check_log_markers.py#L107-L112
- "Four reads; the shared GATT queue orders them" sits above a single read — https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/qml/pages/SensorCalibrationPage.qml#L113-L116

9. Citation slips in a change that otherwise cites carefully: `a2092efc` to `69e4277c` is 77 days, eleven weeks not ten (and `tst_de1device_mmrreads.cpp` says eleven in the same PR); `parse_binary_calibration` is `binary.tcl:1325`, not 1326; the `ResetFactory` note says "above `readCalibration`" but sits after it.

https://github.com/Kulitorum/Decenza/blob/9dc480514b6d7dd36e55be9bfa70f03bef0cedbe/src/ble/de1device.h#L342-L345

<details>

- [Worth a look] No test exercises the real `subscribeAll()`, so flipping the new `CALIBRATION` subscription to `required=true` or deleting it would go unnoticed — the same function whose wiring caused #1819. Pre-existing symmetry: `WATER_LEVELS`, `READ_FROM_MMR` and `TEMPERATURES` are equally uncovered (score: 70)

</details>

skialpine:
### Code review

Found 9 issues.

**1. Quitting or being killed while the wizard is open leaves the machine on a hidden 1 °C profile.** `loadProfile` persists the selection (`setCurrentProfile` + `setSelectedFavoriteProfile(-1)`, profilemanager.cpp:1923-1929) and startup replays it (`loadProfile(m_settings->app()->currentProfile())`, profilemanager.cpp:365). The only undo is `StackView.onRemoved`, which does not fire on app teardown. `test_pressure_calibration.json` is `hidden: "1"` and declares `espresso_temperature: "1.00"`, so the next launch is silently on an invisible 9-bar profile that stops the boiler heating — and on a GHC machine the next button-started shot pulls it. Restore has to survive teardown.

https://github.com/Kulitorum/Decenza/blob/c4c2153953c6d2eeb7b280c3029d00b8753a9574/qml/pages/SensorCalibrationPage.qml#L113-L115

**2. The wizard cannot be completed on a headless machine, and this PR removed the workaround.** The page header asserts "the machine has a GHC and the user starts it there" as a flat fact, but `m_isHeadless` defaults to `true` (de1device.h:768) and the app's own convention is `canStartOperations: DE1Device.isHeadless || DE1Device.simulationMode` (main.qml:47) — on a headless DE1 the app *is* how a shot starts. The page has no start affordance at all. Before `c4c21539` a headless user could leave, start from idle, and return with the test profile still loaded; the new restore closes that path.

https://github.com/Kulitorum/Decenza/blob/c4c2153953c6d2eeb7b280c3029d00b8753a9574/qml/pages/SensorCalibrationPage.qml#L12-L17

**3. `DE1Device::disconnect()` never clears the calibration cache.** The reset block that does clear it lives in `onTransportDisconnected()`, and `disconnect()` deliberately severs the transport signals first (de1device.cpp:608) precisely so that handler does not run. Reached by the BLE↔USB transport switch (main.cpp:2254/2265) and the Android BLE-recovery path (main.cpp:1844). The cache's own comment names this exact failure: "connecting to a second DE1 and showing the first's values — and the wizard's Apply gate is exactly 'has this machine answered'."

https://github.com/Kulitorum/Decenza/blob/c4c2153953c6d2eeb7b280c3029d00b8753a9574/src/ble/de1device.cpp#L590-L593

**4. A red "nothing to correct" error appears after every successful write.** Line 575 resets `entryValue` to `declaredHold` but leaves `entryTouched` true — it is only cleared in `onDeclaredHoldChanged`, which does not fire because the hold did not change. `rejection` becomes the `alreadyAgree` string and renders in `Theme.errorColor` beside the new "Run it again" card. Deterministic.

https://github.com/Kulitorum/Decenza/blob/c4c2153953c6d2eeb7b280c3029d00b8753a9574/qml/pages/SensorCalibrationPage.qml#L573-L577

**5. The restore replays captured state unconditionally, discarding anything that changed while the page was open.** `main.qml:511-520` calls `loadAutoLoadRecipeIfNeeded()` on Sleep→Idle with no page check, and `applyActivatedRecipe()` loads the recipe's profile (maincontroller.cpp:1906). Open the wizard, let the machine wake, leave: the wizard reactivates the recipe that was current minutes ago and throws away the auto-load. Gate on `SensorCalibration.isTestProfileActive(sensor)`.

https://github.com/Kulitorum/Decenza/blob/c4c2153953c6d2eeb7b280c3029d00b8753a9574/qml/pages/SensorCalibrationPage.qml#L136-L148

**6. Restore can land the user on `default` when the profile came from JSON.** `_restoreProfile` reads `baseProfileName`, but `loadProfileFromJson` sets that to `m_currentProfile.title()` — a display title (profilemanager.cpp:1984). For an uninstalled profile (a "use this profile" from shot history) every lookup tier in `loadProfile` misses and it falls through to `loadDefaultProfile()` (profilemanager.cpp:1754-1760). The controller header's claim that `baseProfileName` is "the profile's FILENAME, never the display name" is true of `loadProfile` only, and is what licensed this.

https://github.com/Kulitorum/Decenza/blob/c4c2153953c6d2eeb7b280c3029d00b8753a9574/qml/pages/SensorCalibrationPage.qml#L107-L110

**7. `isTestProfileActive` is an app-side fact; nothing confirms the profile reached the machine.** `loadProfile` sets `m_baseProfileName` (:1913) before `uploadCurrentProfile()` (:1945), and that upload returns early whenever the machine is in an active phase (:2364), deferring to the next Idle/Ready/Sleep/Heating. In that window the wizard renders "Your machine holds 9.00 bar", enables the stepper, and `applyCorrection` will send — while the DE1 still holds the previous profile's frames. That is the failure the class comment says the design exists to prevent.

https://github.com/Kulitorum/Decenza/blob/c4c2153953c6d2eeb7b280c3029d00b8753a9574/src/controllers/sensorcalibrationcontroller.cpp#L170-L176

**8. The singleton reference count is wrong, and was wrong in the commit that fixed it for being wrong.** `c7d90052` changed "14 references across 2 QML files" to "27 references across 5 QML files". Actual: 20 occurrences across 3 files (`SensorCalibrationPage.qml` 16, `SettingsCalibrationTab.qml` 3, `SettingsPage.qml` 1). A hand-maintained count that nothing checks is drift by construction — the number probably should not be there at all.

https://github.com/Kulitorum/Decenza/blob/c4c2153953c6d2eeb7b280c3029d00b8753a9574/src/core/contextsingletons_qml.h#L420-L424

**9. `design.md` still documents the deleted measuring design and an unimplemented restore.** Three spots survived the `c7d90052` doc sweep, and `openspec archive` would carry them into the canonical specs:
- L131: "commented out **ten** weeks later" — 2018-02-27 → 2018-05-15 is 77 days, 11 weeks. `tests/tst_de1device_mmrreads.cpp:375` says "eleven-week life" in the same PR.
- L153: "A run that never holds steady produces nothing… the threshold is tunable from a real run (see Open Questions)" — the steady-hold detector was deleted, and L164 now says Open Questions is `None`.
- L158: rollback offered "by restoring factory", which this same file's "Restoring is NOT implemented" section and the spec both forbid.

https://github.com/Kulitorum/Decenza/blob/c4c2153953c6d2eeb7b280c3029d00b8753a9574/openspec/changes/add-machine-sensor-calibration/design.md#L130-L132

---

<details>
<summary>11 below-threshold issues (50-74)</summary>

- [Worth a look] `_trVersion: TranslationManager.translationVersion` added in two new files; CLAUDE.md forbids it in new code. The rule's rationale genuinely does not cover `Q_INVOKABLE`s that translate internally, but the controller already has a NOTIFY-backed `contextVersion` and `TranslationManager::translationsChanged` exists — connecting them collapses two counters into one (score: 70)
- [Worth a look] `sensorCalibrationPage` added to the phase-handler exemption (main.qml:3666) but not to the #575 disconnect list (main.qml:3568); when `transportPage` was added both lists were updated together (score: 73)
- [Worth a look] A firmware-refused write is only visible at `CAL_DETAIL` (qDebug), below the INFO default of the connections views — LOGGING.md's named failure mode (score: 70)
- [Worth a look] design.md says "Defaults for cafe" resets five `Settings.hardware` preferences; it writes six (`steamTwoTapStop` at SettingsCalibrationTab.qml:933) (score: 70)
- [Worth a look] `bluetooth.tcl:3344` cited twice for the `WriteKey` guard; actual line is 3345 (score: 68)
- [Worth a look] `writeFailed` is never cleared when the user edits the entry, so a stale "your machine did not accept that" persists until the next successful write (score: 65)
- [Worth a look] tasks.md task 2.1 (marked done) describes `restoreFactoryCalibration(target)`, an API that never existed (score: 65)
- [OK to skip] `de1_skin_settings.tcl:2391` omits its `skins/default/` path segment; line is correct and the filename is unique (score: 62)
- [OK to skip] `qFuzzyIsNull` described as "exact equality rather than an invented tolerance"; it is a 1e-12 tolerance (score: 55)
- [OK to skip] The `+1.0` shift in `qFuzzyCompare(*slot + 1.0, …)` is the standard near-zero workaround but carries no comment (score: 52)
- [OK to skip] `gui.tcl:2525-2529` starts partway into the fake-reply block (2518-2529); substantive claim is accurate (score: 52)

</details>

skialpine:
Correction to finding 2 in the review above — I overstated it.

`m_isHeadless = true` is a permissive **pre-read default**, not the codebase assuming machines are headless. Its own comment says so: "Restore the permissive default so a GHC machine's false doesn't carry into the next connection and block its start buttons until (or unless) the fresh GHC MMR read returns." Once the MMR read lands, `canStartFromApp = (ghcStatus == 0 || 1 || 2 || 4)`, and a machine with a GHC resolves to not-headless.

https://github.com/Kulitorum/Decenza/blob/c4c2153953c6d2eeb7b280c3029d00b8753a9574/src/ble/de1device.cpp#L257-L264

Most machines have a GHC, where the hardware is the only start path and the app cannot start a shot at all. The wizard instructing the user to start at the machine is therefore correct for the majority, and a start affordance in the wizard is not warranted — it would have to be gated on `isHeadless` or be a dead button on exactly the machines most users own.

What remains is a one-line comment fix: `SensorCalibrationPage.qml:13` states "the machine has a GHC" as a universal fact when it is per-machine state. Downgrading finding 2 from 90 to below threshold.

The other eight findings are unaffected.

skialpine:
Finding 7 (upload-not-confirmed) is a false positive — withdrawing it.

Its scenario requires opening the wizard while the machine is mid-shot, and there is no path to that. `EspressoPage`'s only exits are `AppShell.idleRequested()` with `stopReason = "manual"`, i.e. stopping the shot — there is no navigate-to-settings affordance during a run. And the phase handler has already replaced the stack with the espresso page, so the user cannot be sitting in Settings when a shot starts either. Reaching the wizard means stopping the shot first, after which `uploadCurrentProfile()` no longer defers.

The scoring rested on `startNavigation()` having no phase guard (main.qml:1157). That is true and irrelevant: the guard is missing at a call site that cannot be reached in the state described.

The one state where the wizard and an active phase coexist is the exemption this PR adds — wizard already open, shot started from the GHC — and there the test profile was uploaded before the shot began, so the displayed hold is what the machine holds.

Final tally: 9 findings reported, 2 withdrawn (2 and 7), 7 real. All 7 fixed in 3aabe5e7.

PR #1882, opened 2026-08-29
Collapse six repeating log sources, and name the MCP tool

skialpine:
## Why

A submitted debug log held 18,769 lines. The raw count is not the problem — idle runs at a **median of 17 lines/hour** (hour buckets over a 19 h window: 24, 9, 16, 16, 20, 7, 12, 11, 18, 30, 11, 14, 22, 13, 22, 17, 46, 44, 83). The problem is concentration: **six line shapes account for 3,419 lines**, almost all of it inside bursts where it crowds out everything else.

Every one of the six repeats byte-identical text, which is the signal that the repetition carries nothing. `LogCollapse` already existed for exactly this and had six callers — none of them these.

| source | lines | shape |
|---|---:|---|
| BLE scan lifecycle (`Scanning…` + `Scan complete`) | 1,791 | bursts, 99% in runs of 20+ |
| refractometer `Hunt active — chaining another scan` | 540 | one per cycle while the review page is open |
| `alive: constant weight` | 567 | replaced a hand-rolled 2 s throttle |
| Decent battery poll | 342 | identical every ~4 min per connection |
| `[DE1][SettingsDrift] giving up after 3 resend attempts` | 60 | **WARN**, 60 in 6.7 s |
| `[MCP] Stale session header` | 30 | **INFO**, once per request for a session's life |

### The scan lifecycle

A cycle is ~7.5 s and logs two identical lines, so a reconnect ladder or an open review page emits ~16 lines/min for as long as the device stays missing. Clustered by time:

| burst | lines | span | rate |
|---|---:|---|---|
| 1 | 230 | 29.1 min | 7.9/min |
| 2 | 158 | 19.7 min | 8.0/min |
| 3 | 46 | 7.2 min | 6.4/min |

All 230 lines in that first burst say the same thing. *"Looked 230 times over 29 minutes and never found it"* is the fact a reader wants, and it is the one thing 230 identical lines do not state.

### The drift ladder is a defect, not noise

`giving up after 3 resend attempts` returns without latching anything, so it is re-entered on every later drifting indication and re-warns each time — **60 byte-identical WARN lines in 6.7 seconds**, about nine per second, the loudest thing in the buffer at any tier.

The wording is what makes it wrong rather than merely noisy: the ladder gives up *resending* and never gives up *warning*, so the terminal line of the narrative repeats forever and trains a reader to skim the one tier that is supposed to mean "look here". Same cry-wolf pattern `BLEManager::scaleRepeatFailure()` exists to prevent, in a file that had no equivalent.

## Flush points

All six are episodic, so all six flush at their run ends — the mistake `logcollapse.h` records four of six existing callers having shipped. Every declaration site states where:

- scan family → `stopScan()`, `setRefractometerHunt()` (both edges)
- battery poll → `onTransportDisconnected()`
- liveness line → `startExtraction()`, `resetForRetare()`
- drift WARN → the three ladder resets
- MCP session → `removeSession()`, `expireSessions()`

## Also

`modern tools/call` recorded that a call happened and not *which one* — 179 lines that could not answer the first question anyone asks of them. It now names the tool.

Three text constants are centralized as a consequence: each is now written at two sites (emit and flush), and `LogCollapse` compares text to detect a repeat, so a reworded copy would silently stop collapsing with nothing failing.

## Deliberately not collapsed

Their text varies per emission, which is what a line carrying information looks like: the SAW settle trace, per-sample weight lines, `flow too low for stop-at-weight check`, R2 raw action packets. Collapsing the settle trace would have destroyed the evidence used to diagnose a shot's stop behaviour.

## Tests

Single-key `flush()` had no coverage while `flushAll()` had two slots, and four new call sites depend on two of its properties: that it **forgets the key** so the next run starts clean, and that an **unknown key reports nothing** rather than fabricating a tally (`stopScan()` flushes three keys unconditionally, so an unknown key is the common case there).

Both were verified by breaking `flush()` and watching them go red:

```
FAIL!  : tst_LogCollapse::flushForgetsTheKeySoTheNextRunStartsClean()
   Actual   (c.flush("k", 4'000).suppressed): 2
   Expected (0)                             : 0
```

The existing `flushAllReturnsEveryPendingTallyAndForgetsTheKeys` stayed **green** through that break, so the new slots cover a shape nothing else did. Both added to an existing file, not a new one.

**117 passed, 0 failed, 0 skipped, `tests_with_warnings: []`.** Gates pass: log markers, MCP tool budget, test-source duplication.

## Not measured yet

The before/after burst count. The 230-to-1 figure is a prediction from the algorithm, not an observation on a build carrying this change — it needs the app running with a device deliberately absent. Worth doing before this is relied on.

No manual entry: this is log-only and changes nothing on screen.

### DIFF
--- a/src/ble/blemanager.cpp
+++ b/src/ble/blemanager.cpp
@@ -35,6 +35,26 @@ namespace {
 // An order of magnitude above the 362 ms in which a browse resolved this scale
 // in the field — the case this exists to recover.
 constexpr int kReconnectBrowseTimeoutMs = WifiScaleDiscovery::kHdsResolveTimeoutMs;
+
+// The scan-lifecycle texts, defined once because each is now written at TWO
+// sites: the cycle that emits it and the flush that closes its run. LogCollapse
+// decides "repeat" by comparing the text, so two copies that must stay
+// byte-identical is a silent-drift hazard of the worst kind — reword one and
+// every cycle reads as a changed line, the collapse quietly stops collapsing,
+// and nothing fails.
+inline QString de1ScanningText()   { return QStringLiteral("Scanning for devices..."); }
+inline QString de1ScanDoneText()   { return QStringLiteral("Scan complete"); }
+inline QString scaleScanDoneText() { return QStringLiteral("Scan complete (background)"); }
+inline QString huntChainText()     { return QStringLiteral("Hunt active — chaining another scan"); }
+
+// Keys identify the RUN; the texts above say what the run reports. They
+// correspond one-to-one today, and they are still separate because keying on the
+// text would merge two runs the moment two sites happened to share wording —
+// which the two "Scan complete" variants above are one edit away from doing.
+constexpr auto kKeyDe1Scanning   = "de1.scanning";
+constexpr auto kKeyDe1ScanDone   = "de1.scanDone";
+constexpr auto kKeyScaleScanDone = "scale.scanDone";
+constexpr auto kKeyHuntChain     = "hunt.chain";
 }  // namespace
 #include "../network/mdnsresolver.h"
 #include "bleepochgate.h"
@@ -1307,11 +1327,40 @@ void BLEManager::doStartScan() {
     // scale fallback that found its scale and stopped. The machine's story is
     // "Found DE1:" below, which is unambiguous; the scan's story belongs to
     // whoever asked for it.
-    de1Debug(QStringLiteral("Scanning for devices..."));
+    //
+    // Collapsed: a reconnect ladder or an open review page repeats this line
+    // every ~7.5 s for as long as the device stays missing. See m_scanCycleLog.
+    logScanCycle(m_scanCycleLog, QLatin1String(kKeyDe1Scanning), de1ScanningText(),
+                 ScanLogSink::De1);
 
     // Scan for BLE devices only
     ensureDiscoveryAgent();
     m_discoveryAgent->start(QBluetoothDeviceDiscoveryAgent::LowEnergyMethod);
+}
+
+void BLEManager::scanCycleDebug(ScanLogSink sink, const QString& message) {
+    switch (sink) {
+    case ScanLogSink::De1:           de1Debug(message); break;
+    case ScanLogSink::Scale:         scaleDebug(message); break;
+    case ScanLogSink::Refractometer: refractometerDebug(message); break;
+    }
+}
+
+void BLEManager::logScanCycle(LogCollapse& collapse, const QString& key,
+                              const QString& text, ScanLogSink sink) {
+    LogCollapse::Collapsed collapsed;
+    if (collapse.shouldLog(key, text, QDateTime::currentMSecsSinceEpoch(), &collapsed))
+        scanCycleDebug(sink, text + LogCollapse::suffix(collapsed));
+}
+
+void BLEManager::flushScanCycle(LogCollapse& collapse, const QString& key,
+                                const QString& text, ScanLogSink sink) {
+    const LogCollapse::Collapsed collapsed =
+        collapse.flush(key, QDateTime::currentMSecsSinceEpoch());
+    // Nothing was suppressed → the run was a single line that already spoke, and
+    // a bare repeat of it would be a second line saying less than the first.
+    if (collapsed.suppressed > 0)
+        scanCycleDebug(sink, text + LogCollapse::suffix(collapsed));
 }
 
 void BLEManager::stopScan() {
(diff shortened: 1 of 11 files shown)


skialpine:
### Code review

Found 7 issues:

1. The constant-weight collapse is flushed at the START of the next run, never at the end of the current one. `flushConstantSampleLog()` is called only from `startExtraction()` (:957) and `resetForRetare()` (:1126), but the run is gated on `m_active`, which is cleared in `endShotCycle()` and `stopExtraction()` — neither of which flushes. So shot N's tally is emitted during shot N+1, and since `pending()` computes `spanMs` as `nowMs - lastEmitMs`, the reported span swallows the whole idle gap ("+50 identical in the preceding 32400 s" after an overnight break). This mirrors where the OLD throttle was reset, which was harmless; flushing there is not.

https://github.com/Kulitorum/Decenza/blob/a4c058d2524aaffbad38f89240f3ef918464bccc/src/machine/weightprocessor.cpp#L1049-L1059

2. The stale-session flush block is copy-pasted verbatim at two sites instead of extracted (CLAUDE.md says "Centralize anything produced at more than one site — never hand-roll it per call, never copy a helper's body... When you find yourself about to write the second copy, stop and extract."). The same PR extracts this exact pattern three other times — `flushScanCycle`, `flushDriftGiveUpLog`, `flushConstantSampleLog` — so the omission is inconsistent with itself.

https://github.com/Kulitorum/Decenza/blob/a4c058d2524aaffbad38f89240f3ef918464bccc/src/mcp/mcpserver.cpp#L2177-L2195

3. That flush also sits INSIDE the per-session reaping loop while `kStaleSessionLogKey` is a single global constant, not per-session. When a sweep reaps more than one session, the first iteration consumes the tally and the line — which reads "for the session just closed" — attributes it to whichever id came first in iteration order. Both `orphaned` and `expired` can hold multiple ids. (The block is also indented at the enclosing function's level while being loop-scoped.)

https://github.com/Kulitorum/Decenza/blob/a4c058d2524aaffbad38f89240f3ef918464bccc/src/mcp/mcpserver.cpp#L2290-L2307

4. Comment names two functions that do not exist. `grep -rn "removeSession\|expireSessions" src/` matches only this comment; the real call sites are the orphan loop in `findOrCreateSession()` and `cleanupExpiredSessions()`. CLAUDE.md: "an un-sourced claim in a comment gets believed and then licenses wrong code."

https://github.com/Kulitorum/Decenza/blob/a4c058d2524aaffbad38f89240f3ef918464bccc/src/mcp/mcpserver.h#L405-L410

5. Two comments say the hunt collapse's run "ends at `setRefractometerHunt(false)`", but the flush is unconditional at the top of `setRefractometerHunt(bool)` and fires on BOTH edges — as the comment at the flush site itself states. Two comments contradict a third describing the same call. Same wording also at `blemanager.cpp:1689`.

https://github.com/Kulitorum/Decenza/blob/a4c058d2524aaffbad38f89240f3ef918464bccc/src/ble/blemanager.h#L1184-L1190

6. The comment directly above the changed block still describes the 2-second throttle this PR removed. `kChangesOnly` has no time window at all — a changed weight emits immediately, an unchanged one is suppressed indefinitely. CLAUDE.md: "When you change how something is read or called, re-read the comments justifying its optimisations: they are claims with expiry dates."

https://github.com/Kulitorum/Decenza/blob/a4c058d2524aaffbad38f89240f3ef918464bccc/src/machine/weightprocessor.cpp#L311-L319

7. Two new comments cite "four of six existing callers got this wrong", but `logcollapse.h` says three of six are episodic and two of those never flushed — at most three could get run-ends wrong, so four is not reconcilable with its own next paragraph. The inconsistency is pre-existing in `logcollapse.h:39`; this PR propagates it into two new sites (also `weightprocessor.cpp:923-924`).

https://github.com/Kulitorum/Decenza/blob/a4c058d2524aaffbad38f89240f3ef918464bccc/src/ble/blemanager.h#L1171-L1177

<details>

- [Worth a look] Two of the four session-removal paths never flush `m_staleSessionLog` — the HTTP DELETE termination at `mcpserver.cpp:890` and the `MaxTotalSessions` eviction at `:2228`. Latent rather than live, since current MCP clients appear to let sessions idle out rather than send DELETE (score: 65)
- [OK to skip] The comment justifying the unconditional `flushDriftGiveUpLog()` placement in `maincontroller.cpp` describes a counter==0-with-pending-tally state that looks unreachable, since every counter reset is already paired with a flush. Placement is harmless either way (score: 50)

</details>

skialpine:
All 7 findings addressed in bb90bb73, plus the 65-scorer (the two session-removal paths that never flushed).

Two were fixed structurally rather than patched:

- **Constant-weight collapse** now flushes in `endShotCycle()` (the disarm chokepoint that also covers aborted-before-flow) instead of `startExtraction()`. The old placement mirrored where the 2 s throttle was cleared, which is exactly why it was wrong — a throttle reset at a run's start is harmless, a collapse flush there dates the window to the next shot.
- **MCP stale-session collapse** is now keyed by session id rather than one global key. That makes "for the session just closed" true, removes the in-loop placement, and lets all four removal paths close their own run through a single extracted `flushStaleSessionLog()`.

On finding 7: I did not invent a corrected count for `logcollapse.h`. The right number isn't recoverable from what that file records, so the two citations now carry the rule without the figure and say why it's omitted. The inconsistency at `logcollapse.h:39` is left for someone who can check the actual caller history.

117 passed, 0 failed, 0 warnings; log-marker, MCP-budget and test-duplication gates green.

PR #1883, opened 2026-08-29
Make the QML marker gate actually work, and mark the unmarked subsystems

skialpine:
## The measurement that prompted this

After #1882 I measured what fraction of the log's *redundancy* it had removed, which I had not done before opening it. On a 3,500-line sample across 7 windows (19% of an 18,755-line buffer), defining noise as **a line whose exact text already appeared**:

```
total sampled       3,500
distinct texts        245
redundant           3,255  = 93% of all lines
#1882 addressed       443  = 14% of the noise
```

The six sources that PR collapsed were a minority. The rest sat in subsystems carrying **no registered marker** — which is also why they were missed: the analysis that picked those six ranked lines by marker, and an unmarked subsystem cannot appear in a per-marker ranking.

## Root cause: the gate had the same blind spot

`check_log_markers.py` rule 4 rejects a bracketed marker in `qml/` that the registry does not declare. Its pattern was:

```python
QML_MARKER_RE = re.compile(r'"\[([A-Z][A-Za-z0-9]*)\]"')
```

A closing quote immediately after the `]`, so it matches `"[Scale]"` and nothing else. Every real log line is `console.log("[Scale] something happened")` — a **space** after the bracket. The rule saw 7 tokens in `qml/` where a start-anchored pattern sees 59, and **46 call sites under 12 unregistered names** were invisible.

A rule that passes because its pattern cannot match the shape it was written for is worse than no rule: the passing run gets read as coverage. Everything below is what the fixed rule then found.

## Five new markers

| marker | the question it answers |
|---|---|
| `AutoSleep` | why the machine did or did not sleep on its own |
| `AutoLoad` | why the loaded profile changed without anyone choosing it |
| `Keyboard` | why a physical-keyboard shortcut did nothing |
| `Weather` | why the weather is wrong or stale |
| `Update` | why the update is not being offered |

The remaining names were file/component names, not subsystems, and fold into the marker that already answers their question: `[Background]`/`[Emoji]` into `Theme`/`Font` (whose registry text already names "background images" and "glyph-coverage problems"), the three screensaver components into `Screensaver`, `[PostShotReview]` into `Refractometer` (both lines are R2 TDS readings), and the three `[Main]` lines individually. `[CustomEditorPopup]` and `[QML]` lose the prefix — a file name is not a subsystem.

## The largest repeater needed no collapse

`[AutoSleep] Inactivity elapsed but inside scheduled stay-awake window` was **489 lines — more than all of #1882 removed.** `sleepCountdownNormal` stays `<= 0` once it reaches zero, so a one-minute timer re-announced the same state for the length of the window (~8 hours).

That is a state, not a periodic report. It now logs on the **edge** into that state, cleared at all six countdown resets. A collapse would have counted repeats that should never have existed — worth noting, because reaching for `LogCollapse` first is what the previous PR would have done.

## Collapsed

`WeatherManager` and `UpdateChecker` move onto the new markers, and their four periodic lines are collapsed (125, 125, 132, 132 occurrences). All four are **periodic, not episodic** — these objects live as long as the process — so none takes a flush, which is the case `logcollapse.h` names as unable to leak. Both files join `COVERED_GLOBS`, so a future bare `qDebug` there cannot re-open the hole.

## Tests

**117 passed, 0 failed, 0 warnings**; log-marker, MCP-budget and test-duplication gates green. Verified against artifact timestamps, not the summary line.

No new test: the gate regex is exercised by the gate itself, which now fails on the tree as it stood before this change and passes after — the falsification is the 46 call sites it flagged.

## Not measured

Same caveat as #1882: the before/after line counts are predictions from the algorithm, not observations on a build carrying this change. The AutoSleep edge fix is the one where the arithmetic is certain (one line per window instead of one per minute).


skialpine:
Folded `Weather` and `Update` into a single `App` marker in e9a338b0, with the distinction kept as tags: `[App][Weather]` and `[App][Update]`.

Registry is now **16 markers**, not 18. `applogging.h` carries the rule for the next one: add a tag for anything of the same kind (licence checks, telemetry, migration housekeeping), promote to its own marker only when its lines are retrieved as a group often enough that filtering `[App]` and reading tags stops being good enough.

The three genuinely machine-facing additions stay separate: `AutoSleep`, `AutoLoad`, `Keyboard`.

117 passed, 0 failed, 0 warnings; all three gates green.

skialpine:
### Code review

Found 6 issues:

1. The rule-4 regex fix INVERTED its coverage rather than widening it. Requiring `[ :]` after the `]` means a string that is entirely a marker no longer matches — and those bare strings are the on-screen log-view filters (`markers: ["[DE1]", "[Bluetooth]"]` at SettingsConnectionsTab.qml:765, :941, :1961), which the rule's own unchanged preamble says it exists to protect: "a rename would otherwise pass rules 1-3 and leave a view silently empty." Verified by injecting an unregistered marker into a bare string: rule 4 passes. A lookahead covers both shapes: `r'"\[([A-Z][A-Za-z0-9-]*)\](?=[ :"])'`

https://github.com/Kulitorum/Decenza/blob/e9a338b0ab856fbaa16f0086425755498e3f97c3/scripts/check_log_markers.py#L241-L243

2. Three new markers are registered with descriptions promising user-facing answers, but every call site is `console.log` — which Qt maps to QtDebugMsg, which webdebuglogger.cpp renders as "DEBUG". SubsystemLogView defaults `minLevel: "INFO"`, so `debug_get_log filter="[AutoSleep]" minLevel="INFO"` returns nothing. (LOGGING.md: "a user-facing line left at DEBUG vanishes from its view.")

https://github.com/Kulitorum/Decenza/blob/e9a338b0ab856fbaa16f0086425755498e3f97c3/src/core/logtags.h#L178-L184

3. Five `qInfo()` sites were silently demoted to DEBUG because applogging.h defines no `APP_INFO_STREAM` — only the statement-form `APP_INFO_STDERR`. Worst effect is a split binary diagnostic: "UpdateRelaunchReceiver fired" stays INFO at :1393, "did NOT fire" drops to DEBUG at :1399, so a reader filtering at INFO cannot tell absence from "it didn't happen". Not mentioned in the PR.

https://github.com/Kulitorum/Decenza/blob/e9a338b0ab856fbaa16f0086425755498e3f97c3/src/core/updatechecker.cpp#L1392-L1400

4. Six sites stream a stray empty literal left over from the mechanical conversion — `APP_WARN_STREAM("Update") << "" << m_errorMessage` — producing a doubled space in the emitted line. On main these were `qWarning() << "UpdateChecker:" << m_errorMessage`.

https://github.com/Kulitorum/Decenza/blob/e9a338b0ab856fbaa16f0086425755498e3f97c3/src/core/updatechecker.cpp#L275-L277

5. The same regex is anchored to a double quote, so a marker opening a single-quoted string or a template literal is still invisible. Both are legal QML/JS; neither occurs in the tree today. Flagged because the PR's own argument is that a gate which cannot match the shape it is written for reads as coverage.

https://github.com/Kulitorum/Decenza/blob/e9a338b0ab856fbaa16f0086425755498e3f97c3/scripts/check_log_markers.py#L241-L243

6. `QDateTime::currentMSecsSinceEpoch()` is newly called here with no direct `#include <QDateTime>`; it compiles only transitively. Not a compiler-catchable defect today, which is the point.

https://github.com/Kulitorum/Decenza/blob/e9a338b0ab856fbaa16f0086425755498e3f97c3/src/core/updatechecker.cpp#L435-L437

<details>

- [OK to skip] The author's PR comment says "Registry is now 16 markers"; the actual count in logtags.h is 17. Description error only — the code and the gate are self-consistent (score: 50)

</details>

skialpine:
All 6 findings fixed in 299a81b1, plus the below-threshold one.

**The regex fix was wrong in the same way the original bug was.** `[ :]` inverted rule 4 rather than widening it. Now a lookahead that consumes nothing and admits a space, colon, or the closing quote — and the opening class accepts single quotes and backticks too, closing finding 5 in the same edit:

```python
["'`]\[([A-Z][A-Za-z0-9-]*)\](?=[ :"'`])
```

Falsified in **both** directions this time: an unregistered marker injected into a bare filter string fails the gate, and so does one injected into a log call. Neither the original pattern nor its first fix could do both, and neither was falsified before shipping — that omission is the actual root cause of this round.

**`APP_INFO_STREAM` now exists.** Its absence had made a tier decision for five call sites; they are back at INFO, and the `fired`/`did NOT fire` pair is no longer split across tiers.

**Tiers now match what the registry promises.** Outcomes are `console.info` (slept, stayed awake, woke, remote request, both auto-load paths, all 10 `[Keyboard]` lines); the countdown's own bookkeeping stays DEBUG, and the AutoSleep registry row now says so explicitly rather than promising more than the tiers deliver.

Also: six stray `<< ""` removed, direct `#include <QDateTime>` added.

Correction to my earlier comment: the registry is **17** markers, not 16.

117 passed, 0 failed, 0 warnings; all three gates green.

Issue #582, opened 2026-03-28
BR: dark blue line at top of screensaver on Lenovo Tab One


jmweiss:
Decenza v1.6.1: dark blue line at top of screensaver on Lenovo Tab One. Confirmed by another user on Decent Diaspora.

<img width="1340" height="800" alt="Image" src="https://github.com/user-attachments/assets/5a9b80e6-4e40-4ce9-895d-2388c59ccf12" />

Does not appear on my Samsung A8.

PR #1886, opened 2026-08-30
Stop wake() re-arming a running watchdog, and count AccessibilityManagers

skialpine:
Both found by reading build 3574's log on the tablet after #1885 shipped — the two things I flagged there as unresolved.

## 1. `wake()` re-arms a running watchdog (fix)

`wake()` arms the watchdog whenever `m_watchdogArmPending` is clear, and `startWatchdog()` resets `m_watchdogRetries` and `m_watchdogUpdatesSeen`. A `wake()` landing while the watchdog already runs therefore hands back the **entire retry budget** and forgets that weight data had been seen, so the next lapse is timed as a first sight (`kWatchdogFirstTimeoutMs`) rather than as a stall.

**The logic is old; what is new is that it became reachable.** Previously the enable dispatched after both wakes — the build-3573 session shows it issuing at 8.19 s against wakes at 200 ms and 500 ms — so the pending flag was still set and `wake()` never armed. #1885 cut the connect-burst contention, and build 3574 issues the enable at 5.040 s, before the 500 ms wake at 5.238 s:

```
[   5.040] [Scale][BLE DecentScale] Watchdog started (initial 1000ms timeout)
[   5.233] [Scale][BLE DecentScale] Sending wake/LCD command again (500ms)
[   5.238] [Scale][BLE DecentScale] Watchdog started (initial 1000ms timeout)
```

`wake()` now declines to arm a watchdog that is already running — the rule `onNotificationsIssued()` already follows for a later enable: restart the countdown, never reset the counters.

## 2. AccessibilityManager instance count (diagnostic, not a fix)

The class is meant to hold exactly one instance: main.cpp's stack object, published via `setQmlInstance()`. `create()` never constructs and the type is not QML-creatable.

Build 3574 shows more than one. `initTts()`'s "Available TTS engines" appears at 0.798 s and again at 1.802 s **with no "initTts() called again" warning between them** — the guard added in #1885 — so both calls found `m_tts` null. A single object cannot do that outside `shutdown()`, which is called once at quit and logs a line that is absent here. "App started" appears once, so it is not a second process either.

**The second construction site is not identified.** The tree has exactly one, no string-based `invokeMethod` reaches `initTts`, and every other reference is a pointer or a forward declaration. The timing puts it inside QML loading — after context-property registration at 0.875 s, before `engine.load()` returns at 3.197 s.

So this counter names the caller rather than guessing at it, and **the guess is the thing it exists to prevent**: a blind guard around an unidentified mechanism is exactly how the #582 diagnostics got written. It is kept permanently rather than as a probe because it asserts an invariant that holds today, costs an `int`, and says nothing at all when the app is well.

Worth stating the stake — this is not only log noise. The second instance builds its own `QTextToSpeech`, a live Android TTS binding with its own `stateChanged` handler, while `setQmlInstance()` publishes only main.cpp's object. A screen-reader user can end up with the locale on one engine and announcements on the other, or with both speaking.

## Tests

Two new slots in `tests/tst_scaleprotocol.cpp` (existing file, no new TU). `aWakeDoesNotReArmARunningWatchdog` spends part of the retry budget, sees data, then wakes, and asserts neither counter moved. **Verified it fails when reverted** — the retries drop 4 → 0.

Local: build clean, **117/117 pass, no warnings**. `check_log_markers.py` and `check_test_source_duplication.py` both OK.

## Verification gap

Same as #1885 — the scale wake path and the AccessibilityManager constructor are both Android-exercised, and the macOS build these results come from does not run them. The counter's whole purpose is to be read in the next beta's log.

### DIFF
--- a/scripts/check_log_markers.py
+++ b/scripts/check_log_markers.py
@@ -136,6 +136,14 @@ COVERED_GLOBS = [
     # re-open that hole.
     "src/weather/weathermanager.cpp",
     "src/core/updatechecker.cpp",
+    # Wholly about screen-reader announcements and the TTS engine behind them:
+    # every line is either the route an announcement took or the setup that
+    # decides which routes exist. Covered because "TalkBack says nothing on this
+    # screen" is answered ONLY by these lines, and the file shipped with two
+    # hand-rolled conventions that had already drifted apart — eight "[a11y] "
+    # bare-qInfo lines and six "AccessibilityManager: " ones — so neither a
+    # marker filter nor a single grep returned the whole story.
+    "src/core/accessibilitymanager.cpp",
 ]
 
 # Files that HOST a registered subsystem's lines alongside unrelated code.
--- a/src/ble/scales/decentscale.cpp
+++ b/src/ble/scales/decentscale.cpp
@@ -630,7 +630,22 @@ void DecentScale::wake() {
         // Measured on a tablet: armed at 6.412 by the 500 ms wake, expired at
         // 7.412, and the enable did not reach the radio until 7.690. The
         // warning fired 278 ms before the question was asked.
-        if (!m_watchdogArmPending) startWatchdog();
+        //
+        // Nor while it is ALREADY RUNNING, which is the opposite ordering and
+        // was reachable the moment the enable started dispatching promptly.
+        // startWatchdog() resets m_watchdogRetries and m_watchdogUpdatesSeen, so
+        // a second arm silently restores the whole retry budget and forgets that
+        // weight data had been seen — the next lapse is then timed as a first
+        // sight (kWatchdogFirstTimeoutMs) rather than a stall. Build 3574's log
+        // shows both arms 198 ms apart: the enable issued at 5.040 s and cleared
+        // the pending flag, then the 500 ms wake() re-armed at 5.238 s. Before
+        // the connect burst got quieter the enable landed after both wakes, so
+        // this path never ran.
+        //
+        // This is the rule onNotificationsIssued() already follows for a later
+        // enable — restart the countdown, never reset the counters.
+        const bool watchdogRunning = m_watchdogTimer && m_watchdogTimer->isActive();
+        if (!m_watchdogArmPending && !watchdogRunning) startWatchdog();
     }
 }
 
--- a/src/ble/scales/decentscale.cpp
+++ b/src/ble/scales/decentscale.cpp
@@ -171,6 +171,26 @@ void DecentScale::onCharacteristicsDiscoveryFinished(const QBluetoothUuid& servi
 
     DECENT_LOG("Characteristics discovered");
     m_characteristicsReady = true;
+
+    // BEFORE setConnected(), which emits connectedChanged() SYNCHRONOUSLY
+    // (scaledevice.cpp) and so hands control to observers while this function is
+    // still mid-way through the connect sequence. One of them re-enters us:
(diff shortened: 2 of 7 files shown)


skialpine:
### Code review

Five agents. Found 3 issues, all fixed in 19f596e8:

1. The guard was hand-rolled at one of three call sites instead of centralized (CLAUDE.md says "Centralize anything produced at more than one site — never hand-roll it per call... When you find yourself about to write the second copy, stop and extract"). `startWatchdog()` is called from `wake()`, `onNotificationsIssued()` and `armWatchdogIfEnableNeverIssues()`; only the first was guarded, leaving the others correct by an invariant nothing asserts. `startHeartbeat()` twenty lines below had already made the right call for the identical bug shape.

https://github.com/Kulitorum/Decenza/blob/9cafb0b1530dd90469c00dd078560166e37cfdc8/src/ble/scales/decentscale.cpp#L664-L676

2. Pre-existing, and worse than the bug this PR set out to fix: a re-entrant `wake()` arms the watchdog before the enable exists. `onCharacteristicsDiscoveryFinished()` calls `setConnected(true)` at the top but does not set `m_watchdogArmPending` until the wake sequence 70 lines later. `setConnected()` emits `connectedChanged()` synchronously and main.cpp's handler calls `wake()` when an LCD restore is pending, so both guards pass and the watchdog arms ~300 ms before the notify-enable is submitted. `kWatchdogFirstTimeoutMs` is 1000 ms, so it expires against a question never asked and re-enables — the duplicate CCCD write #1885 removed the 400 ms repeat to avoid. Ten of those force-disconnect a healthy scale. This also falsified what `noPathArmsTheWatchdogBeforeTheEnableIsIssued` claims: it wires no `connectedChanged` observer, so it passed either way.

https://github.com/Kulitorum/Decenza/blob/9cafb0b1530dd90469c00dd078560166e37cfdc8/src/ble/scales/decentscale.cpp#L172-L176

3. A comment claimed parity with `onNotificationsIssued()` that does not exist — that function restarts the countdown while preserving the counters, whereas the running-watchdog case restarts nothing. Flagged independently by three agents. The behaviour is correct (a `wake()`'s LCD write carries no information about when weight data should arrive), the comment was not. The test now asserts the countdown is not restarted, which no prior assertion covered.

Also fixed while in these files, per "Fix pre-existing violations in any file you touch": `accessibilitymanager.cpp` carried two hand-rolled logging conventions that had drifted apart — eight `[a11y] ` bare-`qInfo` lines and six `AccessibilityManager: ` ones — neither a registered marker, so a marker filter over a submitted log returned nothing and a grep for either returned half the story. Now an `Accessibility` marker, a helper header, all 20 lines converted, and the file in `COVERED_GLOBS`. And `s_instanceCount` is atomic, since the construction site it exists to name is the one we cannot see and cannot be assumed on-thread.

Both new guards verified by reverting them: each turns its own test red.

117/117 pass, build clean, `check_log_markers.py` and `check_test_source_duplication.py` OK.

<details>

- [OK to skip] The instance counter is monotonic ("constructions ever") rather than live-count, so it cannot distinguish two live instances from destroy-then-reconstruct — true, but no such path exists (sole production site is a stack object living for the process), and this is now stated in the comment (score: 60)
- [Worth a look, done] Redundant `m_watchdogArmPending = true` in the wake test, already set by `onCharacteristicsDiscoveryFinished` — removed, so the test exercises the production path (score: 55)

</details>

PR #1887, opened 2026-08-31
Make QML reach the real AccessibilityManager, and gate the trap tree-wide

skialpine:
The instance counter added in #1886 fired on its first build. Build 3575's log:

```
[   2.631] WARN [Accessibility][Lifetime] instance 2 constructed — the app owns exactly one …
```

Root cause is in Qt's own selection logic, not ours. `singletonConstructionMode()` (`qtdeclarative/src/qml/qml/qqmlprivate.h:155-167`) tests default-constructibility **before** it looks for a factory:

```cpp
if constexpr (!is_same_v<T, WrapperT> && HasSingletonFactory<T, WrapperT>)
    return SingletonConstructionMode::FactoryWrapper;   // QML_FOREIGN — safe
if constexpr (is_default_constructible<T>::value)
    return SingletonConstructionMode::Constructor;      // new T   <-- wins
if constexpr (HasSingletonFactory<T>::value)
    return SingletonConstructionMode::Factory;          // create()
```

`AccessibilityManager`'s only constructor was `explicit AccessibilityManager(QObject *parent = nullptr)`. Default-constructible, so Qt took `new T` at `:190` and **never called `create()` at all** — which is why its `qCritical` guard never fired and nothing ever looked wrong.

## What that actually broke

QML's `AccessibilityManager` singleton was an orphan Qt built during `engine.load()`, while `setQmlInstance()` published main.cpp's object to an `s_qmlInstance` nothing read. The split was total:

| Reaches Qt's orphan | Reaches main.cpp's object |
|---|---|
| 149 `announce()` calls | `mcpServer.setAccessibilityManager()` |
| 11 `announceLabel()` | the `announceCoaching` connection |
| every `lastAnnouncedItem` write | |
| the whole Accessibility settings tab | |

Consequences, in order of how much they bite:

1. **The settings tab and `settings_get` disagree for the session.** The tab writes the orphan; MCP reads `accessibility->enabled()` off main.cpp's object.
2. **They clobber each other's saved settings.** `saveSettings()` writes *every* `accessibility/*` key, not just the changed one, so once the two diverge either object stamps its stale copy over the other's.
3. **Two live `QTextToSpeech` engines that cannot interrupt each other** — coaching speaks on one, all QML announcements on the other.
4. **The `m_enabled` guard is per-object**, so disabling accessibility over MCP leaves QML announcements speaking.

Little of this bites a user with accessibility off, which is how it survived. All of it bites a screen-reader user.

## The fix

Drop the constructor's default `parent` so the type is no longer default-constructible and Qt falls through to the factory; main.cpp passes `nullptr` explicitly at the one construction site.

## Keeping it fixed

Two mechanisms, because neither alone suffices:

- A `static_assert(!std::is_default_constructible_v<AccessibilityManager>)` after the class — compile-time, but only where someone knew to write it.
- `scripts/check_qml_singletons.py` in the build-free per-PR job, which covers a **new** singleton whose author never knew the rule exists. Six types carry a factory today; all six are correct. `QML_FOREIGN` wrappers are exempt (`T != WrapperT` takes the FactoryWrapper branch first), and a singleton with **no** factory is meant to be Qt-constructed — the defect is specifically "has a factory AND is default-constructible".

**The check was verified by refusal, not by passing.** Restoring the default parent reproduces the exact failure naming the exact type. Two bugs in the check itself were found that way and fixed: forward declarations (`class QQmlEngine;`) were matching a later class's body and produced fifteen phantom failures, and a substring test for `QML_FOREIGN` silently skipped `MainController` because the word appears in its comments. A gate that reports green over a type it never looked at is worse than no gate.

`QML_GOTCHAS.md` documents the trap beside the existing singleton entry.

Local: build clean, **117/117 pass, no warnings**; `check_log_markers.py`, `check_qml_singletons.py` and `check_test_source_duplication.py` all OK.

## Verification gap

The behaviour change is Android/QML-side and the macOS build does not exercise the screen-reader paths. What the next beta's log should show is the absence of the `instance 2 constructed` warning.


skialpine:
### Code review

Five agents. Found 4 issues, all fixed in 74f010d7:

1. The new check was a second, weaker copy of a scanner this repo already has (CLAUDE.md says "Centralize anything produced at more than one site — never hand-roll it per call... When you find yourself about to write the second copy, stop and extract"). `tests/tst_qmlregistration.cpp` already enumerates every `QML_SINGLETON`, strips comments and strings, reconciles its parse against an independent declaration count, and `QFAIL`s on anything it cannot read.

https://github.com/Kulitorum/Decenza/blob/005e7d01a1e1bb9d1d5dbc4ff8d5cc9e4c0c9b9e/tests/tst_qmlregistration.cpp#L63-L68

2. Three real bugs in that script, two of them silent. `= delete` could never match, so a correctly-deleted default constructor would have FAILED the gate; a defaulted parameter with a comma inside brackets (`QSize s = QSize(0, 0)`) parsed as not-default-constructible and returned exit 0 on the exact defect it exists to catch; and `class Q_DECL_EXPORT Foo` captured the macro as the class name. Its docstring also promised a blindness guard ("says so and fails rather than guessing") that no code implemented.

3. Stale claims left behind by this PR, each a premise the PR itself moved (CLAUDE.md says "Re-derive the justification when the design around it moves... a comment nobody re-derives is how a bad decision keeps getting re-approved"). `accessibilitymanager.cpp` still said the second construction site "is not identified" — in the PR that identifies it — and justified its atomic by "the one we cannot see".

https://github.com/Kulitorum/Decenza/blob/005e7d01a1e1bb9d1d5dbc4ff8d5cc9e4c0c9b9e/src/core/accessibilitymanager.cpp#L64-L68

4. `QML_GOTCHAS.md` presented an abridged paraphrase of `singletonConstructionMode()` as if it were a verbatim quote, dropping a branch and the `::value` suffixes — so it would not compile as written and would not diff against a future Qt.

**What replaced the script is stronger in both directions.** All six factory-bearing singletons now carry `static_assert(!std::is_default_constructible_v<T>)`, so the compiler proves the property instead of a regex guessing at it — the other five had no guard at all, safe only because their constructors happen to require arguments. One slot in `tst_qmlregistration` then asserts a singleton declaring `create()` HAS that assert, which is the case no compile-time check can cover.

Both halves verified by refusal: removing `MachineState`'s assert fails the test slot; giving its constructor a defaulted parameter fails the build with the assert's own message.

Also fixed: `main.cpp`'s migration comment cited line numbers drifted by ~130 and ~360 lines, and `CLAUDE.md`'s gotcha list had no pointer to the new section though that doc's header says every section has one.

117/117 pass, build clean, `check_log_markers.py` and `check_test_source_duplication.py` OK.

<details>

- [OK to skip] The narrative was hand-written in four places with minor drift already — reduced to two (the gotchas doc and the code comments that cite it) when the script was deleted (score: 60)
- [OK to skip] PR body said "qCritical guard"; it is `A11Y_WARN_STDERR` → `qWarning` since #1886. Never landed in a code comment (score: 55)

</details>

PR #1888, opened 2026-08-31
Ask the platform whether a screen reader is running, not whether one attached

skialpine:
Found while validating #1887: accessibility was enabled, VoiceOver was **not running**, and the app said nothing.

## The bug

`AccessibilityManager` suppresses its own `QTextToSpeech` when it believes a screen reader is active, so the two don't talk over each other (#889). The belief was `QAccessible::isActive()` — which Qt documents as returning false *"until a tool such as a screen reader accessed the accessibility framework"*. That's an **attachment** test. Anything touching the platform accessibility API flips it.

So the app handed every announcement to a reader that wasn't there, and the log filled with:

```
[Accessibility][Route] path=platform isActive=true len=21 preview=Accessibility enabled
```

which reads like success. Qt reports nothing about whether an announcement was consumed, so there's no delivery signal to fall back on — the question has to be answered before dispatch or not at all.

## Not a local discovery

Telegram Desktop shipped the identical false positive — [tdesktop#30511](https://github.com/telegramdesktop/tdesktop/issues/30511), a screen-reader banner on every Linux X11 desktop with no reader running, because the `AT_SPI_BUS` atom makes `isActive()` true. They resolved it the same way this does: platform-specific detection where a real answer exists, Qt's generic state elsewhere.

## The fix

`decenzaPlatformScreenReaderActive()` asks each platform the broadest question it can actually answer, returning `nullopt` where there's no better answer than Qt's:

| Platform | Check |
|---|---|
| macOS | `-[NSWorkspace isVoiceOverEnabled]` |
| iOS | `UIAccessibilityIsVoiceOverRunning()` |
| Windows | `SPI_GETSCREENREADER` |
| else | `nullopt` → `QAccessible::isActive()` |

Windows deliberately uses the **flag NVDA, JAWS and Narrator all set** rather than probing for products by name, so a reader this code has never heard of still counts.

**The two failure directions aren't equally bad, and that decides the design.** Claiming a reader is present when none is produces **silence** — the feature dead, with a log line that looks like success. Claiming none is present when one is produces **double speech** — irritating, but the content is heard and the app's voice can be switched off. So an imprecise check should err toward reporting a reader, which is exactly what `SPI_GETSCREENREADER` does (Magnifier sets it without speaking).

Android keeps `isActive()` — TalkBack is what activates accessibility there. Linux keeps it too and is **known wrong** for the tdesktop reason; that's recorded in the header rather than guessed at, because fixing it needs an AT-SPI query nobody here can test today.

## Build plumbing

The `.mm` needed `SKIP_PRECOMPILE_HEADERS` in the test library — the first Objective-C++ source ever compiled into a test target, which `tests/CMakeLists.txt:278` had already predicted would need exactly this. The app target needed no edit: its skip-list is derived from `SOURCES` by matching `.mm` rather than hand-written. No `.cpp` lost its PCH.

## Verification

**macOS, confirmed by hand:** with VoiceOver off the app now takes `path=tts` and speaks; with VoiceOver on it still routes to the platform and VoiceOver speaks. 117/117 tests pass, build clean.

**Not verified:** the Windows and iOS branches compile on no build available here. Windows is the one worth a CI build before merge — say the word and I'll dispatch `windows-release.yml`.

No unit test: the change is inside `isScreenReaderActive()`, which is `virtual` and overridden by the existing accessibility tests, so a test would exercise the override rather than the new code. Platform-guarded code is verified by building on that platform, per the project's rules.


skialpine:
### Code review

Four agents. Found 4 issues, all fixed in 62798f57:

1. **The Windows branch was a regression.** Microsoft documents that Narrator does *not* set `SPI_GETSCREENREADER`. With Narrator running the call succeeds writing `FALSE`, which returned an engaged `optional{false}`, short-circuited `QAccessible::isActive()`, and let the app speak over Windows' default screen reader — the exact overlap #889 removed. Now positive-only: `true` means a reader announced itself, anything else is `nullopt` and Qt decides (which is `main`'s behaviour, so it cannot regress). NVDA and JAWS set the flag and are unaffected.

2. **The tdesktop citation was backwards** (CLAUDE.md: "an un-sourced claim in a comment gets believed and then licenses wrong code"). I wrote that Telegram "resolved it the same way this does". They didn't — an app-side Linux detector was argued down in review as the wrong layer and withdrawn, and the fix went to Qt ([QTBUG-145881](https://bugreports.qt.io/browse/QTBUG-145881)). The issue is evidence that `isActive()` over-reports and that the Linux half belongs upstream, not evidence for this approach. The open question it left is also closed: that Qt change is **not** in 6.11.2 — still open on Gerrit, and `qtbase`'s `dbusconnection.cpp` still sets `m_enabled` unconditionally from the XCB atom.

3. **Two comments contradicted each other on which direction is safe.** The header said over-reporting a reader causes silence and is the bad direction; the `.cpp` called over-reporting "the safe one". The header was right — silence kills the feature, double speech merely irritates — so an imprecise check errs toward `nullopt`.

4. **The fix narrowed when the bug fires but left the log line that hid it unchanged.** `path=platform` read identically whether a platform API confirmed a reader or the known-wrong fallback guessed one. Route lines now carry `source=platform-probe` or `source=qt-fallback`, so a silence report saying `qt-fallback` has named its own cause.

Tests added: the `optional`/`bool` composition is extracted as `decenzaResolveScreenReaderActive()` with its three cases pinned, since engaged-`false` vs `nullopt` is exactly where the Windows bug lived; plus a slot asserting the probe answers where platform code exists and returns `nullopt` where it doesn't. **Verified by refusal** — making macOS return `nullopt` reddens it.

Also fixed: `ACCESSIBILITY.md` still documented `isActive()` as the platform-path trigger and named Narrator as a case it covers; the PCH comment said five `.mm` when there are six; and two claims stated as fact ("the only screen reader on macOS", "no permission prompt") are now marked as observations.

117/117, build clean.

**Not mergeable as-is:** the Windows and iOS branches are compiled by nothing that has run. Windows now carries real logic and warrants `windows-release.yml` against this branch before merge.

PR #1890, opened 2026-08-31
Match de1app: every pressure step carries an 8 mL/s flow limit

skialpine:
## Why

Decent's second machine (Bengle) reaches ~20 mL/s where the DE1 manages 7-8, so a pressure step with no flow limit pours very differently on the two — and profiles are shared freely between owners of both. The Diaspora discussion settled on capping every pressure step at 8 mL/s by default, now shipped in de1app nightly ([`skialpine/de1app@fdd091f3`](https://github.com/skialpine/de1app/commit/fdd091f31209c834ab14aa1c766d334aab215b68)).

Decenza has to brew a given profile the same way de1app does, and did not: of the 77 shipped profiles that contain a pressure step, **74 had at least one step with no flow limit**. This is not only about Bengle — 8 mL/s is inside the DE1's own range, so a profile that would have run at 8.5 mL/s on a fast puck was capped in de1app and uncapped here.

## What changed

- **`Profile::applyDefaultPressureFlowLimit()`** caps unlimited pressure steps and defaults a `settings_2a` profile's scalar `maximum_flow`. Called from `loadProfile()` and `loadProfileFromJson()` — where a profile becomes current — and deliberately nowhere else:
  - **Not in a parse.** A stored profile (a shot's record of what it was pulled with, an import being compared for de-duplication) must read back exactly what it says.
  - **Not after an editor save.** de1app caps at load only, so its plugins write uncapped frames on save and the cap reappears on the next load. `tst_recipeeditorapppath` compares against those plugins and rejected an attempt to cap there.
  - In `loadProfile()` the call sits after every on-disk write in that function, so loading can never persist it. It lands in the user's file only when the user saves.
- **Forced-rise frames carry the limiter** and are named `"forced rise"` — de1app dropped "without limit" once it stopped being true. `ProfileFrame::isForcedRiseName()` accepts both spellings, so profiles written before the rename keep their preinfusion count (and their Stop-at-Volume behaviour).
- **A pressure step's flow limit can no longer be off.** Floors at 0.1, a typed 0 snaps to the default, switching a step from Flow to Pressure snaps a 0 limiter. `ValueInput` gains an opt-in `snapZeroTo` (default 0 = disabled, so no other caller changes) — without it the control's own clamp turns a typed 0 into 0.1 rather than 8, and de1app does both.
- **Editable flow ceiling 8 → 20 mL/s**, so a profile authored on a higher-flow machine opens and saves here without being clamped to a value its author never chose.

Both numbers reach QML through `ProfileManager.defaultPressureFlowLimit` / `.maxSettableFlow`; no QML file spells 8 or 20 itself.

## Oracles regenerated

Three recorded oracles moved with the de1app bump, all limiter-only:

- `tests/data/de1app_packed/` — 4 goldens. de1app's own packer now emits the rise frame's extension frame, byte-identically to ours.
- `tests/data/edit_matrix/` — 107 goldens, but the tool needed fixing first. `de1app_edit_oracle.tcl` set `advanced_shot` and ran the plugin's `prep`/`update_*` without ever running de1app's load path, so it could not see a cap that `select_profile` applies. It now sources `profile.tcl` and calls the real `::profile::apply_default_flow_limit_to_pressure_steps` — the proc itself, not a transcription, which is what keeps it from going stale on the next bump.
- `resources/profiles/` — 4 built-in JSONs via `profile_sync --sync`. Nothing else in the 89 differed.

## Steam changes from the same de1app commit

`fdd091f3` carried three steam changes alongside the flow-limit work. One applies to us and is included here:

- **Rounding, not truncation, on scaled settings.** de1app moved its steam-flow entry from `int()` to `round()`. Decenza had the same defect in **six** places, not one: every setting stored as a scaled integer (steam flow, hot-water flow, and four heater settings) arrived as a real-world double and was written with `static_cast`, which truncates. Binary floating point puts the product under the tick often enough to matter — `0.57 * 100` is `56.999999999999993`, so setting 0.57 mL/s stored 56 and read back as 0.56. All six now route through one `scaledSettingValue()` helper in `src/mcp/mcptools_write.cpp` rather than six separate `qRound`s, since the sites differ only in scale.

The other two do not apply:

- **`steam_flow` default 700 → 70** fixes a de1app bug we never had. 700 meant 7.0 mL/s, which Streamline read as "off", so fresh installs appeared to have steam disabled. Decenza's default is 150 (1.5 mL/s).
- **Streamline's `steam_disabled` / `steam_timeout` reconcile** is skin-local. It exists because de1app carries two representations of steam on/off — Insight's `steam_disabled` flag and Streamline's `steam_timeout == 0` — which drift when switching skins. Decenza has one: `steamDisabled` is a runtime flag driven by the Off pitcher preset, never derived from a timeout. There is no second source to reconcile.

## Not in scope

The same de1app commit's Samsung font/resolution fix has no counterpart here (Decenza does not rewrite screen size from the running resolution), and its Streamline steam on/off fix is skin-local — Decenza's `steamDisabled` is a runtime flag driven by the Off pitcher preset, not derived from the steam timeout. de1app's `steam_flow` default moved 700 → 70 to fix a fresh install reading as steam-off in Streamline; Decenza's default is already 150 (1.5 mL/s).

## Worth knowing

de1app's `convert_float_to_U8P4` clamps its input to 16 then multiplies by 16 — 256, a byte overflow — so nothing above ~15.94 mL/s survives their wire encoding either. Ours saturates cleanly at 15.9375. The 20 mL/s ceiling is real for authoring and round-tripping profiles; actually delivering it presumably comes with Bengle support.

## Testing

117/117 local tests pass, no warnings. `qmllint_check` clean (241/241 analysed). Five new cases in `tst_profile` (cap applied, explicit limit preserved, flow step untouched, `settings_2a` scalar reaches the generated frames, both frame names counted). Three existing assertions encoded the old "no limiter on forced rise" model and were updated.

Manual entry added to the wiki's Profile Editors section.


skialpine:
### Code review

Found 3 issues:

1. The load-time cap leaks into `loadProfile()`'s two on-disk write-backs, and for a profile stored with the flat de1app `max_flow_or_pressure` spelling it is silently persisted into the user's file. `setCurrentProfile()` applies the cap at the hand-over, but the recipe-block strip and the `espresso_temperature` repair both run *below* it and both pass `m_currentProfile` — the capped profile — to `writeProfileBackIfLossless`. `collectParityErrors` only walks keys present in the *before* object, so an **added** `limiter` key is invisible to the gate: parity passes and the file is rewritten. Where the gate does fire (nested `limiter` encoding), the outcome is the other failure — the repair is `Refused` on every load, forever, instead of once. Either way the comment on these lines and the `qInfo` in `setCurrentProfile()` ("not written to disk unless the profile is saved") state the opposite of what happens. Note this is a regression from the centralization commit: before it, the cap ran last, after every write. Fix: pass the uncapped `candidate` to both write-backs.

https://github.com/Kulitorum/Decenza/blob/137d97a7c40661222244ecabdbf65cfc19e075a2/src/controllers/profilemanager.cpp#L1777-L1783

2. `addFrame()` still creates an unlimited pressure step, so the PR's own invariant ("a pressure step's flow limit can no longer be off") has a hole, and the editor now renders that state as `0.00 mL/s` instead of "off". The new frame is `pump = "pressure"` with `maxFlowOrPressure` left at its `0.0` default, and it mutates `m_currentProfile` in place rather than going through `setCurrentProfile()` — the only place the cap runs. On the wire 0 means no limiter is sent at all (the extension frame is only emitted when `maxFlowOrPressure > 0`), so the frame reaches the DE1 uncapped, while the control shows a value below its own `from: 0.1` floor. Before this PR the same state correctly read "off".

https://github.com/Kulitorum/Decenza/blob/137d97a7c40661222244ecabdbf65cfc19e075a2/src/controllers/profilemanager.cpp#L3300-L3312

The display half, which suppresses "off" for pressure steps:

https://github.com/Kulitorum/Decenza/blob/137d97a7c40661222244ecabdbf65cfc19e075a2/qml/pages/ProfileEditorPage.qml#L1393-L1396

3. Two flow ceilings were not migrated to `Profile::kMaxSettableFlow`, reopening the exact drift this PR fixes elsewhere (CLAUDE.md says "Centralize anything produced at more than one site — never hand-roll it per call... each copy is free to change alone, silently, and nothing fails when one does"). `clamp()` and `validate()` were widened to 20 for all four of `pourFlow`, `holdFlow`, `flowEnd`, `preinfusionFlowRate`, and the QML was migrated for the first two — but `preinfusionFlowRate` is still `to: 10` and `flowEnd` still `to: 8`. `ValueInput` clamps on edit, so a high-flow profile's value is silently reduced the moment either slider is touched.

https://github.com/Kulitorum/Decenza/blob/137d97a7c40661222244ecabdbf65cfc19e075a2/qml/pages/SimpleProfileEditorPage.qml#L451-L453

https://github.com/Kulitorum/Decenza/blob/137d97a7c40661222244ecabdbf65cfc19e075a2/qml/pages/SimpleProfileEditorPage.qml#L572-L575

<details>

- [Worth a look] The `checkFlow` comment implies `clamp()` was widened in an earlier change with `validate()` left behind; both were widened inside this PR's own two commits, so the stated history did not happen (score: 70)
- [Worth a look] `"forced rise"` is a much more collidable name than `"forced rise without limit"`, and `countPreinfuseFramesWithForcedRise()` matches it anywhere in the list — a hand-authored simple `.tcl` with a frame of that name would get an inflated `NumberOfPreinfuseFrames`. Needs a hand-edited file to hit (score: 65)
- [OK to skip] The rise frame is named `kForcedRiseName` unconditionally while the limiter is attached only when `maximumFlow > 0`, so the name can outrun reality in memory — but `applyDefaultPressureFlowLimit()` corrects it before the profile brews (score: 50)

</details>

skialpine:
### Code review

Re-review of `08c4a245`. All six issues from the previous round were verified fixed by tracing the code rather than trusting the commit message. Found 1 new issue, since fixed in `82f373bf`:

1. The conditional forced-rise frame naming diverged from de1app and mislabelled a frame that ends up limited. de1app names the frame unconditionally at both sites and gates only the limiter value on `maximum_flow` (`de1plus/profile.tcl:97`, `:163`), and this PR's own spec says the generator "SHALL name these frames `forced rise`". The two mechanisms also interacted: a frame generated with no limiter took the legacy name, then `applyDefaultPressureFlowLimit()` put 8 mL/s on it at load — so the live profile carried a frame named "without limit" whose limiter was 8, which is what gets saved, uploaded to Visualizer, and put in the MCP payload. No test caught it; both name assertions use non-zero-limiter fixtures, so the branch was never exercised.

https://github.com/Kulitorum/Decenza/blob/08c4a245e4f3237547b74b0f4b036d89ac6caab9/src/profile/profile.cpp#L204-L209

Worth recording: that finding scored 50 in the previous round — the lowest of the six. Acting on the marginal one is what introduced the regression.

<details>

- [Worth a look] `docs/CLAUDE_MD/RECIPE_PROFILES.md` still said the hand-over sits "below the notes backfill, after every on-disk write" — true before the centralization, not after; the fix was to pass `candidate` to the two write-backs, not to move the hand-over. It also cited the test covering the encoding upgrade rather than the one covering the write-backs. Fixed in `82f373bf` (score: 70)
- [Worth a look] The two comments justifying the `candidate` hand-over gave opposite accounts of the same mechanism — one said the parity gate is blind to an added limiter, the other that it catches it. Both occur, depending on whether the stored step carries an explicit limiter. Stated once, accurately, in `82f373bf` (score: 65)

</details>

<details>
<summary>1 false positive, recorded for the next reader</summary>

- A reviewer reported that `from: 0.1` on the Simple editor's flow control took the "off" state away from the Flow editor as well. It did not: that bound was already `isFlow ? 0.1 : 0`, so flow mode always had the 0.1 floor and only the pressure arm changed. Verified against `git diff main...HEAD`.

</details>

Verified clean this round: the `candidate` hand-over (`candidate` is an lvalue passed to a by-value parameter, so it survives the internal `std::move`, and nothing between the hand-over and the write-backs mutates `m_currentProfile` beyond a flag); `addFrame()`'s new default and its range fallback; the flow-ceiling single-sourcing across all three editor pages; and the narrowed preinfuse-count fold-in, which leaves `settings_2a` — the type the original Stop-at-Volume bug was reported against — on the forced-rise path.

PR #722, opened 2026-04-12
feat: POC — AI Chat via Claude Code Remote Control

skialpine:
## Summary

- Add "Claude Desktop" Discuss mode that opens a persistent Claude Code Remote Control session for dialing-in conversations with live MCP context
- New `current_dialing_context` MCP resource (bean, grinder, last 3 shots, profile, machine phase) and `get_agent_file` tool for self-bootstrapping CLAUDE.md
- iOS `SFSafariViewController` bridge bypasses universal links on older iPads (iOS 17) where the Claude app lacks the Code tab
- `/mcp/setup` page updates gated behind `QT_DEBUG` — release builds show original page unchanged
- Includes both openspec changes: `poc-remote-control-chat` (active POC) and `add-claude-code-mcp-chat` (parked bridge spec)

## Key design decisions

- **User-managed session lifecycle**: user runs `claude remote-control --name "Decenza_REMOTE" --spawn=session` themselves, pastes session URL into Decenza settings. No QProcess, no config auto-write — works identically on all Decenza platforms.
- **`--spawn=session`**: single-session mode so `--name` applies deterministically (default pool mode names sessions after the hostname)
- **Index 7 append**: "Claude Desktop" added after "None" (index 6) to avoid settings migration for existing users
- **SFSafariViewController on iOS**: bypasses universal link interception so claude.ai/code renders in an in-app webview regardless of installed Claude app version. Auto-dismissed by screensaver to prevent burn-in.
- **All Discuss paths unified**: PostShotReviewPage, ShotDetailPage, and DiscussItem now all route through `Settings.openDiscussUrl()` for consistent platform-specific handling

## Test plan

- [ ] macOS: Select Claude Desktop, paste session URL, tap Discuss → browser opens claude.ai/code session
- [ ] iOS 18+ (iPhone): tap Discuss → SFSafariViewController opens claude.ai/code session in-app
- [ ] iOS 17 (iPad): same — SFSafariViewController bypasses old Claude app without Code tab
- [ ] Screensaver: open Discuss session, wait for screensaver → Safari view auto-dismisses, screensaver renders
- [ ] Other Discuss modes (Claude App, ChatGPT, etc.): behavior unchanged
- [ ] MCP: `current_dialing_context` returns correct bean/grinder/shots; `get_agent_file` returns CLAUDE.md with version
- [ ] Release build: `/mcp/setup` page shows original content (no AI Chat section)
[...]

PR #1892, opened 2026-09-01
Decode the Skale's weight exponent instead of assuming it is -1

skialpine:
## The bug

The Atomax Skale II sends five-byte weight notifications on `EF81`: a flag byte, a signed 24-bit little-endian mantissa, and a signed base-10 exponent. [skalescale.cpp:136](https://github.com/Kulitorum/Decenza/blob/main/src/ble/scales/skalescale.cpp#L136) read bytes 1-2 as a 16-bit mantissa and divided by ten — correct only for exponent `-1`, which is the one value every observed scale sends.

| Frame | Correct | Old parser |
|---|---:|---:|
| `00 D2 04 00 FE` | 12.34 g | **123.4 g** |
| `00 E8 03 00 FF` | 100.0 g | 100.0 g ✓ |
| `00 7B 00 00 01` | 1230 g | **12.3 g** |
| `00 C9 FD FF FF` | −56.7 g | −56.7 g ✓ |

de1app has the same decode (`binary scan ... cus1cu` then `t1 / 10.0`), so this is inherited across all three apps rather than a Decenza mistake.

**No Skale has been observed sending another exponent** — this is a latent misread, not a reported fault. It's worth correcting because the value feeds stop-at-weight, where a silent factor of ten is worse than most bugs.

## The fix

`SkaleScaleProtocol::decodeWeightGrams()` — length check, 24-bit sign extension, exponent applied. Same shape as the Decent scale fix in #1891: dispatch on frame length first, and report what could not be decoded once per shape through the shared `scaleFrameShapeLine()`. The old parser accepted any frame of three bytes or more, so a truncated notification produced a reading; only a five-byte frame does now.

**Not** taken from decaid#722: its 4-byte `/2560` compatibility path. That's decaid's own historical representation, and adding a branch for a format Decenza has never emitted would invent a compatibility requirement.

## Log noise

One line per undecodable frame shape per connection, DEBUG, collapsed and capped by the shared helper, flushed at disconnect. Zero lines on a working scale.

## Tests

Two data tables and one collapse test added to the existing `tst_scaleprotocol.cpp` (no new file); `skalescale.cpp` joins that target. The decode rows are the worked examples from [decaid#712](https://github.com/decentespresso/decaid/issues/712), verified arithmetically before they were written — which caught one row of mine that had the wrong exponent byte.

Suite green: **117 passed, 0 failed, no warnings.**

## Not verified on hardware

There's no Skale here — this is desk-verified against the SDK format and the issue's examples only.

### DIFF
--- a/src/ble/protocol/skalescaleprotocol.h
+++ b/src/ble/protocol/skalescaleprotocol.h
@@ -1,0 +1,61 @@
+#pragma once
+
+#include <QByteArray>
+
+#include <cmath>
+#include <cstdint>
+#include <optional>
+
+// Atomax Skale II weight notifications (characteristic EF81).
+//
+// Frame: [flags, mantissa LE 24-bit signed, exponent int8], weight = mantissa * 10^exponent.
+namespace SkaleScaleProtocol {
+
+inline constexpr qsizetype WeightFrameLength = 5;
+
+// The exponent range this driver will act on. The SDK format permits any int8,
+// but nothing between a microgram and a tonne needs more, and the value reached
+// here feeds stop-at-weight: 10^127 would arrive as a weight, not as an error.
+inline constexpr int MinExponent = -6;
+inline constexpr int MaxExponent = 6;
+
+// Grams, or nullopt when the frame is not one this decoder can read.
+//
+// The exponent is the point. Decenza, de1app (bluetooth.tcl, `binary scan ...
+// cus1cu` then `t1 / 10.0`) and decaid all read bytes 1-2 as a 16-bit mantissa
+// and divided by ten, which is correct ONLY for exponent -1 -- the one value
+// every observed scale happens to send. A frame carrying any other exponent
+// decoded to a weight wrong by a factor of ten or a hundred:
+//
+//   00 D2 04 00 FE   1234 x 10^-2 = 12.34 g,  read as 123.4 g
+//   00 7B 00 00 01    123 x 10^1  = 1230 g,   read as 12.3 g
+//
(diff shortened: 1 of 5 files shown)


skialpine:
Closing unfixed. The decode here is a real protocol misread — `int16` over bytes 1-2 divided by ten is correct only at exponent -1 — but the risk of shipping it outweighs the benefit:

- **Nothing is reported.** decaid#712 says the same about its own scale: the observed R029 shows plausible weights, ordinary notifications use exponent -1, and it explicitly declines to claim current readings are wrong. This fixes frames nobody has seen.
- **We cannot test it.** No Skale here. This touches the only weight path of a supported scale, so getting it wrong means a connected scale that never weighs and a stop-at-weight that never fires.
- **That failure was not hypothetical in this PR.** The first version gated on `size() == 5`, which would have dropped every frame from the scale de1app's simulator models (18 bytes: flag + int16 tenths + markers, `de1plus/blz/blz_sim.tcl:283-292`). Neither the tests nor review-by-reading caught it; it took a reviewer opening the simulator.
- **The length change alone can break a working scale.** Moving from `>= 3` to `>= 5` drops 4-byte frames, and decaid deliberately *retained* a 4-byte decoder as a compatibility path — evidence such frames exist somewhere.

The old decode has field evidence behind it; the replacement has a layout derived from a stripped Atomax binary. With no report and no hardware, working code wins.

**Revisit when** a Skale user reports a weight off by a factor of ten or a hundred, or when a Skale is available to capture from. The analysis is in decentespresso/decaid#712 and #722, and this branch's history has a working patch to start from.

PR #1898, opened 2026-09-01
Make the grind picker open fast, and fix #1713 still living on the web datalist

### LINKED ISSUE #1713: [Bug] Grind settings only showing a single digit (no decimal point)

skialpine:
A user reported the grind wheel taking **several seconds** to open, and the app feeling generally slower. This is what the log and the profiler actually showed, what was done about it, and one unrelated user-facing bug found on the way.

## What was measured, not guessed

**From the user's log** (5,788 shots, Android tablet), the two history reads behind the picker's step:

```
[12263.585] grind step for DF83V = 0.5, derived from 59 distinct numeric setting(s)
[12264.068] grind step for DF83V (rpm) = 100, derived from 12 distinct numeric setting(s)
```

483 ms for the RPM query alone, its twin the same — ~918 ms of query in front of a dialog that does all its work in `onAboutToShow`, before it becomes visible.

**From the QML profiler** (ten opens on a Mac), the rest:

```
GrindRowSource.qml:204  stepGrind   186 ms / 8020 calls, 125 ms self
```

8020 = 10 opens x 802 rows. 63% of the whole open.

## Performance

**1. Covering indexes (migration 40).** The two step queries filter on `equipment_id` with nothing indexed on it, so SQLite scanned every row — and the cost tracks table *bytes*, because `profile_json` and `debug_log` are TEXT on `shots`. The plan moves to `SEARCH shots USING COVERING INDEX`, so the table is never opened. Verified in the shipped configuration: with no `sqlite_stat1` present (the app never runs `ANALYZE`), both queries still plan as covering. 0.15 MB for both.

**2. `SettingsDye::stepGrinderSettingRange()`** — one QML→C++ crossing per wheel instead of 801, resolving the grinder once and parsing the current value once. The per-delta rules live in a shared `stepFromParsed()` so the batch and single-row forms cannot drift.

**3. A row memo** keyed on `(cur, step, brand, model)`. Returning the same reference means QML signals no change, so the Tumbler skips rebuilding a model it already has. The two history-derived paths are deliberately excluded — their rows depend on an input the key does not name, which is the invalidated-on-the-wrong-axis failure CLAUDE.md records for the deleted distinct-value cache.

Plus `GrinderAliases::findEntry` memoized for its remaining per-row caller, and `allGrinders()` returning by const reference — `findEntry` returns a pointer into it, previously valid only because QVector is implicitly shared.

`onAboutToShow` total, measured on the Mac (1,187 shots, debug + sanitizers):

| | before | after |
|---|---|---|
| grind changed | 32–43 ms | **10 ms** |
| grind unchanged | 32–43 ms | **1 ms** |

Plus ~918 ms of query removed at the reporter's data scale.

## Correctness: #1713 was still live on the web surface

`GrindCandidates::stepDecimals()` took the step alone while its comment claimed it mirrored `GrindRowSource._stepDecimals`, which has taken the max with the *value* since #1713. That max is the fix.

With history too thin to derive a step, `grindStep` is 0, the endpoint falls back to `1.0` = zero decimals, and a recorded `1.1` formats to `1`. The user's own setting was absent from the datalist and the n=0 suggestion silently rewrote it — #1713's exact reported shape. The app wheel offered `0.1 / 1.1 / 2.1` for a record whose web datalist offered `0 / 1 / 2`.

Found while reviewing this PR, in the declared C++ twin of the file it rewrites.

## What was tried and dropped

- **Memoizing `_stepDecimals` inside the loop.** Measured: no change. The cost was the crossings, not the work behind them.
- **Hoisting `stepGrind`'s invariant work into a prepared context.** Correct as far as it went, but there is no QML test infrastructure here and it restructured four branches, three of which would have shipped unverified.
- **Cutting `grindWindowSteps: 400`.** Raised twice as the likely lever; the measurement says no. Nothing dominates the remaining 10 ms, so shrinking the wheel's range would change behaviour for every user to save 3–4 ms nobody can feel.

## Honest limits

Neither available machine reproduces "several seconds", so **this is not confirmed to fix the reported symptom**. Established: ~918 ms of query at his data scale, and a 3–4x cut to the rest. The temporary instrumentation on the open path is deliberate — it logs at INFO so it reaches the connections view and any submitted log, and his device is the only one that reproduces the problem. It comes out once his log lands.

## Tests

Four, each verified by breaking the code and watching it go red:

- The query plan is `COVERING INDEX`, not merely that the indexes exist — deliberately without `ANALYZE`, since no shipped database has statistics.
- The batch form is indistinguishable from the single-row form for every `n`, over nine grinder shapes plus the degenerate window.
- The `findEntry` memo cannot serve a stale grinder. Its first draft used two Niche models that are both `NumericWithSuffix`, so it passed with the model dropped from the key; 1Zpresso J-Max and K-Max carry at 30 vs 90 positions/rev and do discriminate.
- The web datalist keeps the precision of the user's own value.

Also centralizes the terminal schema version, a bare literal in 24 assertions across two test files.

Full suite: **117 passed, 0 failed**, no warnings. All three `text-invariants` gates pass.

## Review findings, fixed in-branch

A review of this PR found three comments asserting things that were not true, all now corrected: the `findEntry` memo justified by a call pattern this same PR removes; its `thread_local` rationale naming the ShotServer, which runs on the main thread; and a `GrindPickerDialog` comment claiming a `historyDataChanged` subscription that appears nowhere in `qml/` except in the sentence claiming it.

## Unrelated observation, not acted on

The profiler's largest single item was not the picker: `ScaleWeightItem.qml:58` (`showScaleWarning`), 450 ms total / 270 ms self over 7 calls — ~64 ms per evaluation for a five-term boolean. Debug + sanitizer build, not a hot path, and unexplained. Worth re-measuring in a release build before anyone digs.

### DIFF
--- a/qml/components/GrindPickerDialog.qml
+++ b/qml/components/GrindPickerDialog.qml
@@ -107,9 +107,15 @@ DecenzaDialog {
     //
     // The original trigger was the distinct-value cache warming a moment after
     // the first open. That cache is gone and the reads are live, so that exact
-    // sequence cannot recur — but the snapshot is still right: rowSource.grindStep
-    // now moves on historyDataChanged(), and a shot saved while the picker is
-    // open would otherwise rebuild the model under the user's finger.
+    // sequence cannot recur, and the snapshot is what keeps it that way: nothing
+    // here subscribes to anything, so nothing can move the model mid-gesture.
+    //
+    // An earlier version of this said "rowSource.grindStep now moves on
+    // historyDataChanged()". It does not, and never did — that signal appears
+    // nowhere in qml/ except in the sentence claiming it. grindStep() is a plain
+    // function called from _rebuildRows(), so the rows change only when a handler
+    // asks. Kept as a correction because the wrong version reads as a live
+    // subscription a future edit would try to preserve.
     property var _grindRows: []
     property var _rpmRows: []
 
--- a/qml/components/GrindPickerDialog.qml
+++ b/qml/components/GrindPickerDialog.qml
@@ -114,10 +114,29 @@ DecenzaDialog {
     property var _rpmRows: []
 
     function _rebuildRows() {
-        root._grindRows = root.rowSource
+        // TEMPORARY instrumentation (grind-picker open cost). The reported
+        // symptom is "several seconds" and only ~0.9 s of it is accounted for by
+        // the two history queries, so the split between query, row generation
+        // and view positioning has to be measured rather than guessed. Remove
+        // these four console.info lines once the dominant term is known.
+        var _t0 = Date.now()
+        // Split the ASSIGNMENT out of the call: handing a fresh 801-element
+        // array to the Tumbler rebuilds its view, and that turned out to be the
+        // dominant term once row generation moved to one batch call.
+        var _gRows = root.rowSource
             ? root.rowSource.grindRowsFor(root._pendingGrind) : []
-        root._rpmRows = (root.rowSource && root.rowSource.rpmCapable)
+        var _tBuilt = Date.now()
+        root._grindRows = _gRows
+        var _tGrind = Date.now()
+        var _rRows = (root.rowSource && root.rowSource.rpmCapable)
             ? root.rowSource.rpmRowsFor(parseInt(root._pendingRpm) || 0) : []
+        var _tRpmBuilt = Date.now()
+        root._rpmRows = _rRows
+        console.info("[Equipment] grind picker: _rebuildRows grind-build="
+                     + (_tBuilt - _t0) + "ms grind-assign=" + (_tGrind - _tBuilt)
+                     + "ms rpm-build=" + (_tRpmBuilt - _tGrind)
+                     + "ms rpm-assign=" + (Date.now() - _tRpmBuilt)
+                     + "ms total=" + (Date.now() - _t0) + "ms")
     }
 
     // Index of the current value within a rows array (-1 if none is current).
@@ -195,6 +214,7 @@ DecenzaDialog {
     // taking the middle: the window is wide and clamps at zero, so the middle
     // row is no longer the anchor.
     function _centerWheels() {
+        var _t0 = Date.now()  // TEMPORARY — see _rebuildRows
         var gi = root._currentIndex(root._grindRows)
         root._snapTo(grindTumbler, gi >= 0 ? gi : Math.floor(root._grindRows.length / 2))
         var ri = root._currentIndex(root._rpmRows)
@@ -208,6 +228,9 @@ DecenzaDialog {
         // Remember where we parked it, so a later currentIndex that differs is
         // proof the user moved it themselves.
         root._rpmSnapIndex = rpmTumbler.count > 0 ? rpmIndex : -1
+        console.info("[Equipment] grind picker: _centerWheels " + (Date.now() - _t0)
+                     + "ms (grindRows=" + root._grindRows.length
+                     + " rpmRows=" + root._rpmRows.length + ")")
     }
 
     // All setup happens BEFORE the dialog is visible, so its first frame
@@ -215,6 +238,7 @@ DecenzaDialog {
     // place flicker, no travel. onOpened repeats the (idempotent) snap once
     // in case the ListView finished layout only after showing.
     onAboutToShow: {
+        var _t0 = Date.now()  // TEMPORARY — see _rebuildRows
         root._rpmTouched = false
         root._rpmSnapIndex = -1
         root._pendingGrind = root.currentGrind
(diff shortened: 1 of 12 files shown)


skialpine:
### Code review

Found 3 issues:

1. The `findEntry()` memo is justified by a call pattern this same PR eliminates, and a sibling comment in the same diff records that this exact memo measured no win. The comment says every one of 801 rows "reaches here through `SettingsDye::stepGrinderSetting`" — but after the batching change in this PR, `grindRowsFor()` calls `stepGrinderSettingRange()` once, which calls `findEntry()` once (`settings_dye.cpp`, outside its loop). Meanwhile `settings_dye.cpp` states "a memo on the registry lookup did not touch it... The only thing that helps is not making the calls." The memo does still have a narrower justification — `grinderIsClickIndexed()` reaches `findEntry` per row on the custom-grinder JS fallback — but that is a different function than the comment names. (CLAUDE.md says "Re-derive the justification when the design around it moves... they are claims with expiry dates.")

https://github.com/Kulitorum/Decenza/blob/7cc7ec989d7de51e70e4d0f87836ee3c02e8a9c3/src/core/grinderaliases.h#L356-L363

2. `GrindCandidates::stepDecimals()` takes only the step, while its comment claims it mirrors `GrindRowSource._stepDecimals` — which takes the max with the *value's* decimals, and that max is the #1713 fix ("Step alone was #1713"). Pre-existing and outside this diff, but it is the declared C++ twin of the file this PR rewrites. Scenario: a grinder with thin history returns step 0, the endpoint falls back to `step = 1.0`, `decimals` becomes 0, and a recorded grind of `1.1` is formatted as `1` — the user's own value is absent from the datalist and the n=0 suggestion silently rewrites it. The in-app wheel offers `0.1, 1.1, 2.1` for the same record. (CLAUDE.md: "When you change an in-app page that also exists in the ShotServer (or vice-versa), update the counterpart in the same change so they don't drift.")

https://github.com/Kulitorum/Decenza/blob/7cc7ec989d7de51e70e4d0f87836ee3c02e8a9c3/src/network/grindcandidates.h#L56-L69

3. This comment asserts a signal wiring that does not exist. `historyDataChanged` appears exactly once in all of `qml/` — inside this comment. Nothing in `GrindRowSource.qml`, `GrindPickerDialog.qml`, `GrindField.qml` or `GrindQuickSelectItem.qml` subscribes to it, and `GrindRowSource.qml` says the opposite two files over ("there is nothing to listen for"). Pre-existing, in a file this PR modifies. It will mislead the next reader into thinking a live subscription exists that must be preserved. (CLAUDE.md: "Fix pre-existing violations in any file you touch.")

https://github.com/Kulitorum/Decenza/blob/7cc7ec989d7de51e70e4d0f87836ee3c02e8a9c3/qml/components/GrindPickerDialog.qml#L107-L113

---

<details>

- [Worth a look] Temporary perf instrumentation logs at `console.info` (INFO = the user tier) under the `[Equipment]` marker, whose registered topic is grinder identity, not picker timings — `console.log` would put it on the developer tier. Deliberate per the PR description (it must reach the reporter's submitted log), but the tier rule is audience-based (score: 70)
- [Worth a look] The `thread_local` comment names ShotServer as an off-thread caller; `handleGrindCandidatesApi` appears to run on the main thread (no `moveToThread`, no `QThread::create` wrapper, unlike sibling handlers). `dialing_blocks` does run off-thread, so `thread_local` is still correct — only the stated reason is half wrong (score: 70)
- [Worth a look] The drafting-history parenthetical retained in `shothistorystorage_queries.cpp` records the comment's own past miscount. CLAUDE.md: "a record of drafts and corrections" does not survive the cut. The count above it is now correct, so it can go (score: 60)
- [OK to skip] `logGrinderCensus()` posts three full scans of `shots` to the shared FIFO `SerialDbWorker` at startup, so an early Shot History load queues behind them. Pre-existing, unchanged by this PR, and the fix (detached thread) is a separate concern (score: 60)
- [OK to skip] The census `COUNT(DISTINCT s.id)` dedupes only within a group, so a package carrying two grinder items with different models inflates the logged shot total. Pre-existing and rare (score: 55)
- [OK to skip] `settings_dye.cpp`'s "186 ms over 8020 calls... 63%" gives a number without the device and database size its sibling comments carefully state (score: 55)

</details>

PR #1899, opened 2026-09-01
fix(ble): recover a DE1 link that has gone silent while reporting connected

skialpine:
## The problem

A DE1 link can stop working while `QLowEnergyController` still reports it connected. Measured twice in one user log (`debug-2.log`, builds 3571 and 3572):

| | write abandoned | platform reported `Unconnected` | dead time |
|---|---|---|---|
| Episode 1 | 189.14 | 211.65 | 22.51 s |
| Episode 2 | 1929.62 | 1952.13 | 22.50 s |

In episode 2 there is no DE1 traffic of any kind — inbound or outbound — across that window. Every command sent to the machine in it is discarded, so a stop press or a stop-at-weight stop does nothing, and the log told the user to reconnect the DE1 by hand.

`m_notificationLiveness` already tracked this — every inbound push restarts it — but it was read in exactly one place: `connectToDevice()`, to turn an incoming reconnect into a teardown. **Nothing calls `connectToDevice()` while the link claims to be connected**, so the check never ran in the case it exists for.

## The fix

`evaluateLinkLiveness()` runs the same check on **write outcomes** — events the link produces whether or not it is healthy — and tears the link down when it confirms. The teardown emits `disconnected()` and stops there; the existing reconnect ladder in `main.cpp` owns the reconnect, its backoff and its attempt budget. Recovery lands around 10 s instead of 22.5 s.

Two thresholds, both **provisional** pending on-device measurement of the DE1's raw push cadence (the app's own inbound logging is de-duplicated and cannot supply it):

| | Value | Why |
|---|---|---|
| `NOTIFICATION_STALE_MS` | 30 s → **20 s** | Silence alone. Must stay below the ~22.5 s the platform takes, or it buys nothing over waiting for Qt. |
| `NOTIFICATION_STALE_CORROBORATED_MS` | **5 s** | Silence plus a write that just exhausted its retries. |

The corroborated bound is the one that fires in practice: the only guaranteed periodic write is the 60 s MMR keepalive, so after an abandoned write there is no further evaluation for a minute. An abandoned write is direct evidence the link is broken; silence alone is only evidence against a cadence nobody has measured.

**One guard:** the link is not torn down while the machine is running an operation *and* its writes still succeed — there the app can still stop the machine, and a reconnect would remove that for several seconds. When writes are failing too, nothing is preserved by waiting.

The teardown is posted **queued, never inline**: it runs inside a GATT failure callback, and tearing down the controller that callback is unwinding from is the same re-entrancy the zombie path in `connectToDevice()` already avoids.

Also: stop spending profile-upload retries on a link known not to be carrying writes. In the log the ladder burned two of five attempts that way, each ending in a 10 s ACK timeout, and five consecutive failures reach a dialog asking the user to **power-cycle the DE1** — the wrong remedy for a link fault the machine had no part in.

## User-visible

- A dead DE1 link recovers in ~10 s instead of ≥22.5 s of a machine that silently ignores every command.
- No "power-cycle the DE1" prompt for what is a link fault.
- No "reconnect from the Connections page" instruction for something the app now does itself.

## Scope

**Scale and refractometer paths are untouched.** The scale has its own weight-sample stall detection; the refractometers deliberately have none and must not gain it — an R2 is active only in short bursts on a page already looking for it, so "no inbound data" is its normal idle state and a silence detector would fire on correct behaviour.

## Testing

`shouldRecoverSilentLink()` is split out as a pure function because `isConnected()` needs a real `QLowEnergyController` and cannot be faked in a test — the same reason `BlePriorityDetector` is pure and clock-free. `tst_bletransporterror.cpp` gains a 6-row data-driven case over the thresholds and the busy guard, plus one asserting the silence threshold stays below the platform's own notice (if that regresses, the fix silently stops working while everything still compiles).

Full suite built and run locally — green. Both build-free gates pass (`check_log_markers.py`, `check_test_source_duplication.py`).

## Commits

1. `fix(ble)` — the recovery, plus the `BLE_PROTOCOL.md` section describing it.
2. `docs(ble)` — unrelated but folded in at the maintainer's request: recording that a BALANCED connection-priority latch is **permanent by design** and that the build code in its startup line is provenance, not staleness. The intent was nowhere in the tree, so the line reads like a warning about a stale setting every time a log is reviewed.

## Note

OpenSpec task 3.1 (extend the mock transport) is struck out rather than done: the teardown path is unreachable from a unit test whatever the mock does, so the pure decision seam replaced it.

Planning artifacts: `openspec/changes/recover-de1-write-dead-link/`.


skialpine:
### Code review

Found 7 issues:

1. The mid-operation busy guard can latch ON permanently, in exactly the failure it was written for. `m_machineBusy` is fed only from `MachineState::phaseChanged`, and the phase is derived from DE1 `STATE_INFO` notifications — the same inbound stream whose absence this feature detects. When the link goes quiet mid-operation while still ACKing writes, the phase freezes at `Pouring`, `m_machineBusy` stays true forever, and every later evaluation arrives from `noteWriteSucceeded()` (`writeJustAbandoned == false`), so `shouldRecoverSilentLink()` returns false on every pass. Nothing else can rescue it: the `connectToDevice()` zombie check is never reached while connected, and no write is abandoned because the link ACKs. The app parks on a frozen screen, logging "not tearing it down mid-operation" once a minute. This is the same deadlock the design doc uses to reject deferral ("the signal that would release it travels on the link that died"); the claim that the guard "cannot deadlock, because the writes-failing arm always terminates" holds only when writes fail.

https://github.com/Kulitorum/Decenza/blob/3261ddea2d568d54caf648b88335bac9a67c55fc/src/main.cpp#L2710-L2724

2. The upload retry hold strands the pending upload. It sets `m_profileUploadPending`, calls `updateProfileUploadRetrying()` and returns without re-arming `m_profileUploadRetryTimer`. The comment's premise ("the reconnect path will re-upload") only holds if a disconnect follows. If the link starts ACKing again without dropping — `noteWriteSucceeded()` clears `m_writeDeadLinkReported` — there is no disconnect, so no `initialSettingsComplete` → `applyAllSettings` → `uploadCurrentProfile`. The timer is dead and `m_profileUploadPending` only fires on a *transition* into Idle/Ready/Sleep/Heating, which a machine already sitting at Idle never emits. The profile is silently never uploaded and the next shot runs whatever the DE1 still had loaded. This is a regression: before the hold, the retry would have fired and succeeded once writes recovered.

https://github.com/Kulitorum/Decenza/blob/3261ddea2d568d54caf648b88335bac9a67c55fc/src/controllers/profilemanager.cpp#L182-L190

3. The new `de1LinkFault("link-silent")` reaches the permanent dual-HIGH scale latch, so "Scale and refractometer paths are untouched by this" is true of the files and false of the effect. The wiring is unconditional, and `QtScaleBleTransport::onDe1LinkFault` feeds a detector that latches skip-HIGH app-run-wide and writes through to settings — epoch-scoped, so it persists across restarts until `devices_reset_scale_priority`. That detector's own comment marks it "KNOWN OVER-EAGER (not fixed here, deliberately)", citing #1691 where a single MMR keepalive exhaustion latched it on a desktop. This PR makes a new fault class reachable that can fire from nothing but a successful keepalive plus 20 s of silence, against a cadence nobody has measured.

https://github.com/Kulitorum/Decenza/blob/3261ddea2d568d54caf648b88335bac9a67c55fc/src/main.cpp#L3033-L3037

4. `NOTIFICATION_STALE_MS` is tightened from 30 s to 20 s, and the silence-alone path is made to act, while the prerequisite measurement remains undone. Tasks 5.2 (`[~]`) and 8.5 (`[ ]`) in `openspec/changes/archive/2026-07-23-harden-de1-ble-reliability/tasks.md` recorded 30000 as "chosen conservatively" pending on-device measurement of raw push cadence; 8.5 has been unchecked since July. If the DE1's at-rest cadence at Idle or Sleep exceeds 20 s, every 60 s keepalive success tears down a healthy link into a permanent reconnect loop. The new tests assert only the thresholds' relationship to each other, so nothing fails if the assumption is wrong, and this change re-carries the "provisional" wording forward without adding a task to finally measure it.

https://github.com/Kulitorum/Decenza/blob/3261ddea2d568d54caf648b88335bac9a67c55fc/src/ble/bletransport.h#L403-L416

5. `m_livenessRecoveryPending` produces a false recovery line, two ways. It is set before the queued lambda runs, so if that lambda takes its `!isConnected()` early return no teardown happened yet the flag stays set — and unlike `m_livenessTeardownStarted` it is cleared in neither `disconnect()` nor `onControllerDisconnected()`, so it survives indefinitely and any later unrelated reconnect logs "DE1 is delivering again after the link was reconnected for going silent". Separately, it is consumed in `submitReadyMarker()`, which runs at subscribe-complete time, before any notification has arrived — so the comment's claim that "only the link actually delivering again clears it" does not hold.

https://github.com/Kulitorum/Decenza/blob/3261ddea2d568d54caf648b88335bac9a67c55fc/src/ble/bletransport.cpp#L1027-L1032

6. The comment documenting the signal this PR now acts on still states the opposite contract, unchanged (CLAUDE.md says "Re-derive the justification when the design around it moves... a comment nobody re-derives is how a bad decision keeps getting re-approved"). It reads "Reporting only. Nothing is torn down on this signal", and, above it, that "neither `m_notificationLiveness` below nor the wedge detector ... can see it" — correlating exactly those two is what this change adds. Both sentences are now false.

https://github.com/Kulitorum/Decenza/blob/3261ddea2d568d54caf648b88335bac9a67c55fc/src/ble/bletransport.h#L313-L327

7. `shouldRecoverSilentLink()` re-implements `SerialStall::Detector`, which already exists and is already tested (CLAUDE.md says "Centralize anything produced at more than one site... When you find yourself about to write the second copy, stop and extract"). That detector was split out of `SerialTransport` for the same stated reason given here for the new helper — so it could be tested at all — and its surface is what this PR hand-rolls: `arm()` maps to the liveness baseline, `disarm()` to the invalidate, `noteInbound()` returning true once on the edge that ends a stall to `m_livenessRecoveryPending`, `shouldWarn()` latching one warning per episode to `m_livenessTeardownStarted`, and `staleForMs()` to `elapsed()`. The two thresholds and the busy guard are extensions to it, not reasons for a second copy.

https://github.com/Kulitorum/Decenza/blob/3261ddea2d568d54caf648b88335bac9a67c55fc/src/usb/serialstall.h#L27-L33

---

<details>

- [Worth a look] `src/usb/serialtransport.h` says its threshold is "deliberately double BleTransport's 30 s" and that BleTransport's "is evaluated solely at a reconnect attempt" — both false after this PR (60 s is now triple, and it is evaluated on write outcomes) (score: 70)
- [Worth a look] `m_machineBusy` resets to false on every reconnect, because `DE1Device::connectToDevice()` constructs a new `BleTransport` per attempt; it is only refreshed at the next `phaseChanged`, leaving the guard inert in the window a mid-shot recovery lands in (score: 65)
- [Worth a look] `setMachineBusy`'s doc comment says "espresso, steam, hot water or flush cycle", but the wiring also sets it during Descaling, Cleaning, Refill and Transport; and "mirrors `setShotActive`" overstates it, since that one brackets only an espresso shot (score: 60)
- [Worth a look] The `connectToDevice()` zombie branch is now documented by this PR as unreachable, but is left in place with its original comment still describing it as live protection (score: 60)
- [OK to skip] `m_livenessTeardownStarted`'s comment describes only one of its three reset sites (score: 55)
- [OK to skip] The teardown path itself has no automated coverage; only the pure decision is asserted (score: 50, already acknowledged in tasks.md 3.1)

</details>

skialpine:
### Code review

Re-review after the force-push. Of the previous review's 7 issues, 4 are genuinely fixed (busy-guard deadlock, upload stranding, the `de1LinkFault` reaching the permanent scale latch, and the false "Reporting only" comment), and the threshold concern was addressed by dropping the silence-alone path rather than tightening it. `SerialStall::Detector` duplication remains open by decision.

Found 6 new issues, most introduced by the fixes themselves:

1. The "DE1 is delivering again" recovery line can never fire, and would be wrong if it did. `m_livenessRecoveryPending` is set on the transport that is tearing itself down, but `DE1Device::connectToDevice()` constructs a brand-new `BleTransport` for every reconnect attempt and `deleteLater()`s the old one, so the reader on the reconnected instance always sees `false`. Independently, its consumer sits in `onCharacteristicChanged()`, which `onCharacteristicRead()` calls directly — so a read *response* would satisfy it, which is the false positive its own comment says was designed out. Dead code, two ways.

https://github.com/Kulitorum/Decenza/blob/7727427c8feae9e098aa0d73dce7fe07010fbb6d/src/ble/bletransport.cpp#L1036-L1042

2. `m_uploadHoldReported` is a latch with one reset site, so the hold is logged at most once per app run. It is cleared only when the hold's own timer fires again and finds writes accepting. Every other way a hold ends stops the timer without touching it: the success handler, the disconnect handler, retry exhaustion, `acknowledgeDe1CommunicationFailure()`, and both profile-load resets. After any of those, later hold episodes log nothing — the opposite of the comment's stated intent. Note the same commit gets this right for the structurally identical `m_writeDeadLinkReported`, which is reset in three places for exactly this reason.

https://github.com/Kulitorum/Decenza/blob/7727427c8feae9e098aa0d73dce7fe07010fbb6d/src/controllers/profilemanager.h#L750-L756

3. The upload hold breaks the documented bound on `kUploadRetryMaxMs` and pins the "Reconnecting..." toast indefinitely. That constant's own comment says the 8 s cap "keeps the total wall-clock before the dialog shows down to ~15s". The hold re-arms the same timer at the same constant without incrementing `m_profileUploadRetryAttempts`, so it can run indefinitely and never reach `de1CommunicationFailure`. Meanwhile `updateProfileUploadRetrying()` stays true for the hold's whole duration, and `qml/main.qml` binds the toast to it with no auto-dismiss.

https://github.com/Kulitorum/Decenza/blob/7727427c8feae9e098aa0d73dce7fe07010fbb6d/src/controllers/profilemanager.cpp#L66-L75

4. The reworded dead-link WARN is false for the one case this PR deliberately does not fix. The teardown requires notification silence as well, so a link that discards writes while still pushing is never torn down — `bletransport.h` says so explicitly. For that case "The app reconnects the DE1 itself... no action is needed" is wrong, and the removed sentence naming the Connections page and `devices_connect_de1` was the user's only recourse. Combined with issue 3, that user now gets neither the escalating dialog nor accurate guidance — strictly worse than before this PR.

https://github.com/Kulitorum/Decenza/blob/7727427c8feae9e098aa0d73dce7fe07010fbb6d/src/ble/bletransport.cpp#L977-L985

5. `BLE_PROTOCOL.md` names a function that no longer exists and misstates what restarts the liveness clock. It cites `shouldRecoverSilentLink()`; the shipped name is `shouldRecoverAfterAbandonedWrite()`. It also says the clock is "restarted by every inbound push" — read responses restart it too, via `onCharacteristicRead()` delegating to `onCharacteristicChanged()`. That matters here because this PR is the first thing to act on that clock.

https://github.com/Kulitorum/Decenza/blob/7727427c8feae9e098aa0d73dce7fe07010fbb6d/docs/CLAUDE_MD/BLE_PROTOCOL.md#L125-L127

6. The upload hold is a timer standing in for an event that already exists (CLAUDE.md says "'suppress X until Y has happened' should be a boolean cleared by the Y event, not a timer"). Y here is "the link accepts writes again", and `m_writeDeadLinkReported` flips in exactly two places — `noteWriteSucceeded()` and `forgetWriteFailureState()` — neither of which emits anything. A signal from the transport would resume the upload immediately instead of up to 8 s late.

https://github.com/Kulitorum/Decenza/blob/7727427c8feae9e098aa0d73dce7fe07010fbb6d/src/controllers/profilemanager.cpp#L193-L213

---

<details>

- [Worth a look] `profilemanager.cpp` comment claims "Same treatment as the 'BLE disconnect' reason above" and then, two paragraphs down, explains that exact approach was tried and found insufficient — self-contradicting within one block (score: 65)
- [Worth a look] `bletransport.h` dropped a still-accurate sentence bounding the zombie check's false-positive cost to "one extra reconnect, not a spurious disconnect of a healthy in-use link"; the invariant no longer lives next to the code it protects (score: 60)
- [OK to skip] `tst_bletransporterror.cpp` docstring says the test "says why that was reverted" if a silence-alone caller returns, but its assertions cannot observe a call site and would still pass (score: 55)

</details>

Verified as accurate, since they were checked specifically: the ~5850-evaluations/97.6 h arithmetic, the 22.5 s figures, the busy-guard-unreleasable reasoning, the no-`de1LinkFault` justification, and the 7.5 s episode figure (which is derivable: `WRITE_TIMEOUT_MS + MAX_WRITE_RETRIES x WRITE_RETRY_DELAY_MS` = 7500 ms, matching the log's 7540/7550 ms).

PR #1900, opened 2026-09-01
Stop pre-tare weight samples reaching every downstream consumer

skialpine:
## The bug

A tare the app issues can still be travelling to the scale when the DE1 starts flow — ~50 ms apart in the field logs behind this. The samples arriving in that window carry the **old** zero, and every consumer downstream judged them as post-tare readings. One cause, three WARN families, on every shot poured into a cup the previous pour had left heavy:

```
[SAW][Worker] Sanity check: weight 70.80 g at 0.0 s into extraction — skipping stop-at-weight (likely untared cup)
[SAW][Worker] Spike rejected: weight=0.00 g last=141.10 g          (x3, then the escape hatch)
[SAW][HotWater] Tare not completed — skipping stop-at-weight check, weight=-138.40 g
```

The verdicts were wrong, not merely noisy: the cup had been tared 30 ms earlier, the rejected zero was the app's own tare, and the hot-water tare had not been sent yet.

**No shot mis-stopped.** All three affected shots in the reviewed logs landed on target (42.2/42.3/42.3 g against a 43.0 g target). This is log correctness, not a stop-at-weight defect. Worth knowing when weighing the risk of the change against the risk of leaving it.

## The fix

`markExtractionStart()` no longer clears `m_awaitingTare` outright. It grants a bounded run of arrivals instead, composed as `4 + kTareLandedConfirmations` so raising either input cannot leave it short: 3 stale arrivals seen in the field, one for margin, plus the two the zero itself needs to be believed. The near-zero confirmation ends the wait as soon as the real zero shows up, usually within one or two.

- **Sanity check** gated on that wait — a pre-tare reading is skipped without accusing the user.
- **`UNTARED_CUP_CONFIRM_SAMPLES` 4 → 2.** Its old justification was the #1837 stale-sample race, which the wait now handles where the streak would start. What remains is a single corrupt packet, and two samples answers that — same reasoning as `kTareLandedConfirmations`.
- **The window is anchored to the last granted arrival, not to flow start.** The grace is bounded in *arrivals* and the window in *seconds*, so on a 2 Hz scale six arrivals are 3.0 s and would consume the window whole — a genuinely untared cup would never warn and never raise the popup, on exactly the hardware where the pour is longest.
- **One exit from the wait.** Three `m_awaitingTare = false` sites become `clearAwaitingTare()`, which reports how many pre-tare samples were skipped and emits `tareLanded()`.
- **ShotDataModel** re-anchors on `tareLanded()` rather than `MachineState::tareCompleted`, which fires when the *command* goes out. That is how a stale 141.1 g reading became its filter's anchor and got 0.1 g rejected against it at 534 g/s.
- **Hot water** gets `m_hotWaterTarePending`; its warning throttle moves from a function-local `static` to a member, since a process-wide throttle lets one caller silence another's line.

## Verification

117/117 local, no warnings. Each new test was **red-checked** by disabling the fix it covers rather than trusting a passing run:

- grace set to 0 → 9 tests red, including both new ones
- window anchored to flow start again → `untaredCupStillDetectedOnASlowScale` red with `cupSpy.count() == 0`

## Trade-offs, stated rather than buried

- The per-frame weight exit is deferred up to 3 s on a 2 Hz scale. The full grace only runs when the tare never lands, which is precisely when firing that exit on a pre-tare reading would be wrong.
- If a tare ever lands *after* the grace, the false-positive margin on the popup is now 2 samples where it was 4.

## Review note

A five-agent review ran against an earlier revision of this diff and found the slow-scale regression above; its fix (`m_tareWaitEndedMs`, the `qMax` anchor) is the only production logic here that review did not see.

### DIFF
--- a/src/machine/machinestate.cpp
+++ b/src/machine/machinestate.cpp
@@ -538,6 +538,8 @@ void MachineState::updatePhase() {
 
                 m_tareCompleted = false;
                 m_waitingForTare = false;
+                m_hotWaterTarePending = false;
+                m_lastHotWaterTareWarnMs = 0;
                 if (m_tareTimeoutTimer)
                     m_tareTimeoutTimer->stop();
 
@@ -554,7 +556,12 @@ void MachineState::updatePhase() {
                 // Delay 200ms after resetTimer/startTimer to avoid BLE command contention —
                 // WriteWithoutResponse can silently drop packets when sent in rapid succession.
                 if (m_phase == Phase::HotWater) {
+                    m_hotWaterTarePending = true;
                     QTimer::singleShot(200, this, [this]() {
+                        // Cleared on BOTH paths: the early return leaves no tare coming,
+                        // and a pending flag that outlives its tare would silence the
+                        // warning for the rest of the pour.
+                        m_hotWaterTarePending = false;
                         if (m_phase != Phase::HotWater) return;  // Operation ended before timer fired
                         tareScale();
                         qDebug() << "=== TARE: Hot Water started (200ms after timer cmds) ===";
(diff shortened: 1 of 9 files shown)


skialpine:
### Anchor review

Targeted review of the `m_tareWaitEndedMs` window anchor — the one piece of production logic the earlier five-agent pass did not see.

Sound: the stamp/reset invariant (`m_awaitingTare = true` is written at two sites, both of which zero the anchor; `= false` at one, which disarms the grace), the `qMax` composition, and the streak's may-only-start-inside-the-window rule.

One real gap, now fixed in 6a0e30a4: the anchor was unbounded relative to flow start. Arrivals carry no wall-clock bound, so a stalled feed could put the 3 s window anywhere in the pour, and late is the dangerous direction — past ~10 s a >50 g reading is real coffee on a large target, and the verdict latches. `m_untaredCupSignalled` short-circuits the streak increment, the frozen streak holds `withinWindow` true, and every later heavy sample returns ahead of the SAW stop and the per-frame exit. One false positive would disable stop-at-weight for the remainder of the shot.

Clamped to flow start + 4 s, derived from the grace's own worst case (six arrivals at 2 Hz, ~3.0 s, plus margin), so the window cannot extend past ~7 s of flow. The residual — a feed stalling longer than the window between the last granted arrival and the next — is stated at the call site and left to the scale-stall detector.

Three tests added, each red-checked against its own mutation rather than against the fix as a whole. Worth noting that the clamp test passed without the clamp on its first draft: its sample spacing put the readings outside the window either way. It now carries the arithmetic in its comment.

117/117 local, no warnings.

PR #1901, opened 2026-09-01
Match each ShotSettings read-back to the write it answers

skialpine:
## The bug

Every `setShotSettings()` write queues a read behind itself on the same serial GATT queue, so N rapid writes produce N reports in write order, each carrying the state after **its own** write. The drift detector compared every report against the *newest* commanded values instead, so any burst of writes looked like the DE1 had dropped one.

Profile activation issues three writes ~10 ms apart (`uploadCurrentProfile`, `applySteamSettings`, `applyHotWaterSettings`), which is why this fired constantly.

## Evidence

From a user debug log — 12 `DE1-dropped-write` WARNs across six bursts. The clearest:

```
[3991.369] write: steam=160 dur=43 hw=76 group=88   (#1)
[3991.377] write: steam=0   dur=43 hw=76 group=88   (#2)
[3991.381] write: steam=0   dur=43 hw=72 group=88   (#3)
[3992.179] reported: steam=160 dur=43 hw=76 ...     <- answers #1
[3992.184] WARN DE1-dropped-write: steam heater ON at 160C but we commanded OFF
[3992.405] reported: steam=0 dur=43 hw=76 ...       <- answers #2
[3992.630] reported: steam=0 dur=43 hw=72 ...       <- answers #3, "resolved"
```

In all six bursts the report count equals the non-elided write count, and each reported payload equals the Nth write's payload, in order. No report ever carried a value we had not written. Every episode "resolved after 1 resend" on its own once the last report landed — the resend was not what fixed it.

Cost of the false positive: a hardware accusation at WARN (which field AIs read over MCP and act on), plus a redundant BLE write per episode into a GATT queue already warning about depth 40 during connect.

## The fix

`DE1Device` keeps a FIFO of the values each queued read-back is expected to bring, pops one per report, and exposes it as `expected*()`. An empty FIFO means nothing is in flight, so the last commanded values are the right expectation (the connect-time read). Elided writes push nothing. Cleared on disconnect. `MainController`'s drift check reads `expected*()` instead of `commanded*()`.

## Second, independent defect fixed here

`resendLastShotSettings()` wrote blind — it queued no read of its own. The drift ladder could therefore only advance on a read some earlier write happened to have left in flight; a genuine dropped write with nothing else outstanding would resend and then produce **no terminal line at all**. The log could not show this because a pending read always happened to be there. It now queues its own read-back, pinned by `resendQueuesItsOwnReadBack`.

## Testing

Six slots added to the existing `tests/tst_shotsettings.cpp` (no new TU): burst matching, elided write queues no expectation, unsolicited report falls back to commanded, pre-first-write report has no expectation, resend queues its own read, disconnect clears the queue.

Verified the tests can fail: breaking the pop reproduces the log's exact phantom —

```
FAIL!  : burstOfWritesMatchesEachReportToItsOwnWrite()
   Actual   (expectedSteamTargetC()): 0
   Expected (160.0)                 : 160
```

— which is write 1's report judged against write 3's `steam=0`, i.e. `[3992.184]` above. Restored and green.

Full suite via Qt Creator: **117 passed, 0 failed, 0 warnings**. `check_log_markers.py`, `check_test_source_duplication.py` and `check_mcp_tool_budget.py` all pass.

No manual or release-note change: this is log-only behaviour, nothing on screen changes.

### DIFF
--- a/src/ble/bletransport.cpp
+++ b/src/ble/bletransport.cpp
@@ -233,7 +233,7 @@ void BleTransport::subscribeAll() {
     // SHOT_SETTINGS is intentionally NOT subscribed: the DE1 firmware does
     // not push notifications on writes (confirmed in de1app's de1_comms.tcl).
     // Verification happens via explicit read() after each write in
-    // DE1Device::setShotSettings().
+    // DE1Device::setShotSettings() and resendLastShotSettings().
 
     submitReadyMarker();
 
--- a/src/ble/de1device.cpp
+++ b/src/ble/de1device.cpp
@@ -277,6 +277,8 @@ void DE1Device::onTransportDisconnected() {
     m_commandedGroupTargetC = -1.0;
     m_lastShotSettingsWriteMs = 0;
     m_lastShotSettingsPayload.clear();
+    m_pendingShotSettings.clear();
+    m_expectedShotSettings = {};
     // Clear the MMR dedup cache so a reconnect re-writes real values rather
     // than trusting cached values from the previous session (the DE1 may have
     // power-cycled or had its firmware state reset between sessions).
@@ -2581,6 +2583,9 @@ void DE1Device::setShotSettings(double steamTemp, int steamDuration,
     m_commandedGroupTargetC = groupTemp;
     m_lastShotSettingsWriteMs = QDateTime::currentMSecsSinceEpoch();
     m_lastShotSettingsPayload = data;
+    // Queue what the read below is expected to bring back.
+    m_pendingShotSettings.append({steamTemp, steamDuration, hotWaterTemp,
+                                  hotWaterVolume, groupTemp});
 
     // Trace every write so the timeline of commanded values is visible in the
     // debug log alongside the DE1-reported values from parseShotSettings().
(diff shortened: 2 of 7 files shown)


skialpine:
## Review pass — the FIFO is gone, replaced by echo matching

Five review agents converged on one defect in the first commit, and it was worse than the bug being fixed. Pushed `4671e45f`.

**The positional FIFO rested on an invariant only one transport provides.** `SerialTransport::read()` is a no-op once subscribed (`serialtransport.cpp:115`) and `subscribeAll()` subscribes SHOT_SETTINGS, so on USB no read is queued per write at all — the queue desynced on the first write. On BLE it was breakable too: `clearCommandQueue()` (live during every shot, via `stopOperationUrgent`, `machinestate.cpp:536`, `maincontroller.cpp:4102`) discards our queued read-backs, and a read can be abandoned after retry exhaustion.

A slipped positional queue stays slipped for the whole connection, scoring every later report against the previous write. That is the bug this PR exists to fix, reintroduced as a permanent state where the original drained itself once the burst ended.

**Now:** a report is an in-flight echo when it matches *any* write still awaiting confirmation (matching one drops the writes it superseded); a report matching none is scored against the newest write — which is exactly what `main` does today. The failure mode degrades to current behaviour instead of to a stuck offset. Bounded list, and `clearCommandQueue()` drops it next to the `m_lastMMRValues` invalidation that exists for the identical reason.

## A ladder stall this branch introduced

Giving the resend its own read-back (commit 1) meant it produced exactly one report — which `m_shotSettingsResendInFlight` then consumed as a rate limit *without counting it*. The ladder stalled at "attempt 1 of 3", never reached `kMaxResendAttempts`, and ended on a WARN with no terminal line: the failure-half-only narrative LOGGING.md exists to prevent. The flag is deleted and a still-drifting report advances the ladder, which stays bounded and still advances only on a report.

## Test that wasn't testing

`MockTransport::read()` was `{}`, so `resendQueuesItsOwnReadBack` asserted the bookkeeping, not the read — deleting the `read()` call left all six green. The mock records reads now.

## Verification

Broke the echo match, rebuilt, and the suite reproduced the log's exact phantom:

```
FAIL!  : burstOfWritesScoresEachReportAgainstItsOwnWrite()
   Actual   (expectedSteamTargetC()): 0
   Expected (160.0)                 : 160
```

Restored: **117 passed, 0 failed, 0 warnings**, all three text-invariant gates green.

One process note worth recording: three "successful" builds during this pass compiled a *different clone* of this repo after Qt Creator's active project drifted, and reported green while the objects sat 13 minutes older than the sources. The results above are from a build confirmed against object timestamps in the right tree.

PR #1913, opened 2026-09-02
fix(recipes): stop live steam/hot-water changes from rewriting the recipe

### LINKED ISSUE #1895: [Feature] ONLY Grind Setting adjustments should alter the recipe, not hot water

My daily coffee is latte/water mix (70ml), my wife's is americano. (with water first LOL)
often I start my wife's drink, and see the hot water was set to 70ml, so I tap the 140ml button on the fly and that's that... but unfortunately this change sticks, changes the recipe. The negative result is that often I end up delivering 140 water to my cup by mistake because I often make her cup before making mine... and if I am not watching, I get 2x as much water as I want.

### Proposed solution

IMO ONLY grind setting should permanently change a recipe when making an adjustment on the home screen. The logic is that grind size changes A LOT, particularly as one is dialing in... it is a necessary and frequent part of the process. Changing how much water you prefer in the cup should remain the same in a recipe, as should all other changes made temporarily to the recipe.

Lets say I start with my recipe, 70g water, and I do a shot for my wife first... so I tap the 140ml button to give her that amount. Then I do my shot, but the recipe is now delivering 140 instead of the desired 70ml... I would want it to revert to the programmed recipe settings immediately after doing a shot. Only Grind Settings should permanently change the Recipe. 

BTW I have also tried selecting another recipe, and then going back to the one i want to use in hope that the original values will be reinstated, but they do not, the change to hot water sticks, in the same was as grind setting does.

### Alternatives considered

have duplicates for all recipes, one for me one for my wife... which I may do, but it's quicker to just start the hot water immediately and adjust the output on the fly.

### Additional context

_No response_

skialpine:
Fixes #1895.

## The defect

Selecting a different water vessel on the Hot Water page while a recipe was active re-snapshotted that vessel into the recipe's `hot_water_json` — permanently. The reporter's case: an americano poured at 140 ml for someone else redefined the recipe, so every later americano on it poured 140 ml. Re-selecting the recipe restored nothing, because the stored row itself had changed. Steam pitcher selection, pitcher preset edits, and the steamed-milk weight behaved the same way.

## The change

**A steam pitcher and a water vessel are ingredients, not tweaks.** Changing one while a recipe is active now deactivates the recipe — exactly what the bean, the equipment package and the profile already do.

The recipe's stored block is untouched, so **one tap on its pill brings the whole drink back**. Deactivation is what makes that one tap work: a deselected pill activates on the first tap, where a selected pill would have started a shot.

- Five write-through connections removed: `selectedSteamPitcherChanged`, `steamPitcherPresetsChanged`, `lastSteamMilkGChanged`, `selectedWaterVesselChanged`, `waterVesselPresetsChanged`. `stampActiveRecipeSteam()` / `stampActiveRecipeHotWater()` go with them.
- Two deactivation watchers added beside the bean/equipment/profile ones, on the two **selection** signals only.
- `Recipe::ownsSteamPitcherChoice` / `steamPitcherDiverged` / `ownsWaterVesselChoice` / `waterVesselDiverged` join `profileDiverged` in `recipestorage.h`. Ownership gates first — a recipe naming no pitcher, or with no hot-water block, is never dropped by a change it does not own — and re-selecting the recipe's own pitcher or vessel is a no-op. "Heater off" counts as a choice in both directions.
- Dose, grind and RPM write-through untouched: dial-in measurements of what the user physically did, and the dose stamp is welded to the dose-source-precedence ladder.

### Two things deliberately do NOT deactivate

**A preset edit.** Add, remove, reorder or field-edit — not an ingredient swap, and the recipe's block is a by-value snapshot that deliberately does not follow the preset.

**The steamed-milk weight.** `lastSteamMilkG` is written automatically when a steam session ends ([main.qml:626](https://github.com/Kulitorum/Decenza/blob/main/qml/main.qml#L626)), not by the user. Deactivating on it would drop every milk recipe seconds before its own shot is saved — the drink that most needs its recipe attribution would be the one to lose it.

### Deactivation must not eat its own trigger

`deactivateRecipe()` unwinds the recipe pitcher override, and that unwind re-selects the standing pitcher parked at activation. Deactivating in response to the user's own pitcher pick would restore the parked pitcher over the pick. The user has now chosen a pitcher themselves, so there is no parked "user's own" selection left: the watcher clears it first and the unwind becomes the no-op it should be.

### A regression this removes

The shot-start pre-heat reads steam intent from the recipe row, and `heaterOff` only ever reached that row through the deleted pitcher stamp. An intermediate version of this PR therefore ignored a live "Heater off" pick and relit the boiler at shot start. A deactivated recipe has no intent, so the live veto stands on its own.

## Design history

The first cut made the blocks per-brew overrides on the yield/temperature model — apply the tweak, keep the recipe active, leave its row alone. Correct as far as it went, but silent: the machine kept making a drink the active recipe did not describe, the shot was still attributed to that recipe, and from the idle screen the recipe could not be re-applied at all. Deactivation replaced it.

## Tests

- `tst_RecipeStorage::steamPitcherOwnershipAndDivergence` and `waterVesselOwnershipAndDivergence` — behavioral tests of the four rules: ownership, case-insensitive name match, the Heater-off marker both ways, and the incomplete/dormant hot-water block. Both proven red by breaking the rule under them.
- `tst_RecipeStorage::onlyDialInValuesStampTheActiveRecipe` — keeps the stamps gone. An **allowlist** of the fields that do stamp, asserted to be exactly `{doseG, grindPinned, rpmPinned}`, plus an occurrence count so a call the regex cannot parse fails loudly. A denylist of signal names was the first draft and is defeated by a member-slot connect, a renamed helper, or a signal nobody listed.
- Test functions, not new files — a new `tst_*.cpp` costs ~1.4 s of build forever.

Full suite: **117 passed, 0 failed, 0 warnings.**

## Docs

`docs/CLAUDE_MD/RECIPES.md` and the wiki manual (`e4bf677`) both updated: the pitcher and vessel join the ingredient list that deactivates, and the write-through line shrinks to dose and grind.

## Not verified here

On-machine checks are manual and have not been run: that changing the vessel deselects the pill, that one tap brings it back, that finishing a steam session does *not* deactivate, and that a live "Heater off" pick now keeps the boiler cold through a shot.

OpenSpec change: `recipe-blocks-editor-only`. Archive + spec sync will ride as the final commit on this PR.


fredphoesh:
Hi Jeff, I had a look, and don't think this is a good solution. What I understand is that if I start a shot with a recipe, then when im adding water, i change from 70 to 140ml setting, and when I go back to the home screen, the shot is no longer a recipe. So when I do another shot, the water has not reverted to 70ml. I am now required to tap Recipes, tap the recipe I was using again and do the shot... so this feels worse than it was, more steps, and I would not want the history of the shot using that recipe to be removed due to adding a bit more water on the fly. I what would be ideal is for the recipe to remain a recipe, and next time you press the shot button, the recipe reverts to 70ml, ie back to the actual recipe rather than not using the recipe.
here is a longish video of my experience using it this morning...
https://videoblurb.com/mp/5-800recipe-change-next-shot.mp4

skialpine:
Revert back to a known recipe is what the pin a recipe function is for. When you change an element like water that is part of a recipe it stops being a recipe. That’s now consistent across beans, steam, hot water. The other option was to update the recipe which you did not like eitherand I agree was confusing. The pin a recipe allows you to set a recipe to jump back to when the machine wakes up or after a fixed amount of time. 

You want the machine to know you are done with the overrides setting which is confusing and hard to implement without an event to drive it which is why we have the pin. The only problem with the pin for you is that you have too many recipes you are using that you flip around between, but that is your other option to put the standard changes as separate recipes and then manually switch between them.

fredphoesh:
Hi, I have not tested this scenario, but if I started with my recipe, tapped 140ml for Madi, Then started my shot and changed back to 70ml, would it intelligently reverse the process, ie match the recipe ingredients such that it records as a shot done with a recipe again (as it is identical). That would be halfway there LOL. 
I'd prefer recipe to always be recipe and revert all but dose and grind... but appreciate it may/would be a "ball ache"!
off to shops, speak later...

skialpine:
it does not reverse the process, but clicking on the recipe button to go back is the same number of clicks as changing back to 70. I can work with you to make the pin behavior smarter also where instead of time it maybe could also revery back after a full drink is done, but I have no idea how to figure that out. the thing is sometimes you have a latte set, but then you don't do the milk because the shot was so tasty 8-)

fredphoesh:
Well, it's not the same really because the clicks you do to return to 70ml are done while you are already in the process of adding 140, so you are not stopping the process to go fiddle with settings.
I wonder if the "recipe reset" couldn't be easily done if it was activated by both the end of your shot and by the end of Hot Water adding... That way, if you add water before or after or not at all, you still get the recipe you'd chosen ready to go again.

I appreciate there are users who would not want the tweaks they'd made to a recipe, on-the-fly, to result in that shot to be accepted as being of that recipe... a change is a change,  no longer that recipe... maybe that black and white approach is the purest and least messy logic.

skialpine:
Have you looked at the recipe pin and given it a try? I am thinking enhancing it might be best direction to get what you want.

Issue #1893, opened 2026-09-01
[Bug] press button randomly comes up


fredphoesh:
### What happened?

The message to press the power button seems to come up often, and usually when the button is already on!

### Steps to reproduce

I don't know.

### Decenza version

2.05

### Platform

Android

### Device / OS version

12

### Debug log / shot debug log

[1 sept debug.zip](https://github.com/user-attachments/files/31685407/1.sept.debug.zip)

skialpine:
let me know if you are still seeing this after todays update. if so add a new debug log to this issue

fredphoesh:
Hey Jeff, see the video, it's still happening...

https://github.com/user-attachments/assets/66b0136f-6f60-4ff2-ac2a-8cd2439dbd7e

[2 sept debug.zip](https://github.com/user-attachments/files/31733986/2.sept.debug.zip)

skialpine:
thanks for the log. will take another look. is the grind widget now faster?

skialpine:
are you still seeing this? if so can you add an updated log to this issue?

fredphoesh:
hi. actually i made coffee on 3 occasions today and did not notice... which PROBABLY means it is fixed... I will try remember to be more mindful about it tomorrow...

I did notice my Mark Peru Adaptive recipe shot was not recorded as a shot made with that recipe simply because I added more water for Madi. Its a bit annoing like that because I use the double tap on the Recipe button to filter the history to show only shots made with that recipe, which purely boils down to wanting to check if the Grind Settings showing on my home screen is the grind setting I changed to after doing the shot, or whether I still need to make the adjustment. This flow is ruined by removing the recipe status from such shots... 
On the +ve side of that, re selecting the recipe changes the water back to 70g... but of course that's more convenient and takes no extra time if done on the fly. oh well.

fredphoesh:
Seems fixed! Thanks Jeff!

PR #1915, opened 2026-09-02
fix(machine): stop the standby-switch warning firing at a heating machine

skialpine:
Refs #1893 — deliberately not a closing keyword; leaving the issue open until the reporter confirms.

## The report

"The message to press the power button seems to come up often, and usually when the button is already on!" Firmware **v1363** — inside the range the 1337 gate treats as trustworthy. The reporter sees it on **every tap-to-wake**, and after an app restart. It stays up about three seconds and clears itself, untouched, with the machine working normally either side of it.

## Cause

The DE1 reports substate `Error_NoAC` briefly while it wakes or heats. **A snapshot cannot tell that apart from a genuinely open switch** — an open switch reports `Idle, Error_NoAC` and so does the transient. de1app states this outright: flipping the switch "only moves the SUBstate (e.g. \"Idle, ready\" => \"Idle, Error_NoAC\")".

So duration is the only discriminator. There is no event to use instead: `DE1Device` emits `subStateChanged` only when the value changes (`de1device.cpp:705`, emitted at `:796-797`), so a condition that merely persists generates nothing to listen for.

## The fix

`Error_NoAC` must persist for **6 seconds** before the warning shows. A disconnect or a substate change before that abandons the wait; an episode that clears itself never reaches the screen.

**The 6 s is our own estimate, not borrowed.** The one reported episode ran about three seconds and this is that plus margin. An earlier revision of this branch attributed it to de1app, citing `de1_de1.tcl:946` — that citation was wrong and has been removed everywhere. That `after 6000` is a one-shot post-connect re-check for a machine already latched in `Error_NoAC` (the gap our `firmwareVersionChanged` hook covers), not a settling wait, and de1app shows its warning immediately (`machine.tcl:1354`; its commit `04d3b02e` reports the page appearing "~1ms after the switch is flipped" on firmware 1352). Why Decent need no filter and we appear to is **unexplained**, not reconciled — recorded in the code rather than smoothed over.

An earlier design keyed on the substate an episode arrived from. It was event-based and wrong: on a tap-to-wake the machine is asleep at `Ready` and reports `Error_NoAC` before any heating substate, so it missed the commonest path entirely.

## Diagnostics

Every `[DE1][StandbySwitch]` line carries the episode's **measured duration** alongside the configured interval, at INFO so it reaches the user-facing views. The subsystem previously logged nothing at all, which is why 20,578 lines of the reporter's log could neither confirm nor deny when the warning fired. Since the interval is an estimate from a single report, a field log is the only thing that can correct it — and a "still happening" report can then be told apart from a different mechanism.

## Also in here

- **CLAUDE.md's no-timers rule** gains a narrow hardware carve-out, with the argument for why no event-based rule works here, and what separates it from the debounce the same bullet still forbids: debounce has edges to count; this has two identical values and no distinguishing edge.
- **`m_shotTimer` is stopped on disconnect.** Pre-existing, found one line from the code this change adds — the disconnect early return skips the only `stopShotTimer()` call and `onShotTimerTick()` has no connectivity guard, so a mid-shot BLE drop left the shot clock running and emitting every 100 ms.

## Verification

- Full suite via Qt Creator: **117 passed, 0 failed**, no test warnings.
- **Falsified, twice, on the things that matter.** Deleting the timer's `connect()` turns the suite red; deleting the `>= 1337` firmware gate turns it red. Both would have passed before this round — `elapseNoAcWait` reimplemented the timeout lambda instead of running it, so the mechanism had no coverage at all, and the firmware test had become unfailable because it never elapsed the wait.
- `QTimer::timeout` takes a `QPrivateSignal`, so the helper shortens the real timer and waits for it rather than emitting a synthetic signal. The 6000 ms default is pinned by its own test.
- `scripts/check_log_markers.py` passes; OpenSpec change validates.


skialpine:
### Code review

Found 8 issues:

1. The suppression latch cannot be re-derived after `DE1Device::disconnect()`, so a suppressed blip re-shows the warning on reconnect. The disconnect early-return clears `m_noAcHeaterInduced` but leaves `m_previousSubState` stale. That is only safe because `onTransportDisconnected()` resets the device substate — and `DE1Device::disconnect()` deliberately drops the transport's signals so that never runs, and does not reset `m_subState` itself. On reconnect `previousSubState == Error_NoAC`, the fresh-episode branch is skipped, and the warning fires from a dead connection's substate. Reachable from Android BLE recovery (`src/main.cpp:1885`) and the BLE↔USB switch (`src/main.cpp:2314`, `:2325`). This also violates the pre-existing spec: `openspec/specs/standby-switch-warning/spec.md:50` requires that a stale `Error_NoAC` "SHALL NOT persist across the disconnect and cannot re-show the warning on reconnect unless the machine reports it again".

https://github.com/Kulitorum/Decenza/blob/dae51c95b9a161aaf0f812e01df4253dcbfcb7e6/src/machine/machinestate.cpp#L360-L367

2. Mirror of the same hole: a genuine switch-off is suppressed indefinitely. Machine heating, `disconnect()` runs, both stale values keep `Heating`; on reconnect `connectedChanged` re-seeds `m_previousSubState = Heating` from the previous connection's evidence. A real `Error_NoAC` then latches as heater-induced, and since an open switch holds the substate there, the latch never clears.

https://github.com/Kulitorum/Decenza/blob/dae51c95b9a161aaf0f812e01df4253dcbfcb7e6/src/machine/machinestate.cpp#L384-L394

3. The suppression is blind on the connect/wake path, which is one of the two reported sightings. It accepts only evidence from the immediately preceding call, and on a fresh connect the substate is reset to `Ready` (`src/ble/de1device.cpp:334`) or is the `Ready` default. If the first `STATE_INFO` after connecting is already `Error_NoAC`, `isHeatingSubState(Ready)` is false and the warning shows exactly as before the fix. `DE1Device::isMachineHeating()` carries the heating fact without needing transition history.

https://github.com/Kulitorum/Decenza/blob/dae51c95b9a161aaf0f812e01df4253dcbfcb7e6/src/machine/machinestate.cpp#L386-L390

4. The comment above `isHeatingSubState` was updated to say "three readers" but keeps the two-party language it was written with ("That second read", "both agree"), and still names only the auto-tare failure mode. It does not say that widening the set would now also teach the standby-switch code that a new substate proves AC. Sharper: this PR's reader evaluates `previousSubState` ("did AC exist a moment ago") while the other two evaluate the current substate ("is the puck still dry") — one predicate, two different questions, coinciding only by observation.

https://github.com/Kulitorum/Decenza/blob/dae51c95b9a161aaf0f812e01df4253dcbfcb7e6/src/machine/machinestate.cpp#L316-L331

5. The same comment's "three readers" count is wrong, and the drift it warns about already exists: `DE1Device::isMachineHeating()` hand-writes the same three substates. (CLAUDE.md says "Centralize anything produced at more than one site — never hand-roll it per call, never copy a helper's body.")

https://github.com/Kulitorum/Decenza/blob/dae51c95b9a161aaf0f812e01df4253dcbfcb7e6/src/ble/de1device.cpp#L1407-L1415

6. The comment states "it was that firmware-arrival re-run that put the warning on screen after a restart" as settled fact, while this PR's own `proposal.md` says the subsystem "previously emitted nothing at all, which is why a 20,578-line user log could neither confirm nor deny when the warning fired". It is a reconstruction from MMR timing, not a traced fact. (CLAUDE.md: "an un-sourced claim in a comment gets believed and then licenses wrong code.") "Restart" is also ambiguous — the test named for the restart sighting sets up an in-session blip, and the app-relaunch case is not protected at all (see issue 3).

https://github.com/Kulitorum/Decenza/blob/dae51c95b9a161aaf0f812e01df4253dcbfcb7e6/src/machine/machinestate.cpp#L375-L384

7. The comments describe a "~3 s blip" and say the property is false "while the substate blipped in from a heating one", but the latch has no time bound — it holds for the entire episode. A switch genuinely opened right after heating is suppressed indefinitely. The PR description discloses this narrowing; the code comments read as though the mechanism is self-limiting.

https://github.com/Kulitorum/Decenza/blob/dae51c95b9a161aaf0f812e01df4253dcbfcb7e6/src/machine/machinestate.h#L69-L75

8. Logging tier asymmetry. Shown and cleared are INFO, but the suppression decision — the only line that would explain why a genuine open switch produced no warning — is DEBUG, invisible in the connections views that default to `minLevel INFO`. CLAUDE.md names this exact shape: "a fault reported at WARN whose *resolution* sits at DEBUG: the reader sees only the failure half".

https://github.com/Kulitorum/Decenza/blob/dae51c95b9a161aaf0f812e01df4253dcbfcb7e6/src/machine/machinestate.cpp#L388-L394

Issues 1, 2 and 3 share one root cause worth fixing together: the suppression reasons from the *previous substate*, which does not survive a teardown that leaves the device's own substate stale. Resetting `m_previousSubState` alongside the latch on disconnect, and giving `DE1Device::disconnect()` the substate reset `onTransportDisconnected()` already documents as necessary, addresses 1 and 2. Issue 3 needs the heating fact taken from the machine's current state rather than from transition history.

skialpine:
Reworked after the reporter confirmed they see the warning on **every tap-to-wake**. That condemned the original fix: on a tap-to-wake the machine is asleep at substate `Ready` and reports `Error_NoAC` on the wake transition, before any heating substate — so the heating-origin latch never fired for the commonest path.

The heating latch is replaced by a 6 s settling interval. A snapshot cannot distinguish the two cases at all: an open switch reports `Idle, Error_NoAC` and so does the transient, which de1app states outright — the switch "only moves the SUBstate (e.g. \"Idle, ready\" => \"Idle, Error_NoAC\")". Duration is the only discriminator, and every history-based rule misses whichever entry point it was not written for.

6 s is de1app's own interval (`de1_de1.tcl:946`) rather than a number of our own — Decent write the firmware that emits the substate.

This closes review issues 1, 2, 3 and 7 by construction: nothing reasons from the previous substate any more, so the stale-substate holes across `DE1Device::disconnect()` no longer reach this code, and a switch genuinely opened right after heating now warns after the interval instead of never. Issues 4 and 5 are gone with the third reader — `isPreFlowSubState` keeps its original name and two-reader comment.

It is a timer, which CLAUDE.md otherwise sends to an event-based flag. There is no event available: the DE1 pushes STATE_INFO on change, so nothing arrives to say a condition is still true, and the event-shaped alternative is exactly the origin rule this replaces. The rule gains a narrow carve-out for a hardware signal whose duration is the discriminator, with that argument recorded so the next reader does not re-derive it.

Net: smaller than the version reviewed, one mechanism instead of two.

Verification: build clean, 117 passed / 0 failed, no test warnings. Falsified — with the wait removed `tst_machinestate` goes red, restored it is green. Log-marker gate and OpenSpec validation pass. The three pre-existing standby tests were updated to elapse the wait; they encoded the old warn-immediately model.

skialpine:
### Code review

Second review, against the reworked design (`c9f2b577`). Found 8 issues:

1. **The de1app citation the design rests on is false.** Verified against the source. `de1_de1.tcl:946`'s `after 6000` is documented in the comment directly above it as "the belt-and-braces path for a machine that was ALREADY in Error_NoAC when the app launched — StateInfo is event-driven, so no notification would ever arrive on its own": a one-shot post-connect re-check, not a settling interval. `_update_front_switch_page_now` (`machine.tcl:1354`) shows the page immediately, with no duration check anywhere. **de1app applies no settling to this signal at all.** The claim appears in five places — this comment, the constructor, CLAUDE.md, `proposal.md` and `spec.md` — and the 6 s figure therefore has no provenance. CLAUDE.md's own rule is that a citation is the one thing never to cut; a citation that says something other than what it is cited for is worse than none, because it invites trust instead of checking.

https://github.com/Kulitorum/Decenza/blob/c9f2b5776ea9b698eb9141cfbf18def3cb5d9b37/src/machine/machinestate.cpp#L390-L397

2. **The CLAUDE.md amendment overstates the hardware claim.** It asserts the DE1 "pushes nothing while a condition merely persists" as settled fact, while [`docs/CLAUDE_MD/BLE_PROTOCOL.md:102`](docs/CLAUDE_MD/BLE_PROTOCOL.md) says the opposite is unknown: "the DE1's true minimum push cadence needs on-device measurement". What *is* verifiable is narrower and is our own code, not the machine's: `DE1Device` emits `subStateChanged` only on a value change. The amendment should claim that, with a line number, rather than contradict the project's own BLE docs.

https://github.com/Kulitorum/Decenza/blob/c9f2b5776ea9b698eb9141cfbf18def3cb5d9b37/CLAUDE.md#L99-L102

3. **The timer — the entire mechanism this PR adds — has zero coverage.** `elapseNoAcWait` reimplements the timeout lambda's body instead of firing it, so no test ever delivers a `QTimer::timeout`. Delete the `connect()` in the constructor and the whole suite stays green while the warning can never appear in production. The 6000 ms interval is unpinned too: change it to 600000 and nothing fails. Fix is to emit the real signal and assert the interval.

https://github.com/Kulitorum/Decenza/blob/c9f2b5776ea9b698eb9141cfbf18def3cb5d9b37/tests/tst_machinestate.cpp#L424-L431

4. **`standbySwitchStaysClosedOnOldFirmware` can no longer fail.** It never elapses the wait, so all three rows now pass because `m_noAcSettled` is false, not because of the firmware gate. Delete `&& m_device->firmwareBuildNumber() >= 1337` and it still passes — the only assertion protecting the 1337 rule is now inert. This PR created that.

https://github.com/Kulitorum/Decenza/blob/c9f2b5776ea9b698eb9141cfbf18def3cb5d9b37/tests/tst_machinestate.cpp#L402-L409

5. **The PR description is stale and contradicts the shipped code** on three counts: it describes the history-based fix and says "No timer and no debounce"; it says the `isPreFlowSubState` → `isHeatingSubState` rename shipped, which was reverted; and it says the suppressed-episode line is DEBUG, which is now INFO.

6. **`m_shotTimer` is not stopped in the disconnect branch this PR edits.** `stopShotTimer()` is only reachable from the connected phase-transition logic, which the early return skips, and `onShotTimerTick()` has no connectivity guard — so a mid-shot BLE drop leaves it emitting `shotTimeChanged()` every 100 ms. Pre-existing, but one line from the code this PR added, and CLAUDE.md says to fix pre-existing violations in files you touch.

https://github.com/Kulitorum/Decenza/blob/c9f2b5776ea9b698eb9141cfbf18def3cb5d9b37/src/machine/machinestate.cpp#L370-L379

7. **The claim that the `DE1Device::disconnect()` stale-substate hole "no longer reaches this code" is false.** `disconnect()` still never resets `m_subState`, and `onTransportConnected()` emits `connectedChanged` before a fresh `STATE_INFO` lands, so `updatePhase()` reads the previous connection's substate on every reconnect and can start a 6 s wait seeded by it. The consequence is now bounded — the real `STATE_INFO` clears it in ~100 ms — but the claim is wrong and the path is untested. Either reword it or reset `m_subState` in `disconnect()`.

https://github.com/Kulitorum/Decenza/blob/c9f2b5776ea9b698eb9141cfbf18def3cb5d9b37/src/machine/machinestate.cpp#L411-L415

8. **`standbySwitchWaitsWhateverItArrivedFrom`'s four rows exercise one code path.** There is no branch on the preceding substate any more, so the rows do not discriminate entry points as the comment implies. Defensible as a guard against a future history-based shortcut, but the comment should say that.

https://github.com/Kulitorum/Decenza/blob/c9f2b5776ea9b698eb9141cfbf18def3cb5d9b37/tests/tst_machinestate.cpp#L444-L462

Issues 1 and 3 are the ones that matter: the interval has no justification, and the mechanism that applies it is untested.

<details>

- [Worth a look] The CLAUDE.md carve-out does not say what makes this case *not* the debounce the same bullet still forbids — a bouncing contact is also indistinguishable at a snapshot, with no "still bouncing" event (score: 70)
- [Worth a look] The `STANDBY_INFO` comment says "all three lines" but there are five call sites — a count that goes stale on the next edit (score: 60)
- [Worth a look] `elapseNoAcWait` sits in `private slots:` rather than beside `armPreheat` in `private:`, so QTest registers it as a meta-method it then rejects (score: 55)

</details>

skialpine:
### Code review

Third review, against `9735708d`. Found 6 issues. Notably, most were introduced by the previous round's fixes rather than by the original change.

1. **`standbySwitchWaitIsNotRestartedByAReEvaluation` cannot fail for the regression its own comment describes.** It asserts the timer pointer is unchanged (trivially true) and `isActive()`. Change the guard to `} else if (!m_noAcSettled) {` — restarting the wait on every re-evaluation, exactly the bug named — and it still passes: `QTimer::start()` on an active timer resets the remaining time while `isActive()` reads true either way. It needs `remainingTime()`, or to observe `m_noAcEpisode` restarting.

https://github.com/Kulitorum/Decenza/blob/9735708db4d21cd0d6a54f68b2c99c9e1c50e4fc/tests/tst_machinestate.cpp#L507-L520

2. **`tasks.md` still carries the false de1app attribution the rest of this PR removed.** Task 1 reads "matching de1app's 6 s" while `proposal.md`, `spec.md` and the code comment all now state that de1app's `after 6000` is a post-connect re-check and that the interval is explicitly not theirs. One file was missed in the cleanup, and it contradicts the other three in the same PR.

https://github.com/Kulitorum/Decenza/blob/9735708db4d21cd0d6a54f68b2c99c9e1c50e4fc/openspec/changes/fix-standby-switch-heater-sag/tasks.md#L1-L6

3. **The logging this change exists to add has no test coverage**, despite being an ADDED requirement in its own spec. `STANDBY_INFO` resolves to `qInfo()`, catchable with `QTest::ignoreMessage(QtInfoMsg, ...)` — a pattern already used in this same file at line 247 for `QtWarningMsg`. A regression dropping or garbling any of the three lines passes every test here.

https://github.com/Kulitorum/Decenza/blob/9735708db4d21cd0d6a54f68b2c99c9e1c50e4fc/tests/tst_machinestate.cpp#L245-L249

4. **The `stopShotTimer()` added last round took half of a pairing the file documents three times.** At `:706-711`, `:785-789` and `:865-869`, stopping the shot timer is paired with stopping the *scale's* on-device timer. The new early-return call omits it, on the one path that also emits `espressoCycleEnded()`. A mid-shot DE1 drop now stops the app's clock while the scale — a separate BLE link, still up — keeps counting until the next flow cycle resets it.

https://github.com/Kulitorum/Decenza/blob/9735708db4d21cd0d6a54f68b2c99c9e1c50e4fc/src/machine/machinestate.cpp#L378-L387

5. **The duration-logging requirement this change wrote into its own spec is not met on two of three paths.** `spec.md` says each line SHALL carry the measured duration and interval. The disconnect line carries neither (though `m_noAcEpisode` is still valid three lines above the `invalidate()`); an episode *ended* by a disconnect is never logged at all, losing exactly the real-episode measurements the logging exists for; and the "warning cleared" line structurally cannot carry it, since the episode is invalidated ~20 lines before that transition is computed. The comment claiming "Every line below carries the episode's measured length in ms" is false. Separately, a shown warning clearing via substate change emits two INFO lines for one event, in two non-matching phrasings.

https://github.com/Kulitorum/Decenza/blob/9735708db4d21cd0d6a54f68b2c99c9e1c50e4fc/src/machine/machinestate.cpp#L455-L465

6. **"the only `stopShotTimer()` call" is a wrong count, repeated in two places.** There are two other call sites (`:706`, `:865`), both also skipped by the early return — so the conclusion holds but the stated fact does not.

https://github.com/Kulitorum/Decenza/blob/9735708db4d21cd0d6a54f68b2c99c9e1c50e4fc/src/machine/machinestate.cpp#L380-L386

<details>

- [Worth a look] The CLAUDE.md carve-out's "this is not debounce" argument is contestable: level-signal debounce IS "start a timer on the edge, trust the value only if it holds", which is structurally what this does. There is an edge here — what is missing is a second, *distinguishing* edge. The carve-out may still be right, but this argument does not establish it, and it is about to harden into house doctrine (score: 70)
- [Worth a look] The comment sentence "Recorded because an earlier revision of this comment cited that `after 6000` AS a settling wait" is drafting history about the comment itself — the category CLAUDE.md says to cut (score: 60)
- [OK to skip] `m_noAcEpisode.isValid() ? elapsed() : -1` — the `-1` arm is unreachable, since `standbySwitchOpen` implies `m_noAcSettled` implies a valid episode (score: 55)
- [OK to skip] `elapseNoAcWait` leaves the instance's interval at 1 ms rather than restoring 6000; harmless today since each test builds a fresh fixture (score: 50)

</details>

Issue #630, opened 2026-04-03
bug: original Decent Scale (non-HDS) fails XOR checksum validation on every packet


skialpine:
## Problem

The original Decent Scale (v1, non-HDS) sends weight packets (type `0xCE`) where the XOR checksum in byte 6 does not match the calculated checksum. Every packet is dropped with:

```
[BLE DecentScale] Invalid checksum on type 0xce, dropping packet
```

This results in the scale showing `Weight: 0.0 g` permanently — the scale is connected but no weight data gets through.

## Root Cause

Checksum validation was added in #560 for the HDS (Decent Scale v2). The original Decent Scale may not compute the XOR checksum the same way, or may send a garbage byte in position 6. Both scales advertise as "Decent Scale" over BLE — there is no name-based way to distinguish them.

The de1app (`binary.tcl:1381`) parses the XOR field but **never validates it**, so it works with both scale versions.

**Relevant code:** `src/ble/scales/decentscale.cpp:174-180`

## Proposed Fix

Auto-detect scales that don't support checksums:

1. Track consecutive checksum failures per connection
2. After N consecutive failures (e.g., 5) on weight packets (`0xCE`/`0xCA`), disable checksum validation for the remainder of the connection
3. Log a warning once: `"Checksum validation disabled — scale may be original Decent Scale (non-HDS)"`
4. Continue parsing weight data normally

This adapts automatically without requiring a settings toggle or scale type detection.

## Reproduction

Connect an original Decent Scale (v1) to Decenza. Observe the debug log flooding with checksum failures and weight stuck at 0.0g.

PR #1919, opened 2026-09-02
Decode the original Decent Scale's 10-byte v1.2 weight frames

skialpine:
## The report

Build 3582: an original Decent Scale (not HDS) spins `watchdog: no initial weight data` ten times, disconnects, reconnects, and repeats. Ten is exactly `kWatchdogMaxRetries`.

## Cause

#1891 made two changes that combine badly on a v1.2 scale:

1. `DecentScale::parseWeightData()` now requires a non-`0x25` notification to be **exactly 7 bytes**, else it logs "Undecodable frame" and returns false.
2. `handleData()` now tickles the watchdog **only when a frame decoded** — previously it tickled on any READ notification.

The original Decent Scale's v1.2 firmware notifies weight in **ten** bytes. de1app has dispatched on that length since forever: `decent_scale_weight_read_spec_v12` (`de1plus/binary.tcl:350-364`), selected at `:1422-1427`. The layout is the 7-byte one with minutes/seconds/milliseconds inserted after the weight short, two unused bytes, and the XOR moved to byte 9. Weight stays at bytes 2-3.

So every weight frame is now dropped: no weight, nothing feeds the watchdog, ten retries, disconnect loop.

Before #1891 it worked by accident. The guard was `data.size() < 7`, so the weight parsed out of bytes 2-3; the checksum compared byte 6 and failed; five consecutive failures tripped the v1 auto-disable (#630); and the watchdog was fed by every notification regardless of what it contained.

Decaid never hit this — `lib/src/models/device/impl/decent_scale/scale.dart` has no length check and no checksum validation, and its notification watchdog resets on any notification.

## Fix

- `notifiedFrameLengthExact(command, len)` answers the question a notification can actually ask: it carries exactly one frame, so the length is known. 7 always, 41 for `0x25`, and 10 **for weight types only**.
  - Weight only, because de1app decodes nothing else from the v1.2 layout, and the 7-byte LED response's battery and firmware bytes have no known v1.2 position — guessing would report a wrong battery level rather than nothing.
- `notifiedFrameLength()` is unchanged for the USB stream framer, where a buffer legitimately holds more than one frame so the length is ambiguous — and where the original scale cannot appear at all, having no USB.
- Bounding the checksum to the frame's own length also stops a good v1.2 frame spending the v1 auto-disable budget: byte 9 is checked, not byte 6.
- `0xCE` / `0xCA` are now named constants shared by both transports.

## Tests

Three added to `tst_scaleprotocol.cpp`:

- a 10-byte v1.2 weight frame decodes to the right grams
- its checksum is read at byte 9 — seven such frames produce no "Invalid checksum" and no auto-disable
- a 10-byte frame of another type stays undecodable

Full suite: 117 passed, 0 failed.

Not verified on hardware — I have no original Decent Scale. The reporter's log would confirm the 10-byte shape.

### DIFF
--- a/src/ble/protocol/decentscaleprotocol.h
+++ b/src/ble/protocol/decentscaleprotocol.h
@@ -112,6 +112,25 @@ inline bool checksumMatches(const QByteArray& data, qsizetype frameLen) {
     return static_cast<uint8_t>(frame[frameLen - 1]) == calculateXor(frame);
 }
 
+// Offset of the first valid 7-byte weight packet in `data`, or -1.
+//
+// This SCANS rather than frames, because its callers -- the USB probes, which
+// identify a scale by seeing one weight packet -- open a port mid-stream and so
+// start at an arbitrary byte. A framer that assumed data[0] began a frame would
+// miss the scale entirely on a port that happened to be mid-packet.
+inline qsizetype indexOfWeightPacket(const QByteArray& data) {
+    for (qsizetype i = 0; i + StandardFrameLength <= data.size(); i++) {
+        if (static_cast<uint8_t>(data[i]) != static_cast<uint8_t>(PacketHeader))
+            continue;
+        const uint8_t type = static_cast<uint8_t>(data[i + 1]);
+        if (type != TypeWeight && type != TypeWeightAlt)
+            continue;
+        if (checksumMatches(data.sliced(i, StandardFrameLength), StandardFrameLength))
+            return i;
+    }
+    return -1;
+}
+
 // HDS LED responses carry the firmware triple in their last two bytes: the
 // major version is BCD packed and the minor and patch values occupy one nibble
 // each. Keep this transport-neutral because USB and Bluetooth receive the same
--- a/src/ble/protocol/decentscaleprotocol.h
+++ b/src/ble/protocol/decentscaleprotocol.h
@@ -25,8 +25,20 @@ inline constexpr char PacketHeader = 0x03;
 inline constexpr uint8_t TypeLedResponse = 0x0A;
 inline constexpr uint8_t TypeAdsDebug = 0x25;  // 41 bytes, checksum in byte 40
 
+// Weight frame types. 0xCE is what the firmware notifies; 0xCA is accepted too
+// because de1app has always accepted both (binary.tcl:1437).
+inline constexpr uint8_t TypeWeight = 0xCE;
+inline constexpr uint8_t TypeWeightAlt = 0xCA;
+
 inline constexpr qsizetype StandardFrameLength = 7;
 inline constexpr qsizetype AdsDebugFrameLength = 41;
+// Decent Scale v1.2 firmware notifies weight in TEN bytes, not seven: the
+// 7-byte layout with minutes/seconds/milliseconds inserted after the weight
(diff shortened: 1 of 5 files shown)


skialpine:
### Code review

Found 4 issues:

1. Citation is off by one and lands on a blank line. `binary.tcl:1437` is blank; the `if {$recv(command) == 0xCE || $recv(command) == 0xCA}` check is at 1438. Same wrong start line is reused at L83. (CLAUDE.md says "The one thing never to cut is a SOURCE... because the next reader can check it")

https://github.com/Kulitorum/Decenza/blob/30e815c078d778ac6f099206cd0da7863514dc55/src/ble/protocol/decentscaleprotocol.h#L28-L30

2. `notifiedFrameLength()`'s doc now prescribes the bug this PR fixes. "A packet-oriented caller that wants exactness must compare the result against its own frame size" is exactly the deleted `frameLen != data.size()` line, and following it rejects the 10-byte v1.2 frame, since this function only ever returns 7 or 41. No caller does this any more. It should point at `notifiedFrameLengthExact()`. (CLAUDE.md says comments are "claims with expiry dates" and "When you change how something is read or called, re-read the comments justifying its optimisations")

https://github.com/Kulitorum/Decenza/blob/30e815c078d778ac6f099206cd0da7863514dc55/src/ble/protocol/decentscaleprotocol.h#L52-L57

3. File banner still says the file is a 7-byte protocol, after this PR added a second frame shape to the same namespace. A reader starting at the top has no signal `V12WeightFrameLength` exists.

https://github.com/Kulitorum/Decenza/blob/30e815c078d778ac6f099206cd0da7863514dc55/src/ble/protocol/decentscaleprotocol.h#L8-L11

4. Centralization pass stopped five lines short. This PR replaced `0xCE`/`0xCA` with named constants at L383, but L388 still hand-rolls `0x0A` — and `DecentScaleProtocol::TypeLedResponse` is already used for that same value at L346 of this file. (CLAUDE.md says "When you touch code that already has copies, collapse them in that pass")

https://github.com/Kulitorum/Decenza/blob/30e815c078d778ac6f099206cd0da7863514dc55/src/usb/usbdecentscale.cpp#L387-L389

<details>

- [Worth a look] Two more citations imprecise: "dispatched at :1422-1427" starts on `unset -nocomplain recv` and a blank line (real dispatch is 1424-1427), and "unpacks any 10-byte frame... (binary.tcl:1437-1460)" cites a range that only backs the second half of the claim — the "any 10-byte frame" half is at 1424-1427 via a different proc. Same one-line fix as finding 1 (score: 72)
- [OK to skip] `usbscalemanager.cpp:447,661` still hand-rolls `b0 == 0x03 && (b1 == 0xCE || b1 == 0xCA)` plus a copied XOR loop — a third and fourth copy of what this PR just centralized, but that file is untouched by this PR (score: 50)
- [OK to skip] The 10-byte frame feeds the shared `m_consecutiveChecksumFailures` counter rather than being insulated like the 41-byte ADS frame. Only bites if v1.2 hardware miscomputes byte 9 AND a valid-checksum 7-byte frame interleaves to reset the counter; the LED response is checksum-exempt and a button press needs a human, so there is no frequent resetter during idle streaming (score: 50)

</details>

PR #1936, opened 2026-09-12
Re-upload the profile after a cleared queue leaves the DE1 unknown

skialpine:
## What

A profile upload interrupted by flow starting was classified as non-retryable and then forgotten, leaving the DE1 holding an unknown profile with nothing to correct it. Found by reading 50 hours of Android debug log (2.0.5 build 3590, SM-X210); it occurred once there without reaching a shot.

## The bug

When flow starts, `MachineState` clears the DE1 command queue, which aborts any upload in flight with reason `"command queue cleared during upload"`. That reason is deliberately non-retryable — a timer retry would fight the running shot. But the handler only restored `m_profileUploadPending` when a *newer* profile change had queued behind the upload (`m_uploadPendingAfterInFlight`):

```cpp
if (!isRetryableUploadFailure(reason)) {
    if (hadPending) { m_profileUploadPending = true; }
    return;
}
```

`uploadCurrentProfile()` clears that flag on entry, so in the common case — a deferred upload firing on wake, then an auto-flush clearing the queue — nothing had queued behind it and the flag was lost. The `phaseChanged` handler then dead-ends on its first line (`if (!m_profileUploadPending) return;`), and because the branch returns before `updateProfileUploadRetrying()` and the retry counter, no banner and no dialog appear either. The DE1 keeps whatever it holds until the user next touches a profile.

**Why that matters.** Frames are indexed slot writes (`frame[0] = FrameToWrite`, `profile.cpp:2298`) with no commit step, under a header that already declared the new frame *count* (`profile.cpp:2276`). An interrupted batch can therefore leave the machine executing that count over a **mix of new frames and the previous profile's frames** in the slots that never arrived. The group-head button starts extraction with the app not in the loop, so the app cannot gate the shot — the only defence is the invariant that the DE1 always holds a complete, known profile.

The `"will re-arm"` rationale in the original comment was wrong from the start (added in #748), not something that regressed later.

## The fix

`isRetryableUploadFailure()` becomes `classifyUploadFailure()`, returning `Retry` / `ReArm` / `Defer` from one table of reason prefixes, so the strings are matched in exactly one place rather than at a second site in the handler. `"command queue cleared"` is now `ReArm`: still no timer retry, so it never fights the running shot, but the pending flag is set unconditionally and `phaseChanged` re-uploads once the machine is back to Idle/Ready — after the flow, never during it. Supersede, BLE disconnect and firmware flash stay `Defer`, where another path already owns the re-upload.

The recovery is logged at INFO to pair with the WARN `DE1Device` already emits on the failure. The connections views default to `minLevel INFO`, so a DEBUG line would have shown a reader only the failure half.

Two comments in the same handler corrected while here: the retry counter is reset only on success (not on "any non-retryable reason", as it claimed), and `clearCommandQueue()` is reached from `stopOperationUrgent()` as well as from flow start, so the INFO line no longer says "starting".

## Testing

Three tests added to `tst_profilemanager` (new functions in an existing file, no new target):

- `queueClearFailureReArmsPendingUpload` — the lost invariant.
- `queueClearedUploadRetriesOnceFlowEnds` — the field sequence end to end: asleep, wake to Idle, upload starts, queue cleared, nothing written during Pouring, re-upload on return to Idle.
- `supersededFailureDoesNotReArmPendingUpload` — boundary guard, so a future change cannot widen `ReArm` to every non-retryable reason.

Both new behaviour tests were confirmed to **fail without the fix** (`ReArm` temporarily reverted to `Defer`):

```
FAIL!  : queueClearFailureReArmsPendingUpload() 'm_profileUploadPending' returned FALSE
FAIL!  : queueClearedUploadRetriesOnceFlowEnds() Compared values are not the same
```

With the fix restored: **117/117 test targets pass, 0 failures, 0 skipped, no warnings** (Qt Creator, macOS Debug). All seven `text-invariants.yml` gates pass locally.

## Known, not addressed

`ReArm` arms the pending flag at the start of the flow that cleared the queue, so it stays armed for that whole flow. `Phase::Heating` passes both the `phaseChanged` allow-list and `uploadCurrentProfile()`'s `isActivePhase` block list, and `State::Steam` with a non-flow substate maps to `Phase::Heating` — so in principle a mid-steam substate change could fire an upload during steam.

Deliberately not guarded. The window is pre-existing (`uploadCurrentProfile()` already armed the flag mid-flow when blocked by an active phase), `Heating` cannot leave the allow-list because it is the normal wake path that five of six deferred uploads in the log resolved on, and nobody has shown the DE1 actually emits such a substate mid-steam. Guarding an unobserved state would be speculative. Recorded here so it can be fixed with evidence if it ever shows up in the field.

## Scope note

An earlier revision of this PR also gated the refractometer auto-connect in `blemanager.cpp::onDeviceDiscovered()` on `m_refractometerHunt`, to stop a failed R2 discovery holding the one-at-a-time GATT queue for ~6 s. That has been dropped: every such stall in the 50-hour log landed while the machine was idle, and the gate as written introduced a regression of its own (the duplicate-device `return` runs before the auto-connect decision, so an R2 first seen while the hunt was off stayed locked out for the rest of that scan). Not worth carrying a new regression to fix a stall nothing was waiting on. `blemanager.cpp` is untouched by this PR.

### DIFF
--- a/src/controllers/profilemanager.cpp
+++ b/src/controllers/profilemanager.cpp
@@ -72,33 +72,51 @@ static constexpr int kUploadRetryBaseMs = 1000;
 static constexpr int kUploadRetryMaxMs = 8000;
 static constexpr int kMaxUploadRetryAttempts = 5;
 
-// Reasons returned by DE1Device::profileUploaded(false, reason) that should
-// NOT trigger an auto-retry. The rest (frame sequence mismatch, ACK timeout)
-// are treated as retryable.
+// How a failed upload must be followed up. Reasons come from
+// DE1Device::profileUploaded(false, reason); matching them in one table keeps
+// the prefixes off the handler, where a second match site would be free to
+// drift from this one.
 //
 // We use startsWith() rather than exact equality so DE1Device can include
 // variable details after a stable prefix — for example "frame sequence
 // mismatch (expected [0x00, 0x01], got [0x00, 0x00])" carries the hex
-// payload in the same string. The exact retryable/non-retryable prefix
-// text is locked down by tst_profileupload.cpp's `.at(1).toString()`
-// assertions, so any future rename of a reason string in
-// finishProfileUpload() will break those tests loudly before it can
-// silently flip classification here.
-static bool isRetryableUploadFailure(const QString& reason) {
+// payload in the same string. The exact prefix text is locked down by
+// tst_profileupload.cpp's `.at(1).toString()` assertions, so any future
+// rename of a reason string in finishProfileUpload() will break those tests
+// loudly before it can silently flip classification here.
+enum class UploadFailureFollowUp {
+    Retry,    // Transient — arm the backoff timer.
+    ReArm,    // What the DE1 now holds is unknown — re-upload at the next safe moment.
+    Defer,    // Another path already owns the re-upload.
+};
+
+static UploadFailureFollowUp classifyUploadFailure(const QString& reason) {
     // Superseded: a newer upload is already in flight — let it own the outcome.
-    if (reason.startsWith(QStringLiteral("superseded"))) return false;
-    // Queue cleared: a shot/steam/hot-water just started, clearing the queue
-    // intentionally. The next uploadCurrentProfile() will re-arm.
-    if (reason.startsWith(QStringLiteral("command queue cleared"))) return false;
+    if (reason.startsWith(QStringLiteral("superseded")))
+        return UploadFailureFollowUp::Defer;
+    // Queue cleared: flow started and cleared the queue mid-upload. A timer
+    // retry would fight the shot, but Defer is wrong too — nothing else
+    // re-uploads, and what the DE1 holds is now unknown. Frames are indexed
+    // slot writes with no commit step (profile.cpp:2298) under a header that
+    // already declared the new frame COUNT (profile.cpp:2276), so an
+    // interrupted batch leaves the machine executing that count over a MIX of
+    // new frames and the previous profile's in the slots that never arrived —
+    // and the group-head button extracts that mix without the app in the loop.
+    // ReArm restores the invariant via phaseChanged, once the machine is
+    // Idle/Ready again: after the flow, never during it.
+    if (reason.startsWith(QStringLiteral("command queue cleared")))
+        return UploadFailureFollowUp::ReArm;
     // BLE disconnect: the reconnect path (initialSettingsComplete ->
     // applyAllSettings -> uploadCurrentProfile) already re-uploads when the
     // link comes back. Retrying on a timer would race with that.
-    if (reason.startsWith(QStringLiteral("BLE disconnect"))) return false;
+    if (reason.startsWith(QStringLiteral("BLE disconnect")))
+        return UploadFailureFollowUp::Defer;
     // Firmware flash: DE1Device dropped the call because a firmware update is in
     // progress. The reconnect path re-uploads once the flash completes and the
     // DE1 reconnects — no timer retry needed, and it would just flood the log.
-    if (reason.startsWith(QStringLiteral("firmware flash"))) return false;
-    return true;
+    if (reason.startsWith(QStringLiteral("firmware flash")))
+        return UploadFailureFollowUp::Defer;
+    return UploadFailureFollowUp::Retry;
 }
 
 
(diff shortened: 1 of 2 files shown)


skialpine:
### Code review

Found 6 issues:

1. Adding a second hunt gate makes the auto-connect decision once per device per scan, so an R2 first seen while the hunt is OFF is locked out for the rest of that scan (bug due to the duplicate-device `return` at `blemanager.cpp:1474` running before `autoConnect` is computed at `:1488`). Sequence: hunt ON and scanning, user leaves the review page, user powers the R2 on and its first advert is recorded with `autoConnect == false`, user re-opens the page within the same scan — `tryDirectConnectToRefractometer()` then declines with "a scan is already in flight" and every later advert hits the dedup `return`. Self-heals when the scan re-chains, so the cost is a delay on the one page whose purpose is taking the reading, not a permanent failure. Evaluate `autoConnect` before the dedup return, or clear `m_refractometerDevices` on the hunt's OFF→ON edge.

https://github.com/Kulitorum/Decenza/blob/69109c052d029ea5fb62cf27085b5fa332a89bbf/src/ble/blemanager.cpp#L1486-L1490

2. The pre-existing comment claiming `tryDirectConnectToRefractometer()` is "the single chokepoint every auto-reconnect path funnels through" is now false, and this PR is the reason (CLAUDE.md says "Fix pre-existing violations in any file you touch — do not dismiss them as 'pre-existing'", and calls stale/contradictory comments a recurring defect class here). Two comments in this file now contradict each other.

https://github.com/Kulitorum/Decenza/blob/69109c052d029ea5fb62cf27085b5fa332a89bbf/src/ble/blemanager.cpp#L2506-L2509

3. The new comment cites `tryDirectConnectToRefractometer() (:2522)`, but line 2522 is unrelated prose (`// used to arrive as one line.`) — the `if (!m_refractometerHunt)` gate is at `:2533` and the function starts at `:2502` (CLAUDE.md says "The one thing never to cut is a SOURCE ... because the next reader can check it"). A wrong citation is worse than none. The two `profile.cpp` citations in the same PR both check out, so this is an isolated slip.

https://github.com/Kulitorum/Decenza/blob/69109c052d029ea5fb62cf27085b5fa332a89bbf/src/ble/blemanager.cpp#L1481-L1484

4. `ReArm` arms the pending flag at the *start* of the flow that cleared the queue, and `Phase::Heating` is allowed through both the `phaseChanged` allow-list (`:149`) and `uploadCurrentProfile()`'s `isActivePhase` block list (which omits it). `State::Steam` with a non-flow substate such as Refill maps to `Phase::Heating` (`machinestate.cpp:544-551`), so a mid-steam refill can emit `phaseChanged` and push a full header + frame batch while the machine is in `State::Steam` — the contention `machinestate.cpp:642-647` clears the queue to avoid. Conditional: not confirmed that the DE1 emits that substate mid-steam. Prior to this PR the flag was armed on this path only when a newer profile edit had queued behind the upload, so the window is meaningfully wider now.

https://github.com/Kulitorum/Decenza/blob/69109c052d029ea5fb62cf27085b5fa332a89bbf/src/controllers/profilemanager.cpp#L203-L207

5. The new INFO line says the upload was "cleared by the machine starting", but `clearCommandQueue()` is also the stop path — `stopOperationUrgent()` calls it at `de1device.cpp:1509`, so a stop-at-weight interrupting an upload logs a start. CLAUDE.md notes these logs are read by users and their AI assistants. Widening to "cleared by the machine" is true on both paths.

https://github.com/Kulitorum/Decenza/blob/69109c052d029ea5fb62cf27085b5fa332a89bbf/src/controllers/profilemanager.cpp#L211-L215

6. Pre-existing comment in the handler this PR rewrites says "Success — or any non-retryable reason like \"superseded\" — resets the counter", but the non-retryable branch returns early without touching `m_profileUploadRetryAttempts`; the only reset on this path is in the `success` branch at `:190`. The new three-way `Retry`/`ReArm`/`Defer` split makes the stale claim more conspicuous.

https://github.com/Kulitorum/Decenza/blob/69109c052d029ea5fb62cf27085b5fa332a89bbf/src/controllers/profilemanager.cpp#L159-L163

Checked and clean: no upload storm or loop on the ReArm path (each cycle needs a fresh `clearCommandQueue()` with an upload in flight, bounded by the 10 s timeout); skipped retry state on the early return matches the existing `superseded`/`BLE disconnect`/`firmware flash` behaviour; manual pairing is unaffected (`connectToRefractometer()` never consults the hunt); the four-way ternary is logically correct; git history confirms `ReArm` does not recreate the #1466 stacking, which was fixed at the transport layer, and that `onDeviceDiscovered()` was an oversight in #1618 rather than a deliberate exemption.

<details>

- [OK to skip] The classifier comment's "two call sites" framing — `classifyUploadFailure()` has exactly one call site, and so did its predecessor, so the wording could send a reader hunting for a second one; reads as forward-looking rationale rather than a false claim (score: 50)

</details>

skialpine:
Updated at `307c3d2f`, addressing the review above.

**Dropped the refractometer gate entirely** rather than repairing it. That removes findings 1, 2 and 3 at the source instead of adding code to fix a regression in a fix for a problem that was not biting: every R2 stall in the log landed at idle, and the unit is powered on only some of the time. `blemanager.cpp` is untouched now.

**Finding 5** (log said "cleared by the machine starting" but `stopOperationUrgent()` takes the same path) and **finding 6** (stale claim that non-retryable reasons reset the retry counter) are fixed. The below-threshold "two call sites" wording is reworded too.

**Finding 4** (mid-steam upload via `Phase::Heating`) is deliberately not guarded, and is written up under "Known, not addressed" in the PR body. It is pre-existing, `Heating` cannot leave the allow-list because it is the normal wake path, and the triggering substate has not been observed.

Re-verified after the change: 117/117 test targets pass, no warnings, all seven text-invariant gates green, and the built test binary confirmed to carry the new tests and no trace of the reverted gate.

PR #1938, opened 2026-09-13
Add Belka PORTAL live EC and temperature integration

StandardDani:
## Summary

Add Belka PORTAL as an independent BLE sensor alongside the DE1 and scale. Raw EC and outlet temperature appear during extraction and remain available in shot review/history, aligned to the shot clock with gaps when measurements are interrupted.

PORTAL is selected through **Settings → Connections → Scales / Refractometer → Scan for Devices**. Connection progress and errors appear after explicit selection; saved controls persist after verified connection or restored pairing; users without PORTAL retain their existing layout, section heading and shared scan action. Saved pairing and optional display synchronization are independent of the scale/refractometer. Graph visibility uses the existing legend/settings system. Dragging the history/review cursor shows enabled PORTAL EC (raw) and outlet-temperature values, without interpolating across recorded gaps. A droplet beside the status-bar scale shows fresh live raw EC and lets owners retry an offline PORTAL, including while the DE1 waits for tank refill; saved layouts and the non-owner bar remain unchanged. Stale or disconnected measurements revert to status text. Live points append to persistent native renderers, with separate segments for measurement gaps; colors use the normal dark/light theme palettes. The device display receives the known graph-on/off commands at extraction start/finalization; BLE acknowledgements do not imply saved device sessions.

The decoder and command bytes reference the MIT-licensed [public PORTAL driver](https://github.com/variegated-coffee/variegated-rs/tree/a9cf6e324481ee70e399839bb1f8a68ef2050086/crates/variegated-belka-portal-trouble-driver). No proprietary APK/code is included. EC stays explicitly raw: units/compensation are not verified, and it is not TDS or hydraulic conductance. Existing shots remain compatible through an optional sample-blob field, without a schema migration. Private test-app packaging, signing, credentials and backups are excluded.

## Validation

- Complete native suite: **118/118 CTest entries pass**, with ASan/UBSan and both canaries enabled (Qt 6.11.2, macOS arm64, Clang 23 with Xcode libc++ 20 headers). The newer libc++ 22/23 headers triggered a sanitizer abort in an existing AI test; the Xcode header configuration passes without changing that application code or disabling checks.
- Android arm64 test APK builds successfully from these sources with the existing isolated private packaging. That packaging is outside this PR; the contribution retains the upstream app ID, updater and signing configuration.
- Full QML module: **245/245 files pass** the diagnostics gate. All repository text-invariant checks pass; OpenSpec strict validation passes.
- Decoder, shared discovery, connection lifecycle, display writes, extraction timing, sample persistence, old-shot compatibility, settings backup/import and legend/cache regression coverage. Deliberate mutations are detected and restored sources pass.
- Follow-up regressions reproduce Qt teardown without a disconnect signal, native asynchronous cancellation, subscription failure and real first-data/freshness timers, optional display errors followed by the generic transport error, ordered graph-on/off acknowledgements, shot-finalization/manual-control release, lazy transport creation and malformed-packet log suppression. Three deliberately reintroduced lifecycle defects each fail their regression; restored sources pass.
- Offscreen layout validation loads the real EspressoPage, ShotGraph and PORTAL overlay with inert backend fixtures at 1340×800, 1340×800 with 1.3× scale, and 1024×768. Both original overlap mutations fail the geometry checks. The fixture also checks persistent line-object identity across 6,000 appends, gaps, reset, page-entry replay and the non-owner layout. Native renderers are inert stand-ins in this geometry/dispatch fixture; this is not a tablet performance measurement or hardware screenshot.

The history cursor also has a regression against the shipping JavaScript lookup, including zero values, interpolation, boundaries and interruption gaps; removing the gap guard fails that test. Local offscreen checks cover the real history graph/readout and status indicator, visibility toggles, Fahrenheit conversion, accessibility announcements, owner gating and reconnect dispatch. Backend objects and the adjacent scale widget are inert fixtures; physical touch, radio and GPU verification remain open.

Hardware testing on a Samsung Android tablet confirms discovery/connection with the DE1 and scale present, live curves, saved-history curves after app restart, and the graph on PORTAL's own display. Two short extractions contain 107 and 125 stored PORTAL samples, with approximately 98 ms median notification spacing. A diagnostics packet is covered by a fixture against the screen's rounded values.

## Review corrections

The [first review](https://github.com/Kulitorum/Decenza/pull/1938#issuecomment-5654733357) and [follow-up review](https://github.com/Kulitorum/Decenza/pull/1938#issuecomment-5655817454) are addressed:

- One enum drives PORTAL activity/reading state. Qt disconnect completes synchronously; native cancellation remains explicit. The transport is created only on first selection.
- Only notifications feed measurements; the optional read is removed. Subscription failure and absent initial data end setup. Failed display writes preserve an established stream and cannot discard a later command acknowledgement.
- Connection progress/errors remain visible after first selection, and display-sync failures are visible without opening diagnostics. Users who never select PORTAL retain the original shared UI.
- Actual link faults always log, repeated routine events collapse, and shot end/faults flush counts. Malformed examples include length and hex without logging each changed payload.
- Local settings restore remains live; capture and graph-on/off follow the shot/settling clock. Live rendering appends incrementally, with shared validity/EC bounds and customizable colors in the app, web editor and palette generator.
- Visualizer-compatible history JSON is the default; both local exporters explicitly opt into PORTAL. The theme JSON null guard and stale comments are corrected.

The owner’s tablet log exposed another connection-policy defect: the DE1 stayed in `Refill` after startup, and that phase prevented every subsequent PORTAL connection attempt. Refill now permits reconnect; extraction and other active-operation guards remain intact. A controller/discovery replay covers startup, Refill, radio loss, reconnect via both cached and newly discovered devices, manual settings selection and protection during pouring. Restoring the old Refill gate fails these regressions; restored sources pass. Local QML checks cover live EC changes, zero, stale/offline/pending states and the raw-value accessibility label; restoring the fixed connected label fails the readout check.

## Why this is a draft

Remaining hardware acceptance: normal stop-at-weight shots, controlled PORTAL loss/reconnect and machine sleep/wake, restoring a PORTAL-containing backup, and checking the latest layout and reconnect fixes on the tablet. Apple cancellation changes compile on macOS but need separate physical-device verification. Persistent sessions on PORTAL itself have not been checked. The unverified custom JSON field is deliberately omitted from Visualizer uploads; local export/record-import retains PORTAL samples. The general file picker still imports Tcl shot files. Multi-shot comparison table values do not include PORTAL yet.

The wiki manual entry is prepared locally; it needs the maintainer's separate wiki publication workflow. OpenSpec acceptance and pre-merge archival are tracked in `openspec/changes/add-belka-portal/`.

## Hardware screenshots

These are real Android captures from the working test build, **before** the readout/header and view-selector/axis overlap corrections included in this PR. They demonstrate captured data and persistence; corrected layout verification on hardware is pending.

| During extraction | Saved shot review |
| --- | --- |
| Live DE1, scale and PORTAL curves | Saved PORTAL EC and temperature curves |


skialpine:
### Code review

Found 11 issues:

1. The PORTAL settings panel is shown to every user, including those who have never discovered a PORTAL. `PortalDevicePanel` has no `visible:` gate, so Settings -> Connections gains a "Belka PORTAL" header, a "Disconnected" readout, disabled Connect/Forget buttons, a sync switch and a diagnostics button. Few users will own this device, so it should be hidden until one has been discovered or saved, the way the refractometer panel is gated. (maintainer requirement: the app UI must not change at all until a PORTAL has been discovered by a Bluetooth scan and connected)

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/qml/pages/settings/SettingsConnectionsTab.qml#L1774-L1776

2. The shared "Scan for Devices" button is now disabled whenever `BelkaPortal.machineBusy` is true, which is every machine phase except Disconnected, Sleep, Idle, Heating and Ready. Users with no PORTAL can no longer scan for a DE1, scale or refractometer from this button during steam, flush, refill, descale and so on. (maintainer requirement: no behaviour change for users without the device)

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/qml/pages/settings/SettingsConnectionsTab.qml#L1085-L1087

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/main.cpp#L2283-L2286

3. The section header changes for every user from `settings.bluetooth.scalesRefractometer` ("Scales / Refractometer") to a new key `settings.bluetooth.scalesSensors` ("Scales & Sensors"). The new key has no translations, so translated users now see English. Please keep the existing header. (maintainer requirement: no UI change for users without the device)

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/qml/pages/settings/SettingsConnectionsTab.qml#L1011-L1013

4. PORTAL pairing and the sync preference bypass the Settings facade: raw `AppSettings` keys `portal/address`, `portal/name` and `portal/syncDisplay` are hand-typed in `main.cpp` and twice in `SettingsSerializer`. Beyond the convention this causes a bug. `importFromJson` runs while the app is running (backup restore, ShotServer restore, data migration) and writes the keys to disk, but `BelkaPortalDevice` only read them at startup. The next `connectDevice()` - including the automatic reconnect after every steam or flush - emits `savedDeviceChanged`, and `main.cpp` then writes the old in-memory values back over the restored ones. The refractometer goes through live `Settings` setters and does not have this problem. (CLAUDE.md says "Settings go in their domain sub-object, not on `Settings` directly." and "Centralize anything produced at more than one site")

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/main.cpp#L2244-L2256

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/core/settingsserializer.cpp#L440-L448

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/ble/belkaportaldevice.cpp#L156-L160

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/CLAUDE.md#L115-L122

5. Every measurement packet is logged at DEBUG, about 10 lines per second while connected, including when the machine is idle. That evicts the 500-line in-memory log in under a minute and adds about 10 lines per second to every shot's saved debug log. For comparison, the Decent scale logs each frame shape once per connection through `LogCollapse`, and the DiFluid R2 logs only malformed packets. Please do the same: log the first packet's shape once per connection and route malformed packets through `LogCollapse`. `lastPacket` and `packetCount` already expose the raw bytes in the diagnostics panel. (maintainer requirement: PORTAL logging should be no more than a scale or R2; LOGGING.md says "changing a repeat counter or moving a line to DEBUG is not suppression")

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/ble/belkaportaldevice.cpp#L211-L215

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/docs/CLAUDE_MD/LOGGING.md#L111-L124

6. Repeating state lines are not collapsed. "No measurement for more than 5 seconds" and "Receiving measurements" alternate without limit on a marginal link. The busy-phase disconnect/reconnect cycle also logs "Connecting", "Characteristics ready" and "Disconnected" around every steam, flush or hot-water run while PORTAL is not streaming. Please route these through `LogCollapse::kChangesOnly`. (LOGGING.md says "A periodic line that says nothing must say nothing")

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/ble/belkaportaldevice.cpp#L228-L250

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/ble/belkaportaldevice.cpp#L126-L136

7. The new `[PORTAL]` marker has no helper header; `PORTAL_DEBUG/INFO/WARN` are defined inline in the .cpp. (LOGGING.md says "Then a helper header aliasing `DECENZA_SUBSYS_LOG*` for the new marker")

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/ble/belkaportaldevice.cpp#L7-L11

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/docs/CLAUDE_MD/LOGGING.md#L201-L204

8. The live graph rebuilds all PORTAL data for every new sample. `portalSamplesVariant()` builds a fresh `QVariantList` of maps for the whole shot on each read, and it is read twice (by `ShotGraph` and by the `EspressoPage` legend). `PortalGraphOverlay` then recomputes both segment lists and the EC range, and its Repeaters recreate every line delegate. The cost grows with shot length and runs on the main thread during extraction. Please measure it on a tablet, or append incrementally as the other live series do. (CLAUDE.md says "the main thread is carrying BLE traffic and live telemetry, so the budget is much tighter than at idle")

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/models/shotdatamodel.h#L47-L49

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/qml/components/PortalGraphOverlay.qml#L16-L41

9. PORTAL policy lives in `main.cpp` lambdas that no test can reach: the capture gating, and a hand-written list of "busy" machine phases. The 5-second staleness threshold is also defined three times: `staleAfterMs`, a literal `5.0` in `ShotDataModel::addPortalSample`, and the log text. Please move the policy into a testable class and share one constant. (CLAUDE.md says "Centralize anything produced at more than one site")

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/main.cpp#L2255-L2289

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/ble/belkaportaldevice.cpp#L16-L18

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/models/shotdatamodel.cpp#L516-L518

10. The two PORTAL trace colours are literal hex values. Every neighbouring trace colour goes through `_c("...", Settings.theme.customThemeColors.... || "#...")`, so these two cannot be customised and are missing from the web theme editor and `CustomEditorPopup`. (bug due to `readonly property color portalEcColor: isDarkMode ? "#26d7bb" : "#007d68"`)

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/qml/Theme.qml#L926-L933

11. Two comments are made stale by this change. `settings_graph.cpp` still says "Eleven properties", but there are now thirteen. `GraphSeries.qml` still says "the two visibility gates" and documents only `advanced`/`postShotOnly`, but this PR adds a third, `portal`. (CLAUDE.md says "Re-derive the justification when the design around it moves.")

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/src/core/settings_graph.cpp#L57-L60

https://github.com/Kulitorum/Decenza/blob/a5f1e70964fa9e886ac0be715a6d390d3045f416/qml/components/GraphSeries.qml#L5-L22

---

<details>

- [Worth a look] At startup, the saved-PORTAL reconnect calls `BLEManager::scanForDevices()`, which is the user-scan entry point. For users with a saved PORTAL, every launch restarts the startup scan, resets the scale-failure dialog guards and starts WiFi/USB probing (score: 60)
- [Worth a look] A display-command write that times out emits neither `characteristicWritten` nor `error`, so `displayCommandStatus` stays "requested" and the manual Show/End graph buttons stay disabled until reconnect (score: 55)
- [Worth a look] `CoreBluetoothScaleBleTransport::didDisconnectPeripheral` ignores which peripheral disconnected. PORTAL's quick busy-to-idle disconnect and reconnect can let a late callback clear the new connection's state and strand the device until Forget (score: 55)
- [Worth a look] The last-shot image cache key includes the PORTAL visibility flags for every shot; include them only when the shot has PORTAL samples (score: 50)
- [Worth a look] Visualizer uploads now carry `decenza_portal_samples`, which tasks.md lists as unverified with the service; consider leaving it out until verified (score: 50)
- [OK to skip] PORTAL samples are stored as one object per sample rather than in the columnar `pointsToJsonObject` format; level-9 compression absorbs most of the size difference (score: 50)

</details>

skialpine:
### Follow-up review of d406e54

Thanks for the quick turnaround. This commit resolves the earlier review:
- **UI for non-PORTAL users:** PORTAL UI stays hidden until a device is saved, and the scan button and heading are restored.
- **Settings and logic:** settings moved into `SettingsHardware` and restore works live; policy moved into `PortalController` with one shared threshold.
- **Logging:** the per-packet log is gone and `portallogging.h` was added.
- **Performance and theming:** live graph appends are incremental; trace colours go through `_c()`.
- **Integrations:** the last-shot cache key ignores shots without PORTAL data; the Visualizer upload leaves the samples out; `gattOperationFailed` was added.

The per-PR text gates pass on this head.

A second, broader pass found the problems below. The first four break PORTAL itself, or change the app for people who do not own one.

## Blocking

**1. On Android, Windows and Linux, PORTAL cannot reconnect after any disconnect until the app restarts.**
- **The flag:** `m_waitingForDisconnect` is set whenever the link is up, and it is cleared only by the transport's `disconnected` signal.
- **Why it never clears:** `QtScaleBleTransport::disconnectFromDevice()` drops every controller signal and never emits `disconnected`; its own comment at `qtscalebletransport.cpp:700` says so. The flag therefore stays set after Disconnect, Forget, a transport error, the machine going busy during connect, or a restore that swaps devices. From then on `connectDevice()`, `tryReconnect()` and the scan fallback all return early. The readout stays on "Disconnecting", and the display-command status can stay on "requested".
- **Why the tests miss it:** the fake transport emits `disconnected` from inside `disconnectFromDevice()`, which the real Qt transport does not do.
- **Suggestion:** replace the state string plus `m_requested` / `m_valid` / `m_waitingForDisconnect` with one `Q_ENUM State`, and treat the Qt transport's own disconnect as the end of the disconnect. The comment and spec text claiming "reconnect waits for disconnect completion" only hold on CoreBluetooth.

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/ble/belkaportaldevice.cpp#L184-L201

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/ble/transport/qtscalebletransport.cpp#L148-L183

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/tests/tst_belkaportal.cpp#L42-L48

**2. One failed display write, or a failed optional read, disconnects PORTAL for the rest of the shot.**
- **What the transports do:** on a write, read or operation error, the Qt transport calls `failGattOperation()` and then also emits `error("Service error: ...")`. CoreBluetooth emits `error` for failed writes and reads as well.
- **What PORTAL does:** it connects `error` directly to `fail()`, which disconnects. So if graph-on fails at extraction start, capture stops for the shot, busy phases block reconnect, and graph-off is never sent.
- **Why the tests miss it:** `failedDisplayWriteDoesNotStopMeasurements` emits only `gattOperationFailed`, not the `error` that follows it.
- **Fix:** only call `fail()` for errors from discovery or the measurement subscription.

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/ble/belkaportaldevice.cpp#L76-L78

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/ble/transport/qtscalebletransport.cpp#L859-L893

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/ble/transport/corebluetooth/corebluetoothscalebletransport.mm#L570-L577

**3. If notification setup fails on Android, PORTAL can sit on "Waiting for measurements" forever.**
- **Ignored failure:** the `gattOperationFailed` handler only looks at the command characteristic, so a timed-out subscription to 0x7410 is ignored.
- **Silent rejection:** when the device rejects the notification-enable write, the Qt transport logs a warning worded for a *scale* and returns without emitting `error`.
- **No fallback:** the health timer only starts after the first valid packet, so nothing ends the wait.
- **Fix:** handle `key == measurement` as a subscription failure, and give "waiting" a deadline that ends when the first notification arrives.

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/ble/belkaportaldevice.cpp#L69-L76

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/ble/transport/qtscalebletransport.cpp#L866-L881

**4. Every macOS and iOS user now starts a second CoreBluetooth manager at launch, with or without a PORTAL.**
- **Where:** `main.cpp` constructs `CoreBluetoothScaleBleTransport` unconditionally, and its constructor creates a `CBCentralManager` immediately.
- **Why it matters:**
  - This undoes the lazy creation `BLEManager` documents for simulator-mode launches.
  - Unlike `applebtstate.mm`, it omits `CBCentralManagerOptionShowPowerAlertKey: @NO`, so the system Bluetooth-off alert may appear on launch. Please confirm on a device.
  - Every Bluetooth state change writes a `[Scale]` "Central state" log line, even for users with no scale or PORTAL.
- **Fix:** create the transport lazily on the first `connectDevice()`, as scales do. The Qt transport is already lazy, so Android is unaffected. (maintainer requirement: no behaviour change for users without the device)

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/main.cpp#L2238-L2244

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/ble/transport/corebluetooth/corebluetoothscalebletransport.mm#L602-L604

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/ble/blemanager.cpp#L189-L193

## Logging

**5. The log collapsing now hides real events.**
- **Too wide an episode:** `kChangesOnly` with a fixed text per key means "disconnected", "stale", "streaming", "malformed" and "display-failed" each print once, then stay silent until a manual disconnect, a device change, or app exit.
- **Real drops disappear:** busy-cycle disconnects and unexpected link loss share one key. After the first drop, later drops during shots never reach the log, and neither does recovery after a fault. LOGGING.md: "An episodic source must flush at its run end".
- **Malformed packets:** the WARN has no length or bytes, so a firmware format change cannot be diagnosed from a submitted log.
- **Suggestion:** always log unexpected disconnects and errors, and flush counts at a natural boundary (reconnect or shot end). Keep collapsing the routine stale and busy-cycle lines. Include length and hex in the malformed line.
- **Tests:** they currently pin `"+5999 identical"` and `<= 8` lines. Asserting "no line repeats before the flush" is less brittle.

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/ble/belkaportaldevice.cpp#L370-L386

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/docs/CLAUDE_MD/LOGGING.md#L129-L133

## Should fix

**6. When a first connection fails, the user sees nothing.** The pairing is saved only after capability discovery, and the panel and readout stay hidden until then. A timeout, missing service or transport error leaves the tapped row unchanged, with no "Connecting" or "Connection failed". Tapping a second PORTAL while one is connecting is also ignored silently. Gating the panel on `savedAddress.length > 0 || BelkaPortal.state !== "disconnected"` keeps it hidden for anyone who never taps a PORTAL.

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/qml/pages/settings/SettingsConnectionsTab.qml#L1774-L1779

**7. The end-of-shot wiring is not tested.** If the `shotProcessingReady` connection breaks, `m_extractionActive` stays true after the first shot. Connect, Forget and the display buttons then do nothing, and graph-off is never sent. `captureFollowsRealExtractionClockAndFinalization` still passes, because `sensorTime()` already returns nothing after `endShot()`. Please announce the 0x7420 characteristic in that test and assert the graph-on and graph-off writes, plus one manual write afterwards.

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/controllers/portalcontroller.cpp#L33-L37

**8. `includePortal` defaults to `true`.** Only the upload call site passes `false`, so any new upload path leaks the unverified extension. Please default it to `false` and have the two local exporters opt in. The header comment ("visualizer-compatible") also no longer describes the default output.

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/network/visualizeruploader.h#L212-L216

**9. The display-command status can read "failed" after a write succeeded.** Graph-on and graph-off can both be in flight. If the first is abandoned, the counter drops to 0 and the second write's acknowledgement is ignored. Tracking the last completed write, rather than a count, avoids this.

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/ble/belkaportaldevice.cpp#L63-L76

**10. `ShotServer::buildThemeJson()` dereferences `m_settings` one line before its null guard.**

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/src/network/shotserver_theme.cpp#L16-L21

**11. The PORTAL colours are missing from the in-app theme editor and the theme generator.** Only the web editor and the widget swatches have them, while `design.md` and `tasks.md` say the theme editors are covered.

https://github.com/Kulitorum/Decenza/blob/d406e54a8d6d6a8233d21326f961710849371849/qml/pages/settings/SettingsThemesTab.qml#L45-L56

<details>
<summary>Minor: comments, docs and suggestions</summary>

- **`portalcontroller.h`** says it "Owns capture and connection policy", but the reconnect and busy policy live in the driver.
- **`portalcontroller.cpp:58`** says saved devices piggyback on the startup scan. That scan only runs with a saved BLE DE1 or scale, so for example a USB DE1 user gets no startup attempt.
- **`tests/CMakeLists.txt:1333`** says "no brew-time writes", but the tests assert graph writes at extraction start and end.
- **`tasks.md:18` and `:55`** cite offscreen UI checks that are not in the tree; please mark them as local checks.
- **`settings_graph.cpp:59`** still says "88 lines". The `SettingsHardware` class comment still says "sent to the DE1 firmware".
- **`GraphSeries.keys`** has no users left. **`visualizeruploader.cpp:1222`** has an orphan comment. The LOGGING.md row omits the `[PORTAL][BLE]` tag.
- **Duplicated rules:** sample validity is implemented in both `PortalSamples::fromVariant` and `ShotDataModel::addPortalSample`. The EC axis seeds (0 / 0.1) and the 1.1 padding are repeated in C++ and QML.
- **Untested:** the `isPortal()` classifier and BLEManager's early return. The freshness timer is never exercised; every stale test calls `expireReading()` directly.
- **User-facing errors:** raw transport strings such as `!!! CONTROLLER ERROR ...` reach the panel untranslated. A failed automatic display sync appears only inside the collapsed diagnostics.

</details>

PR #1943, opened 2026-09-14
fix: non-espresso profiles ignore the espresso brew-by-ratio stop-at-weight

### LINKED ISSUE #1941: Tea/non-espresso profile keeps the espresso brew-by-ratio stop-at-weight (steep cut short)
# Bug: switching to a non-espresso (tea/water) profile keeps the espresso stop-at-weight

## Summary
Selecting a **tea (or other non-espresso) profile** does not clear the espresso
**brew-by-ratio** yield anchor, so the tea steep is stopped at `ratio × dose` grams
instead of running by the tea profile's own frames. The native de1app treats
stop-at-weight as profile-owned and its tea profiles ship `final_desired_shot_weight 0`
(no weight stop); Decenza diverges here.

## Repro
1. Dial in an espresso profile with a **ratio** yield anchor (e.g. 1:2.5).
2. Without changing the yield, select a tea-portafilter profile (e.g. "Rishi Turmeric Chai").
3. Pull. The steep stops on weight at ~`2.5 × dose` g (observed on-device: **48.8 g**,
   inherited from an ~19 g espresso dose × 2.5), cutting a ~96 s steep short.

## Root cause
Stop-at-weight resolves through the yield ladder in `ProfileManager::targetWeight()`:
`YieldSpec::resolveGrams(brewYieldMode, brewYieldOverride, dose, profile.targetWeight())`.
A **ratio** anchor is deliberately *kept* across profile loads
(`SettingsBrew::clearProfileScopedBrewOverrides()` early-returns for ratio mode —
add-yield-ratio-anchor Decision 8, "1:2 is 1:2 on any profile"). That assumption holds
for espresso but not for a tea/water profile, which has no dose-ratio concept — so the
ladder resolves the leftover espresso ratio and never consults the tea profile's own
`target_weight` (which is imported as 0 = no stop, matching de1app).

- `clearProfileScopedBrewOverrides()` keeps ratio, clears absolute + temperature.
- `weightprocessor.cpp` gates the stop on `m_targetWeight > 0`, so 0 = no weight stop.
- The de1app importer maps `final_desired_shot_weight[_advanced]` → `target_weight`
  (`de1apptclfields.cpp`), so the tea profile's 0 is carried through correctly; it is
  simply never reached while a ratio anchor is active.

## de1app (native) behavior for comparison
`select_profile` (vars.tcl) loads each profile's **own** `final_desired_shot_weight` via
`load_settings_vars`. There is no persistent cross-profile ratio anchor. Every shipped tea
profile carries `final_desired_shot_weight 0` / `beverage_type tea_portafilter`, so
selecting one disables the weight stop. de1app also explicitly resets per-profile state on
switch "so the value doesn't leak from whichever profile was selected before."

## Proposed fix
A dose-ratio yield only has meaning for espresso. Gate the ratio anchor to espresso-class
profiles at the single resolution point; non-espresso profiles honor their own
`target_weight` (0 = no stop). The anchor itself is left intact, so returning to an
espresso profile restores brew-by-ratio.

```cpp
// ProfileManager::targetWeight(), after the m_shotLatched early-return:
if (brew->brewYieldMode() == YieldSpec::modeRatio()
    && !Profile::isEspressoBeverageType(m_currentProfile.beverageType()))
    return m_currentProfile.targetWeight();
```

with a `Profile::isEspressoBeverageType(t)` helper (empty ⇒ espresso, so no change for
existing espresso profiles). Absolute anchors are already cleared on load, so only the
surviving ratio case needs the guard.

Regression test: a live 1:2.5 ratio + 18 g dose resolves to 45 g on an espresso profile,
0 g (its own target) on a tea profile, and returns to 45 g when an espresso profile is
reloaded — proven red without the guard.

loganross2001:
Fixes #1941.

## Problem
Selecting a tea (or other non-espresso) profile does not clear the espresso
**brew-by-ratio** yield anchor, so the tea steep is stopped at `ratio × dose` grams
instead of running by the profile's own frames. Observed on-device: a tea profile
stopped on weight at ~48 g, inherited from a ~19 g espresso dose × 2.5.

## Root cause
Stop-at-weight resolves through `ProfileManager::targetWeight()`:
`YieldSpec::resolveGrams(brewYieldMode, brewYieldOverride, dose, profile.targetWeight())`.
A ratio anchor is deliberately kept across profile loads
(`SettingsBrew::clearProfileScopedBrewOverrides()` early-returns for ratio mode —
add-yield-ratio-anchor Decision 8, "1:2 is 1:2 on any profile"). That holds for espresso
but not for a tea/water profile, which has no dose-ratio — so the ladder resolves the
leftover espresso ratio and never reaches the tea profile's own `target_weight` (imported
as 0 = no stop; `weightprocessor.cpp` treats target 0 as "no weight stop").

## Fix
Gate the ratio anchor to espresso-class profiles at the single resolution point;
non-espresso profiles honor their own `target_weight`. The anchor is left intact, so
returning to an espresso profile restores brew-by-ratio. Absolute anchors are already
cleared on load, so only the surviving ratio case needs the guard.

Matches the native de1app, where stop-at-weight is profile-owned (`select_profile` loads
each profile's own `final_desired_shot_weight`) and every shipped tea profile carries
`final_desired_shot_weight 0`.

## Test
`tst_profilemanager` — a live 1:2.5 ratio + 18 g dose resolves to 45 g on an espresso
profile, 0 g (its own target) on a tea profile, 250 g on a tea profile with an absolute
stop, and back to 45 g when an espresso profile is reloaded. Proven red without the guard.


skialpine:
ran with your fix, but changed some of the behavior to me more intuitive.

Issue #1942, opened 2026-09-14
Filter the profile list by beverage category (Coffee / Tea & Water / Maintenance)


loganross2001:
### Problem
The profile selector's only quick filter is by *source/ownership* (Selected, Decent
Built-in, Downloaded, User Created, All). With espresso, tea-portafilter, pour-over and
maintenance profiles all mixed together, finding the one you want — especially when you
regularly brew more than just espresso (e.g. tea via the tea portafilter) — means
scrolling or searching every time.

### Request
Add a top-level filter by **beverage category**, surfaced as a segmented row above the
search box on the profile selector (which is shared by the picker and Settings):

```
[ Coffee | Tea & Water | Maintenance | All ]     ← new
[ Selected ▼ ]  [ Search… ]                       ← existing ownership filter + search
```

Three categories, AND-combined with the existing ownership filter:
- **Coffee** ← `espresso`, `pourover`, `filter`, `manual` (+ empty/unknown, which already
  defaults to espresso)
- **Tea & Water** ← `tea_portafilter`, `tea` (+ any water/rinse type)
- **Maintenance** ← `cleaning`, `descale`, `calibrate`

### Why this is cheap
No new tagging is needed — every profile already carries a `beverage_type`, and the
selector already computes a per-profile display category from it. This is a filter over
data that already exists. It also pairs naturally with the tea stop-at-weight issue
(separate report): once tea is a first-class, easy-to-find category, non-espresso brewing
becomes a normal path rather than an edge case.

### Notes / open choices for the maintainer
- Default category on open (Coffee keeps the common case clutter-free; All is the most
  neutral), and whether to remember the last-used category and ownership filter across
  opens (today the ownership filter resets each open).
- Whether `manual`/utility profiles belong under Coffee or a fourth category.

(Implemented and running in a community fork; happy to open a PR if the direction and the
category mapping look right.)

skialpine:
Shipped in #1948 (profile picker rebuild: search, composable filters including a beverage group — Espresso / Filter / Tea / Maintenance — favorites-only membership, bean-ranked recommendations) and #1949 (follow-ups: Favorites Order dialog size, wizard highlights the recipe's profile, picker pinned to the top while the recommended row arrives). Both squash-merged to main: bdd9a2dd and 5d94f1b1.

PR #1944, opened 2026-09-14
feat: filter the profile list by beverage category (Coffee / Tea & Water / Maintenance)

loganross2001:
Addresses #1942.

## What
Adds a top-level **beverage-category filter** to the profile selector (shared by the
picker and Settings): a segmented row of **Coffee · Tea & Water · Maintenance · All**
above the search box, AND-combined with the existing ownership filter.

## Why
The only quick filter today is by source/ownership (Selected, Built-in, Downloaded, …).
With espresso, tea-portafilter, pour-over and maintenance profiles mixed together,
finding the one you want takes scrolling or searching every time — especially once you
brew more than espresso.

## How
Driven entirely by the existing `beverage_type` tag — no re-tagging. Mapping:
- **Coffee** ← `espresso`, `pourover`, `filter`, `manual` (+ empty/unknown, which already
  defaults to espresso)
- **Tea & Water** ← `tea_portafilter`, `tea` (+ `water`)
- **Maintenance** ← `cleaning`, `descale`, `calibrate`

The category predicate and search are applied uniformly to whichever list the ownership
filter returns; as a side benefit, the search box now filters in **every** ownership view
(previously it only worked under "All Profiles"). Defaults to Coffee on open.

## Open choices for the maintainer
- Default category on open (Coffee keeps the common case clean; All is most neutral), and
  whether to remember the last-used category/ownership filter across opens.
- Whether `manual`/utility profiles belong under Coffee or a fourth category.

Pairs with #1941 (tea stop-at-weight): once tea is an easy-to-find first-class category,
non-espresso brewing becomes a normal path.


skialpine:
Its not a bad idea, but. 

1 It does not seem to know about themes, background and light and dark mode, It only looks ok in light mode with standard background.
2 makes an awkward interface feel more awkward.

Somehow I think there is a better way. Maybe mixing your idea here with the which profile from the new recipe window, What do you think about that? 

image

skialpine:
I am going to attempt to improve this along the lines I called out above. I don't like profile selection either.
Will ask you to review the changes. I am going to use the filter idea.

skialpine:
Superseded by #1948, which rebuilt the profile picker with a beverage chip group (Espresso / Filter / Tea / Maintenance, driven by the same beverage_type tag) alongside search, source and favorites filters, and by #1949's follow-ups. Thanks for the push on this — the category split you proposed is what shipped, with Filter as its own bucket so downloaded filter profiles are one tap away.

PR #1955, opened 2026-09-19
Keep a DE1's connect setup through a queue clear

skialpine:
## Problem

On an SM-X210 running build 3596, the tablet held a Bluetooth link to the DE1 for 16 hours but the app never finished connecting. Writes (charger keepalives, a profile upload) succeeded, but no machine state arrived: MCP reported `connected: true` with `phase: "Disconnected"`. Scans found nothing, because a connected DE1 does not advertise.

The log shows why:

```
[   7.275] INFO  [DE1][BLE] Characteristics ready: 18 registered
[   7.368] DEBUG [Scale][BLE DecentScale] Disabling LCD (scale stays powered)
[   7.372] DEBUG [Bluetooth][GattQueue] dropped 11 operation(s) for a torn-down transport
[   7.376] DEBUG [Screensaver][main] goToScreensaver called, type: videos
```

A sleep request about 100 ms after characteristics were ready called `clearQueue()`, which dropped the entire connect setup: 6 notification subscriptions, the ready marker and 4 initial reads. `DE1 CONNECTED` was never logged, `DE1Device` stayed `connecting`, and every reconnect path skips a device that is connecting.

## Fix

**`clearQueue()` clears commands, not the connect setup.** `BleGattQueue::Operation` gains a `connectSetup` tag, set on characteristic discovery, the subscriptions, the ready marker and the initial reads. `clearQueue()` calls the new `BleGattQueue::clearCommands()`, which is `forget()` except that it keeps tagged operations, queued or in flight. `forget()` is unchanged for teardown. A sleep or stop during connect still goes to the front and is sent first, and the setup completes behind it.

**A sleep during connect is not undone by the connect-time wake.** `goToSleep()` still sends immediately, as on `main`, but if it happens while connecting it sets a flag, and `onTransportConnected()` skips its usual `requestState(Idle)`. Any Idle request (a wake) clears the flag, as does `disconnect()`. A flag that is never consumed can only suppress an Idle that never comes; it cannot lose a sleep.

Net change against `main`: 9 files, +124/-14.

## Why this and not the earlier commits

The first three commits on this branch kept `clearQueue()` dropping the setup and repaired the damage afterwards: a teardown and reconnect, then a requeue, then deferring the sleep until connected. Each review found new bugs in that repair machinery (dropped follow-up writes, the machine woken after a sleep, a sleep that could be deferred forever). This commit replaces all of it by not dropping the setup in the first place. The PR is squash-merged, so only the net change lands.

## Not a 2.0.6 regression

Builds 3595 and 3596 differ only by #1938 (Belka PORTAL), which does not touch this path. The window has existed since the shared GATT queue and ready marker landed in #1820.

## Tests

- `tst_blecommandqueue::clearKeepsConnectSetupAndDropsCommands`: queued discovery and ready marker survive a clear while the two commands around them are dropped and counted; an in-flight discovery survives too, with its operation clock still running. Fails with `clearQueue()` reverted to `forget()` (4 dropped instead of 2).
- `tst_de1device_headless::sleepWhileConnectingIsNotUndoneByTheConnectWake`: a sleep during connect is written at once and no Idle follows on connect (fails without the flag: two state writes instead of one); a wake in between supersedes the sleep (fails without the wake clearing the flag).

The subscriptions and initial reads cannot be queued without a `QLowEnergyService`, so discovery and the ready marker stand in for the setup in the headless test.

Full suite: 118/118 passed locally (macOS, Debug).

## Not covered

- A sleep issued before characteristics are ready is still skipped by the transport, exactly as on `main`.
- The connect completing on real hardware after a mid-connect sleep needs a check on the tablet.


skialpine:
### Code review

Found 3 issues:

1. The requeued setup wakes the machine straight after the sleep it was protecting (bug due to `onTransportConnected()` requesting Idle unconditionally). `goToSleep()` during setup clears the queue, `clearQueue()` requeues the setup, the Sleep write goes out first, and then the requeued ready marker emits `connected()`, which calls `requestState(DE1::State::Idle)`. With the screensaver already up, nothing re-sends sleep, so the DE1 is left awake. The same happens on app exit, where the drain wait lets the requeued setup and the Idle go out before quit.

https://github.com/Kulitorum/Decenza/blob/b5284b0cb25d3e240b982721d287c156b0cf8662/src/ble/bletransport.cpp#L305-L318

https://github.com/Kulitorum/Decenza/blob/b5284b0cb25d3e240b982721d287c156b0cf8662/src/ble/de1device.cpp#L168-L171

2. A clear while characteristic discovery is still queued leaves the link stuck in the same way (bug due to recovery keying only on the ready marker). `submitDiscovery()` queues the discover operation on the shared queue, where it can wait behind another device. `clearQueue()`'s `forget()` drops it and stops the operation timer, so `discoverDetails()` never runs, `subscribeAll()` never runs, `m_readyMarkerPending` is never set, and nothing requeues. Some `goToSleep()` callers (MQTT, relay, ShotServer) do not gate on the DE1 being connected.

https://github.com/Kulitorum/Decenza/blob/b5284b0cb25d3e240b982721d287c156b0cf8662/src/ble/bletransport.cpp#L1288-L1315

3. The new test cannot fail on the requeue it is named after (CLAUDE.md says "A test has to be able to fail, has to catch something the suite does not"). The transport is headless, so `subscribeAll()` returns at `if (!m_service) return;`. The INFO log it asserts is emitted before `subscribeAll()` is called, and the write-survival assertion does not depend on it. Deleting the `subscribeAll();` call leaves the test green.

https://github.com/Kulitorum/Decenza/blob/b5284b0cb25d3e240b982721d287c156b0cf8662/tests/tst_blecommandqueue.cpp#L378-L394

https://github.com/Kulitorum/Decenza/blob/b5284b0cb25d3e240b982721d287c156b0cf8662/src/ble/bletransport.cpp#L200-L202

skialpine:
### Code review

Found 5 issues:

1. After a required-stream failure, sleep is never sent (bug due to deferring on `m_connecting`). `failRequiredStream()` drops the ready marker without tearing the link down, so `connected()` never fires and `m_connecting` stays true while `DE1Device.connected` is true. QML auto-sleep, the screensaver and the app-exit path call `goToSleep()`, which now defers forever. Before this PR the sleep went out on the working link.

https://github.com/Kulitorum/Decenza/blob/dbfca86c893158d177101614d54e111563e011e7/src/ble/de1device.cpp#L1563-L1571

https://github.com/Kulitorum/Decenza/blob/dbfca86c893158d177101614d54e111563e011e7/src/ble/bletransport.cpp#L1275-L1297

2. The deferred sleep is announced at INFO but sent at DEBUG (docs/CLAUDE_MD/LOGGING.md: the tier is chosen by audience, and a reader who sees only the first half concludes it never recovered). The connections view shows INFO and above, so it shows "it will be sent once the connection is ready" and never that it was sent.

https://github.com/Kulitorum/Decenza/blob/dbfca86c893158d177101614d54e111563e011e7/src/ble/de1device.cpp#L168-L172

3. A USB DE1 found while a BLE connect is in flight leaves `m_connecting` stuck true, so sleep is never sent for the session (bug due to deferring on `m_connecting`). `SerialTransport::open()` emits `connected()` before `de1Discovered`, so `onTransportConnected()` never runs for USB, and the USB handler calls `disconnect()` only when `isConnected()`, which is false mid-connect.

https://github.com/Kulitorum/Decenza/blob/dbfca86c893158d177101614d54e111563e011e7/src/main.cpp#L2327-L2337

4. The deferred sleep drops the initial reads (bug due to `goToSleep()` calling `clearQueue()` from `onTransportConnected()`). `subscribeAll()` queues VERSION, STATE_INFO, WATER_LEVELS and SHOT_SETTINGS behind the ready marker. The marker's `connected()` runs the deferred `goToSleep()`, whose clear drops them, and neither requeue flag is set by then. VERSION is read only here, so the firmware version is missing for that session.

https://github.com/Kulitorum/Decenza/blob/dbfca86c893158d177101614d54e111563e011e7/src/ble/bletransport.cpp#L237-L247

https://github.com/Kulitorum/Decenza/blob/dbfca86c893158d177101614d54e111563e011e7/src/ble/de1device.cpp#L1578-L1582

5. The new comments say the clear "would drop the connect's setup" (CLAUDE.md: re-derive a comment's justification when the design around it moves). This PR's `clearQueue()` now requeues a dropped setup stage, so only the second reason, the connect-time wake to Idle, still holds. Same claim in the new test's comment.

https://github.com/Kulitorum/Decenza/blob/dbfca86c893158d177101614d54e111563e011e7/src/ble/de1device.cpp#L1562-L1566

https://github.com/Kulitorum/Decenza/blob/dbfca86c893158d177101614d54e111563e011e7/tests/tst_de1device_headless.cpp#L143-L147

---

<details>

- [Worth a look] A wake requested mid-connect (MQTT, /api/power/wake, relay, MCP) does not clear `m_sleepPendingAfterConnect`, so the deferred sleep's clear drops the queued Idle and the machine sleeps right after being woken. One-line fix in `requestState(Idle)` or `wakeUp()` (score: 72)
- [OK to skip] `clearBeforeTheLinkIsReadyKeepsTheNextWrite` still cannot observe the `subscribeAll()` requeue headless; the test name and comment now say so, and the discovery requeue has its own failable test (score: 50)

</details>

skialpine:
### Code review

Found 1 issue:

1. The comment at the end of `dropFor()` still describes only teardown (CLAUDE.md says re-derive a comment's justification "when the design around it moves"). `dropFor()` now backs both `forget()` and the new `clearCommands()`, which runs on a live link and deliberately keeps the connect setup. "The whole point of releasing on teardown is that a dead link does not hold the stack" is false for that caller.

https://github.com/Kulitorum/Decenza/blob/3ec03f5254cc20a29fbd408c150bd4d37c1c7e6b/src/ble/blegattqueue.cpp#L328-L331

No bugs found in the net change. Checked the operation timer when a kept setup operation holds the slot, setup tagging and re-entrancy, callers that rely on a full clear (flow start, stop-at-weight, sleep), and the sleep-while-connecting flag across reconnects, USB and the simulator.

PR #1956, opened 2026-09-19
Send the last sleep or wake requested during a DE1 connect

skialpine:
## Problem

After #1955, a sleep requested while the DE1 was connecting made the connect skip its usual wake to Idle. A wake during that same connect was lost: `ScreensaverPage.wake()` only called `DE1Device.wakeUp()` when the DE1 was connected, so while it was still connecting the wake never reached `DE1Device`. The connect then skipped its wake and the machine stayed asleep behind an awake app.

Seen on an SM-X210 on build 3597 (2026-09-19, session starting 11:47:45):

```
[   4.314] goToScreensaver called                     (connecting; characteristics not ready until 10.188)
[   6.359] Waking from screensaver
[  11.556] INFO  [DE1][Device] Not waking the DE1 on connect: it was put to sleep while connecting
[  11.932] DEBUG [DE1][WaterLevel] 537 ml (20.6 mm) first reading [Sleep]
```

The same log shows a second, pre-existing cost. A sleep sent before the characteristics are ready is written, fails five retries, and is reported as a DE1 link fault, which latched the scale to BALANCED for the rest of the app run:

```
[   6.911] WARN  [DE1][BLE] Write FAILED after 5 retries (uuid=0000a002, 1 bytes)
[   6.957] WARN  [Scale][BLE QtTransport] Scale connection-priority: app-run skip-HIGH latch SET (trigger=de1-fault-cluster)
```

## Fix

While connecting, the latest sleep or wake is the state the connect sends.

- `goToSleep()` records it. Before the characteristics are ready it writes nothing, so there is no doomed write and no link fault. Once the link can carry writes it still sends at once, so a connect that never completes (a required-stream failure, USB taking over) cannot hold the sleep.
- `wakeUp()` before the characteristics are ready cancels the pending sleep and writes nothing; the connect then sends Idle.
- `onTransportConnected()` sends Sleep if the last request was sleep, otherwise Idle as before. It is a plain state write, not `goToSleep()`, so the initial reads queued behind the setup are kept.
- `ScreensaverPage.wake()` calls `wakeUp()` while connecting as well as when connected.

## Tests

`tst_de1device_headless::connectSendsTheLastSleepOrWakeRequestedWhileConnecting`:
- sleep before the link is ready: nothing written, then Sleep on connect;
- sleep then wake before the link is ready: nothing written, then Idle on connect;
- sleep while connecting with the link already carrying writes: Sleep sent at once.

Verified red with the sleep hold reverted (a write before the link is ready) and with the wake handling reverted (an Idle written before the link is ready). Full suite: 118/118 passed locally (macOS, Debug).

The `ScreensaverPage.qml` change is not covered by a unit test and needs checking on the tablet: launch, let the screensaver come up during the connect, wake it before the connect finishes, and the machine should come out of sleep once connected.


skialpine:
### Code review

Found 3 issues:

1. A held sleep that a wake cancels leaves no trace in the log (docs/CLAUDE_MD/LOGGING.md: "Don't announce an intent as if it were an outcome"). `goToSleep()` logs at INFO that the sleep "will be sent once the connection is ready"; if `wakeUp()` then cancels it before the link is ready it returns silently, and the connect sends Idle with no line saying the sleep was dropped.

https://github.com/Kulitorum/Decenza/blob/4e7dfa044136b48407474ea68c6b4b0c1fcf7123/src/ble/de1device.cpp#L1572-L1577

https://github.com/Kulitorum/Decenza/blob/4e7dfa044136b48407474ea68c6b4b0c1fcf7123/src/ble/de1device.cpp#L1597-L1604

2. The comment justifying the plain Sleep write on connect gives a reason that no longer holds (CLAUDE.md: re-derive a comment's justification "when the design around it moves"). Since #1955, `clearQueue()` keeps operations tagged `connectSetup`, and `subscribeAll()` tags the initial reads, so `goToSleep()`'s clear would not drop them.

https://github.com/Kulitorum/Decenza/blob/4e7dfa044136b48407474ea68c6b4b0c1fcf7123/src/ble/de1device.cpp#L169-L176

3. When a sleep is held and the machine was awake at connect, the screensaver flickers (bug due to the Sleep write being queued behind the connect setup). The machine's first STATE_INFO reports Idle before the queued `requestState(Sleep)` lands, so ScreensaverPage's `onStateChanged` wakes the scale and calls `idleFromScreensaverRequested()`; the Sleep phase then re-shows the screensaver. End state is correct.

https://github.com/Kulitorum/Decenza/blob/4e7dfa044136b48407474ea68c6b4b0c1fcf7123/qml/pages/ScreensaverPage.qml#L921-L935

---

<details>

- [OK to skip] A held sleep survives BleTransport's internal service-discovery retry and is sent when that retry connects. That retry is the same connect attempt the sleep was requested during, so sending it matches the request (score: 70)
- [OK to skip] A held sleep is dropped if the connect attempt fails and the reconnect ladder starts a new one (`disconnect()` clears it). Before this PR that sleep was a failed write and also lost; keeping it across attempts is a design choice, not a one-line fix (score: 65)
- [OK to skip] During BleTransport's internal service-discovery retry `m_connecting` is false, so the old immediate-write path applies. Pre-existing and rare (Android "DE1 service not found") (score: 65)
- [OK to skip] A sleep sent in the ready window (characteristics ready, before the ready marker) is written again by the connect. A second Sleep to a sleeping DE1 is harmless (score: 50)

</details>

PR #1957, opened 2026-09-19
Log every DE1 state request, and don't re-send a sent sleep

skialpine:
Follow-up to #1956, from the seven sessions logged on the SM-X210 after that build shipped.

## What the field log showed

Six of the seven sessions behaved correctly, including the exact regression #1956 fixed: a wake during a connect now reaches the machine ~1 s after the link is ready (sessions at 12:32:43 and 12:34:26).

One session (12:33:20) still showed the reported symptom — the machine stayed asleep across a wake and only woke 14 s later, after a second sleep/wake cycle. That session is also the only one of the seven that put **two** Sleeps on the wire.

## The defect

`goToSleep()` set `m_sleepRequestedWhileConnecting` whenever `m_connecting` was true, including the branch where the link is already ready and it writes the sleep immediately. `onTransportConnected()` then read that flag and sent the sleep a second time.

That branch is reachable only between "Characteristics ready" and the ready marker — 0.75 s in the failing session, which the tap landed inside.

The flag was carrying two meanings. Split:

- `m_sleepRequestedWhileConnecting` — the connect must not wake the machine.
- `m_sleepAwaitingReadyLink` — the sleep has not been written, so the connect owes it.

Only the second makes the connect send.

**I cannot show that the duplicate Sleep is what swallowed the wake.** Both Sleeps went out ~3.7 s before the wake write, and I have no mechanism by which that makes the machine ignore a later Idle. The correlation is suggestive, not proof, so this does not claim to close the stuck-asleep report.

## Why the log couldn't answer it

`requestState()` logged nothing. A sleep appears in the log twice over; the app asking the machine to wake left no trace, so a missing wake can't be told from one that was never sent. Every `REQUESTED_STATE` write is now a DEBUG line, which is what makes a third round of field testing worth doing.

## Test

Extended `connectSendsTheLastSleepOrWakeRequestedWhileConnecting()`'s third fixture — the link-ready-but-still-connecting case — to drive the connect to completion. It previously stopped at the immediate write, which is why the double-send got through.

Reaching that required `MockTransport::emitConnectedSim()`: `setConnectedSim(true)` returns early when the transport is already connected, so it cannot announce a connect on a link that is already carrying writes — precisely the window this bug lives in.

Verified red before green: `Compared lists have different sizes. Actual size: 2, Expected size: 1`.

Full suite: 118 passed, 0 failed, no warnings. `check_log_markers.py` and `check_test_source_duplication.py` both clean.

## Not in the release notes

The logging is invisible to users, and the stuck-asleep fix is unproven, so neither belongs in the v2.0.6 notes as written.


skialpine:
### Code review

Found 2 issues, both now fixed in [fb970f9f](https://github.com/Kulitorum/Decenza/commit/fb970f9f).

1. `onTransportDisconnected()` cleared neither sleep flag, so a sleep held for a connect that FAILED suppressed the next connect's wake. `disconnect()` was the only teardown that cleared them, and `BleTransport` retries on the same object — a failed attempt reaches `onTransportDisconnected()` and never `disconnect()`. This was a regression introduced by the first commit: previously the stale flag made the next connect re-send Sleep, which at least slept the machine; after the split it could send nothing at all.

https://github.com/Kulitorum/Decenza/blob/ac91c1ee2b6b39a3b0c4e8d8b1b47ba5e9f7a9c1/src/ble/bletransport.cpp#L112-L137

2. `REQUESTED_STATE` is written from three places — `requestState()`, `stopOperationUrgent()` and `goToSleep()` — and the first commit logged only one, so the PR's claim that every state write is logged was untrue: the urgent SAW stop and every ordinary sleep stayed invisible. The log also sat above `requestState()`'s own `!m_transport` guard, so a dropped request would read as "app asked, machine ignored" — the exact misdiagnosis the line exists to prevent. All three now route through one writer that logs at the write, below the guards. (CLAUDE.md: "Centralize anything produced at more than one site — never hand-roll it per call.")

https://github.com/Kulitorum/Decenza/blob/ac91c1ee2b6b39a3b0c4e8d8b1b47ba5e9f7a9c1/src/ble/de1device.cpp#L1544-L1550

Both are covered by tests verified red-then-green. The new `aFailedConnectDoesNotCarryItsSleepIntoTheNext()` fails without the fix with `Actual "\x00"` (Sleep) against `Expected "\x02"` (Idle). Full suite 118/118.

Also corrected without separate findings: `wakeUp()`'s "the sleep will not be sent" INFO now keys off the unsent sleep rather than the suppression flag, so it cannot claim to cancel a sleep already on the wire; the comment above the connect's sleep branch now describes all three outcomes rather than two; and `goToSleep()` no longer claims the write discharges the sleep, since `writeUrgent()` is queue position and not delivery.

<details>

- [OK to skip] An abandoned urgent write loses the held sleep, since the flag is discharged at enqueue — accepted tradeoff against a connect that re-sent a Sleep every time; now stated in the comment rather than fixed with a re-arm in `onWriteAbandoned()` (score: 60)
- [OK to skip] `goToSleep()`'s profile-upload early return sets neither connecting flag, so a sleep requested while connecting and uploading lets the connect wake first — uploads only start after the connect completes, so it is unreachable (score: 50)

</details>

